Issue #22, opened 2018-12-11
Schema must be aware of the document id


Kerollmops:
Currently the schema know which fields of the document should be indexed and stored. Using the `SchemaBuilder` struct we declare each field in the order of importance we want and the document id is unknown to the finally built schema.

The advantage of this solution is that it will be impossible to serialize a document that does not have an id. The serializer will not be able to generate the appropriate key and therefore not be able to store it in the database.

We must absolutely keep some kind of "raw" API that will allow us to store a document under any id and with any attributes.

How should we modify the `SchemaBuilder` API ? What if the user want its id to be stored and indexed ?

We will probably need to transform the id (i.e. hashing it) to be able to save the in the database, when a user will need to identify a document (i.e. deleting it), he will refer to its original id, not the transformed one.

Kerollmops:
We though about an API that could look like this:

```rust
let mut schema = SchemaBuilder::with_identifier("id");
schema.new_attribute("id", STORED | INDEXED);
schema.new_attribute("title", STORED | INDEXED);
schema.new_attribute("description", STORED | INDEXED);
schema.new_attribute("image", STORED);
schema.build()
```

PR #44, opened 2018-12-22
Create a real DocumentId type

Kerollmops:
Forcing the `DocumentId` to be something internal will permit to avoid possible miss comparisons to be done with other types. Avoiding users to compute it themself and possibly fuck it up.

In a near future the internal `u64` will be private therefore it will be impossible to construct it outside of the library, it will entirely computed by the library (i.e. #22).

### DIFF
--- a/src/database/document_key.rs
+++ b/src/database/document_key.rs
@@ -19,7 +19,7 @@ impl DocumentKey {
 
         let mut wtr = Cursor::new(&mut buffer[..]);
         wtr.write_all(b"doc-").unwrap();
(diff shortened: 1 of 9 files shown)


src/database/document_key.rs
```diff
@@ -88,7 +89,8 @@ impl DocumentKeyAttr {
     }
 
     pub fn document_id(&self) -> DocumentId {
-        (&self.0[4..]).read_u64::<NativeEndian>().unwrap()
```
Kerollmops: Prefer implementing the `io::Read` and `io::Write` traits on `DocumentId`.

PR #59, opened 2019-01-01
Cache some criteria

Kerollmops:
Allow to add some cache to criteria that can take advantage of it.


Kerollmops:
it's not worth it.

PR #111, opened 2019-02-20
Add a config per index

qdequele:


src/database/config.rs
```diff
@@ -0,0 +1,85 @@
...
+
+impl<'de> Deserialize<'de> for RankingOrdering {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where D: Deserializer<'de>
+    {
+        let s = String::deserialize(deserializer)?;
+        Ok(match s.as_str() {
+            "asc" => RankingOrdering::Asc,
+            "dsc" => RankingOrdering::Dsc,
+            _ => RankingOrdering::Dsc
```
Kerollmops: It is not correct to accept any string without any error, if the string does not match "asc" or "dsc" so we need to return at least `D::Error::custom("invalid ordering")`.

src/database/config.rs
```diff
@@ -0,0 +1,85 @@
...
+        })
+    }
+}
+
+impl<'de> Deserialize<'de> for RankingOrdering {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where D: Deserializer<'de>
+    {
+        let s = String::deserialize(deserializer)?;
+        Ok(match s.as_str() {
```
Kerollmops: We should be case insensitive on this. We could use [`str::eq_ignore_ascii_case`](https://doc.rust-lang.org/std/primitive.str.html#method.eq_ignore_ascii_case).

src/database/config.rs
```diff
@@ -0,0 +1,85 @@
...
+impl Config {
+    pub(crate) fn default() -> Config {
+        Config {
+            stop_words: None,
+            ranking_order: None,
+            distinct_field: None,
+            ranking_rules: None,
+            access_token: None,
+        }
+    }
```
Kerollmops: The `#[derive(Default)]` syntax would have produced the same output, please remove this function and replace it by the derived one.

src/database/config.rs
```diff
@@ -0,0 +1,85 @@
...
+        if let Some(distinct_field) = new.distinct_field {
+            self.distinct_field = Some(distinct_field);
+        };
+        if let Some(ranking_rules) = new.ranking_rules {
+            self.ranking_rules = Some(ranking_rules);
+        };
+        if let Some(access_token) = new.access_token {
+            self.access_token = Some(access_token);
+        };
+    }
```
Kerollmops: This could be rewritten like so:

```rust
self.access_token = new.access_token.or(self.access_token);
```

Kerollmops: Nope, it couldn't, due to borrowing rules.

PR #130, opened 2019-03-16
Moving to a workspace

Kerollmops:
This introduces a new era of MeiliDB: splitting the whole project into sub-crates.

It will help scaling for new features, keeping the functionality enclosed and keeping them from spreading into other parts of the engine.

This update currently miss two main sub-crates: `meilidb-data` and `meilidb-network`. The former is glued with `meilidb-search` (currently named `meilidb`) and will be split in the future. The latter, on the other side, doesn't even exists for the moment.


Kerollmops:
Closing is favor of #147

PR #167, opened 2019-07-01
Allow the schema to accept new attributes

Kerollmops:


Kerollmops:
This is outdated and no longer relevant as we now have a full ACID key-value store, no more `ArcSwap` needed.

PR #180, opened 2019-09-03
Change the STORED attribute property by DISPLAYED

Kerollmops:
This pull request make the database stores all documents attributes, this is the first step toward alterable schemas.

Altering schemas would allow user to:
  1. change indexing properties of documents fields *after* the first indexation
  2. adding new fields to documents
  3. reordering fields


nemosupremo:
I found this issue through a search; and I was wondering if Meilisearch still supports not storing the original document?

Reason being is I'm trying to build a low overhead search solution where the actual content is stored in a separate database. The actual search content is rather large (over a TB, raw), so not duplicating this content saves disk, ram, and money.

PR #201, opened 2019-09-26
Add more methods for updates process

qdequele:
Add a method to get the current processed update id & next updates in queue

### DIFF
--- a/meilidb-data/src/database/index/mod.rs
+++ b/meilidb-data/src/database/index/mod.rs
@@ -359,6 +359,24 @@ impl Index {
         SynonymsDeletion::new(self)
     }
 
+    pub fn current_update_id(&self) -> Result<Option<u64>, Error> {
+        if let Some((key, _)) = self.updates_index.iter()?.next() {
+            return Ok(Some(key.as_ref().try_into().map(u64::from_be_bytes).unwrap()))
+        }
+        Ok(None)
+    }
+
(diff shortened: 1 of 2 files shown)


meilidb-data/src/database/index/mod.rs
```diff
@@ -359,6 +359,24 @@ impl Index {
         SynonymsDeletion::new(self)
     }
 
+    pub fn current_update_id(&self) -> Option<u64> {
+        if let Some((key, _)) = self.updates_index.iter().unwrap().next() {
+            return Some(key.as_ref().try_into().map(u64::from_be_bytes).unwrap())
```
Kerollmops: Please do not unwrap the `iter()` method, prefer using a `Result<Option<u64>, Error>` return type and use the `?` operator.

meilidb-data/src/database/index/mod.rs
```diff
@@ -359,6 +359,24 @@ impl Index {
...
+        }
+        None
+    }
+
+    pub fn enqueued_updates_ids(&self) -> Vec<u64> {
+        let mut updates = Vec::new();
+
+        for (key, _) in self.updates_index.iter().unwrap() {
+            let update_id = key.as_ref().try_into().map(u64::from_be_bytes).unwrap();
```
Kerollmops: Please do not unwrap the `iter()` method, prefer using a `Result<Vec<u64>, Error>` return type and use the `?` operator.

PR #232, opened 2019-10-23
Support splitted words

Kerollmops:
for a query containing `"serachengine"`, a document containing `"search engine"` will match but not one containing `"search slow engine"` for example.


Kerollmops:
Fixes #232

PR #279, opened 2019-11-12
Slogan and Resume proposition

bidoubiwa:


bidoubiwa:
Any problem with this PR ?

PR #291, opened 2019-11-13
Make changes to speedup the update system

Kerollmops:
We do not clean up index when deleting documents currently.


Kerollmops:
No more relevant.

PR #298, opened 2019-11-15
Create a dataset where the release_date is a numeric timestamp

bidoubiwa:
```bash
cat movies.json | jq '.[]' |jq -c 'try (.release_date |= if . == null or . == "" then . else (. | strptime("%Y-%m-%d") | mktime )  end) catch (.)'  >> movies_ranged.json
```

Added `-c` for minify version.
Since try and catch doesnt work on mapping function I had to append a file with each object.
Because of that, the final file is not formatted well.
Missing wrapping '`[]`' around all objects.
Missing comas after each object.
vim command to add a comma at the end of each object `'%s/}/},/g'`


Kerollmops:
Can you please replace the current dataset instead of adding a new one ? please.

bidoubiwa:
It's done

Kerollmops:
Why do this new dataset have 6 times more lines than the old one ?

bidoubiwa:
`cat movies.json | jq '.[]' |jq -c 'try (.release_date |= if . == null or . == "" then . else (. | strptime("%Y-%m-%d") | mktime )  end) catch (.)'  >> movies_ranged.json`

Added `-c` for minify version.
Since try and catch doesnt work on mapping function I had to append a file with each object.
Because of that, the final file is not formatted well.
Missing wrapping '`[]`' around all objects.
Missing comas after each object.
vim command to add a comma at the end of each object `'%s/}/},/g'`

PR #324, opened 2019-11-22
Replace Azure CI by Github Actions

tpayet:
- Add gh actions for cargo check using rust nightly
- Add readme about actions workflows
- Add basic Dockerfile
- Add action workflow for docker publish
- Change check action to test action


.github/workflows/nightly-test.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  check:
+    name: MeiliDB
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Install latest nightly
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
```
Kerollmops: We now compile on stable, would it be possible to change that?

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM rustlang/rust:nightly
```
Kerollmops: Same, would it be possible to make it compile on stable?

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+
+WORKDIR /meilisearch
+
+COPY . .
+
+RUN cargo build
+
+CMD cargo run
```
Kerollmops: Wouldn't it be preferable to compile using `cargo build --release`, copy the `/target/release/meilidb-http` binary into _another_ sub image. This way we avoid compiling it every times it run and avoid keeping useless dependencies like `cargo`, `gcc`.

.github/workflows/README.md
```diff
@@ -0,0 +1,11 @@
+# GitHub actions workflow for MeiliDB
+
+> **Note:**
+> - The repository is currently compiling with Rust Nighlty, we aim to compile on Rust stable as soon as possible.
```
Kerollmops: We compile with Rust stable now

Dockerfile
tpayet: yup ;)

.github/workflows/README.md
```diff
@@ -0,0 +1,11 @@
...
+> - We do not use [cache](https://github.com/actions/cache) yet but we could use it to speed up CI
+
+## Workflow
+
+-  On each pull request, we are triggering `cargo test`.
```
Kerollmops:
```suggestion
- On each pull request, we are triggering `cargo test`.
```

Kerollmops:
You do not have to add the `MEILI_NO_ANALYTICS` env variable because analytics are only send when using `cargo run`, my bad!

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# Run
+FROM    alpine:3.10
+
+RUN     apk update --quiet
+RUN     apk add libressl
+RUN     apk add build-base
+
+COPY    --from=compiler /meilisearch/target/release/meilidb-http .
+
+CMD     ./meilidb-http
```
Kerollmops: Did you try to run that? I remember that alpine isn't shipped with a libc and that was a pain in the ass to make Rust work without it and use musl instead. I would prefer using Debian.

tpayet: This was not about the docker image because we are not using the docker image for the test. It was the GitHub actions that was not using the latest rust version, and I just fixed it :)

Kerollmops: Yes of course, that's just the Github Review interface that make it like those two comments are related :)

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN     apk add build-base
+RUN     apk add libressl-dev
+
+RUN     curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+
+WORKDIR /meilisearch
+
+COPY    . .
+
+ENV     RUSTFLAGS="-C target-feature=-crt-static"
```
Kerollmops: Why did you specified that line? To make it act like a fully static binary? Does it work?

README.md
```diff
@@ -1,6 +1,7 @@
+
```
Kerollmops: Why did you add a new line here? :)

Kerollmops:
Would it be possible to change the main readme to tell that we use stable rust now?

Dockerfile
Kerollmops: Ok, that make it compile in fact

PR #343, opened 2019-11-27
Change the update loop to be more explicit on index clear

Kerollmops:


meilisearch-core/src/database.rs
```diff
@@ -55,7 +55,23 @@ fn update_awaiter(
     index: Index,
 ) -> MResult<()> {
     let mut receiver = receiver.into_iter();
-    while let Some(UpdateEvent::NewUpdate) = receiver.next() {
+    while let Some(event) = receiver.next() {
+
+        // if we receive a *MustClear* event, clear the index and break the loop
+        if let UpdateEvent::MustClear = event {
```
Kerollmops: As you can see here, this comment is perfectly aligned with the `MustClear` type just below.

PR #367, opened 2019-12-05
Allow users to send csv files from stdin in examples

Kerollmops:


bidoubiwa:
@Kerollmops Could you explain to me how this works?

PR #368, opened 2019-12-05
Add publish action to gemfury for apt pkg

tpayet:
Upload the .deb package on gemfury apt repository.

Usage to install on debian-based OS:
```bash
$ echo "deb [trusted=yes] https://apt.fury.io/meilisearch/ /" > /etc/apt/sources.list.d/fury.list
$ apt update && apt install meilisearch-http
$ meilisearch
Server is listening on: http://127.0.0.1:7700
```

I wonder if I should add this usage in the README.md but it may add some complexity.
@Kerollmops Shoud we have an INSTALL.md file somewhere at the root of the project or in the misc/ directory?

PS: When merging this PR, you should add the GEMFURY_PUSH_TOKEN secrets in the repository settings

### DIFF
--- a/.github/workflows/publish-deb-pkg.yml
+++ b/.github/workflows/publish-deb-pkg.yml
@@ -1,4 +1,4 @@
-name: Publish debian package to GitHub release
+name: Publish deb pkg to GitHub release & apt repository
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


Kerollmops:
I think we should add those lignes on the README, even if it seems to much ;)

When you make the `.service` PR, you talk to me, we will probably create an INSTALL.md and only keep the most common installation intructions on the main README.

For the moment I would prefer you add the instruction instead of nothing 😺 

Thank you !

tpayet:
Okay, I put the usage in the README 👍

PR #373, opened 2019-12-05
Use POST instead of DELETE method to delete stops-word

### LINKED ISSUE #371: Use POST instead of DELETE to delete a stop-word
Because we need to pass a body

curquiza:
closes #371


Kerollmops:
Nice thank you!

bidoubiwa:
This is going to be added in the next release: v0.9.0 but it's also contradictory with 
https://github.com/meilisearch/MeiliSearch/issues/417. 
Is this going to be overridden?

bidoubiwa:
documentation : meilisearch/documentation#107

PR #375, opened 2019-12-06
Docker command updated

nithinkashyapn:
Docker does not allow Uppercase letters, throws this error 

`docker: invalid reference format: repository name must be lowercase.`


Kerollmops:
Ho! Thank you was a mistake :)

PR #384, opened 2019-12-09
Change regexp in install script

curquiza:


download-latest.sh
```diff
@@ -1,12 +1,12 @@
...
 # COLORS
-RED="\033[31m"
-GREEN="\033[32m"
-DEFAULT="\033[0m"
+RED='\033[31m'
+GREEN='\033[32m'
+DEFAULT='\033[0m'
 
 # GLOBALS
-REGEXP_SEMVER='v[^0-9]*\([0-9]*\)[.]\([0-9]*\)[.]\([0-9]*\)\([0-9A-Za-z-]*\)'
```
Kerollmops: What were those `[.]`?

Kerollmops: Ok I understand, please keep them like that as this is specifying the raw `.`, if you remove the brackets you specify any character. And therefore `v0x6f8` would be accepted.

curquiza: Brackets are usefull when there are many characters. It means `one of the characters in the brackets`.

Kerollmops: Yes, but if you remove them therefore you use the `.`, the `.` means `any character you want`. You don't want that, you want the raw `.` character, specifying `[.]` means, `one of the characters in the brackets` and therefore, as there is only one character, `match a point please`.

curquiza: No, I don't think so:
```bash
$ echo '"v0x6f8.12.1"' | grep '\"v\([0-9]*\).\([0-9]*\).\([0-9]*\)\"'
[nothing]
$ echo '"v0.12.1"' | grep '\"v\([0-9]*\).\([0-9]*\).\([0-9]*\)\"'
"v0.12.1"
```

curquiza: Maybe I did not understand

curquiza: (I put back the brackets btw)

Kerollmops: I think you did not understand, here the example showing what I want to explain:
```bash
echo '"v0x12x1"' | \grep '\"v\([0-9]*\).\([0-9]*\).\([0-9]*\)\"'
"v0x12x1"
```

Kerollmops: Your previous example showed that `"v0x6f8.12.1"` did not match but it was because you had too many numbers specified.

Kerollmops:
Thank you, great Job! Merging :)

PR #415, opened 2019-12-17
Use a main read transaction instead of a write one

### LINKED ISSUE #413: Blocking setting update request
The current settings update route is blocking due to the fact that we use a write transaction on the main database which must never occur in an HTTP function.

https://github.com/meilisearch/MeiliSearch/blob/59f74dabe7fb582c966d0aa64aab5b64b7a5ac32/meilisearch-http/src/routes/setting.rs#L67-L107

Kerollmops:
Closes #413


bidoubiwa:
No documentation needed

PR #433, opened 2019-12-19
Set the indexes info in the create_index function

Kerollmops:
Adding essential data like the name, the creation, and update date at creation time allows us to open an index created using the `from_file` example with the HTTP server.


bidoubiwa:
No documentation needed

PR #435, opened 2019-12-20
Add more debug timings

Kerollmops:
I fix the time measurement of the FST search and added debug on the automatons loop.


bidoubiwa:
No documentation needed

PR #439, opened 2019-12-30
Fix a blocking channel, appearing like a deadlock

Kerollmops:
I used a bounded crossbeam channel to send update notifications and awaited the receiver to receive the notification, when the channel was full the update system appeared like it was deadlocked because processing updates is slower than pushing them in the update database, therefore the channel fills up faster than it empties.


bidoubiwa:
No documentation needed

PR #444, opened 2020-01-09
Add cjk filter for korean language

djKooks:
Hello. I'm not sure of your roadmap so this may could be useless...

I've update `is_cjk` module because seems filter for 'korean' language is missing. Could you check? Seems there are several more cjk unicodes that are not being filtered...
If you've not updated because of your plan, please ignore it.

Thank you.

### DIFF
--- a/meilisearch-tokenizer/src/lib.rs
+++ b/meilisearch-tokenizer/src/lib.rs
@@ -4,15 +4,18 @@ use slice_group_by::StrGroupBy;
 use std::iter::Peekable;
 
 pub fn is_cjk(c: char) -> bool {
-    (c >= '\u{2e80}' && c <= '\u{2eff}')
-        || (c >= '\u{2f00}' && c <= '\u{2fdf}')
-        || (c >= '\u{3040}' && c <= '\u{309f}')
-        || (c >= '\u{30a0}' && c <= '\u{30ff}')
+    (c >= '\u{2e80}' && c <= '\u{2eff}')  // CJK Radicals Supplement
+        || (c >= '\u{2f00}' && c <= '\u{2fdf}') // Kangxi radical
+        || (c >= '\u{3040}' && c <= '\u{309f}') // Japanese Hiragana
+        || (c >= '\u{30a0}' && c <= '\u{30ff}') // Japanese Katakana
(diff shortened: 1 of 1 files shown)


tpayet:
Hey, thank you for this contribution :)
It is definitely in our roadmap and will be merged as soon as @Kerollmops can take a closer look to this.

djKooks:
@tpayet thanks!
I'll check more if there is other unicode related with CJK.

Kerollmops:
Hey,
Would it be possible to rebase this PR and let me merge or would you like to check for more CJK kanji related problem?

djKooks:
@Kerollmops thanks for your check!
I've just update some filter by referring below:
- https://en.wikipedia.org/wiki/Korean_language_and_computers
- http://www.rikai.com/library/kanjitables/kanji_codes.unicode.shtml

One point want you to check is `Full-width roman characters` in second link...but as I know it is not commonly used outside of japan.
Could you check?

Thanks.

Kerollmops:
I think this not a problem to split by `Full-width roman characters`, I think I will merge that.
What could be wrong with these characters? Is the range you defined invalid?

By the way would it be possible you rebase on master instead of merging master into this branch, please? If it is too much I will merge like that, no worries :)

djKooks:
@Kerollmops thanks for review.

djKooks:
@Kerollmops not sure I've done correctly...could you check? Or I could just create new PR(because there are only 1 file changed...). Feel free to suggest~

Kerollmops:
You should try to remove the last merge commit by running `git reset --hard HEAD~` IIRC.
You rebased correctly, the only thing missing is the useless merge commit :)

djKooks:
@Kerollmops seems commits are getting dirty 😢 so updated new PR for same change. Could you check #456 ?

n826vnl8:
Has unicode already been supported?  I just tried and it seems still not working :)

n826vnl8:
I checked the is_cjk function in tokenizer and it seems that it's not called for all cjk chars during index request though is_cjk implementation is correct.

Kerollmops:
@n826vnl8 Is it not just related to the highlight system, in the screenshots above it seems like the documents are detected bu the highlight is miss placed.

Am I missing something?

PR #456, opened 2020-01-30
Update CJK filter

djKooks:
Update CJK filter


Kerollmops:
Thank you very much for your work!

PR #489, opened 2020-02-28
Use distinct on search

qdequele:


meilisearch-core/src/update/settings_update.rs
```diff
@@ -60,9 +63,13 @@ pub fn apply_settings_update(
 
     match settings.distinct_attribute {
         UpdateState::Update(v) => {
+            schema.set_ranked(&v)?;
             index.main.put_distinct_attribute(writer, &v)?;
         },
         UpdateState::Clear => {
+            if let Some(ranked_field) = index.main.distinct_attribute(writer)? {
+                schema.remove_ranked(&ranked_field);
+            }
```
Kerollmops: What if the distinct attribute is also used like a criterion? (`asc(myattribute)`).
If you remove the distinct attribute from the ranked attributes list it will invalidate the criterion.

meilisearch-http/src/helpers/meilisearch.rs
```diff
@@ -205,6 +207,23 @@ impl<'a> SearchBuilder<'a> {
 
         query_builder.with_fetch_timeout(self.timeout);
 
+        // Apply Distinct if it's wished
+
```
Kerollmops:
```suggestion
```

PR #496, opened 2020-03-02
Fix some issues before v0.9

qdequele:


meilisearch-http/src/option.rs
```diff
@@ -17,9 +17,6 @@ pub struct Opt {
     pub master_key: Option<String>,
 
     /// This environment variable must be set to `production` if your are running in production.
-    /// Could be `production` or `development`
-    /// - `production`: Force api keys
-    /// - `development`: Show logs in "info" mode + not mendatory to specify the api keys
```
Kerollmops: > If the development flag is set it will show more logs, and the master key can be avoided which implies that there is no security on the updates routes. This is useful to debug when integrating the engine in another service.

qdequele:
fix #505

meilisearch-http/src/option.rs
qdequele: I fix it: 
Capture d’écran 2020-03-06 à 10 16 40

PR #509, opened 2020-03-05
Fix internal schema

qdequele:


meilisearch-schema/src/error.rs
```diff
@@ -6,6 +6,7 @@ pub type SResult<T> = Result<T, Error>;
 #[derive(Debug)]
 pub enum Error {
     FieldNameNotFound(String),
+    SchemaAlreadyHaveIdentifier,
```
Kerollmops:
```suggestion
    IdentifierAlreadyPresent,
```

meilisearch-core/src/error.rs
```diff
@@ -87,7 +87,7 @@ impl fmt::Display for Error {
         match self {
             Io(e) => write!(f, "{}", e),
             IndexAlreadyExists => write!(f, "index already exists"),
-            MissingIdentifier => write!(f, "schema cannot be built without identifier"),
+            MissingPrimaryKey => write!(f, "schema cannot be built without primary key"),
```
Kerollmops:
```suggestion
            MissingPrimaryKey => write!(f, "schema cannot be built without a primary key"),
```

meilisearch-core/src/serde/mod.rs
```diff
@@ -57,7 +57,7 @@ impl fmt::Display for SerializerError {
                 f.write_str("serialized document does not have an id according to the schema")
             }
             SerializerError::InvalidDocumentIdType => {
-                f.write_str("documents identifiers can be of type integer or string only composed of alphanumeric characters, hyphens (-) and underscores (_).")
+                f.write_str("documents primary keys can be of type integer or string only composed of alphanumeric characters, hyphens (-) and underscores (_).")
```
Kerollmops:
```suggestion
                f.write_str("a document primary key can be of type integer or string only composed of alphanumeric characters, hyphens (-) and underscores (_).")
```

meilisearch-http/src/helpers/tide.rs
```diff
@@ -55,7 +55,7 @@ impl RequestExt for Request<Data> {
     fn url_param(&self, name: &str) -> SResult<String> {
         let param = self
             .param::<String>(name)
-            .map_err(|_| ResponseError::bad_parameter("identifier", name))?;
+            .map_err(|e| ResponseError::bad_parameter(name, e))?;
```
Kerollmops: Why do you change that? Does this mean the identifier is now the name of it directly (like "objectId" or "documentId")?

meilisearch-http/src/helpers/tide.rs
```diff
@@ -69,10 +69,10 @@ impl RequestExt for Request<Data> {
         Ok(index)
     }
 
-    fn identifier(&self) -> SResult<String> {
+    fn document_id(&self) -> SResult<String> {
         let name = self
-            .param::<String>("identifier")
-            .map_err(|_| ResponseError::bad_parameter("identifier", "identifier"))?;
+            .param::<String>("documentId")
+            .map_err(|_| ResponseError::bad_parameter("documentId", "primaryKey"))?;
```
Kerollmops: Why is it different than the previous Request? Don't we need to use the primaryKey name used by the user here?

meilisearch-http/src/routes/document.rs
```diff
@@ -134,22 +133,24 @@ async fn update_multiple_documents(mut ctx: Request<Data>, is_partial: bool) ->
...
                 None => return Err(ResponseError::bad_request("Could not infer a schema")),
             },
         };
-        let settings_update = SettingsUpdate {
-            identifier: UpdateState::Update(id),
-            ..SettingsUpdate::default()
-        };
-        index.settings_update(&mut update_writer, settings_update)?;
+
+        if schema.set_primary_key(&id).is_ok() {
```
Kerollmops: What must happen if there is an error here?

meilisearch-http/src/routes/index.rs
```diff
@@ -224,15 +231,19 @@ pub async fn update_index(mut ctx: Request<Data>) -> SResult<Response> {
...
-    if let Some(identifier) = body.identifier {
-        if let Ok(Some(_)) = index.main.schema(&writer) {
-            return Err(ResponseError::bad_request(
-                "The index identifier cannot be updated",
-            ));
+    if let Some(id) = body.primary_key.clone() {
+        if let Some(mut schema) = index.main.schema(&mut writer)? {
+            match schema.primary_key() {
+                Some(_) => {
+                    return Err(ResponseError::bad_request("The index primary key cannot be updated"));
```
Kerollmops:
```suggestion
                    return Err(ResponseError::bad_request("The primary key cannot be updated"));
```

meilisearch-http/src/routes/index.rs
```diff
@@ -224,15 +231,19 @@ pub async fn update_index(mut ctx: Request<Data>) -> SResult<Response> {
...
+                },
+                None => {
+                    if let Ok(_) = schema.set_primary_key(&id) {
+                        index.main.put_schema(&mut writer, &schema)?;
+                    }
```
Kerollmops: What must happen if there is an error here?

meilisearch-http/src/routes/setting.rs
```diff
@@ -63,11 +59,7 @@ pub async fn get_all(ctx: Request<Data>) -> SResult<Response> {
             .iter()
             .map(|s| (*s).to_string())
             .collect::<HashSet<String>>();
-        if attrs.is_empty() {
-            None
-        } else {
-            Some(attrs)
-        }
+        Some(attrs)
```
Kerollmops: I would prefer you remove this useless `Some` wrapping and put it at Settings creation, like the others.

meilisearch-http/src/routes/setting.rs
```diff
@@ -50,11 +50,7 @@ pub async fn get_all(ctx: Request<Data>) -> SResult<Response> {
             .iter()
             .map(|s| (*s).to_string())
             .collect::<Vec<String>>();
-        if attrs.is_empty() {
-            None
-        } else {
-            Some(attrs)
-        }
+        Some(attrs)
```
Kerollmops: I would prefer you remove this useless `Some` wrapping and put it at Settings creation, like the others.

meilisearch-http/src/helpers/tide.rs
qdequele: It's because in the route the parameter is called `document_id`

e.g. `/indexes/:index/documents/:document_id`

meilisearch-http/src/helpers/tide.rs
qdequele: No, it's just because previously the result wasn't expected. 

The error should be `Bad parameter {name}; {err}` and not `Bad parameter identifier; {name}`

PR #534, opened 2020-03-20
Automate homebrew publish

curquiza:
- Add homebrew push to github action
- Restrict the push to stable versions

GitHub action used: https://github.com/mislav/bump-homebrew-formula-action

### DIFF
--- a/.github/workflows/publish-deb-brew-pkg.yml
+++ b/.github/workflows/publish-deb-brew-pkg.yml
@@ -1,15 +1,14 @@
-name: Publish deb pkg to GitHub release & apt repository
+name: Publish deb pkg to GitHub release & apt repository & Homebrew
 
 on:
   push:
     tags:
-      - '*'
+      - 'v[0-9]+.[0-9]+.[0-9]+'
 
 jobs:
-  publish:
+  debian:
     name: Publish debian packagge
     runs-on: ubuntu-latest
-
     steps:
     - uses: hecrj/setup-rust-action@master
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-deb-brew-pkg.yml
```diff
@@ -28,3 +27,13 @@ jobs:
         tag: ${{ github.ref }}
     - name: Upload debian pkg to apt repository
       run: curl -F package=@target/debian/meilisearch.deb https://${{ secrets.GEMFURY_PUSH_TOKEN }}@push.fury.io/meilisearch/
+
+  homebrew:
+    name: Bump Homebrew formula
+    runs-on: ubuntu-latest
```
Kerollmops: Why do you specify ubuntu-latest here? Don't we want the binary to be compiled on Mac?
Or will Homebrew compile it for the different osx versions and that is only the platform on which we test the binary?

curquiza: The ubuntu is not used to compile the mac version. It's used to apply this command of brew:
``` bash
$ brew bump-formula-pr --strict meilisearch --url='https://github.com/meilisearch/MeiliSearch/archive/v0.9.0.tar.gz'
The same one I used manually to create a PR on HomeBrew repository. This kind of PR:
https://github.com/Homebrew/homebrew-core/pull/51906/files

The compilation from source code is done on HomeBrew's side.
```

Kerollmops: Nice, thank you for the details!

PR #539, opened 2020-03-24
html sanitize

emresaglam:
Added a function to sanitize the html for meilisearch-http/public/interface.html
This is for browser side only. And also minimal sanitizer that takes care of < and > (&lt; and &gt; respectively) 

Merci beaucoup!

### DIFF
--- a/meilisearch-http/public/interface.html
+++ b/meilisearch-http/public/interface.html
@@ -147,7 +147,7 @@
   </body>
 
   <script>
-    function sanitizeHTMLEntities (str) {
+    function sanitizeHTMLEntities(str) {
         if (str && typeof str === 'string') {
             str = str.replace(/</g,"&lt;");
             str = str.replace(/>/g,"&gt;");
@@ -177,7 +177,7 @@
         lastRequest.open("GET", theUrl, true);
         lastRequest.onload = function (e) {
             if (lastRequest.readyState === 4 && lastRequest.status === 200) {
-                let sanitizedResponseText = lastRequest.responseText;
+                let sanitizedResponseText = sanitizeHTMLEntities(lastRequest.responseText);
                 let httpResults = JSON.parse(sanitizedResponseText);
                 results.innerHTML = '';
 
--- a/meilisearch-http/public/interface.html
+++ b/meilisearch-http/public/interface.html
@@ -148,11 +148,11 @@
 
   <script>
     function sanitizeHTMLEntities (str) {
-        if(str && typeof str === 'string') {
+        if (str && typeof str === 'string') {
             str = str.replace(/</g,"&lt;");
             str = str.replace(/>/g,"&gt;");
-          }
-    return str;
+        }
+        return str;
     }
 
     function httpGet(theUrl) {
--- a/meilisearch-http/public/interface.html
+++ b/meilisearch-http/public/interface.html
@@ -151,6 +151,8 @@
         if (str && typeof str === 'string') {
             str = str.replace(/</g,"&lt;");
             str = str.replace(/>/g,"&gt;");
+            str = str.replace(/&lt;em&gt;/g,"<em>");
+            str = str.replace(/&lt;\/em&gt;/g,"<\/em>");
         }
         return str;
     }
--- a/meilisearch-http/public/interface.html
+++ b/meilisearch-http/public/interface.html
@@ -147,6 +147,14 @@
   </body>
 
   <script>
+    function sanitizeHTMLEntities (str) {
+        if(str && typeof str === 'string') {
+            str = str.replace(/</g,"&lt;");
+            str = str.replace(/>/g,"&gt;");
+          }
+    return str;
+    }
+
     function httpGet(theUrl) {
         var xmlHttp = new XMLHttpRequest();
         xmlHttp.open("GET", theUrl, false); // false for synchronous request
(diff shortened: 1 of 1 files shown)


meilisearch-http/public/interface.html
```diff
@@ -147,6 +147,14 @@ <h2 class="subtitle">
   </body>
 
   <script>
+    function sanitizeHTMLEntities (str) {
+        if(str && typeof str === 'string') {
+            str = str.replace(/</g,"&lt;");
+            str = str.replace(/>/g,"&gt;");
+          }
+    return str;
```
Kerollmops:
```suggestion
        if (str && typeof str === 'string') {
            str = str.replace(/</g,"&lt;");
            str = str.replace(/>/g,"&gt;");
        }
        return str;
```

meilisearch-http/public/interface.html
```diff
@@ -169,7 +177,8 @@ <h2 class="subtitle">
         lastRequest.open("GET", theUrl, true);
         lastRequest.onload = function (e) {
             if (lastRequest.readyState === 4 && lastRequest.status === 200) {
-                let httpResults = JSON.parse(lastRequest.responseText);
+                let sanitizedResponseText = lastRequest.responseText;
+                let httpResults = JSON.parse(sanitizedResponseText);
```
Kerollmops: Whooops... Don't you forget to call your sanitize function, here?

emresaglam: Duh! I'm an idiot. Good catch. I think I gave you a wrong version of my changes :( I did it quickly before joining a meeting. Feel free to call the function. Or I can change it after my endless meetings. :)

Kerollmops: Take your time, change it when you have time, no problem :)

emresaglam:
All right. Done! :) This was my first time dealing with a change request. I learned something new! :D

Kerollmops:
Hey,

I tested your changes and everything is good but there is a big problem, the highlighting no more works, the engine adds `<em>` tags around the words parts that must be highlighted by the frontend and this change replaces every less-than (`<`) and greater-than (`>`) characters and that breaks the `<em>` tags the engines adds.

Screenshot_2020-03-25 MeiliSearch

So there is two possible solutions to work around that problem:
  1. The clean one would be to sanitize the text before the engine starts adding the highlighting tags, and the tags would enclose an already sanitized text.
  2. The web frontend detects the highlighting tags and ignore those when sanitizing.

The first solution is the best one but the process will be longer, we will work on that one I think.

meilisearch-http/public/interface.html
```diff
@@ -147,6 +147,14 @@ <h2 class="subtitle">
   </body>
 
   <script>
+    function sanitizeHTMLEntities(str) {
+        if (str && typeof str === 'string') {
+            str = str.replace(/</g,"&lt;");
+            str = str.replace(/>/g,"&gt;");
```
Kerollmops: Ok nevermind, I found an ugly temporary solution to the previous highlighting problem.

```suggestion
            str = str.replace(/</g,"&lt;");
            str = str.replace(/>/g,"&gt;");
            str = str.replace(/&lt;em&gt;/g,"<em>");
            str = str.replace(/&lt;\/em&gt;/g,"<\/em>");
```

emresaglam: Yes ugly :D I totally get your point on this. Covering every <> maybe is an overkill. Also again doing it on the browser is just a part of the solution. I feel like there should be a solution for the backend like we discussed before. 

You might look into Caja (framework by google) for limiting the attack surface on the web interface. They did some good job on this type of issues. Mine was really a monkey patch compared to what should be done. 

I use meilisearch to do a lot of searches on vulnerability data, and it's full of Cross Site Scripting examples and that's how I got into all of this. Sanitizing <> was enough for me, I can totally see that this won't work for everybody. :)

I'll commit the suggestion.

Kerollmops: Thank you for your research on MeiliSearch that is really cool!

As I write in [my previous comment](https://github.com/meilisearch/MeiliSearch/pull/539#issuecomment-603791604) I will work on fixing those security issues in the backend.

This is, as you said, a monkey [temporary] patch :)

Kerollmops:
BTW would it be possible to rebase your branch on master? :)

```
git checkout html-sanitize
git fetch
git rebase origin/master
```

emresaglam:
Oh mon Dieu! Now I have to learn rebase! :D Stay tuned... Emre Learning new things.

Kerollmops:
I can help you, you first need to add a remote of our repository, this way git is aware of the state of the banches that are on our repo, then you will be able to execute the command lines I wrote in [my previous comment](https://github.com/meilisearch/MeiliSearch/pull/539#issuecomment-604052491).

```
git remote add forked-remote https://github.com/meilisearch/MeiliSearch.git
```

Kerollmops:
Whoops, that isn't a rebase, but don't worry we will keep it like that :)

Kerollmops:
Thank you for your work, that is great :)

nodesocket:
Doesn't it make sense to use a built-in package to do the html string encoding?

https://docs.rs/htmlescape/0.3.1/htmlescape/fn.encode_minimal.html

Kerollmops:
That is a great idea, we will look at this, thank you!

PR #541, opened 2020-03-25
Adds number of hits in search result

### LINKED ISSUE #469: Search result count?
It occurs to me that the search API lacks a result count, and simply has hits / offset / limit / processingTimeMs and query. This makes it hard to build query pagination as there's no way to determine whether there are more results and how many pages of results there are. It would be great if this could be included

MarinPostma:
Adds support for the number of hits to be returned with a search.
Two fields are added to the search result: 
- `exhaustiveNbHits`
- `nbHits`

for now, only an approximate result is returned, hence the `exhaustiveNbHits` is always set to false.

fixes #469

### DIFF
--- a/meilisearch-http/src/helpers/meilisearch.rs
+++ b/meilisearch-http/src/helpers/meilisearch.rs
@@ -360,7 +360,7 @@ pub struct SearchHit {
     pub matches_info: Option<MatchesInfos>,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct SearchResult {
     pub hits: Vec<SearchHit>,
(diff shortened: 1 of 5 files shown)


meilisearch-http/src/helpers/meilisearch.rs
```diff
@@ -223,12 +223,12 @@ impl<'a> SearchBuilder<'a> {
         }
 
         let start = Instant::now();
-        let docs =
-            query_builder.query(reader, &self.query, self.offset..(self.offset + self.limit));
+        let result = query_builder.query(reader, &self.query, self.offset..(self.offset + self.limit));
         let time_ms = start.elapsed().as_millis() as usize;
+        let (docs, nb_hits) = result.map_err(|e| Error::SearchDocuments(e.to_string()))?;
```
Kerollmops: I would prefer to keep those times lines together.

```suggestion
        let result = query_builder.query(reader, &self.query, self.offset..(self.offset + self.limit));
        let (docs, nb_hits) = result.map_err(|e| Error::SearchDocuments(e.to_string()))?;
        let time_ms = start.elapsed().as_millis() as usize;
```

PR #548, opened 2020-03-27
Stringify nested JSON object

tendant:
Show nested JSON object as string instead of showing "[object Object]". 

Also to avoid XSS, the content should be escaped. Below function can be used for escaping. 

https://github.com/lodash/lodash/blob/master/escape.js


Kerollmops:
Hey @tendant,

About the escaping we [already did it when we receive the response content](https://github.com/meilisearch/MeiliSearch/blob/022aeac808b9bc37423b6899281f1d977a4763fc/meilisearch-http/public/interface.html#L150-L158), so it must do the job, don't you think?

Thank you for your PR :)

tendant:
Yes. That's a good way to escape the content. So we don't need to worry about it everywhere. In such case, this code change doesn't need to care about escape. It should be good to merge.

Looks like "<" and ">" have been escaped. 

From loadash/escape code, it also escapes """ and "'". (double and single quote)

PR #567, opened 2020-03-30
Construct a Set using the from_dirty method

### LINKED ISSUE #566: Query crashes process
If I do a search like `/indexes/{index}/search?q=lala` meilisearch crashes showing this error:
```
thread 'async-std/executor' panicked at 'called `Result::unwrap()` on an `Err` value: NotDedup', meilisearch-core/src/query_tree.rs:534:46
```
It is specific to the query 'lala' everything else I tried works as expected.  
I'm using the latest version from the master branch built with the included Dockerfile.

Sadly I know next to nothing about Rust...
What can I do to narrow this down and provide a better bug report?

The json dataset I'm using is about 350MB in size.
I can share it (its will become public information soon anyway...) if that helps.

Kerollmops:
This commit fixes #566 by ensuring that the slice of matches is ordered and deduplicated.

I think this bug was affecting all double words (i.e. "lala", "kiki", "chouchou") who matches something in the documents when half split (i.e. "la", "ki", "chou").

@piccaso I would be glad if you test this PR before I merge it in master :)

### DIFF
--- a/meilisearch-core/src/query_tree.rs
+++ b/meilisearch-core/src/query_tree.rs
@@ -531,7 +531,7 @@ pub fn traverse_query_tree<'o, 'txn>(
                     let docids = SetBuf::new(docids).unwrap();
                     debug!("{:2$}docids construction took {:.02?}", "", before.elapsed(), depth * 2);
 
-                    let matches = Cow::Owned(SetBuf::new(matches).unwrap());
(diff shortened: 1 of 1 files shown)


piccaso:
Thank you for the quick response!  
I just tested it and `?q=lala` works as expected now.

Issue #571, opened 2020-03-31
Unable to serialize nested null values


ppamorim:
Hi,

The JSON below causes Meili to not serialize the data:

```json
[{"id":0,"foo":{"bar":null}}]
```

I am currently working on a fix + test for this.

PR #572, opened 2020-03-31
Add support of nested null

ppamorim:
Hi,

Issue: https://github.com/meilisearch/MeiliSearch/issues/571

The test below seems to hang, so please do not merge it yet. This problem is caused with the JSON below:

```json
[{"id":0,"foo":{"bar":null}}]
```

What should Meili do? Simply ignore the value or stringify it?

@Kerollmops Wil the function `fn serialize_unit(self) -> Result<Self::Ok, Self::Error>` from `convert_to_string.rs` or `convert_to_number.rs` be called in this case?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,4 +3,4 @@
   - Add the number of hits in search result (#541)
   - Add support for aligned crop in search result (#543)
   - Sanitize the content displayed in the web interface (#539)
-  - Add support of nested null and boolean values
+  - Add support of nested null and boolean values (#571 and #568)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,4 +3,4 @@
   - Add the number of hits in search result (#541)
   - Add support for aligned crop in search result (#543)
   - Sanitize the content displayed in the web interface (#539)
-  - Add support of nested null values
+  - Add support of nested null and boolean values
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,3 +3,4 @@
   - Add the number of hits in search result (#541)
   - Add support for aligned crop in search result (#543)
   - Sanitize the content displayed in the web interface (#539)
+  - Add support of nested null values
--- a/meilisearch-http/tests/documents_add.rs
+++ b/meilisearch-http/tests/documents_add.rs
@@ -79,7 +79,7 @@ fn check_add_documents_with_nested_boolean() {
 
 // Test issue https://github.com/meilisearch/MeiliSearch/issues/571
 #[test]
-fn check_add_documents_with_nested_sequence() {
+fn check_add_documents_with_nested_null() {
     let mut server = common::Server::with_uid("tasks");
 
     // 1 - Create the index with no primary_key
@@ -91,7 +91,12 @@ fn check_add_documents_with_nested_sequence() {
 
     // 2 - Add a document that contains a null in a nested object
 
-    let body = serde_json::from_str("[{\"id\":0,\"foo\":{\"bar\":null}}]").unwrap();
+    let body = json!([{ 
+        "id": 0, 
+        "foo": { 
+            "bar": null
+        } 
+    }]);
 
     let url = "/indexes/tasks/documents";
     let (response, status_code) = server.post_request(&url, body);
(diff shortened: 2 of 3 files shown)


meilisearch-http/tests/documents_add.rs
```diff
@@ -76,3 +76,33 @@ fn check_add_documents_with_nested_boolean() {
...
+    assert_eq!(status_code, 202);
+    let update_id = response["updateId"].as_u64().unwrap();
+    server.wait_update_id(update_id);
+
+    // 3 - Check update sucess
+
+    let (response, status_code) = server.get_update_status(update_id);
+    assert_eq!(status_code, 200);
+    assert_eq!(response["status"], "processed");
+}
```
Kerollmops:
```suggestion
}

```

meilisearch-http/tests/documents_add.rs
```diff
@@ -76,3 +76,33 @@ fn check_add_documents_with_nested_boolean() {
...
+    // 1 - Create the index with no primary_key
+
+    let body = json!({ "uid": "tasks" });
+    let (response, status_code) = server.create_index(body);
+    assert_eq!(status_code, 201);
+    assert_eq!(response["primaryKey"], json!(null));
+
+    // 2 - Add a document that contains a null in a nested object
+
+    let body = serde_json::from_str("[{\"id\":0,\"foo\":{\"bar\":null}}]").unwrap();
```
Kerollmops:
```suggestion
    let body = json!([ { "id": 0, "foo":{ "bar": null } } ]);
```

meilisearch-http/tests/documents_add.rs
```diff
@@ -76,3 +76,33 @@ fn check_add_documents_with_nested_boolean() {
     assert_eq!(status_code, 200);
     assert_eq!(response["status"], "processed");
 }
+
+// Test issue https://github.com/meilisearch/MeiliSearch/issues/571
+#[test]
+fn check_add_documents_with_nested_sequence() {
```
Kerollmops:
```suggestion
fn check_add_documents_with_nested_null() {
```

ppamorim:
> Thank you for this test :)
> 
> It is preferable to run your test by compiling MeiliSearch in release, this is probably why you think the test is hanging.
> 
> ```
> cargo test --release -- check_add_documents_with_nested_null
> ```
> 
> I think the error message you get is related to the unit `()` type not the `none` one, therefore I think that if you return an empty string from [the `serialize_unit` method](https://github.com/meilisearch/MeiliSearch/pull/569/files#diff-535c295b9f990f6d78f998f54e856d07R90-R92) the test will pass.

Would be nicer to return Ok(none) in case the content is null?

Kerollmops:
The thing is the `ConvertToString` serializer doesn't return an optional String but a String, returning an empty String would have the same effect as changing the whole `Serializer` signature to return an option string.

ppamorim:
@Kerollmops Would it imply in any side effect where `serialize_unit` should trigger `Err(SerializerError::UnserializableType { type_name: "()" })`? Would be nice list a JSON where these cases can be triggered or not for each error case.

Kerollmops:
I don't think we will handle those errors to the user, this is not an error that most users want to handle anyway. I would like to avoid adding parameters and flags for this kind of niche cases.

MeiliSearch is community driven anyway, so if most of the users want to handle this kind of error we will probably add a flag to handle it. However I think handleling `null`s by ignore the content and indexing literally nothing is a good workaround.

ppamorim:
@Kerollmops Maybe in the future we can create a list of fields that were ignored in the JSON.

Kerollmops:
If you want to ignore top-level fields you can already [define the searchable fields](https://docs.meilisearch.com/references/searchable_attributes.html) a.k.a. the fields that the engine will consider and index.

ppamorim:
I think I should replace `"".to_string()` to `String::from("")`.

ppamorim:
*Update CHANGELOG, not readme :)

CHANGELOG.md
```diff
@@ -3,3 +3,4 @@
   - Add the number of hits in search result (#541)
   - Add support for aligned crop in search result (#543)
   - Sanitize the content displayed in the web interface (#539)
+  - Add support of nested null values
```
Kerollmops: Thank you, would it be possible to also specify that we support boolean values too? I forgot to change it on the other PR :)

```suggestion
  - Add support of nested null and boolean values
```

Kerollmops:
That is a really great job, thank you!

PR #573, opened 2020-03-31
Mark the Changelog CI job as non-fatal

Kerollmops:


Kerollmops:
Unfortunately github shows the job like it passes ✅ instead of ❌.

PR #576, opened 2020-03-31
Fix benchmark

ppamorim:
Salut,

I did some changes in the file `search_benchmark.rs` to try to make it work again since it's broken when I try to run it. The current version is trying to access a TOML file but it was replaced with a JSON file. Some functions were changed and I had to change them too.

Merci

### DIFF
--- a/meilisearch-core/benches/search_benchmark.rs
+++ b/meilisearch-core/benches/search_benchmark.rs
@@ -4,12 +4,13 @@ extern crate assert_matches;
 
 use std::sync::mpsc;
 use std::path::Path;
-use std::{fs, fs::File, io::BufReader};
+use std::fs::File;
+use std::io::BufReader;
 use std::iter;
 
 use meilisearch_core::Database;
 use meilisearch_core::{ProcessedUpdateResult, UpdateStatus};
-use meilisearch_core::settings::{Settings, SettingsUpdate, UpdateState};
+use meilisearch_core::settings::{Settings, SettingsUpdate};
 use meilisearch_schema::Schema;
 use serde_json::Value;
 
@@ -41,14 +42,13 @@ fn prepare_database(path: &Path) -> Database {
 
     let mut update_writer = db.update_write_txn().unwrap();
     let _update_id = index.settings_update(&mut update_writer, settings_update).unwrap();
-
     update_writer.commit().unwrap();
 
     let mut additions = index.documents_addition();
 
     let json: Value = {
         let path = concat!(env!("CARGO_MANIFEST_DIR"), "/../datasets/movies/movies.json");
-        let movies_file = fs::File::open(path).expect("find movies");
+        let movies_file = File::open(path).expect("find movies");
         serde_json::from_reader(movies_file).unwrap()
     };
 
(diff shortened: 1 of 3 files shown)


ppamorim:
Now it's causing this error, I have to investigate what is wrong with the schema setup:

```
thread 'main' panicked at 'assertion failed: `Some(Failed { content: ProcessedUpdateResult { 
update_id: 1, update_type: DocumentsAddition { number: 19653 }, error: Some("schema cannot be
 built without a primary key"), duration: 0.021727323, enqueued_at: 2020-04-
01T00:23:09.226109Z, processed_at: 2020-04-01T00:23:09.345778Z } })` does not match
 `Some(UpdateStatus::Processed { content }) if content.error.is_none()`', meilisearch-
core/benches/search_benchmark.rs:93:5
````

**Comment by ppamorim** [comment]
@Kerollmops Maybe include the benchmark as step in the tests?

**Comment by Kerollmops** [comment]
I have already thought about doing that, the problem is that it is a really bad thing to do on the provided servers of Github. We will not get valid answers about regressions and more importantly, doing so would consume many ours of free server CPU time.

I would prefer to do that on our own servers and only when requested.

**Comment by ppamorim** [comment]
@Kerollmops Now the branch is clean and safe. Could you please check it and merge if possible?

File: meilisearch-core/benches/search_benchmark.rs [hunk 1]
```diff
@@ -4,11 +4,13 @@ extern crate assert_matches;
 
 use std::sync::mpsc;
 use std::path::Path;
-use std::fs;
+use std::{fs, fs::File, io::BufReader};
```
qdequele, line 7: Split this into 2 lines. 
Use only `File` instead of `fs::File` at the line 51.
```suggestion
use std::fs::File;
use std::io::BufReader;
```

meilisearch-core/benches/search_benchmark.rs
```diff
@@ -24,15 +26,22 @@ fn prepare_database(path: &Path) -> Database {
...
+        let file = File::open(path).unwrap();
+        let reader = BufReader::new(file);
+        let settings: Settings = serde_json::from_reader(reader).unwrap();
+        settings.into_update().unwrap()
     };
 
     let mut update_writer = db.update_write_txn().unwrap();
-    let _update_id = index.schema_update(&mut update_writer, schema).unwrap();
+    let _update_id = index.settings_update(&mut update_writer, settings_update).unwrap();
+
```
qdequele, line 44: This space is not necessary.
```suggestion
```

meilisearch-core/benches/search_benchmark.rs
```diff
@@ -4,11 +4,13 @@ extern crate assert_matches;
 
 use std::sync::mpsc;
 use std::path::Path;
-use std::fs;
+use std::{fs, fs::File, io::BufReader};
 use std::iter;
 
 use meilisearch_core::Database;
 use meilisearch_core::{ProcessedUpdateResult, UpdateStatus};
+use meilisearch_core::settings::{Settings, SettingsUpdate, UpdateState};
```
Kerollmops, line 12:
```suggestion
use meilisearch_core::settings::{Settings, SettingsUpdate};
```

PR #578, opened 2020-04-01
Add sentry probe

qdequele:
The probe will be available only on production if analytics has not been deactivated.


meilisearch-http/src/main.rs
```diff
@@ -28,8 +28,21 @@ pub fn main() -> Result<(), MainError> {
...
+            if !opt.no_analytics {
+                let _sentry = sentry::init((
+                    "https://5ddfa22b95f241198be2271aaf028653@sentry.io/3060337",
+                    sentry::ClientOptions {
+                        release: sentry::release_name!(),
+                        ..Default::default()
+                    },
+                ));
+                sentry::integrations::panic::register_panic_handler();
+                sentry::integrations::env_logger::init(None, Default::default());
```
Kerollmops, line 41: Are you sure that it works correctly? I remember that the scope in which you register the panic handler counts.

meilisearch-http/Cargo.toml
```diff
@@ -32,17 +32,18 @@ mime = "0.3.16"
 pretty-bytes = "0.2.2"
 rand = "0.7.2"
 rayon = "1.2.0"
+sentry = { version = "0.18.0", features = ["with_rustls", "with_env_logger"] }
```
Kerollmops, line 35: The sentry dependency is really heavy in term of number of added dependencies, isn't there a way to reduce that? (765 additions, that's a lot)

README.md
```diff
@@ -204,6 +204,7 @@ Hey! We're glad you're thinking about contributing to MeiliSearch! If you think
 
 Once a day, events are being sent to our Amplitude instance so we can know how many people are using MeiliSearch.<br/>
 Only information about the platform on which the server runs is stored. No other information is being sent.<br/>
+We also use Sentry to report us crashs or errors. To get more information about what iformation Sentry collect, please visit the [Sentry privacy policy](https://sentry.io/privacy/).
```
Kerollmops, line 207:
```suggestion
We also use Sentry to make us crash and error reports. If you want to know more about what Sentry collects, please visit their [privacy policy website](https://sentry.io/privacy/).
```

meilisearch-http/Cargo.toml
qdequele: I haven't figured out which features to remove.

PR #579, opened 2020-04-02
Update dependencies

Kerollmops:
Updates most of the dependencies we use in all MeiliSearch workspaces.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,18 +2,18 @@
 # It is not intended for manual editing.
 [[package]]
 name = "ahash"
-version = "0.2.17"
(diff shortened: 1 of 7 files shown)


Kerollmops:
Note that I cannot update `async-std` to v1.5.0 and `http` to v0.2.1 because `http-service` uses those two deps at outdated versions. This crate is maintained by Aaron Turon, so not small chances to be updated soon.

qdequele:
> @qdequele Quick question: Is `http-service` necessary? Do we use that to mock requests?

I need it on test (`/meilisearch-http/tests/common.rs`)

PR #588, opened 2020-04-05
Fix typo in README

techieshark:
Non-plural would be more usual in English. I assume "performances" was a typo.


Kerollmops:
Thank you, that makes sense!

PR #592, opened 2020-04-06
Implements query filters

MarinPostma:
This PR implements the query filters. This is still a work in progress.

It enables a simple query language, consisting of 3 keywords: `NOT`, `OR` and `AND`.
It allows filtering on document fields in the following manner:
```
     v--- the comparison operator, either `=`, `>`, `<`, ...
field:value [OP field:value]*
 ^--- the field name in the document
```
`OP` is the logic operator ( `NOT`, `OR` and `AND`)
`value`, together with the comparison operator define the condition that must hold for the candidate.

The type is inferred at filtering time. When the filter is compiled at the beginning of a query, it is given all the types it can be parsed into (either `bool`, `string` or `number`). This is because no schema is enforced at indexing.

Filtering on string fields support **ONLY**  the `=` operator. It will look for an exact **case insensitive** match.
- exemple: `american ninja` will match `American Ninja`

_implementation detail_ : the unicase library is used for cheap case insensitive comparisons.

Filtering on bool fields support **ONLY**  the `=` operator, that will check for equality of booleans.

Filtering on numbers fields support  the `=`, `<` and `>` operators.

**Note on performance**:
For now, there are no restriction to what a user can filter on. the way meilisearch works makes it rather quick if the user have required a fairly low amount of documents and many matches the filter. However if very few documents match, it becomes a lot slower. In the future, we plan to implement bloom filters to improve the speed of subsequent searches, and introduces facets, for high performance candidate reduction upstream of the search.

**TODO**:
- [x] dynamic number conversion for comparisons
- [ ] Bloom filters


meilisearch-core/src/error.rs
```diff
@@ -33,6 +37,17 @@ impl From<io::Error> for Error {
     }
 }
 
+impl From<PestError<Rule>> for Error {
+    fn from(error: PestError<Rule>) -> Error {
+        Error::FilterParseError(error.renamed_rules(|r| match *r {
+                    Rule::or => "OR".to_string(),
+                    Rule::and => "AND".to_string(),
+                    Rule::not => "NOT".to_string(),
+                    _ => "other".into()
```
Kerollmops, line 46:
```suggestion
                    other => other.to_string()
```

meilisearch-core/src/error.rs
```diff
@@ -33,6 +37,17 @@ impl From<io::Error> for Error {
     }
 }
 
+impl From<PestError<Rule>> for Error {
+    fn from(error: PestError<Rule>) -> Error {
+        Error::FilterParseError(error.renamed_rules(|r| match *r {
```
Kerollmops, line 42:
```suggestion
        Error::FilterParseError(error.renamed_rules(|r| {
            match r {
```

meilisearch-core/src/filters/condition.rs
```diff
@@ -0,0 +1,228 @@
...
+use std::str::FromStr;
+use std::cmp::Ordering;
+
+type ConditionType = Ordering;
+
+/// As Meilisearch does not enforce a schema at indexing, we need to infer type at runtime.
+/// We do it once when the filter is constructed and match every possible types it can be
+/// parsed into.
+///
+/// For now we only support filtering on numbers, booleans, and strings.
```
Kerollmops, line 20:
```suggestion
/// We need to infer type when the filter is constructed and match every possible types it can be
/// parsed into.
```

meilisearch-core/src/filters/condition.rs
```diff
@@ -0,0 +1,228 @@
...
+                        "true" => Some(true),
+                        "false" => Some(false),
+                        _ => None,
+                    };
+                    let number = Number::from_str(value.as_str()).ok();
+                    ConditionValue {
+                        string, boolean, number
+                    }
+                },
+            _ => unreachable!(),
```
Kerollmops, line 44:
```suggestion
            Rule::string | Rule::word => {
                let string =  value.as_str();
                let boolean = match value.as_str() {
                    "true" => Some(true),
                    "false" => Some(false),
                    _ => None,
                };
                let number = Number::from_str(value.as_str()).ok();
                ConditionValue { string, boolean, number }
            },
            _ => unreachable!(),
```

meilisearch-core/src/filters/condition.rs
```diff
@@ -0,0 +1,228 @@
...
+        };
+        value
+    }
+
+    pub fn as_str(&self) -> &str {
+        self.string.as_ref()
+    }
+
+    pub fn as_num(&self) -> Option<&Number> {
```
Kerollmops, line 53:
```suggestion
    pub fn as_number(&self) -> Option<&Number> {
```

meilisearch-core/src/filters/condition.rs
```diff
@@ -0,0 +1,228 @@
...
+                                 key.as_str(),
+                                 schema.names().collect::<Vec<_>>().join(", ")
+                             ),
+                },
+                key.as_span()))?;
+    let value = ConditionValue::new(&items.next().unwrap());
+    Ok((field, value))
+}
+
+// /!\ undefined behavior with big numbers
```
Kerollmops, line 88: I wouldn't say this is undefined, as it is completely defined, it is a lack of precision that is deterministic.

meilisearch-core/src/filters/condition.rs
```diff
@@ -0,0 +1,228 @@
...
+        let condition = ConditionType::Equal;
+        Ok(Self { field, condition, value })
+    }
+
+    pub fn test<'r>(
+        &self,
+        reader: &'r RoTxn<MainT>,
+        index: &Index,
+        document_id: DocumentId,
+    ) -> Result<bool, Error> {
```
Kerollmops, line 133:
```suggestion
    pub fn test(
        &self,
        reader: &RoTxn<MainT>,
        index: &Index,
        document_id: DocumentId,
    ) -> Result<bool, Error> {
```

meilisearch-core/src/filters/condition.rs
```diff
@@ -0,0 +1,228 @@
...
+    //    i64   u64  f64  i64  u64  f64
+        (Some(lhs), _, _, Some(rhs), _, _) => lhs.partial_cmp(&rhs),
+        (_, Some(lhs), _, _, Some(rhs), _) => lhs.partial_cmp(&rhs),
+        (_, _, Some(lhs), _, _, Some(rhs)) => lhs.partial_cmp(&rhs),
+        (_, _, _, _, _, _) => None,
+    }
+}
+
+impl<'a> Condition<'a> {
+    pub fn lesser(
```
Kerollmops, line 101:
```suggestion
    pub fn less(
```

meilisearch-core/src/filters/condition.rs
```diff
@@ -0,0 +1,228 @@
...
+            },
+            Some(Value::Bool(b)) => if let Some(value) = self.value.as_bool() {
+                let res = match self.condition {
+                    ConditionType::Equal => b == value,
+                    _ => false
+                };
+                Ok(res)
+            } else {
+                Ok(false)
+            },
```
Kerollmops, line 158:
```suggestion
            Some(Value::Bool(b)) => {
                if let Some(value) = self.value.as_bool() {
                    if self.condition == ConditionType::Equal {
                        return Ok(b == value)
                    }
                }
                Ok(false)
            },
```

meilisearch-core/src/filters/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+    Condition(Condition<'a>),
+    Or(Box<Self>, Box<Self>),
+    And(Box<Self>, Box<Self>),
+    Not(Box<Self>),
+}
+
+impl<'a> Filter<'a> {
+    pub fn parse(expr: &'a str, schema: &'a Schema) -> FilterResult<'a> {
+        let lexed = FilterParser::parse(Rule::prgm, expr.as_ref());
+        //eprintln!("tokens: {:?}", lexed);
```
Kerollmops, line 31:
```suggestion
```

meilisearch-http/src/helpers/meilisearch.rs
```diff
@@ -62,7 +63,19 @@ impl fmt::Display for Error {
...
     fn from(error: meilisearch_core::Error) -> Self {
-        Error::Internal(error.to_string())
+        use meilisearch_core::pest_error::LineColLocation::*;
+        match error {
+            meilisearch_core::Error::FilterParseError(e) => {
+                let(line, offset) = match e.line_col {
+                    Span((line, _), (offset, _)) => (line, offset),
+                    Pos((line, offset)) => (line, offset),
+                };
+                let message = format!("parsing error on line {} at offset {}: {}", line, offset, e.variant.message());
```
Kerollmops, line 73:
```suggestion
                let (line, column) = match e.line_col {
                    Span((line, _), (column, _)) => (line, column),
                    Pos((line, column)) => (line, column),
                };
                let message = format!("parsing error on line {} at column {}: {}", line, column, e.variant.message());
```

meilisearch-http/src/helpers/meilisearch.rs
```diff
@@ -171,39 +184,15 @@ impl<'a> SearchBuilder<'a> {
...
-                }
-                (_, _) => (),
-            }
+        if let Some(filter_expression) = &self.filters {
+            let filter = Filter::parse(filter_expression, &schema)?;
+            query_builder.with_filter(move |id| {
+                let index = &self.index;
+                let reader = &reader;
+                let filter = &filter;
+                filter.test(reader, index, id).expect("filter error")
```
Kerollmops, line 193:
```suggestion
                filter.test(reader, index, id).unwrap_or(false)
```

I know that it is not ideal but I would prefer avoid crashing a search if one filter fail.

meilisearch-core/Cargo.toml
```diff
@@ -11,6 +11,7 @@ bincode = "1.1.4"
 byteorder = "1.3.2"
 chrono = { version = "0.4.9", features = ["serde"] }
 compact_arena = "0.4.0"
+cow-utils = "0.1.2"
```
Kerollmops, line 14: Do we always need this dependency?

meilisearch-core/Cargo.toml
```diff
@@ -20,19 +21,23 @@ heed = "0.6.1"
...
+lazy_static = "1.4.0"
 levenshtein_automata = { version = "0.1.1", features = ["fst_automaton"] }
 log = "0.4.8"
 meilisearch-schema = { path = "../meilisearch-schema", version = "0.9.0" }
 meilisearch-tokenizer = { path = "../meilisearch-tokenizer", version = "0.9.0" }
 meilisearch-types = { path = "../meilisearch-types", version = "0.9.0" }
 once_cell = "1.2.0"
 ordered-float = { version = "1.0.2", features = ["serde"] }
+pest = {git = "https://github.com/MarinPostma/pest.git"} 
+pest_derive = "2.0"
```
Kerollmops, line 33: Can't we move this dependency to the `dev-dependencies` section?

meilisearch-core/Cargo.toml
```diff
@@ -20,19 +21,23 @@ heed = "0.6.1"
...
 itertools = "0.8.2"
+lazy_static = "1.4.0"
 levenshtein_automata = { version = "0.1.1", features = ["fst_automaton"] }
 log = "0.4.8"
 meilisearch-schema = { path = "../meilisearch-schema", version = "0.9.0" }
 meilisearch-tokenizer = { path = "../meilisearch-tokenizer", version = "0.9.0" }
 meilisearch-types = { path = "../meilisearch-types", version = "0.9.0" }
 once_cell = "1.2.0"
 ordered-float = { version = "1.0.2", features = ["serde"] }
+pest = {git = "https://github.com/MarinPostma/pest.git"}
```
Kerollmops, line 32: Don't forget to move back to the original crate when the merge is effective.

meilisearch-core/src/filters/condition.rs
```diff
@@ -0,0 +1,228 @@
+use super::parser::Rule;
+use crate::{store::Index, DocumentId, MainT};
+use heed::RoTxn;
+use crate::error::Error;
+use meilisearch_schema::{FieldId, Schema};
+use pest::{
+    error::{Error as PestError, ErrorVariant},
+    iterators::Pair,
+};
```
Kerollmops, line 9:
```suggestion
use pest::error::{Error as PestError, ErrorVariant};
use pest::iterators::Pair;
```

meilisearch-core/src/filters/condition.rs
```diff
@@ -0,0 +1,228 @@
...
+use serde_json::{ Value, Number };
```
Kerollmops, line 10:
```suggestion
use serde_json::{Value, Number};
```

meilisearch-http/src/helpers/meilisearch.rs
MarinPostma: Oopsie, I thought I had changed it to warn, fixing it.

meilisearch-core/Cargo.toml
MarinPostma: No we indeed don't, since we now use unicase

meilisearch-core/Cargo.toml
MarinPostma: It appears so, rust complains if I put it as a devdep

meilisearch-core/src/filters/condition.rs
MarinPostma: quoting the book:
>Casting from a float to an integer will round the float towards zero
>
>  NOTE: currently this will cause Undefined Behavior if the rounded value cannot be  represented by the target integer type. This includes Inf and NaN. This is a bug and will be fixed.

and 

> Casting from an integer to float will produce the floating point representation of the integer, rounded if necessary (rounding strategy unspecified)

What do you think?

meilisearch-core/src/error.rs
```diff
@@ -33,6 +37,25 @@ impl From<io::Error> for Error {
...
+                Rule::and => "AND".to_string(),
+                Rule::not => "NOT".to_string(),
+                Rule::string => "string".to_string(),
+                Rule::word => "word".to_string(),
+                Rule::greater => "field>value".to_string(),
+                Rule::lesser => "field<value".to_string(),
+                Rule::eq => "field:value".to_string(),
+                Rule::key => "key".to_string(),
+                _ => "other".to_string(),
+            }
```
Kerollmops, line 54:
```suggestion
            let s = match r {
                Rule::or => "OR",
                Rule::and => "AND",
                Rule::not => "NOT",
                Rule::string => "string",
                Rule::word => "word",
                Rule::greater => "field>value",
                Rule::lesser => "field<value",
                Rule::eq => "field:value",
                Rule::key => "key",
                _ => "other",
            };
            s.to_string()
```

meilisearch-core/src/filters/condition.rs
```diff
@@ -0,0 +1,219 @@
...
+use pest::error::{Error as PestError, ErrorVariant};
+use pest::iterators::Pair;
+use serde_json::{Value, Number};
+use std::str::FromStr;
+use std::cmp::Ordering;
+
+type ConditionType = Ordering;
+
+/// We need to infer type when the filter is constructed and match every possible types it can be
+/// parsed into.
```
Kerollmops, line 15:
```suggestion
/// We need to infer type when the filter is constructed
/// and match every possible types it can be parsed into.
```

meilisearch-core/src/filters/condition.rs
```diff
@@ -0,0 +1,219 @@
+use super::parser::Rule;
+use crate::{store::Index, DocumentId, MainT};
+use heed::RoTxn;
+use crate::error::Error;
+use meilisearch_schema::{FieldId, Schema};
+use pest::error::{Error as PestError, ErrorVariant};
+use pest::iterators::Pair;
+use serde_json::{Value, Number};
+use std::str::FromStr;
+use std::cmp::Ordering;
```
Kerollmops, line 10: Could you separate the std deps form the other ones, and sorted them, please?

meilisearch-core/src/filters/condition.rs
Kerollmops: Yeah, keep it but please, remove this `/!\` :)

meilisearch-core/src/filters/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+use condition::Condition;
+use heed::RoTxn;
+use crate::error::Error;
+use crate::{DocumentId, MainT, store::Index};
+use meilisearch_schema::Schema;
+use parser::{PREC_CLIMBER, FilterParser};
+use pest::{
+    iterators::{Pair, Pairs},
+    Parser,
+};
```
Kerollmops, line 15:
```suggestion
use pest::iterators::{Pair, Pairs},
use pest::Parser;
```

meilisearch-core/src/filters/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+    #[test]
+    fn invalid_syntax() {
+        assert!(FilterParser::parse(Rule::prgm, "field = id").is_err());
+        assert!(FilterParser::parse(Rule::prgm, "field:hello hello").is_err());
+        assert!(FilterParser::parse(Rule::prgm, "field:hello OR OR").is_err());
+        assert!(FilterParser::parse(Rule::prgm, "OR field:hello").is_err());
+        assert!(FilterParser::parse(Rule::prgm, r#"field:"hello world"#).is_err());
+        assert!(FilterParser::parse(Rule::prgm, r#"field:'hello world"#).is_err());
+        assert!(FilterParser::parse(Rule::prgm, r#"field >= 10"#).is_err());
+        assert!(FilterParser::parse(Rule::prgm, r#"field <= 10"#).is_err());
```
Kerollmops, line 89: Couldn't we support those two operations?

meilisearch-http/src/helpers/meilisearch.rs
```diff
@@ -171,39 +184,18 @@ impl<'a> SearchBuilder<'a> {
...
+                let index = &self.index;
+                let reader = &reader;
+                let filter = &filter;
+                fn warn_filter_error(e: meilisearch_core::Error) -> Result<bool, ()> {
+                    log::warn!("unexpected error during filtering: {}", e);
+                    Ok(false)
                 }
-                (_, _) => (),
-            }
+                filter.test(reader, index, id).or_else(warn_filter_error).unwrap()
```
Kerollmops, line 197: I would prefer you do a simple match here :)

meilisearch-core/src/filters/condition.rs
```diff
@@ -108,6 +116,23 @@ impl<'a> Condition<'a> {
...
+    }
+
+    pub fn leq(
+        item: Pair<'a, Rule>,
+        schema: &'a Schema,
+    ) -> Result<Self, Error> {
+        let (field, value) = get_field_value(schema, item)?;
+        let condition = ConditionType::LessEqual;
+        Ok(Self { field, condition, value })
+    }
```
Kerollmops, line 135:
```suggestion
    }

```

meilisearch-core/src/filters/condition.rs
```diff
@@ -134,14 +159,21 @@ impl<'a> Condition<'a> {
             Some(Value::Number(n)) => { 
                 if let Some(value) = self.value.as_number() {
                     if let Some(cmp_result) = compare_numbers(&n, value) {
-                        return Ok(cmp_result == self.condition)
+                        let res =  match self.condition {
+                            ConditionType::Equal => cmp_result == Ordering::Equal,
+                            ConditionType::GreaterEqual => cmp_result == Ordering::Equal || cmp_result == Ordering::Greater,
+                            ConditionType::LessEqual => cmp_result == Ordering::Equal || cmp_result == Ordering::Less,
+                            ConditionType::Greater=> cmp_result == Ordering::Greater,
+                            ConditionType::Less=> cmp_result == Ordering::Less,
```
Kerollmops, line 167:
```suggestion
                            ConditionType::Equal => cmp_result == Ordering::Equal,
                            ConditionType::GreaterEqual => cmp_result == Ordering::Equal || cmp_result == Ordering::Greater,
                            ConditionType::LessEqual => cmp_result == Ordering::Equal || cmp_result == Ordering::Less,
                            ConditionType::Greater => cmp_result == Ordering::Greater,
                            ConditionType::Less => cmp_result == Ordering::Less,
```

Kerollmops: Would it be possible to change `cmp_result` by a shorter name to make the lines shorter.

meilisearch-core/src/filters/mod.rs
```diff
@@ -56,7 +55,9 @@ impl<'a> Filter<'a> {
             |pair: Pair<Rule>| match pair.as_rule() {
                 Rule::eq => Ok(Filter::Condition(Condition::eq(pair, schema)?)),
                 Rule::greater => Ok(Filter::Condition(Condition::greater(pair, schema)?)),
-                Rule::lesser => Ok(Filter::Condition(Condition::lesser(pair, schema)?)),
+                Rule::lesser => Ok(Filter::Condition(Condition::less(pair, schema)?)),
```
Kerollmops, line 58: Why is this Rule names `lesser` and not `less`?

meilisearch-core/src/error.rs
```diff
@@ -33,6 +37,26 @@ impl From<io::Error> for Error {
...
+    fn from(error: PestError<Rule>) -> Error {
+        Error::FilterParseError(error.renamed_rules(|r| {
+            let s  = match r {
+                Rule::or => "OR",
+                Rule::and => "AND",
+                Rule::not => "NOT",
+                Rule::string => "string",
+                Rule::word => "word",
+                Rule::greater => "field>value",
+                Rule::lesser => "field<value",
```
Kerollmops, line 50: Don't you forget to add `geq` and `leq`?

Kerollmops: I would prefer you list all the variants in the last `_` other branch, this way we avoid missing any.

meilisearch-http/src/error.rs
```diff
@@ -74,7 +74,7 @@ impl IntoResponse for ResponseError {
             ResponseError::Internal(err) => {
                 error!("internal server error: {}", err);
                 error(
-                    String::from("Internal server error"),
+                    format!("Internal server error:{:?}", err),
```
Kerollmops, line 77: That was done on purpose, we don't want to expose the internal server error to the user.
The error message can be seen in the logs if needed.

meilisearch-core/src/error.rs
MarinPostma: yes indeed, I forgot them.

Many rules are actually silent, and never generate tokens, for all rules but those two, there is a rewrite rule.

meilisearch-core/src/error.rs
```diff
@@ -33,6 +37,28 @@ impl From<io::Error> for Error {
...
+                Rule::eq => "field:value",
+                Rule::leq => "field<=value",
+                Rule::geq => "field>=value",
```
Kerollmops, line 53: Cannot you please add a space between the operators and the operands?

Issue #584, opened 2020-04-03
Private key has no access to version route


bidoubiwa:
Private key is described in the documentation the following way : 
```
The Private key grants access to all routes except the /keys routes.
```

When trying out every route with the private key I stumbled upon an error on the 
`GET /version` route.

the key given in the example below is the `private` key of my Meilisearch Server.

```bash
$ curl -X GET 'http://127.0.0.1:7700/version' -H 'X-MEILI-API-KEY: 8c222193c4dff5a19689d637416820bc623375f2ad4c31a2e3a76e8f4c70440d ' | jq

{
  "message": "Invalid API key: 8c222193c4dff5a19689d637416820bc623375f2ad4c31a2e3a76e8f4c70440d"
}
```

Is this the desired behavior? In which case the documentation should be changed to be in line with this behavior.

**EDIT**
This is also the case for the following routes: 
`GET /stats`
`GET /sys-info`
`GET /sys-info/pretty`
Private key has access to none of those routes.

PR #601, opened 2020-04-09
Change tide to actix-web

qdequele:
**Todo:**
- [x] Finish changing all handlers
    - [x] Document
    - [x] Index
    - [x] Health
    - [x] Key
    - [x] Search
    - [x] Setting
    - [x] Stats
    - [x] Stop Word
    - [x] Synonym
- [x] Change Params tuples to Struct (test)
- [x] Add middleware CORS+Compression
- [x] Add authorization middleware
- [x] Make all test work
- [x] Make the error handling simple

**Out of scope work done:**
- [x] Remove search multi index

**Before review:**
- [x] Check that handlers return correct status code
- [x] Check all `unwrap()`
- [x] Check usages of `.clone()`
- [x] Revert `index.document()` to the older code
- [x] Check all test
- [x] Check that no code/feature has been added. Only the changes needed for the PR


meilisearch-http/src/helpers/meilisearch.rs
```diff
@@ -225,26 +166,24 @@ impl<'a> SearchBuilder<'a> {
 
         let start = Instant::now();
         let result = query_builder.query(reader, &self.query, self.offset..(self.offset + self.limit));
-        let (docs, nb_hits) = result.map_err(|e| Error::SearchDocuments(e.to_string()))?;
+        let (docs, nb_hits) = result.map_err(ResponseError::search_documents)?;
```
MarinPostma, line 169: ok, this is why you created setter... Yet, isn't it a bit overkill?

meilisearch-http/src/helpers/meilisearch.rs
```diff
@@ -225,26 +166,24 @@ impl<'a> SearchBuilder<'a> {
...
 
             let document: IndexMap<String, Value> = self
                 .index
-                .document(reader, fields.as_ref(), doc.id)
-                .map_err(|e| Error::RetrieveDocument(doc.id.0, e.to_string()))?
-                .ok_or(Error::DocumentNotFound(doc.id.0))?;
+                .document(reader, attributes.as_ref(), doc.id)
+                .map_err(|e| ResponseError::retrieve_document(doc.id.0, e))?
+                .ok_or(ResponseError::internal(
+                    "Impossible to retrieve a document id returned by the engine",
```
MarinPostma, line 185: I'm not sure to understand the error message here. This error occurs when there was no document for that id?

qdequele: This error is an internal error. It will be triggered if we have data corruption. The search method told me there is a document in the database but it's impossible to retrieve it.

meilisearch-http/src/error.rs
```diff
@@ -1,183 +1,173 @@
...
-    fn from(err: serde_json::Error) -> ResponseError {
-        ResponseError::internal(err)
+    pub fn retrieve_document(doc_id: u64, err: impl fmt::Display) -> ResponseError {
+        ResponseError::RetrieveDocument(doc_id, err.to_string())
+    }
+
+    pub fn search_documents(err: impl fmt::Display) -> ResponseError {
+        ResponseError::SearchDocuments(err.to_string())
     }
 }
```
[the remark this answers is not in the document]
qdequele: These constructors are used to on the rest of the database to not handle `.to_string()` 100 times and not do copy everywhere you need to raise an error.

meilisearch-http/src/main.rs
```diff
@@ -49,14 +49,21 @@ pub fn main() -> Result<(), MainError> {
...
+    HttpServer::new(move || {
+        create_app(&data)
+            .wrap(
+                Cors::new()
+                    .send_wildcard()
+                    .allowed_header("x-meili-api-key")
+                    .finish(),
+            )
+            .wrap(middleware::Logger::default())
+            .wrap(middleware::Compress::default())
```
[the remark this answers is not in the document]
qdequele: If I do that, I will have an unnecessary top layer on my meilisearch-http library. In this case, it is easy to override this one on the private MeiliSearch repository. Moreover, the type to respond to create_app it much much more complex if I add the compress middleware.

meilisearch-http/src/main.rs
```diff
@@ -1,24 +1,25 @@
...
 #[cfg(target_os = "linux")]
 #[global_allocator]
 static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
 
-pub fn main() -> Result<(), MainError> {
+#[tokio::main]
+async fn main() -> Result<(), MainError> {
     let opt = Opt::from_args();
+    let local = tokio::task::LocalSet::new();
+    let _sys = actix_rt::System::run_in_tokio("server", &local);
```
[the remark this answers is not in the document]
qdequele: What do you mean by `since we're running single threaded`?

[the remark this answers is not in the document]
qdequele: You have right, I don't have to use this trick yet. I will remove it and re-add it in the "Add sentry probe" PR. This trick is required if you not only use actix on your Tokio thread pool.

meilisearch-http/src/helpers/meilisearch.rs
qdequele: Not only, look at the message above.

meilisearch-http/src/lib.rs
```diff
@@ -8,3 +8,109 @@ pub mod option;
...
+        .service(routes::synonym::delete)
+        .service(routes::stats::index_stats)
+        .service(routes::stats::get_stats)
+        .service(routes::stats::get_version)
+        .service(routes::stats::get_sys_info)
+        .service(routes::stats::get_sys_info_pretty)
+        .service(routes::health::get_health)
+        .service(routes::health::change_healthyness)
+        .wrap(helpers::Authentication::Admin)
+        .service(routes::key::list)
```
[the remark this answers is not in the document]
qdequele: Actually, I think it's a little less clear. My goal is that we find all the information above the function absolutely. So we have the full path, the operator, the input types and so on.

meilisearch-http/src/error.rs
```diff
@@ -1,183 +1,173 @@
...
+            Self::MissingAuthorizationHeader => StatusCode::FORBIDDEN,
+            Self::MissingFilterValue => StatusCode::BAD_REQUEST,
+            Self::MissingHeader(_) => StatusCode::UNAUTHORIZED,
+            Self::NotFound(_) => StatusCode::NOT_FOUND,
+            Self::OpenIndex(_) => StatusCode::BAD_REQUEST,
+            Self::RetrieveDocument(_, _) => StatusCode::BAD_REQUEST,
+            Self::SearchDocuments(_) => StatusCode::BAD_REQUEST,
+            Self::UnknownFilteredAttribute => StatusCode::BAD_REQUEST,
+        }
     }
```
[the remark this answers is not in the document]
qdequele: Yeah, why not :)

meilisearch-http/src/main.rs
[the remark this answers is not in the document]
qdequele: I remove it for now

meilisearch-http/src/lib.rs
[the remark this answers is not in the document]
MarinPostma: Ok, I understand your point. Yet I think this function is very hard to read, and I see no reason to leak handlers in the publlic scope. We could keep the full routes as you want, and leave the configuration to the relevant module. This is how Lemmy [does it](https://github.com/LemmyNet/lemmy/blob/master/server/src/main.rs#L47-L55). Makes for much cleaner code IMHO. 

Also the only order that matters for middleware is their relative order, thus i see no reason not to put them all together somewhere, so they are in an obvious place to find and not spread around.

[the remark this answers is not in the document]
qdequele: Yeah! I will do like Lemmy! Yes, apparently, they wrap themself.  
>  The first middleware layer exposed to a Request is the outermost layer-- the last registered in the builder chain. Consequently, the first middleware registered in the builder chain is the last to execute during request processing.

https://docs.rs/actix-web/1.0.8/actix_web/struct.App.html#method.wrap
If I understand correctly.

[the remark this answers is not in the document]
MarinPostma: I understood the same thing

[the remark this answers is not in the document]
qdequele: With your proposal, I should test if it will not break the authentication middleware. I will test it.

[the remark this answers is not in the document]
qdequele: As expected I could not use `.wrap()` on ` &mut ServiceConfig` so I can't do the same thing they did.

meilisearch-http/src/error.rs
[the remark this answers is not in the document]
qdequele: The real question is: Is it better to use `impl Display` or `impl ToString`? @Kerollmops a piece of advice?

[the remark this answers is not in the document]
Kerollmops: This is written in the documentation of `ToString`, you should always impl `Display`, `ToString` derives from it.

https://doc.rust-lang.org/std/string/trait.ToString.html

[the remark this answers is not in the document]
qdequele: So `pub fn create_index(err: impl fmt::Display) -> ResponseError {` is fine?

meilisearch-http/src/error.rs
```diff
@@ -130,12 +130,11 @@ impl aweb::error::ResponseError for ResponseError {
             | Self::RetrieveDocument(_, _)
             | Self::SearchDocuments(_)
             | Self::UnknownFilteredAttribute => StatusCode::BAD_REQUEST,
-            Self::DocumentNotFound(_)
-            | Self::IndexNotFound(_)
-            | Self::NotFound(_) => StatusCode::NOT_FOUND,
+            Self::DocumentNotFound(_) | Self::IndexNotFound(_) | Self::NotFound(_) => {
+                StatusCode::NOT_FOUND
+            }
```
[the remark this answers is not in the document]
qdequele: I prefer keeping the standard format from cargo fmt

meilisearch-http/src/routes/search.rs
```diff
@@ -28,44 +29,46 @@ struct SearchQuery {
...
-    if let Some(attributes_to_crop) = query.attributes_to_crop {
-        let crop_length = query.crop_length.unwrap_or(200);
+    if let Some(attributes_to_crop) = &params.attributes_to_crop {
+        let crop_length = params.crop_length.unwrap_or(200);
         if attributes_to_crop == "*" {
             let attributes_to_crop = schema
                 .displayed_name()
                 .iter()
-                .map(|attr| (attr.to_string(), crop_length))
+                .map(|attr| ((*attr).to_string(), crop_length))
```
Kerollmops, line 71:
```suggestion
                .map(|attr| (attr.to_string(), crop_length))
```

meilisearch-http/src/routes/search.rs
```diff
@@ -77,12 +80,12 @@ pub async fn search_with_url_query(ctx: Request<Data>) -> SResult<Response> {
...
     }
 
-    if let Some(attributes_to_highlight) = query.attributes_to_highlight {
+    if let Some(attributes_to_highlight) = &params.attributes_to_highlight {
         let attributes_to_highlight = if attributes_to_highlight == "*" {
             schema
                 .displayed_name()
                 .iter()
-                .map(|s| s.to_string())
+                .map(|s| (*s).to_string())
```
Kerollmops, line 88:
```suggestion
                .map(|s| s.to_string())
```

meilisearch-http/src/routes/stats.rs
```diff
@@ -51,29 +67,25 @@ struct StatsResult {
...
                 let number_of_documents = index.main.number_of_documents(&reader)?;
 
                 let fields_frequency = index.main.fields_frequency(&reader)?.unwrap_or_default();
 
-                let is_indexing = ctx
-                    .state()
+                let is_indexing = data
                     .is_indexing(&update_reader, &index_uid)?
-                    .into_internal_error()?;
+                    .unwrap_or_default();
```
Kerollmops, line 88: Why don't we return an error here? This must not happen!

meilisearch-http/src/routes/stats.rs
```diff
@@ -115,20 +125,18 @@ struct VersionResponse {
     pkg_version: String,
 }
 
-pub async fn get_version(ctx: Request<Data>) -> SResult<Response> {
-    ctx.is_allowed(Admin)?;
-    let response = VersionResponse {
+#[get("/version", wrap = "Authentication::Private")]
```
Kerollmops, line 128: Is it wanted to downgrade the privacy to private instead of admin? I remember we talked about this and I think this is safe.

meilisearch-http/src/routes/stats.rs
```diff
@@ -200,28 +209,29 @@ pub(crate) fn report(pid: Pid) -> SysInfo {
...
+        .map(|(_, n)| n.get_received())
+        .sum::<u64>();
+    info.global.output_data = sys
+        .get_networks()
+        .into_iter()
+        .map(|(_, n)| n.get_transmitted())
+        .sum::<u64>();
 
-    if let Some(process) = sys.get_process(pid) {
+    if let Some(process) = sys.get_process(data.server_pid) {
```
Kerollmops, line 223: Did you add a new `server_pid` field to the `Data` struct? Why?

Kerollmops: Where does this `pid` came from before?

meilisearch-http/src/routes/stats.rs
```diff
@@ -297,21 +308,25 @@ pub(crate) fn report_pretty(pid: Pid) -> SysInfoPretty {
...
+    );
+    info.global.output_data = convert(
+        sys.get_networks()
+            .into_iter()
+            .map(|(_, n)| n.get_transmitted())
+            .sum::<u64>() as f64,
+    );
 
-    if let Some(process) = sys.get_process(pid) {
+    if let Some(process) = sys.get_process(data.server_pid) {
```
Kerollmops, line 324: Same here.

meilisearch-http/src/routes/stop_words.rs
```diff
@@ -1,63 +1,83 @@
...
+use crate::routes::{IndexParam, IndexUpdateResponse};
 use crate::Data;
 
-pub async fn get(ctx: Request<Data>) -> SResult<Response> {
-    ctx.is_allowed(Private)?;
-    let index = ctx.index()?;
-    let db = &ctx.state().db;
-    let reader = db.main_read_txn()?;
+pub fn services(cfg: &mut web::ServiceConfig) {
+    cfg.service(get).service(update).service(delete);
```
Kerollmops, line 12:
```suggestion
    cfg.service(get)
        .service(update)
        .service(delete);
```

This feels easier to read.

meilisearch-http/src/routes/stop_words.rs
```diff
@@ -1,63 +1,83 @@
...
+    let index = data
+        .db
+        .open_index(&path.index_uid)
+        .ok_or(ResponseError::index_not_found(&path.index_uid))?;
+    let reader = data.db.main_read_txn()?;
     let stop_words_fst = index.main.stop_words_fst(&reader)?;
     let stop_words = stop_words_fst.unwrap_or_default().stream().into_strs()?;
 
-    Ok(tide::Response::new(200).body_json(&stop_words).unwrap())
+    Ok(HttpResponse::Ok().json(stop_words))
```
Kerollmops, line 31: Why this instead of the `web::Json()` function?

meilisearch-http/src/routes/synonym.rs
```diff
@@ -1,29 +1,39 @@
...
+use crate::error::ResponseError;
+use crate::helpers::Authentication;
+use crate::routes::{IndexParam, IndexUpdateResponse};
 use crate::Data;
 
-pub async fn get(ctx: Request<Data>) -> SResult<Response> {
-    ctx.is_allowed(Private)?;
-    let index = ctx.index()?;
+pub fn services(cfg: &mut web::ServiceConfig) {
+    cfg.service(get).service(update).service(delete);
```
Kerollmops, line 14:
```suggestion
    cfg.service(get)
        .service(update)
        .service(delete);
```

meilisearch-http/src/routes/synonym.rs
```diff
@@ -33,50 +43,57 @@ pub async fn get(ctx: Request<Data>) -> SResult<Response> {
         }
     }
 
-    Ok(tide::Response::new(200).body_json(&synonyms).unwrap())
+    Ok(HttpResponse::Ok().json(synonyms))
```
Kerollmops, line 46: Same here and for every similar places, please don't mixup `web::Json()` and `HttpResponse::Ok().json()`, choose one and use it everywhere. I prefer the `HttpResponse::Ok().json()` because it is consistent with `HttpResponse::Accepted` and others.

meilisearch-http/src/helpers/authentication.rs
```diff
@@ -0,0 +1,100 @@
...
+    type Request = ServiceRequest;
+    type Response = ServiceResponse<B>;
+    type Error = Error;
+    type InitError = ();
+    type Transform = LoggingMiddleware<S>;
+    type Future = Ready<Result<Self::Transform, Self::InitError>>;
+
+    fn new_transform(&self, service: S) -> Self::Future {
+        ok(LoggingMiddleware {
+            acl: (*self).clone(),
```
Kerollmops, line 35:
```suggestion
            acl: self.clone(),
```

meilisearch-http/src/helpers/authentication.rs
```diff
@@ -0,0 +1,100 @@
...
+    type Error = Error;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>>>>;
+
+    fn poll_ready(&mut self, cx: &mut Context) -> Poll<Result<(), Self::Error>> {
+        self.service.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: ServiceRequest) -> Self::Future {
+        let mut svc = self.service.clone();
+        let data = req.app_data::<Data>().unwrap();
```
Kerollmops, line 63: What is this `unwrap` for?

meilisearch-http/src/helpers/authentication.rs
```diff
@@ -0,0 +1,100 @@
...
+            Authentication::Private => {
+                data.api_keys.master.as_deref() == Some(auth_header)
+                    || data.api_keys.private.as_deref() == Some(auth_header)
+            }
+            Authentication::Public => {
+                data.api_keys.master.as_deref() == Some(auth_header)
+                    || data.api_keys.private.as_deref() == Some(auth_header)
+                    || data.api_keys.public.as_deref() == Some(auth_header)
+            }
+        };
```
Kerollmops, line 90: This is really clean, this is nice!

meilisearch-http/src/helpers/meilisearch.rs
```diff
@@ -174,17 +115,17 @@ impl<'a> SearchBuilder<'a> {
         if let Some(filters) = &self.filters {
```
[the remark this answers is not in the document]
qdequele: yes I need to rebase.

meilisearch-http/src/routes/stats.rs
qdequele: Yes, it's an open issue https://github.com/meilisearch/MeiliSearch/issues/584

meilisearch-http/src/routes/stats.rs
qdequele: Previously the `pid` come from `ctx.state().server_pid`. It's exactly the same here. I haven't modified the Data struct.

meilisearch-http/src/routes/stop_words.rs
qdequele: For `meilisearch-http`, I have respected the `cargo fmt` format. Do you think we must stop respect it or add an exception for this one?

meilisearch-http/src/routes/synonym.rs
qdequele: same here

meilisearch-http/src/routes/stop_words.rs
qdequele: Like that, I always see at the end of the function the status code returned. And it's the same return as all handlers.

meilisearch-http/src/routes/synonym.rs
qdequele: totally agree. I will change the `web::Json()` to `HttpResponse::Ok().json()` everywhere.

meilisearch-http/src/helpers/authentication.rs
qdequele: If this one unwraps it's means that actix-web is broken. Should absolutely never appear. I give to actix as my `Data` as `app_data`. If it can't retrieve it when I ask for it, that means all server has an issue and we must urgently do something. I prefer to be sure that the case will be handle on the test regardless of the quality of the tests performed.

meilisearch-http/src/routes/stats.rs
qdequele: Previously the pid come from ctx.state().server_pid. It's exactly the same here. I haven't modified the Data struct.

meilisearch-http/src/routes/document.rs
```diff
@@ -123,34 +141,37 @@ struct UpdateDocumentsQuery {
...
+                .first()
+                .and_then(find_primary_key)
+                .ok_or(ResponseError::bad_request("Could not infer a primary key"))?,
         };
 
-        let mut writer = db.main_write_txn()?;
-        schema.set_primary_key(&id).map_err(ResponseError::bad_request)?;
+        let mut writer = data.db.main_write_txn()?;
+
+        schema.set_primary_key(&id)?;
```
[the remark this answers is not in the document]
qdequele: No, Good point!

meilisearch-http/src/routes/index.rs
```diff
@@ -155,47 +170,50 @@ pub async fn create_index(mut ctx: Request<Data>) -> SResult<Response> {
...
+        .db
+        .create_index(&uid)
+        .map_err(ResponseError::create_index)?;
 
-    let mut writer = db.main_write_txn()?;
-    let name = body.name.unwrap_or(uid.clone());
+    let mut writer = data.db.main_write_txn()?;
+
+    let name = body.name.clone().unwrap_or(uid.clone());
     created_index.main.put_name(&mut writer, &name)?;
```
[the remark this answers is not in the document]
qdequele: It does not work can only send `&name` on `put_name ` and I use `body` and `uid` after in the function.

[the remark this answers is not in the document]
Kerollmops: If the `put_name` take a `&str` therefore you can give it a `&str` and not a `&String` as you are doing here.

```suggestion
    let name = body.name.as_ref().unwrap_or(&uid);
    created_index.main.put_name(&mut writer, name)?;
```

meilisearch-http/src/routes/stats.rs
Kerollmops: I am aware of this issue, it doesn't tell if we need to change the privacy level, but if you want to change it I am not against that.

meilisearch-http/src/routes/stop_words.rs
Kerollmops: Keep it like you have done, in one line, that's ok.

meilisearch-http/src/helpers/authentication.rs
Kerollmops: That's cool, could you add a little explanation comment above it? please.

qdequele: Yes :)

meilisearch-http/src/routes/stats.rs
qdequele: If I change it now, we'll be up to date with the doc.

meilisearch-http/src/routes/search.rs
```diff
@@ -94,144 +98,19 @@ pub async fn search_with_url_query(ctx: Request<Data>) -> SResult<Response> {
...
-    if let Some(filters) = query.filters {
-        search_builder.filters(filters);
+    if let Some(filters) = &params.filters {
+        search_builder.filters(filters.to_string());
     }
 
-    if let Some(timeout_ms) = query.timeout_ms {
+    if let Some(timeout_ms) = params.timeout_ms {
         search_builder.timeout(Duration::from_millis(timeout_ms));
     }
```
Kerollmops, line 107: This parameter doesn't exists anymore [even in the doc](https://docs.meilisearch.com/references/search.html#search-in-an-index). Could you remove it, please?

meilisearch-http/src/helpers/authentication.rs
Kerollmops: I wasn't joking, this is easy to read and understand! GG!

meilisearch-http/src/helpers/meilisearch.rs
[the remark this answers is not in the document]
Kerollmops: Don't forget to rebase :)

meilisearch-http/src/routes/document.rs
```diff
@@ -83,29 +109,21 @@ pub async fn get_all_documents(ctx: Request<Data>) -> SResult<Response> {
...
-            if let Ok(Some(document)) = index.document(&reader, Some(&attributes), document_id) {
-                response_body.push(document);
-            }
-        }
-    } else {
-        for document_id in documents_ids {
-            if let Ok(Some(document)) = index.document(&reader, None, document_id) {
-                response_body.push(document);
-            }
+    let mut response = Vec::<Document>::new();
```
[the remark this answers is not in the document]
qdequele: If I do that I should specify the type on:
```rust
 if let Ok(Some(document)) = index.document::<Document>(&reader, None, document_id) {
     response_body.push(document);
}
```

What did you prefer?

meilisearch-http/src/routes/index.rs
```diff
@@ -116,86 +146,74 @@ struct IndexCreateRequest {
...
     }
 
-    let db = &ctx.state().db;
-
-    let uid = match body.uid {
+    let uid = match &body.uid {
         Some(uid) => {
             if uid
                 .chars()
                 .all(|x| x.is_ascii_alphanumeric() || x == '-' || x == '_')
```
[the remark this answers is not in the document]
qdequele: cargo fmt don't want :(

meilisearch-http/src/helpers/authentication.rs
qdequele: thanks :)

meilisearch-http/src/helpers/meilisearch.rs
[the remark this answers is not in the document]
qdequele: Tell me when I can rebase?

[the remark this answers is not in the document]
Kerollmops: Now, if you are ready ;)

meilisearch-http/src/routes/document.rs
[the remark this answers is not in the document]
Kerollmops: Keep it like it is, didn't mind that, sorry.

meilisearch-http/src/routes/index.rs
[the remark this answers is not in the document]
Kerollmops: Is the complete codebase `cargo fmt`ted, already ?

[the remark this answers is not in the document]
qdequele: Yes, since a long time. only the meilisearch-http part

meilisearch-http/src/routes/search.rs
```diff
@@ -25,38 +27,39 @@ struct SearchQuery {
...
     let schema = index
         .main
         .schema(&reader)?
-        .ok_or(ResponseError::open_index("No Schema found"))?;
+        .ok_or(ResponseError::internal("Impossible to retrieve the schema"))?;
 
-    let query: SearchQuery = ctx
-        .query()
-        .map_err(|_| ResponseError::bad_request("invalid query parameter"))?;
+    let mut search_builder = index.new_search(params.q.clone());
```
Kerollmops, line 51: I am wondering about this clone, is it necessary?

meilisearch-http/tests/search.rs
```diff
@@ -1,23 +1,18 @@
...
-use once_cell::sync::Lazy;
 use serde_json::json;
 
 mod common;
 
-static GLOBAL_SERVER: Lazy<Mutex<common::Server>> = Lazy::new(|| {
-    let mut server = common::Server::with_uid("movies");
-    server.populate_movies();
-    Mutex::new(server)
-});
```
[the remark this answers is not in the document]
qdequele: Yes, it's the second one, I haven't be able to do call populate_movies that is async.

meilisearch-http/src/routes/search.rs
qdequele: Yes, I think. new_search take a String. I could change it but to do that I need to change the SearchBuilder.

qdequele:
@Kerollmops done.

Kerollmops:
So we are now on actix-web, congratulation 🎉

PR #607, opened 2020-04-15
Add '@' char as a tokenizer separator

tpayet:
When we are indexing email adresses, users should be able to search through the domain name.
This is not possible currently because we are not splitting on `@`.

I've not come up with other common usage of the `@`, if you have any in mind we might consider classifying it as a `Soft` separator, a `Hard` separator or not a separator at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v0.10.1 (unreleased)
+
+  - Add '@' character as tokenizer separator (#607)
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## v0.10.1
```
Kerollmops, line 1:
```suggestion
## v0.10.1 (unreleased)
```

PR #610, opened 2020-04-16
Added string case-insensitive contains '*=' symbol to filter language

NoahSaso:
Title.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
-## v0.10.1
+## v0.10.2
 
-  - Added filter string condition case-insensitive contains represented by `field _= value`
+  - Added filter string condition case-insensitive contains represented by `field *= value`
 
 ## v0.10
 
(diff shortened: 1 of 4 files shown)


MarinPostma:
Hello @NoahSaso, thank you for your contribution! We're looking into it. In the meantime, can you update the error message translation in `meilisearch-core/src/error.rs:42`?

NoahSaso:
@MarinPostma, I found a use case in my own project and thought it could be helpful! Updated translation.

Kerollmops:
Hey @NoahSaso,

First of all, thank you for your work! That is really clean!

I am wondering if this is a feature we want to add to the engine:
  1. It could impact the speed of the search too much.
  2. We will release facetted searching (aggregations) that will solve the use case you probably have.
  3. Introducing a new operator is probably not that good.

Anyway, I would like too ear more about the use case you have, please, just to see if another feature could fit you.

PR #616, opened 2020-04-20
filters on arrays

### LINKED ISSUE #611: Support string equality in arrays
The current filter engine supports case insensitive string equality but not case insensitive string equality inside of arrays (array of string). This is one particular feature that I though would be immediately available with the new filter syntax.

This feature as been highlighted by someone on crisp (visitor34965):
> How would I implement tags filtering with this?

If we support this we can make the engine way more expressive than currently, we can support tag filtering when documents can have multiple tags.

I would also like to see a "common usage" section in the filtering documentation page, to explain the usage you can do of string matching another string or string matching inside of an array of string, this way people searching for "filtering tags" or anything similar can find this.

About the potential performance issue: if the documents have big arrays and no strings in any documents contains the filtered tag this is bad. This will be kind of solved when we will introduce facetting, any attribute that is composed of a string or an array of strings can be used as a facetted field and the complexity will be drastically reduced to a simple pre-search intersection on ids.

```JSON
[
  { "id": 1, "colors": ["red", "green", "blue"] },
  { "id": 2, "colors": ["red"] },
  { "id": 2, "colors": "green" }
]
```

```SQL
colors = "red" # matches 1 and 2
colors = "green" # matches 2 and 3
colors = "red" OR colors = "green" # matches 1, 2, and 3
```

MarinPostma:
resolve #611

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## v0.10.1 (unreleased)
 
   - Add '@' character as tokenizer separator (#607)
+  - Add support for filtering on arrays of strings (#611)
 
 ## v0.10
 
--- a/meilisearch-core/src/filters/condition.rs
+++ b/meilisearch-core/src/filters/condition.rs
@@ -10,7 +10,7 @@ use pest::iterators::Pair;
 use serde_json::{Value, Number};
 use super::parser::Rule;
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 enum ConditionType {
     Greater,
     Less,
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## v0.10.1 (unreleased)
 
   - Add '@' character as tokenizer separator (#607)
+  - Add support for filtering on arrays (#611)
```
Kerollmops, line 4:
```suggestion
  - Add support for filtering on arrays of strings (#611)
```

meilisearch-core/src/filters/condition.rs
```diff
@@ -179,22 +186,24 @@ impl<'a> Condition<'a> {
...
-                        return Ok(res)
+                        return res
                     } 
                 } 
-                Ok(false)
+                false
             },
             Some(Value::Bool(b)) => {
                 if let Some(value) = self.value.as_bool() {
                     return match self.condition {
```
Kerollmops, line 196: Could you set a variable and return it after the match? like above.

CHANGELOG.md
Kerollmops: Could you change your commit message to this text, please?

PR #623, opened 2020-04-22
Fixes the content overflow in the web interface in chrome.

### LINKED ISSUE #609: Web interface has a css issue in chrome
The value of the fields is not correctly cut in chrome. 

Chrome:
Screenshot 2020-04-15 at 15 47 14

Firefox: 
Screenshot 2020-04-15 at 15 47 10

lironhl:
Fixes #609,
After the css changes:
### Firefox
firefox
### Chrome
chrome

The difference [according to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) may be because of

> ### overflow-wrap
> ### break-word
>   The same as the anywhere value, with normally unbreakable words allowed to be broken at arbitrary points **if there are no otherwise acceptable break points in the line** ... .

### DIFF
--- a/meilisearch-http/public/interface.html
+++ b/meilisearch-http/public/interface.html
@@ -79,6 +79,7 @@
         box-sizing: border-box;
         padding-left: 10px;
         color: rgba(0,0,0,.9);
+        overflow-wrap: break-word;
       }
     </style>
   </head>


Kerollmops:
That is a great PR, I don't even have to make a review, I merge!
Thank you :)

Issue #625, opened 2020-04-22
Tracking issue: Facets


MarinPostma:
This is the tracking issue for the implementation of faceting in Meilisearch.

## Context

Filtering  was thought as two part release. Some time ago we released filtering (https://github.com/meilisearch/MeiliSearch/pull/592), which allows users to create filter documents on arbitrary fields. While this is very flexible it has some performance issues, and when possible, it is preferable to use faceting.

Faceting, the subject of this tracking issue, differs from filtering in that instead of filtering results in the outbound set of documents, it restricts the documents to be filtered on *upstream*. In order to do that, we must declare _facets_ at indexing. Facets are document attributes that indicate which category a document belongs to.

Filtering and faceting aim at being complementary.

###  Setting up facets 

Facet that can be used are set up thanks to the `[POST]/indexes/:index_uid/settings` route. We add a new attribute `attributesForFaceting` that accepts a `[String]`. It's default value is `null`.

Any call to settings with a value for the attribute `attributesForFaceting` overwrites the currently set faceting attributes. Passing an empty array resets all faceting attributes.

A re-indexing when the value for this setting is changed.

A call to  `[POST]/indexes/:index_uid/settings` should list the currently set facets.

**TODO:**
- [x] update `POST` route to accept new attribute  `attributesForFaceting`
- [x] update `GET` route to include facet information in its response

### Document addition and deletion 

On document addition and deletion, the facet indexation must be updated:
**TODO:**
- [x] add new document references to the facet index on document addition
- [x] remove document references from the facet index on document deletion

### Search query 

When doing a search query, a user can specify:
- facets to filter on with the facet syntax (explained below), thanks to the optional parameter `facetFilters`.
- the facets for which to retrieve the matching count for, with the `facets` optional parameter.

`facetFilters` accept arguments of the form `["facetName:facetValue"]` or `[["FacetName:FacetValue"]]` specifying the facets to be used, with the following semantic:

```
[["color:red", "color:blue"], "kind:t-shirt"] <=> ("color:red" OR "color:blue") AND "kind:t-shirt"
```
inner arrays elements are `OR`ed together, outer array elements are `AND`ed together.

Array depth must not be greater than 2, and be at least 1:

* `[["color:red"]]` => correct syntax
* `["color:red"]` => correct syntax
* `"color:red"` => incorrect syntax

Errors are sent back to the user if:
* the syntax is invalid
[...]

PR #631, opened 2020-04-23
Facet filters

MarinPostma:
#625


MarinPostma:
will do remaining facet count in another PR ;)

meilisearch-core/src/store/facets.rs
```diff
@@ -0,0 +1,57 @@
...
+
+impl Facets {
+    // we use sdset::SetBuf to ensure the docids are sorted.
+    pub fn put_facets(&self, writer: &mut RwTxn<MainT>, facet_key: FacetKey, doc_ids: &Set<DocumentId>) -> ZResult<()> {
+        self.facets.put(writer, &facet_key, doc_ids)
+    }
+
+    pub fn document_ids<'txn>( &self, reader: &'txn RoTxn<MainT>, facet_key: &FacetKey,) -> Result<Option<Cow<'txn, [DocumentId]>>, Error> {
+        let res = self.facets.get(reader, &facet_key).map_err(Error::from);
+        res
```
Kerollmops, line 27:
```suggestion
        self.facets.get(reader, &facet_key)
```

meilisearch-core/src/store/facets.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub(crate) facets: heed::Database<FacetKey, CowSlice<DocumentId>>,
+}
+
+impl Facets {
+    // we use sdset::SetBuf to ensure the docids are sorted.
+    pub fn put_facets(&self, writer: &mut RwTxn<MainT>, facet_key: FacetKey, doc_ids: &Set<DocumentId>) -> ZResult<()> {
+        self.facets.put(writer, &facet_key, doc_ids)
+    }
+
+    pub fn document_ids<'txn>( &self, reader: &'txn RoTxn<MainT>, facet_key: &FacetKey,) -> Result<Option<Cow<'txn, [DocumentId]>>, Error> {
```
Kerollmops:
```suggestion
    pub fn facet_ids<'txn>( &self, reader: &'txn RoTxn<MainT>, facet_key: &FacetKey) -> Result<Option<Cow<'txn, [DocumentId]>>, Error> {
```

Kerollmops: BTW I don't unsderstand why this function doesn't just return a `ZResult<Cow<'txn, Set<DocumentId>>>`?

meilisearch-core/src/store/main.rs
```diff
@@ -188,6 +190,14 @@ impl Main {
         }
     }
 
+    pub fn attributes_for_faceting<'txn>(&self, reader: &'txn heed::RoTxn<MainT>) -> ZResult<Option<Cow<'txn, [FieldId]>>> {
+        self.main.get::<_, Str, CowSlice<FieldId>>(reader, ATTRIBUTES_FOR_FACETING)
+    }
+
+    pub fn put_attributes_for_faceting(self, writer: &mut heed::RwTxn<MainT>, attributes: &[FieldId]) -> ZResult<()> {
+        self.main.put::<_, Str, CowSlice<FieldId>>(writer, ATTRIBUTES_FOR_FACETING, &attributes.to_vec())
```
Kerollmops, line 198:
```suggestion
        self.main.put::<_, Str, CowSlice<FieldId>>(writer, ATTRIBUTES_FOR_FACETING, attributes)
```

meilisearch-core/src/store/mod.rs
```diff
@@ -207,7 +213,7 @@ pub struct Index {
     pub docs_words: DocsWords,
     pub prefix_documents_cache: PrefixDocumentsCache,
     pub prefix_postings_lists_cache: PrefixPostingsListsCache,
-
+    pub facets: Facets,
```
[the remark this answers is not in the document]
Kerollmops: And put it in the same place than when you cerated it, just above the synonyms one.

meilisearch-core/src/store/mod.rs
```diff
@@ -420,6 +413,7 @@ pub fn create(
         updates: Updates { updates },
         updates_results: UpdatesResults { updates_results },
         updates_notifier,
+        facets: Facets { facets },
```
Kerollmops, line 416: Same, here, please separate it from the "updates" related databases.

meilisearch-core/src/update/documents_addition.rs
```diff
@@ -143,6 +163,12 @@ pub fn apply_documents_addition<'a, 'b>(
     };
 
     // 3. index the documents fields in the stores
+    // add facets for docs facets TODO move in a function
+    if let Some(attributes_for_facetting) = index.main.attributes_for_faceting(writer)? {
+        let facet_map = facets::facet_map_from_docs(&schema, &documents_additions, attributes_for_facetting.as_ref())?;
+        index.facets.add(writer, facet_map)?;
+    }
```
Kerollmops, line 170: If I understand that part correctly, the `facet_map_from_docs` function retrieves all the facet names/values from the list of `IndexMap` representing the documents?

meilisearch-core/src/update/documents_addition.rs
```diff
@@ -289,6 +246,11 @@ pub fn reindex_all_documents(writer: &mut heed::RwTxn<MainT>, index: &store::Ind
     let mut indexer = RawIndexer::new(stop_words);
     let mut ram_store = HashMap::new();
 
+    if let Some(ref attributes_for_facetting) = index.main.attributes_for_faceting(writer)? {
+        let facet_map = facets::facet_map_from_docids(&index, &documents_ids_to_reindex, writer, &attributes_for_facetting)?;
+        index.facets.add(writer, facet_map)?;
+    }
+    // ^-- merge? --v
```
Kerollmops, line 253: What does this comment means?

meilisearch-core/src/update/settings_update.rs
```diff
@@ -102,6 +102,19 @@ pub fn apply_settings_update(
...
     }
 
+    // TODO smarter reindexing
+    match settings.attributes_for_faceting {
+        UpdateState::Update(attrs) => {
+            apply_attributes_for_faceting_update(writer, index, &mut schema, &attrs)?;
+            must_reindex = true;
+        },
+        UpdateState::Clear => {
+            apply_attributes_for_faceting_update(writer, index, &mut schema, &[])?;
```
Kerollmops, line 112: Does a clear must set an empty array here? Is the first ever written value an empty array or the value doesn't exist at all?

meilisearch-core/src/update/settings_update.rs
```diff
@@ -131,6 +144,19 @@ pub fn apply_settings_update(
     Ok(())
 }
 
+fn apply_attributes_for_faceting_update(writer: &mut heed::RwTxn<MainT>,
+    index: &store::Index,
+    schema: &mut Schema,
+    attributes: &[String]) -> MResult<()> {
+
```
Kerollmops, line 151:
```suggestion
fn apply_attributes_for_faceting_update(
    writer: &mut heed::RwTxn<MainT>,
    index: &store::Index,
    schema: &mut Schema,
    attributes: &[String]
) -> MResult<()>
{
```

meilisearch-core/src/update/settings_update.rs
```diff
@@ -131,6 +144,19 @@ pub fn apply_settings_update(
...
+    let mut attribute_ids = HashSet::new();
+    for name in attributes {
+        attribute_ids.insert(schema.insert(name)?);
+    }
+    index.main.put_attributes_for_faceting(writer, &attribute_ids.into_iter().collect::<Vec<_>>())?;
```
Kerollmops, line 156: Not sure if the value that we store must be an `Array` or an `HashSet`...

meilisearch-http/src/helpers/meilisearch.rs
```diff
@@ -82,6 +85,11 @@ impl<'a> SearchBuilder<'a> {
         self
     }
 
+    pub fn add_facet_fitlers(&mut self, filters: FacetFilter) -> &SearchBuilder {
```
Kerollmops, line 88:
```suggestion
    pub fn add_facet_filters(&mut self, filters: FacetFilter) -> &SearchBuilder {
```

meilisearch-http/src/routes/search.rs
```diff
@@ -28,6 +30,7 @@ struct SearchQuery {
     attributes_to_highlight: Option<String>,
     filters: Option<String>,
     matches: Option<bool>,
+    facet_filters: Option<String>,
```
Kerollmops, line 33: I am not sure about that, wouldn't it be possible to accept an `Vec<Either<Vec<String>, String>>` here?
We would benefit from the serde error messages when parsing the user request.
Is it because query parameter parsing doesn't works correctly with `Either` and `Vec`s?

meilisearch-http/src/routes/setting.rs
```diff
@@ -91,6 +91,19 @@ async fn get_all(
...
+            Some(attrs
+                .iter()
+                .map(|field| schema
+                    .name(*field)
+                    .unwrap_or(&"unknown field name")
+                    .to_string())
+                .collect())
+        }
+        _ => None,
+    };
```
Kerollmops, line 105: I'm not sure if all this logic must be done at all, we should be able to retrieve an `HashSet` of the `attributes_for_faceting` and not check them. The update system must be correclty handling field removals and facets clean-ups.

meilisearch-schema/Cargo.toml
```diff
@@ -11,3 +11,4 @@ indexmap = { version = "1.3.2", features = ["serde-1"] }
 serde = { version = "1.0.105", features = ["derive"] }
 serde_json = { version = "1.0.50", features = ["preserve_order"] }
 toml = { version = "0.5.6", features = ["preserve_order"] }
+zerocopy = "0.3.0"
```
Kerollmops, line 14: Why have you added this dependency? Is it related to the `attributes_for_faceting` that is an array and not an `HashSet` now?

meilisearch-schema/src/lib.rs
```diff
@@ -36,7 +37,8 @@ impl Into<u16> for IndexedPos {
     }
 }
 
-#[derive(Serialize, Deserialize, Debug, Copy, Clone, Default, PartialOrd, Ord, PartialEq, Eq, Hash)]
+#[derive(Serialize, Deserialize, Debug, Copy, Clone, Default, PartialOrd, Ord, PartialEq, Eq, Hash, AsBytes, FromBytes)]
```
Kerollmops, line 40: Hum... Ok, I think we should revert these changes and only keep the serde traits as we will use an `HashSet` instead of a slice.

meilisearch-http/Cargo.toml
```diff
@@ -23,6 +23,7 @@ actix-web = "2"
 actix-web-macros = "0.1.0"
 chrono = { version = "0.4.11", features = ["serde"] }
 crossbeam-channel = "0.4.2"
+either = { version = "1.5.3", features = ["serde"] }
```
Kerollmops, line 26: I think you added this dependency when you thought about using `Either` in the `UserQuery::facet_filters` field, but you were not able to make that work and forget to remove it from there.

meilisearch-core/src/query_builder.rs
```diff
@@ -93,36 +76,66 @@ impl<'c, 'f, 'd> QueryBuilder<'c, 'f, 'd> {
...
+                            ands.push(sdset::SetBuf::new_unchecked(docids.into_owned()));
+                        }
+                    }
+                };
+            }
+            let ands = ands.iter().map(sdset::SetBuf::as_ref).collect::<Vec<_>>();
+            Some(sdset::multi::OpBuilder::from_vec(ands).intersection().into_set_buf())
+        } else {
+            None
+        };
```
Kerollmops, line 105: I am not sure this is the best implementation of this algorithm, we should thing about that when you got time.

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,345 @@
...
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            )));
+        }
+        let value = &string[split_index + 1..];
+        let value = match serde_json::from_str::<Value>(value) {
+            Ok(Value::String(s)) => s,
+            Ok(v) => v.to_string(),
+            _ => value.to_string(),
+        };
```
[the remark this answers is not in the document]
MarinPostma: we want to support "foo:\"hello world\"" and  "foo:'hello world'", instead of re implementing a small parser to do unquoting, I use serde's

meilisearch-http/src/routes/setting.rs
MarinPostma: we store atrributes as `FieldId`, they must be converted back to strings so the user can read them.

meilisearch-core/src/update/settings_update.rs
```diff
@@ -102,6 +102,19 @@ pub fn apply_settings_update(
...
 
+    // TODO smarter reindexing
+    match settings.attributes_for_faceting {
+        UpdateState::Update(attrs) => {
+            apply_attributes_for_faceting_update(writer, index, &mut schema, &attrs)?;
+            must_reindex = true;
+        },
+        UpdateState::Clear => {
+            apply_attributes_for_faceting_update(writer, index, &mut schema, &[])?;
+            must_reindex = true;
```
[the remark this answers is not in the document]
MarinPostma: simply clearing facet_store in this case should be enough in this case indeed.

meilisearch-core/src/update/settings_update.rs
MarinPostma: no, we can indeed just delete the entry

meilisearch-http/src/routes/search.rs
MarinPostma: I tried, and unfortunately, it doesn't work... :( would have saved me some work

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,345 @@
...
+        attributes_for_faceting: &[FieldId],
+    ) -> Result<Self, Error> {
+        use Error::FacetError;
+
+        let split_index = string.find(":").ok_or(FacetError(format!(
+            "invalid facet: {}, facets should be \"facetName:facetvalue\"",
+            string
+        )))?;
+        let key = &string[..split_index];
+        let field_id = schema.id(key).ok_or(FacetError(format!(
```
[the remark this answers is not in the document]
MarinPostma:
```
pub fn ok_or<E>(self, err: E) -> Result<T, E>  
```
ok_or seems especially made for that purpose... We don't need to perform the extra computation offered by `ok_or_else`

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,345 @@
...
+            _ => value.to_string(),
+        };
+        Ok(Self::new(field_id, &value))
+    }
+}
+
+impl<'a> heed::BytesEncode<'a> for FacetKey {
+    type EItem = FacetKey;
+
+    fn bytes_encode(item: &'a Self::EItem) -> Option<Cow<'a, [u8]>> {
```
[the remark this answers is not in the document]
MarinPostma: uhm.. it appears that it's required:

```
implicit elided lifetime not allowed here
```

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,345 @@
...
+        if item.1.len() > 512 - 2 {
+            None
+        } else {
+            let mut buffer = Vec::with_capacity(2 + item.1.len());
+            let field_id_as_u16: u16 = item.0.into();
+            let field_as_bytes: [u8; 2] = field_id_as_u16.to_be_bytes();
```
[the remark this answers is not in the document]
MarinPostma: rustc does not seem to be able to do the conversion from `FieldId` to u16 and then to bytes in one go. This was done on purpose. If you have a more elegant solution, I'm all ears :)

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,345 @@
...
+            Some(Cow::Owned(buffer))
+        }
+    }
+}
+
+impl<'a> heed::BytesDecode<'a> for FacetKey {
+    type DItem = FacetKey;
+
+    fn bytes_decode(bytes: &'a [u8]) -> Option<Self::DItem> {
+        let field_id_u16 = u16::from_be_bytes([bytes[0], bytes[1]]);
```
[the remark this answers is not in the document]
MarinPostma: Not sure how to use it, let's discuss that

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,345 @@
...
+        let field_id = FieldId(field_id_u16);
+        let string = unsafe { std::str::from_utf8_unchecked(&bytes[2..]) };
+        Some(FacetKey::new(field_id, &string))
```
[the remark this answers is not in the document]
MarinPostma: that's right!

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,345 @@
...
+            let field_as_bytes: [u8; 2] = field_id_as_u16.to_be_bytes();
+            buffer.extend_from_slice(&field_as_bytes);
+            buffer.extend_from_slice(item.1.as_bytes());
+            Some(Cow::Owned(buffer))
+        }
+    }
+}
+
+impl<'a> heed::BytesDecode<'a> for FacetKey {
+    type DItem = FacetKey;
```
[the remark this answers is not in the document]
MarinPostma: I will look into that, I'm not one hundred percent sure I don't need facet key to own it's string at some point, I believe I used to have it as a &str

meilisearch-core/src/facets.rs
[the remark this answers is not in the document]
Kerollmops: If we use `ok_or` the `format!` is always computed however if we use `ok_or_else` the format is only computed on purpose.

meilisearch-core/src/facets.rs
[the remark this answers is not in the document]
Kerollmops: Ok but neither `'hello world'` or `hello` (without quotes) are valid JSON.

meilisearch-core/src/facets.rs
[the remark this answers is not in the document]
Kerollmops: Ho, sorry forgot about that.
```suggestion
impl heed::BytesEncode<'_> for FacetKey {
    type EItem = FacetKey;

    fn bytes_encode(item: &Self::EItem) -> Option<Cow<[u8]>> {
```

meilisearch-core/src/facets.rs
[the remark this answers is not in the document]
Kerollmops: Yes, or something like a `Cow` inside would be great and performant.

meilisearch-core/src/store/facets.rs
Kerollmops: Yeah, we should store a `Set<DocumentId>` because the purpose of these ids is to perform set operations on them. Also I think you are right and we could call this function `facet_documents_ids`, it fit better.

meilisearch-core/src/store/facets.rs
```diff
@@ -0,0 +1,57 @@
...
+#[derive(Clone, Copy)]
+pub struct Facets {
+    pub(crate) facets: heed::Database<FacetKey, CowSlice<DocumentId>>,
+}
+
+impl Facets {
+    // we use sdset::SetBuf to ensure the docids are sorted.
+    pub fn put_facets(&self, writer: &mut RwTxn<MainT>, facet_key: FacetKey, doc_ids: &Set<DocumentId>) -> ZResult<()> {
+        self.facets.put(writer, &facet_key, doc_ids)
+    }
```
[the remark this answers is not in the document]
Kerollmops: Same here, something like `put_facet_documents_ids` would be better.

meilisearch-http/src/routes/setting.rs
Kerollmops: Are you sure this is `unwrap_or` that you want to use or `ok_or` or `expect` maybe?

MarinPostma: I don't want to panic, or return an error. I think I will just filter out values I can't find.

MarinPostma: truth is that it should not happen, or you have more worrying problem too

meilisearch-core/src/facets.rs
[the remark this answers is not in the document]
MarinPostma: Uhmm you're right about the former. The later I expect it to fail with 
```
_ => value.to_string(),
```
and it still allows me to parse foo:"hello world". we only need one though.

this branch me not be usefull:
```
 Ok(v) => v.to_string(),
```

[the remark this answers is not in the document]
MarinPostma: we could also do something like that:

```rust
        let mut indices = value.char_indices();
        let first = indices.next().unwrap();
        let last = indices.last().unwrap();
        let value =  match (first, last) {
            ((s, '\''), (e, '\'')) |
            ((s, '\"'), (e, '\"')) => value[s + 1..e].to_string(),
            _ => value.to_string(),
        };

```

meilisearch-core/src/store/facets.rs
MarinPostma: How can I get heed to return me a Cow<Set<DocumentId>>??

meilisearch-core/src/update/documents_addition.rs
MarinPostma: ah yes, for now I have factorized code to create the facet map in a separate function. However it could use some of the work done in the block under to save some calls to store. It's a performance vs readability/maintanability tradeoff. I left this comment here to remind us that a non obvious  merge is possible, if we decide that the performance gain is worth it.

meilisearch-core/src/update/documents_addition.rs
MarinPostma: exactly, i can then use this map to either add or remove documents

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,342 @@
...
+        };
+        Self(field_id, value)
+    }
+
+    fn from_str(
+        s: &str,
+        schema: &Schema,
+        attributes_for_faceting: &[FieldId],
+    ) -> Result<Self, FacetError> {
+        let mut split = s.splitn(2, ":");
```
Kerollmops, line 99:
```suggestion
        let mut split = s.splitn(2, ':');
```

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,342 @@
...
+        let key = split
+            .next()
```
Kerollmops, line 101:
```suggestion
            .next()
            .trim()
```

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,342 @@
...
+            return Err(FacetError::AttributeNotSet {
+                found: key.to_string(),
+                expected:  attributes_for_faceting
+                    .iter()
+                    .filter_map(|&id| schema.name(id))
+                    .map(str::to_string)
+                    .collect::<Vec<_>>() })
+        }
+        let value = split
+            .next()
```
Kerollmops, line 117:
```suggestion
            .next()
            .trim()
```

meilisearch-core/src/facets.rs
Kerollmops: I think we should better specify `splitn(3, _)` here and check if the 3rd doesn't exists `split.next().is_none()`.
This way we can detect syntax error like `["colors:blue:green"]`.

Kerollmops: There also is a little issue that we should note somewhere: what if the user write this `["colors:\"top:blue\""]`?
We will split on the second `:` and we must not do that.

Therefore you should not `splitn(3)` like I proposed in my comment aboce, but we should implement a better parser (using Pest) in a next PR, later, later.

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,342 @@
...
+            .ok_or_else(|| FacetError::InvalidFormat(s.to_string()))?;
+        // unquoting the string if need be:
+        let mut indices = value.char_indices();
+        let value =  match (indices.next(), indices.last()) {
+            (Some((s, '\'')), Some((e, '\''))) |
+            (Some((s, '\"')), Some((e, '\"'))) => value[s + 1..e].to_string(),
+            _ => value.to_string(),
+        };
```
Kerollmops, line 125: Like I said before we should use a better parser like pest, we can't handle syntx errors like `["colors:'blue\""]`.
Also we do not support quoting the field name.

meilisearch-core/src/facets.rs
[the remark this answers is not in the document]
Kerollmops: You should probably use it like:
```rust
let (id_bytes, value_bytes) = bytes.split_at(2);
let id = OwnedType::<BEU16>::bytes_decode(id_bytes)?;
let id = id.get();
```

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,342 @@
...
+    facet_map: &mut HashMap<FacetKey, Vec<DocumentId>>,
+    field_id: FieldId,
+    value: Value,
+    document_id: DocumentId,
+) -> Result<(), FacetError> {
+    let value = match value {
+        Value::String(s) => s,
+        Value::Object(_) | Value::Array(_) => return Err(FacetError::InvalidDocumentAttribute(document_id)),
+        v => v.to_string(),
+    };
```
Kerollmops, line 163: Ok so this part of the code shows that we support facetting on number by transforming them in strings, We should not support them at all and return an error.

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,342 @@
...
+    let key = FacetKey::new(field_id, value);
+    facet_map.entry(key).or_insert(Vec::new()).push(document_id);
```
Kerollmops, line 165:
```suggestion
    facet_map.entry(key).or_insert_with(Vec::new).push(document_id);
```

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,342 @@
...
+}
+
+pub fn facet_map_from_docids<'txn>(
+    rtxn: &'txn heed::RoTxn<MainT>,
+    index: &crate::Index,
+    document_ids: &[DocumentId],
+    attributes_for_facetting: &[FieldId],
+) -> Result<HashMap<FacetKey, Vec<DocumentId>>, Error> {
+    let mut facet_map = HashMap::new();
+
```
Kerollmops, line 176:
```suggestion
```

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,342 @@
...
+        Value::Object(_) | Value::Array(_) => return Err(FacetError::InvalidDocumentAttribute(document_id)),
+        v => v.to_string(),
+    };
+    let key = FacetKey::new(field_id, value);
+    facet_map.entry(key).or_insert(Vec::new()).push(document_id);
+    Ok(())
+}
+
+pub fn facet_map_from_docids<'txn>(
+    rtxn: &'txn heed::RoTxn<MainT>,
```
Kerollmops, line 170: When you don't bound any returned type to the lifetime of the transaction you can avoid giving it a name.
```suggestion
pub fn facet_map_from_docids(
    rtxn: &heed::RoTxn<MainT>,
```

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,342 @@
...
+    let mut facet_map = HashMap::new();
+
+    for document_id in document_ids {
+        for result in index
+            .documents_fields
+            .document_fields(rtxn, *document_id)?
+        {
+            let (field_id, bytes) = result?;
+            if attributes_for_facetting.contains(&field_id) {
+                match serde_json::from_slice::<Value>(bytes)? {
```
Kerollmops, line 184:
```suggestion
                match serde_json::from_slice(bytes)? {
```

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,342 @@
...
+        for (field_id, name) in &attributes_for_facetting {
+            if let Some(value) = document.get(*name) {
+                match value {
+                    Value::Array(values) => {
+                        for v in values {
+                            add_to_facet_map(&mut facet_map, *field_id, v.clone(), *id)?;
+                        }
+                    }
+                    v => add_to_facet_map(&mut facet_map, *field_id, v.clone(), *id)?,
+                };
```
Kerollmops, line 219:
```suggestion
                }
```

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,342 @@
...
+            FacetKey::new(id, "foo bar".to_string())
+        );
+        assert_eq!(
+            FacetKey::from_str("hello:'foo bar'", &schema, &facet_list).unwrap(),
+            FacetKey::new(id, "foo bar".to_string())
+        );
+        // weird case
+        assert_eq!(
+            FacetKey::from_str("hello:blabla:machin", &schema, &facet_list).unwrap(),
+            FacetKey::new(id, "blabla:machin".to_string())
```
Kerollmops, line 251: Ok, this is just what I talked about earlier, noted!

meilisearch-core/src/query_builder.rs
```diff
@@ -88,41 +71,74 @@ impl<'c, 'f, 'd> QueryBuilder<'c, 'f, 'd> {
...
+            Some(facets) => {
+                let mut ands = Vec::with_capacity(facets.len());
+                let mut ors = Vec::new();
+                for f in facets.deref() {
+                    match f {
+                        Either::Left(keys) => {
+                            ors.reserve(keys.len());
+                            for key in keys {
+                                let docids = self.index.facets.facet_document_ids(reader, &key)?.unwrap_or_default();
+                                ors.push(docids.into_owned());
```
Kerollmops, line 89: I don't understand why we owned it here, it is a Cow that is bounded by the lifetime of the transaction here.
```suggestion
                                ors.push(docids);
```

meilisearch-core/src/query_builder.rs
```diff
@@ -88,41 +71,74 @@ impl<'c, 'f, 'd> QueryBuilder<'c, 'f, 'd> {
     }
 
     pub fn query(
-        self,
+        mut self,
         reader: &heed::RoTxn<MainT>,
         query: &str,
         range: Range<usize>,
     ) -> MResult<(Vec<Document>, usize)> {
+        let facets_docids = match self.facets.take() {
```
Kerollmops, line 79:
```suggestion
        self,
        reader: &heed::RoTxn<MainT>,
        query: &str,
        range: Range<usize>,
    ) -> MResult<(Vec<Document>, usize)> {
        let facets_docids = match self.facets {
```

meilisearch-core/src/query_builder.rs
```diff
@@ -88,41 +71,74 @@ impl<'c, 'f, 'd> QueryBuilder<'c, 'f, 'd> {
...
+                                let docids = self.index.facets.facet_document_ids(reader, &key)?.unwrap_or_default();
+                                ors.push(docids.into_owned());
+                            }
+                            let sets: Vec<_> = ors.iter().map(|s| s.as_set()).collect();
+                            ands.push(sdset::multi::OpBuilder::from_vec(sets).union().into_set_buf());
+                            ors.clear();
+                        }
+                        Either::Right(key) =>{
+                            match self.index.facets.facet_document_ids(reader, &key)? {
+                                Some(docids) => ands.push(docids.into_owned()),
```
Kerollmops, line 97:
```suggestion
                                Some(docids) => ands.push(docids),
```

meilisearch-core/src/query_builder.rs
```diff
@@ -88,41 +71,74 @@ impl<'c, 'f, 'd> QueryBuilder<'c, 'f, 'd> {
...
+                                // no candidates for search, early return.
+                                None => return Ok((vec![], 0)),
+                            }
+                        }
+                    };
+                }
+                let ands: Vec<_> = ands.iter().map(sdset::SetBuf::as_ref).collect();
```
Kerollmops, line 104:
```suggestion
                let ands: Vec<_> = ands.iter().map(|s| s.as_set()).collect();
```

meilisearch-core/src/store/facets.rs
```diff
@@ -0,0 +1,54 @@
...
+        self.facets.get(reader, &facet_key)
+    }
+
+    /// updates the facets  store, revmoving the documents from the facets provided in the
+    /// `facet_map` argument
+    pub fn remove(&self, writer: &mut RwTxn<MainT>, facet_map: HashMap<FacetKey, Vec<DocumentId>>) -> ZResult<()> {
+        for (key, document_ids) in facet_map {
+            if let Some(old) = self.facets.get(writer, &key)? {
+                let to_remove = SetBuf::from_dirty(document_ids);
+                let old = Set::new_unchecked(old.as_ref());
```
Kerollmops, line 35:
```suggestion
```

meilisearch-core/src/store/mod.rs
```diff
@@ -440,6 +437,7 @@ pub fn open(
     let prefix_postings_lists_cache_name = prefix_postings_lists_cache_name(name);
     let updates_name = updates_name(name);
     let updates_results_name = updates_results_name(name);
+    let facets_name = facets_name(name);
```
Kerollmops, line 440: Please move this line just after `prefix_documents_cache`.

meilisearch-core/src/store/mod.rs
```diff
@@ -482,6 +480,10 @@ pub fn open(
         Some(updates_results) => updates_results,
         None => return Ok(None),
     };
+    let facets = match env.open_database(Some(&facets_name))? {
+        Some(facets) => facets,
+        None => return Ok(None),
+    };
```
Kerollmops, line 486: Please move these lines just after `prefix_documents_cache`.

meilisearch-core/src/store/mod.rs
```diff
@@ -495,6 +497,7 @@ pub fn open(
         updates: Updates { updates },
         updates_results: UpdatesResults { updates_results },
         updates_notifier,
+        facets: Facets { facets },
```
Kerollmops, line 500: Please move this line just after `prefix_documents_cache`.

meilisearch-core/src/update/documents_addition.rs
```diff
@@ -289,6 +245,11 @@ pub fn reindex_all_documents(writer: &mut heed::RwTxn<MainT>, index: &store::Ind
     let mut indexer = RawIndexer::new(stop_words);
     let mut ram_store = HashMap::new();
 
+    if let Some(ref attributes_for_facetting) = index.main.attributes_for_faceting(writer)? {
+        let facet_map = facets::facet_map_from_docids(writer, &index, &documents_ids_to_reindex, &attributes_for_facetting)?;
+        index.facets.add(writer, facet_map)?;
+    }
+    // ^-- merge? --v
```
Kerollmops, line 252: What this comment means again?

meilisearch-http/src/routes/search.rs
Kerollmops, line 33: As I understand this line a user cannot send a valid JSON array? He must send a string containing the JSON array itself? I think we must look into serde and probably create our own enum type to be able to accept real JSON arrays.

meilisearch-schema/src/lib.rs
Kerollmops, line 40:
```suggestion
#[derive(Debug, Copy, Clone, Default, PartialOrd, Ord, PartialEq, Eq, Hash)]
#[derive(Serialize, Deserialize)]
#[derive(AsBytes, FromBytes)]
```

meilisearch-core/src/facets.rs
MarinPostma: yes I thought of that, and there is this other case that's breakign too: `['\"foo:bar\":hello]` that I don;t take into account at all

meilisearch-core/src/query_builder.rs
MarinPostma: yes right, but then we gottta deref on line 92:
```rust
let sets: Vec<_> = ors.iter().map(Cow::deref).collect();
```

meilisearch-http/src/routes/search.rs
MarinPostma: I'm not sure to understand what you mean, but the user sends an array in an url-encoded format, that upon decoding happen to be JSON-like.  If we use serde to parse it, we'll have a "syntactically" correct array, but when I parse it I actually do a lot more than what serde can do, I'd have to go through it once more to make parse keys, and make sure the array is correct. I'm not sure to see how this would be an improvement over what we are doing right now

meilisearch-core/src/update/documents_addition.rs
MarinPostma: what facet map does could be done inside the loop under the comment, and save us a few db reads. this would be done at the price of less readability, and would make it harder to move this logic out if we wanted it. This comment sole purpose is to tell us that this optimization is possible, is we want it, as it is not trivial.

meilisearch-core/src/facets.rs
```diff
@@ -158,30 +161,31 @@ pub fn add_to_facet_map(
 ) -> Result<(), FacetError> {
     let value = match value {
         Value::String(s) => s,
-        Value::Object(_) | Value::Array(_) => return Err(FacetError::InvalidDocumentAttribute(document_id)),
+        Value::Object(_)
+        | Value::Array(_)
+        | Value::Number(_) => return Err(FacetError::InvalidDocumentAttribute(document_id)),
         v => v.to_string(),
     };
```
Kerollmops: I think we must accept only strings and array or strings? Here you are converting Null and Booleans to strings, but that's invalid.

meilisearch-core/src/facets.rs
```diff
@@ -143,10 +145,11 @@ impl<'a> heed::BytesDecode<'a> for FacetKey {
     type DItem = FacetKey;
 
     fn bytes_decode(bytes: &'a [u8]) -> Option<Self::DItem> {
-        let field_id_u16 = u16::from_be_bytes([bytes[0], bytes[1]]);
-        let field_id = FieldId(field_id_u16);
-        let string = Str::bytes_decode(&bytes[2..])?;
-        Some(FacetKey(field_id, string.to_string()))
+        let (id_bytes, value_bytes) = bytes.split_at(2);
+        let id = OwnedType::<BEU16>::bytes_decode(id_bytes)?;
```
Kerollmops, line 149: Could you use the same codec in the encode part please?

meilisearch-core/src/store/mod.rs
```diff
@@ -45,7 +45,7 @@ use crate::settings::SettingsUpdate;
 use crate::{query_builder::QueryBuilder, update, DocIndex, DocumentId, Error, MResult};
 
 type BEU64 = zerocopy::U64<byteorder::BigEndian>;
-type BEU16 = zerocopy::U16<byteorder::BigEndian>;
+pub type BEU16 = zerocopy::U16<byteorder::BigEndian>;
```
Kerollmops, line 48: This isn't necessary as this is used in the same crate and the `type` is at the top of the lib.

meilisearch-core/src/facets.rs
MarinPostma: I think filtering on a bool is a very common and legitimate thing to do, we should allow that. And we probably should not error out on null, is is very common to have null values where something else is expected. maybe ignoring them is better suited if we don't want them around

meilisearch-core/src/store/mod.rs
MarinPostma:
```
rror[E0603]: type alias `BEU16` is private                                                                                                                                   
  --> meilisearch-core/src/facets.rs:17:19                                                                                                                                    
   |                                                                                                                                                                          
17 | use crate::store::BEU16;                                                                                                                                                 
   |                   ^^^^^ this type alias is private                                                                                                                       
   |                                                                                                                                                                          
note: the type alias `BEU16` is defined here                                                                                                                                  
  --> meilisearch-core/src/store/mod.rs:48:1                                                                                                                                  
   |                                                                                                                                                                          
48 | type BEU16 = zerocopy::U16<byteorder::BigEndian>;                                                                                                                        
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^     
```

Kerollmops: Ho, sorry this is because is is declared inside of store.rs and not lib.rs, my bad!

meilisearch-core/src/facets.rs
Kerollmops: Yeah,, good point, we should ignore nulls but I maintain that we must only support strings and arrays of string, this PR will just support those two types only. We will support numbers including booleans in a next PR.

MarinPostma:
don't merge !

MarinPostma:
regarding document reindexation (meilisearch-core/src/update/documents_addition.rs:250), what facet map does could be done inside the loop under the comment, and save us a few db reads. this would be done at the price of less readability, and would make it harder to move this logic out if we wanted it. This comment sole purpose is to tell us that this optimization is possible, is we want it, as it is not trivial.

meilisearch-core/src/facets.rs
```diff
@@ -161,10 +171,9 @@ pub fn add_to_facet_map(
 ) -> Result<(), FacetError> {
     let value = match value {
         Value::String(s) => s,
-        Value::Object(_)
-        | Value::Array(_)
-        | Value::Number(_) => return Err(FacetError::InvalidDocumentAttribute(document_id)),
-        v => v.to_string(),
+        // ignore null
+        Value::Null => return Ok(()),
+        _ => return Err(FacetError::InvalidDocumentAttribute(document_id)),
```
Kerollmops, line 176: Would be great if you show the invalid field and the fact that we only support strings and arrays of strings only.

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,355 @@
...
+                            }
+                            filter.push(Either::Left(inner));
+                        }
+                        bad_value => return Err(FacetError::UnexpectedToken { expected: &["String", "Array"], found: bad_value.to_string() }),
+                    }
+                }
+                return Ok(Self(filter));
+            }
+            bad_value => Err(FacetError::UnexpectedToken { expected: &["Array"], found: bad_value.to_string() }),
+        }
```
Kerollmops, line 76: Could you create a contructor for the unexpected token, this will reduce the amount of code here.

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    pub fn key(&self) -> FieldId {
+        self.0
+    }
+
+    pub fn value(&self) -> &str {
+        &self.1
+    }
+
+    // TODO improve parser
```
Kerollmops, line 101: Could you fill a new comment on the facet tracking issue please?

meilisearch-core/src/error.rs
```diff
@@ -176,6 +176,16 @@ pub enum FacetError {
...
 }
 
+impl FacetError {
+    pub fn unexpected_token(expected: &'static [&'static str], found: String) -> FacetError {
+        FacetError::UnexpectedToken{ expected, found }
+    }
+
+    pub fn attribute_not_set(expected: Vec<String>, found: String) -> FacetError {
+        FacetError::AttributeNotSet{ expected, found }
+    }
```
Kerollmops, line 186: We can reduce the line length of the other file by avoiding calling `to_string` and transform it there.

```suggestion
    pub fn unexpected_token(expected: &'static [&'static str], found: impl ToString) -> FacetError {
        FacetError::UnexpectedToken{ expected, found: found.to_string() }
    }

    pub fn attribute_not_set(expected: Vec<String>, found: impl ToString) -> FacetError {
        FacetError::AttributeNotSet{ expected, foundL found.to_string() }
    }
```

meilisearch-core/src/facets.rs
```diff
@@ -0,0 +1,353 @@
...
+                                }
+                            }
+                            filter.push(Either::Left(inner));
+                        }
+                        bad_value => return Err(FacetError::unexpected_token(&["Array", "String"], bad_value)),
+                    }
+                }
+                return Ok(Self(filter));
+            }
+            bad_value => Err(FacetError::unexpected_token(&["String"], bad_value)),
```
Kerollmops, line 73:
```suggestion
            bad_value => Err(FacetError::unexpected_token(&["Array"], bad_value)),
```

meilisearch-http/Cargo.toml
```diff
@@ -26,6 +26,7 @@ actix-web = "2"
...
```
Kerollmops, line 29: Are you using this dependency in `meilisearch-http`?

PR #635, opened 2020-04-24
Bug fix/highlight longest area

### LINKED ISSUE #265: Make the highlighting a bit better again
Currently when the same area must be highlighted we only choose to use the first one but it would be better to use the biggest area found.

https://github.com/meilisearch/MeiliDB/blob/026464b2e493f3bdab1b43f5fd9dc2f0f7032f58/meilidb-http/src/helpers/meilidb.rs#L492-L525

lironhl:
Fixes #265,

### The Problem
Let's assume we have one document in the index that looks like this
`{
    "title": "Ice"
}`

When searching "Ice Ic", `fn calculate_highlights()` will return `<em>Ic</em>e` for the `title` attribute.

### About the solution
1. The solution assumes `fn calculate_highlights()` gets sorted Vec\<MatchPosition\> (the sorting takes place beforehand at `fn calculate_matches()`), I created additional test because of that assumption.
2. The solution uses `group_by` from `Itertools`.

I'm new to open source, if this PR needs improvements let me know please.

### DIFF
--- a/meilisearch-http/src/helpers/meilisearch.rs
+++ b/meilisearch-http/src/helpers/meilisearch.rs
@@ -17,6 +17,7 @@ use meilisearch_tokenizer::is_cjk;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use siphasher::sip::SipHasher;
+use slice_group_by::GroupBy;
 
 #[derive(Debug)]
 pub enum Error {
@@ -526,7 +527,12 @@ fn calculate_highlights(
                 let value: Vec<_> = value.chars().collect();
                 let mut highlighted_value = String::new();
                 let mut index = 0;
-                for m in matches {
+
+                let longest_matches = matches
+                    .linear_group_by_key(|m| m.start)
+                    .map(|group| group.last().unwrap());
+
+                for m in longest_matches {
                     if m.start >= index {
                         let before = value.get(index..m.start);
                         let highlighted = value.get(m.start..(m.start + m.length));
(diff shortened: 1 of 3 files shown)


meilisearch-http/src/helpers/meilisearch.rs
```diff
@@ -526,7 +527,15 @@ fn calculate_highlights(
...
-                for m in matches {
+
+                let longest_matches = matches
+                    .iter()
+                    .group_by(|m| m.start)
+                    .into_iter()
+                    .map(|(_key, group)| group.last().unwrap())
+                    .collect::<Vec<_>>();
+
+                for m in longest_matches {
```
Kerollmops, line 538: I think you can avoid collecting this into a vec and directly iterate trought them.

```suggestion
                    .map(|(_key, group)| group.last().unwrap());

                for m in longest_matches {
```

lironhl:
@Kerollmops Thanks! I'll look into `slice_group_by`.

lironhl:
I forgot to mention a certain behavior you should be aware of, when [`matches`](https://docs.meilisearch.com/guides/advanced_guides/search_parameters.html#matches) option in search is on.
It returns all the matched areas of a certain word, and not the longest area.
Should it stay that way?

Kerollmops:
Yes, I think that could stay that way, we will eventually remove the matches at some point.
Thank you for your PR, would it be possible that your squash it into two commits, one for the change and one for the tests, please?

lironhl:
@Kerollmops Done.
Why is `matches` available now?
Why would you remove it?

Kerollmops:
We want to remove the matches to reduce the index size and only compute them when we need to return the documents.

Kerollmops:
Thank you for this PR, I merge it now :)

PR #645, opened 2020-04-28
Remove call to map_size

### LINKED ISSUE #643: Error "There is not enough space on the disk" on Windows x64
## Error
I'm getting the following error when running MeiliSearch on a Windows Server 2019 x64:
```
Zlmdb(Io(Os { code: 112, kind: Other, message: "There is not enough space on the disk." }))
```

I traced the source of the error to the following call stack:
```
1. heed::EnvOpenOptions::open(path)
2. meilisearch_core::Database::open_or_create(path)
```

Is this expected behaviour, should the server have 200 gigs of storage space to be able to run MeiliSearch?

## Platform
OS: Windows Server 2019
Crate version: master branch
Note: I'm using meilisearch as an embedded database by referencing the meilisearch-core crate directly from github.

## Workaround
I managed to circumvent the error by commenting out the call to `map_size` on the `EnvOpenOptions` in `database.rs`.
https://github.com/meilisearch/MeiliSearch/blob/899559a060aff9051cfc34b187b7a74f8c71c18d/meilisearch-core/src/database.rs#L144-L153

Wazner:
Fixes #643 

This removes the call to `map_size`, which seemingly makes the database autogrow instead of preallocate the 2x 100Gb on Windows Server machines.

This change has _not_ been tested on large scale databases or with production usage patterns.

### DIFF
--- a/meilisearch-core/src/database.rs
+++ b/meilisearch-core/src/database.rs
@@ -142,13 +142,11 @@ impl Database {
 
         fs::create_dir_all(&main_path)?;
         let env = heed::EnvOpenOptions::new()
-            .map_size(100 * 1024 * 1024 * 1024) // 100GB
             .max_dbs(3000)
             .open(main_path)?;
 
         fs::create_dir_all(&update_path)?;
         let update_env = heed::EnvOpenOptions::new()
-            .map_size(100 * 1024 * 1024 * 1024) // 100GB
             .max_dbs(3000)
             .open(update_path)?;
 


Kerollmops:
This change cannot be made as LMDB doesn't support auto-growth. This call to `map_size` is mandatory [as it determines the maximum size the database can grow to](http://www.lmdb.tech/doc/group__mdb.html#gaa2506ec8dab3d969b0e609cd82e619e5).

Another solution to fix #643 would be to introduce two options to the engine, one to specify the size in bytes for the main database and one for the update database, this way it could work on Windows too!

If you want to introduce those two flags please read the documentation link carefully and do it in another PR :)

BTW thank you for your interrest in MeiliSearch!

PR #646, opened 2020-04-28
Add support for configuring lmdb map size

### LINKED ISSUE #643: Error "There is not enough space on the disk" on Windows x64

Wazner:
Fixes #643 

This adds options for configuring the map size via command-line and environment variables.
Currently it takes a plain numeric value.
This is a breaking change seeing as I changed the signature of an existing function. Although the crate is not published on crates.io, it could affect people depending on the git repository directly.
If this is a problem, a new method `open_or_create2` could be added, which can be removed and replace the original in the next major version.

If desired I can add human-friendly parsing with https://crates.io/crates/human-size so you can configure "1 GiB" instead of "1073741824".

### DIFF
--- a/meilisearch-core/src/database.rs
+++ b/meilisearch-core/src/database.rs
@@ -135,20 +135,34 @@ fn update_awaiter(
     Ok(())
 }
 
+pub struct DatabaseOptions {
+    pub main_map_size: usize,
+    pub update_map_size: usize
+}
+
+impl Default for DatabaseOptions {
+    fn default() -> DatabaseOptions {
+        DatabaseOptions {
+            main_map_size: 100 * 1024 * 1024 * 1024, // 100GB
+            update_map_size: 100 * 1024 * 1024 * 1024 // 100GB
+        }
+    }
+}
+
 impl Database {
-    pub fn open_or_create(path: impl AsRef<Path>) -> MResult<Database> {
+    pub fn open_or_create(path: impl AsRef<Path>, options: DatabaseOptions) -> MResult<Database> {
         let main_path = path.as_ref().join("main");
         let update_path = path.as_ref().join("update");
 
         fs::create_dir_all(&main_path)?;
         let env = heed::EnvOpenOptions::new()
-            .map_size(100 * 1024 * 1024 * 1024) // 100GB
+            .map_size(options.main_map_size)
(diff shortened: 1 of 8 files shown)


meilisearch-core/src/database.rs
```diff
@@ -135,20 +135,34 @@ fn update_awaiter(
     Ok(())
 }
 
+pub struct DatabaseOptions {
+    pub main_map_size: usize,
+    pub update_map_size: usize
+}
```
Kerollmops, line 141: Would it be possible to implement structopt on this struct, declare the `default_values` and `env` variable values here, then add one field to the MeiliSearch `Opt` struct with [the flatten serde atribute](https://serde.rs/field-attrs.html#flatten)?

Wazner: That is possible. But that would mean `StructOpt` would leak into meilisearch-core. Currently all command-line arguments and environment variables are neatly written in `option.rs` of `meilisearch-http`.

Kerollmops: Yeah, you are probably right, that's better this way, thank you for pointing this!

Kerollmops:
Thank you, I hope this fixes the Windows related problems, did you try it on your own Windows machine?

Wazner:
Yes. It works on my Windows server, albeit in a test environment, have not tried it in production yet.

PR #660, opened 2020-05-07
Update release process for stable releases

curquiza:
- Fix: the script `download-latest.sh` checks what is the latest tag based on release tags only (and not on all the tags as it was done before)
- The CI to publish the latest docker image runs only for releases (not for pre-releases, not for draft-releases and not for a simple tag) and checks the release is indeed the latest
  Ex: previous tag -> `v0.10.1`
        new tag -> `v0.8.12`
        The new tag should not be the latest: the CI will not run for the release `v0.8.2`
- The CI to publish to Brew and APT runs only for releases (not for pre-releases, not for draft-releases and not for a simple tag)

I'm not sure I put the script `is-latest-release.sh` at the right place.
Maybe we can create a `scripts/` folder at the root of MeiliSearch repository to put all the scripts, included `download-latest.sh`.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,4 +1,4 @@
-# GitHub actions workflow for MeiliDB
+# GitHub Actions Workflow for MeiliSearch
 
 > **Note:**
 
@@ -6,12 +6,14 @@
 
 ## Workflow
 
-- On each pull request, we are triggering `cargo test`.
-- On each tag, we are building:
-    - the tagged docker image
+- On each pull request, we trigger `cargo test`.
+- On each tag, we build:
+    - the tagged Docker image and publish it to Docker Hub
     - the binaries for MacOS, Ubuntu, and Windows
-    - the debian package
-- On each stable release, we are build the latest docker image.
+    - the Debian package
+- On each stable release (`v*.*.*` tag):
+    - we build the `latest` Docker image and publish it to Docker Hub
+    - we publish the binary to Hombrew and Gemfury
 
 ## Problems
 
--- a/.github/workflows/publish-docker-latest.yml
+++ b/.github/workflows/publish-docker-latest.yml
@@ -1,8 +1,7 @@
 ---
 on:
-  push:
-    tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
+  release:
+    types: [published]
 
 name: Publish latest image to Docker Hub
 
@@ -10,8 +9,12 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
+      - name: Check if current release is latest
+        run: echo "##[set-output name=is_latest;]$(sh .github/is-latest-release.sh)"
+        id: release
       - name: Publish to Registry
+        if: steps.release.outputs.is_latest == 'true'
         uses: elgohr/Publish-Docker-Github-Action@master
         with:
           name: getmeili/meilisearch
--- a/.github/is-latest-release.sh
+++ b/.github/is-latest-release.sh
@@ -1,12 +1,15 @@
 #!/bin/sh
 
-# COLORS
-RED='\033[31m'
-GREEN='\033[32m'
-DEFAULT='\033[0m'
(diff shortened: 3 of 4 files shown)


curquiza:
@Kerollmops Should I update the `CHANGELOG.md`, or you don't want this fix to appear in the changelogs?

.github/workflows/publish-docker-latest.yml
```diff
@@ -1,17 +1,20 @@
...
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
+      - name: Check if current release is latest
+        run: echo "##[set-output name=is_latest;]$(sh .github/is-latest-release.sh)"
+        id: release
       - name: Publish to Registry
+        if: steps.release.outputs.is_latest == 'true'
```
Kerollmops, line 17: I'm not sure `steps.release.outputs.is_latest` is what we want here, this is not based on the semver format.

download-latest.sh
```diff
@@ -82,45 +45,11 @@ failure_usage() {
...
-    else
-        semverLT $tag $latest
-        if [ $? -eq 1 ]; then
-            latest="$tag"
-        fi
-    fi
-done
+# MAIN
+os="$(get_os)"
+latest="$(sh .github/get-latest-stable-release.sh)"
```
Kerollmops, line 50: However I don't understanf why you split the scripts into two parts, will that always works using the curl command?
This script depends on a script located at a path, but the user doesn't have this path...

.github/workflows/README.md
```diff
@@ -1,17 +1,19 @@
...
 > - We do not use [cache](https://github.com/actions/cache) yet but we could use it to speed up CI
 
 ## Workflow
 
-- On each pull request, we are triggering `cargo test`.
-- On each tag, we are building:
-    - the tagged docker image
+- On each pull request, we trigger `cargo test`.
+- On each tag, we build:
+    - the tagged Docker image
```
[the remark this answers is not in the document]
curquiza: No, we publish it, that's indeed not clear! I update that!

.github/workflows/publish-docker-latest.yml
curquiza: Yes, the script `is-latest-release.sh` check about the semver format. Why did you think it's not?

download-latest.sh
curquiza: I thought it would work because the user does:
```bash
$ curl -L https://install.meilisearch.com | sh
```
and `https://install.meilisearch.com` is a redirection to `https://raw.githubusercontent.com/meilisearch/MeiliSearch/master/download-latest.sh` (so to the GitHub repository where there are the other scripts)

curquiza: And I split it because I use the `get-latest-stable-release.sh` in another script `is-latest-release.sh` used in GHA. I did not want to duplicate the code.

Kerollmops: I understand why you split it, but the line I commented does not work now, this will try to run `sh` with a script located at `.github/get-latest-stable-release.sh` which doesn't exist at all for the user. The script is no more self contained now.

.github/workflows/publish-docker-latest.yml
Kerollmops: I don't understand, where did you wrote this variable `steps.release.outputs.is_latest`?

Kerollmops: HO! this `echo "##[set-output name=is_latest;]$(sh .github/is-latest-release.sh)"` is the way you write into some variables in the github actions!? OK nice!

curquiza: I set the `is_latest` variable here:
```yml
        run: echo "##[set-output name=is_latest;]$(sh .github/is-latest-release.sh)"
        id: release
```

download-latest.sh
curquiza: Shit, so we have to duplicate the code?

Kerollmops: I think this is the best thing that we can do here, sadly.

curquiza: Snif! Changing that!

PR #664, opened 2020-05-07
add sentry probe

qdequele:


meilisearch-http/src/main.rs
```diff
@@ -27,6 +36,12 @@ async fn main() -> Result<(), MainError> {
                         .into(),
                 );
             }
+
+            #[cfg(feature = "sentry")]
+            if !opt.no_analytics {
+                sentry::integrations::panic::register_panic_handler();
+                sentry::integrations::env_logger::init(None, Default::default());
+            }
```
Kerollmops, line 44: `cfgs` works on if statements? Is it real?

qdequele: I ran it with the `--no-default-features` and without, and in both cases, it works as expected

Kerollmops: Hum, waht version of rust are you using?

qdequele: 1.43.0

PR #681, opened 2020-05-14
enables debug without sentry

MarinPostma:
enables building meili in debug without sentry by default. 
- workaround tokio bug
- probably better stats if debug build are not taken into account by sentry


Kerollmops:
We can enable sentry on debug again if we bump the sentry version to `0.18.1`.
https://github.com/getsentry/sentry-rust/releases/tag/0.18.1

PR #690, opened 2020-05-17
bump sentry

MarinPostma:
partially reverts #681 

made me thinking, do we want sentry on debug build anyway?


Kerollmops:
I think it would be better to remove sentry from the debug build as you said, because when we experiment we can make the server crash and we don't want sentry to report those errors.

What do you think @qdequele?

MarinPostma:
so, closing?

Kerollmops:
No, edit your commit and just keep the part where you updated the depenciy, please :)

PR #694, opened 2020-05-18
Take achitecture into account in download-latest

curquiza:
- Get architecture and download the right binary
- Before: it only accepted Debian and Ubuntu distribution. Now: since the Linux binary works on CentOS and Fedora, I removed the check on the distribution name, and accept all the Linux.

More details:
- Of course, it downloads nothing on FreeBSD.
- I wanted to test on my raspberry, but the archi is `armv7l` and not `armv8`.

### DIFF
--- a/download-latest.sh
+++ b/download-latest.sh
@@ -90,12 +90,12 @@ get_latest() {
             fi
             i=1
         elif [ $i -eq 1 ]; then # Checking draft boolean
-            if [ "$release_info" == "true" ]; then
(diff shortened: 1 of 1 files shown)


Kerollmops:
So, you tested this om the scaleway machine I bring to you?

curquiza:
Yes I did! 🙂 It works 👌

Kerollmops:
It seems good, thank you curqui :)

PR #699, opened 2020-05-20
Added tini process manager and entrypoint decl.

mattjtodd:
The main process in the Dockerfile `CMD` and images generated from it is not passing through linux signals correctly so shutdown is not correctly instigated when the container is stopped.

This introduces the `tini` init process manager to the `entrypoint` decl. which addresses this. 

More info about this can be found here.

https://github.com/krallin/tini

It would also be possible to use a feature of the docker runtime command to introduce this behaviour, but this would be less portable, for example:

```
docker run -it --rm --init \
    -p 7700:7700 \
    -v $(pwd)/data.ms:/data.ms \
    getmeili/meilisearch
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,11 +18,12 @@ RUN     $HOME/.cargo/bin/cargo build --release
 # Run
 FROM    alpine:3.10
 
-RUN     apk update --quiet
-RUN     apk add libgcc
+RUN     apk add -q --no-cache libgcc tini
 
 COPY    --from=compiler /meilisearch/target/release/meilisearch .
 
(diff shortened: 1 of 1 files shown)


Kerollmops:
Hey @mattjtodd,

Sorry for the late response, I appreciate your interrest in MeiliSearch and would like to merge this PR. Would it be possible you rebase it on master, this way it will be up-to-date and the CI will also test the Docker file generation?

mattjtodd:
@Kerollmops rebased and pushed 😄

Kerollmops:
Thank you, very much for your interrest in MeiliSearch :)

PR #700, opened 2020-05-21
Add openssl-dev pkg as dependencies in Dockerfile

tpayet:
Since Actix or Sentry the Docker image can not be built.

Should I add in the GitHub actions a test wether we can build or not the Docker image when anyone makes a pull request?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ FROM    alpine:3.10 AS compiler
 RUN     apk update --quiet
 RUN     apk add curl
 RUN     apk add build-base
+RUN     apk add openssl-dev
 
 RUN     curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 


Kerollmops:
> Should I add in the GitHub actions a test wether we can build or not the Docker image when anyone makes a pull request?

I am not, sure, it takes many time to build, we didn't released anything with actix for now and we would have found ot that the Docker image couldn't be built when we were trying to release, so. It seems ok.

Could you add a line about actix in the Changelog file please?
Something in the line of:
```
- Change the HTTP framework, moving from tide to actix-web #601
```

qdequele:
I have a fix to remove openssl from all the package by tuning sentry features

Kerollmops:
Yeah, absolutelly but what is the downside of this sentry tuning? If all sentry requests becomes unsafe: non-ssl secured, I don't think that this is good. Or is it a feature that makes sentry use rustls?

qdequele:
Yes, in fact, it already uses rustls but we haven't removed the feature that includes also openssl on request

tpayet:
Yep, I dont mind adding openssl dependency, I was afraid because the image was getting bigger but apparently it's because of the whole sentry feature, not just openssl ^^

On a different subject @Kerollmops, I can see that the test that are launched on each pull request take up to 18 minutes, while building the docker image is maximum 14min.
I think the different GitHub actions are launched in parallel so that won't slow down the time needed for each PR.

Kerollmops:
@qdequele, for go on and remove this openssl dependency :)

@thomas, ok for adding a new test to the CI, waiting for your PR :)

tpayet:
I will update the changelog on #705

PR #706, opened 2020-05-22
Bump the fst crate version to 0.4

Kerollmops:
This is another step towards making MeiliSearch easier to use and faster.

### DIFF
--- a/meilisearch-core/src/query_builder.rs
+++ b/meilisearch-core/src/query_builder.rs
@@ -186,7 +186,7 @@ mod tests {
     use std::collections::{BTreeSet, HashMap};
     use std::iter::FromIterator;
 
-    use fst::{IntoStreamer, Set};
+    use fst::IntoStreamer;
     use meilisearch_schema::IndexedPos;
     use sdset::SetBuf;
     use tempfile::TempDir;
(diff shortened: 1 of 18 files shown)


Kerollmops:
I have open a PR on the fst crate which introduces the `map_data` method on the fst types themselves, this ay we can get rid of the `FstExt` trait I introduced, if it get merged obviously ;)

https://github.com/BurntSushi/fst/pull/108

meilisearch-core/src/query_builder.rs
```diff
@@ -283,10 +279,7 @@ mod tests {
                 .put_synonyms(&mut writer, word.as_bytes(), &new_alternatives)
                 .unwrap();
 
-            let synonyms = match self.index.main.synonyms_fst(&writer).unwrap() {
-                Some(synonyms) => synonyms,
-                None => fst::Set::default(),
-            };
+            let synonyms = self.index.main.synonyms_fst(&writer).unwrap();
```
MarinPostma, line 282: why are those unwrap safe?

Kerollmops: Yeah, this is in the tests module :)

PR #714, opened 2020-05-26
fix null facets in response

MarinPostma:
remose "facet: null" in response for searches without facets


Kerollmops:
Thank you, merging!

PR #723, opened 2020-05-27
Bitpack the postings lists

Kerollmops:
This PR will bring compaction to the postings lists of the engine, this way we can make MeiliSearch less disk space hungry!


Kerollmops:
Closing because it is not more relavnt (we will use [RoaringBitmap](https://docs.rs/roaring)).

PR #732, opened 2020-05-29
Allow users to input an API Key to search into private data

Kerollmops:
This PR adds a field where the user can input the private or master key to allow the dashboard to access to private MeiliSearch instances. The API key is saved in the localStorage for more convenience.

Capture d’écran 2020-05-29 à 10 49 42


Kerollmops:
> (I will not give my opinion for the design because pretty sure you asked Quentin for that)

I didn't asked @qdequele for the design, as this is not the final dashboard anyway, and I don't really think of something better.

> But if I delete the master key and I retype it (without reloading the page) it adds again the same indexes.

Yeah, this is a known "bug", the indexes list is refreshed when the call to `/indexes` returns a valid response. We don't have to worry about this because this is not a securoty issue.

curquiza:
> I didn't asked @qdequele for the design

Ok, so the design is ok for me, but not sure I'm the best here to judge that 😂 

> Ho! You mean it append the index names! Ok, will fix that, sorry.

Yes, that's what I meant! Thanks! 🙂

Kerollmops:
The design has just been validated IRL by @qdequele!

Issue #727, opened 2020-05-27
Pass API KEY in query parameters instead of header?


jiminy-billy-bob:
Currently the API KEY must be passed in the request header.
This seems the logical way to do it. But there might be some performance issue with this.

When sending requests to MeiliSearch from the browser, a preflight `OPTIONS` request is sent before the actual `GET` request, because of the presence of the `x-meili-api-key` header.
This means that two requests are sent to the server, the second one waiting for the first one to come back. This essentially double the time it takes between triggering the search request and getting the results.
I have a 150-200ms ping between my machine and my MeiliSearch server, the search results actually arrive 300-400ms later! The difference is significant, especially for a search-as-you-type experience.

The solution would be to make [simple requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Simple_requests), which means no custom headers. The API key would be passed as a query parameter.

What do you guys think?

I'm no HTTP expert, I randomly stumbled upon this while struggling with some CORS issues, so I might be interpreting things the wrong way. Please correct me if I'm mistaken.

MarinPostma:
I'm definitely not an HTTP expert either, but if I understand right what you say, using the basic authorization header would solved this problem?

```
 'authorization: Basic my_key'
```

I remember that we had discussions about this, but I don't remember what we said about this. It's an interesting topic to investigate in any case!

jiminy-billy-bob:
> using the basic authorization header would solved this problem?

No, it looks like it would not solve this problem.
The "Authorization" header is not in [this list](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Simple_requests). It is even explicitly said a bit lower on the same page:
> However, if the request is one that triggers a preflight due to the presence of the Authorization header in the request, ...

MarinPostma:
Thanks for that information, we're discussing a solution.

MarinPostma:
closing as #735 solves the problem

PR #735, opened 2020-05-29
Post search route

MarinPostma:
create a post route for search queries, to prevent preflight queries due to the presence of a x-meili-api-key in the header.

thanks to @jiminy-billy-bob for pointing this out in #727 

bellow is a graph of the improvement in search time when using an API key thanks to this:
image

zones where the 2 are equal is when backspacing in the query bar and the browser rely on cache. Y-coordinates in ms

https://github.com/MarinPostma/meilisearch-network-bench


jiminy-billy-bob:
Just to make sure I understand:

- Using a POST request instead of GET prevents the browser from sending a preflight request?
- Or now we need to send the API Key as part of the POST request body instead of a header?

MarinPostma:
so the trick is you can set a TTL for the preflight check (eg 10 min for chrome and 24h for firefox) as long as the url remain the same. This is the case for post but not for get, so we can keep authentication in the header and cache the OPTION request for some time when using POST

jiminy-billy-bob:
Interesting! So there will be an OPTION request sent out only every 10 minutes? That sounds reasonable.

MarinPostma:
yes :)

Kerollmops:
Don't forget to create a PR to the documentation repository to add this to the doc, please link this PR in the one on the doc.

PR #752, opened 2020-06-04
new: Updated sysinfo crate in meilisearch-http/Cargo.toml. Fixes #740

### LINKED ISSUE #740: Build fails on FreeBSD/ HardenedBSD
```
    Updating crates.io index
    Updating git repository `https://github.com/MarinPostma/pest.git`
thread 'main' panicked at 'byte index 2 is not a char boundary; it is inside 'ϗ' (bytes 1..3) of ` tp`', src/libcore/str/mod.rs:2219:
```

dmilith:


Kerollmops:
Please do a `cargo update` first to update the lock file, could you squash that in the same commit ;)

PR #791, opened 2020-06-22
Create tests for error codes

### LINKED ISSUE #836: Bad code error when trying to infer a primary-key
- I create an index with no primary key set
- I try to add documents with no inferrable attributes
- MeiliSearch API returns:
```json
{
    "message": "Could not infer a primary key",
    "errorCode": "bad_request",
    "errorLink": "https://docs.meilisearch.com/error/bad_request"
}
```
According to the [future documentation](https://github.com/meilisearch/documentation/pull/421/files), the `errorCode` should be `missing_primary_key` instead of `bad_request` 🙂

- [x] Fix in the code
- [x] Add tests

Tested on `master` from this commit https://github.com/meilisearch/MeiliSearch/commit/2f6c55ef789818263ba999fc950ac4a3bd16ef82

MarinPostma:
- create tests for error codes
-  fix primary key error that returned internal error instead of the correct error
- bits of documentation for error
- change a bunch of error type, for better accuracy, @curquiza, @eskombro, @bidoubiwa  you may want to take a look at `meilisearch-error/src/lib.rs`
- fix #836

### DIFF
--- a/meilisearch-error/src/lib.rs
+++ b/meilisearch-error/src/lib.rs
@@ -102,7 +102,7 @@ impl Code {
             PrimaryKeyAlreadyPresent => ErrCode::invalid("primary_key_already_present", StatusCode::BAD_REQUEST),
 
             // invalid document
-            MaxFieldsLimitExceeded => ErrCode::invalid("max_field_limit_exceeded", StatusCode::BAD_REQUEST),
+            MaxFieldsLimitExceeded => ErrCode::invalid("max_fields_limit_exceeded", StatusCode::BAD_REQUEST),
             MissingDocumentId => ErrCode::invalid("missing_document_id", StatusCode::BAD_REQUEST),
 
             // error related to facets
--- a/meilisearch-error/src/lib.rs
+++ b/meilisearch-error/src/lib.rs
@@ -87,10 +87,11 @@ impl Code {
         match self {
             // index related errors
             // create index is thrown on internal error while creating an index.
-            CreateIndex => ErrCode::invalid("index_creation_failed", StatusCode::BAD_REQUEST),
+            CreateIndex => ErrCode::internal("index_creation_failed", StatusCode::BAD_REQUEST),
             IndexAlreadyExists => ErrCode::invalid("index_already_exists", StatusCode::BAD_REQUEST),
             // thrown when requesting an unexisting index
-            IndexNotFound => ErrCode::invalid("index_not_found", StatusCode::NOT_FOUND), InvalidIndexUid => ErrCode::invalid("invalid_index_uid", StatusCode::BAD_REQUEST),
+            IndexNotFound => ErrCode::invalid("index_not_found", StatusCode::NOT_FOUND),
+            InvalidIndexUid => ErrCode::invalid("invalid_index_uid", StatusCode::BAD_REQUEST),
             OpenIndex => ErrCode::internal("index_not_accessible", StatusCode::INTERNAL_SERVER_ERROR),
 
             // invalid state error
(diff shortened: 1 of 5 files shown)


meilisearch-error/src/lib.rs
```diff
@@ -85,26 +85,34 @@ impl Code {
 
         match self {
             // index related errors
-            CreateIndex => ErrCode::invalid("create_index", StatusCode::BAD_REQUEST),
+            // create index is thrown on internal error while creating an index.
+            CreateIndex => ErrCode::internal("create_index", StatusCode::BAD_REQUEST),
+            // thrown when an user want to creae an index twice with the same uid
             IndexAlreadyExists => ErrCode::invalid("index_already_exists", StatusCode::BAD_REQUEST),
+            // thrown when reuesting an unexisting index
```
MarinPostma, line 92:
```suggestion
            // thrown when requesting an unexisting index
```

meilisearch-error/src/lib.rs
```diff
@@ -85,26 +85,34 @@ impl Code {
...
             // invalid state error
             InvalidState => ErrCode::internal("invalid_state", StatusCode::INTERNAL_SERVER_ERROR),
+            // thrown when no primary key has been set
             MissingPrimaryKey => ErrCode::internal("missing_primary_key", StatusCode::INTERNAL_SERVER_ERROR),
-            PrimaryKeyAlreadyPresent => ErrCode::internal("primary_key_already_present", StatusCode::INTERNAL_SERVER_ERROR),
+            // error thrown when trying to set an already existing primary key
+            PrimaryKeyAlreadyPresent => ErrCode::invalid("primary_key_already_present", StatusCode::BAD_REQUEST),
 
             // invalid document
             MaxFieldsLimitExceeded => ErrCode::invalid("max_field_limit_exceeded", StatusCode::BAD_REQUEST),
```
curquiza, line 105:
```suggestion
            MaxFieldsLimitExceeded => ErrCode::invalid("max_fields_limit_exceeded", StatusCode::BAD_REQUEST),
```

MarinPostma: > For missing_primary_key, I assume that means MS cannot infer the primary key. Could we rename the code accordingly? Like could_not_infer_primary_key slightly_smiling_face So that we can clearly understand the difference between this error and the missing_document_id one

no objection on my end. @Kerollmops, @bidoubiwa, @eskombro ?

eskombro: I like the idea, I don't understand why we talk about a `document_id` instead of the normal term we use, primary key. I propose:

`could_not_infer_primary_key`
`missing_primary_key_in_document`

eskombro:
I'm sorry if I'm repeating myself, but for me it is not clear what `invalid_state` means, and I didn't see test for it either :)

MarinPostma:
@eskombro This is not really testable, this happens on data corruption that leaves the db in an invalid state. I only tested error that can happen with an inappropriate use of meilisearch. Another PR tests a bunch of error thrown in the update system (#792)

eskombro:
So is it basically an internal server error?

MarinPostma:
```rust
InvalidState => ErrCode::internal("invalid_state", ...
```
yes, this is labeled as such. There isn't much we can do to help, but ask to wipe the db and re-index...

curquiza:
Could we rename it as `db_invalid_state` if it concerns only the DB 🙂

MarinPostma:
I'd rather we kept it that way, because it could hide other kind of invalid states that are not db related

PR #792, opened 2020-06-22
Error codes in updates

MarinPostma:
Make error reporting in updates consistent with other error messages.

also:
- fix potential infinite loop in tests
- fix reporting of missing document id on document addition to return the correct error

### DIFF
--- a/meilisearch-http/tests/common.rs
+++ b/meilisearch-http/tests/common.rs
@@ -111,17 +111,19 @@ impl Server {
 
 
     pub async fn wait_update_id(&mut self, update_id: u64) {
-        loop {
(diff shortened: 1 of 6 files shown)


meilisearch-http/tests/common.rs
```diff
@@ -111,17 +111,19 @@ impl Server {
 
 
     pub async fn wait_update_id(&mut self, update_id: u64) {
-        loop {
+        // try 10 times to get status, or panic to not wait forever
+        for _ in 1..10 {
```
Kerollmops, line 115: This will not loop for 10 rounds but only 9 :)

MarinPostma: of course...

Issue #551, opened 2020-03-27
Automatic snapshots


qdequele:
Enable to snapshot the Meilisearch into another folder. The snapshots must be throttled and fully automatic. 

The user can specify the folder in which to do the snapshot by giving an environment variable. The snapshot will be done each x minutes if there is a document or settings update. The snapshot is a compacted and compressed view of the key-value store. 

To go further, we could imagine that the user, instead of specifying a folder, could give an URL of the format [scheme]://[host]/[path]. With that kind of URL the user can specify a folder on a local or remote disk, an S3 server, and more. But this is much more complex.

PR #829, opened 2020-07-02
implement snapshoting

LegendreM:
related to #551.

This pull request permit user to create periodically a snapshot of MeiliSearch database via a command line and launch meiliSearch from a snapshot with another command

### schedule a snapshot
`--snapshot-path <DIRECTORY_PATH>`:
this will periodically create a snapshot `<DB_NAME>.tar.gz` in the specified directory

### change period between 2 snapshot creation
`--snapshot-interval-sec <GAP_IN_SEC>`
choose the time gap between 2 snapshot

### start meilisearch from a snapshot
`--load-from-snapshot <FILE_PATH>`
this will use the snapshot stored at `<FILE_PATH>` to initialize MeiliSearch database,

`--ignore-snapshot-if-db-exists` if set and if a db already exists,
this will skip snapshot importation and continue process with actual db instead of quitting process by returning an Error

`--ignore-missing-snapshot` if set and if no snapshot exists at provided path,
this will skip snapshot importation and continue process with actual db instead of quitting process by returning an Error

### DIFF
--- a/meilisearch-http/src/option.rs
+++ b/meilisearch-http/src/option.rs
@@ -1,7 +1,7 @@
+use std::{error, fs};
 use std::io::{BufReader, Read};
 use std::path::PathBuf;
 use std::sync::Arc;
-use std::{error, fs};
 
 use rustls::internal::pemfile::{certs, pkcs8_private_keys, rsa_private_keys};
 use rustls::{
@@ -93,6 +93,28 @@ pub struct Opt {
     /// SSL support tickets.
     #[structopt(long, env = "MEILI_SSL_TICKETS")]
     pub ssl_tickets: bool,
+
+    /// Defines the path of the snapshot file to import.
+    /// This option will, by default, stop the process if a database already exist or if no snapshot exists at
+    /// the given path. If this option is not specified no snapshot is imported.
+    #[structopt(long, env = "MEILI_LOAD_FROM_SNAPSHOT")]
+    pub load_from_snapshot: Option<PathBuf>,
+
+    /// The engine will ignore a missing snapshot and not return an error in such case.
+    #[structopt(long, requires = "load-from-snapshot", env = "MEILI_IGNORE_MISSING_SNAPSHOT")]
+    pub ignore_missing_snapshot: bool,
+
+    /// The engine will skip snapshot importation and not return an error in such case.
+    #[structopt(long, requires = "load-from-snapshot", env = "MEILI_IGNORE_SNAPSHOT_IF_DB_EXISTS")]
+    pub ignore_snapshot_if_db_exists: bool,
+
+    /// Defines the directory path where meilisearch will create snapshot each snapshot_time_gap.
+    #[structopt(long, env = "MEILI_SNAPSHOT_PATH")]
+    pub snapshot_path: Option<PathBuf>,
+
+    /// Defines time interval, in seconds, between each snapshot creation.
+    #[structopt(long, requires = "snapshot-path", default_value = "86400", env = "MEILI_SNAPSHOT_INTERVAL_SEC")]
+    pub snapshot_interval_sec: u64,
 }
 
 impl Opt {
--- a/meilisearch-http/src/snapshot.rs
+++ b/meilisearch-http/src/snapshot.rs
@@ -1,0 +1,124 @@
+use crate::Data;
+use crate::error::Error;
+
+use flate2::Compression;
+use flate2::read::GzDecoder;
+use flate2::write::GzEncoder;
+use log::error;
+use std::fs::{create_dir_all, File};
+use std::io;
+use std::path::Path;
+use std::thread;
+use std::time::{Duration};
+use tar::{Builder, Archive};
+use tempfile::TempDir;
+
+fn pack(src: &Path, dest: &Path) -> io::Result<()> {
+    let f = File::create(dest)?;
+    let gz_encoder = GzEncoder::new(f, Compression::default());
+
+    let mut tar_encoder = Builder::new(gz_encoder);
+    tar_encoder.append_dir_all(".", src)?;
+    let gz_encoder = tar_encoder.into_inner()?;
+
+    gz_encoder.finish()?;
+
+    Ok(())
+}
+
+fn unpack(src: &Path, dest: &Path) -> Result<(), Error> {
+    let f = File::open(src)?;
+    let gz = GzDecoder::new(f);
+    let mut ar = Archive::new(gz);
+
+    create_dir_all(dest)?;
+    ar.unpack(dest)?;
+
+    Ok(())
+}
+
+pub fn load_snapshot(
+    db_path: &str,
+    snapshot_path: &Path,
+    ignore_snapshot_if_db_exists: bool,
+    ignore_missing_snapshot: bool
+) -> Result<(), Error> {
+    let db_path = Path::new(db_path);
+
+    if !db_path.exists() && snapshot_path.exists() {
+        unpack(snapshot_path, db_path)
+    } else if db_path.exists() && !ignore_snapshot_if_db_exists {
+        Err(Error::Internal(format!("database already exists at {:?}", db_path)))
+    } else if !snapshot_path.exists() && !ignore_missing_snapshot {
+        Err(Error::Internal(format!("snapshot doesn't exist at {:?}", snapshot_path)))
+    } else {
+        Ok(())
+    }
+}
+
+pub fn create_snapshot(data: &Data, snapshot_path: &Path) -> Result<(), Error> {
+    let tmp_dir = TempDir::new()?;
+
+    data.db.copy_and_compact_to_path(tmp_dir.path())?;
+
+    pack(tmp_dir.path(), snapshot_path).or_else(|e| Err(Error::Internal(format!("something went wrong during snapshot compression: {}", e))))
+}
+
+pub fn schedule_snapshot(data: Data, snapshot_dir: &Path, time_gap_s: u64) -> Result<(), Error> {
+    if snapshot_dir.file_name().is_none() { 
+        return Err(Error::Internal("invalid snapshot file path".to_string()));
+    }
+    let db_name = Path::new(&data.db_path).file_name().ok_or_else(|| Error::Internal("invalid database name".to_string()))?;
+    create_dir_all(snapshot_dir)?;
+    let snapshot_path = snapshot_dir.join(format!("{}.tar.gz", db_name.to_str().unwrap_or("data.ms")));
+    
+    thread::spawn(move || loop { 
+        thread::sleep(Duration::from_secs(time_gap_s));
+        if let Err(e) = create_snapshot(&data, &snapshot_path) {
+            error!("Unsuccessful snapshot creation: {}", e);
+        }
+    });
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::io::prelude::*;
+    use std::fs;
+
+    #[test]
+    fn test_pack_unpack() {
+        let tempdir = TempDir::new().unwrap();
+
+        let test_dir = tempdir.path();
+        let src_dir = test_dir.join("src");
+        let dest_dir = test_dir.join("complex/destination/path/");
+        let archive_path = test_dir.join("archive.tar.gz");
+
+        let file_1_relative = Path::new("file1.txt");
+        let subfolder_relative = Path::new("subfolder/");
+        let file_2_relative = Path::new("subfolder/file2.txt");
+        
+        create_dir_all(src_dir.join(subfolder_relative)).unwrap();
+        File::create(src_dir.join(file_1_relative)).unwrap().write_all(b"Hello_file_1").unwrap();
+        File::create(src_dir.join(file_2_relative)).unwrap().write_all(b"Hello_file_2").unwrap();
+
+        
+        assert!(pack(&src_dir, &archive_path).is_ok());
+        assert!(archive_path.exists());
+        assert!(load_snapshot(&dest_dir.to_str().unwrap(), &archive_path, false, false).is_ok());
+
+        assert!(dest_dir.exists());
+        assert!(dest_dir.join(file_1_relative).exists());
+        assert!(dest_dir.join(subfolder_relative).exists());
+        assert!(dest_dir.join(file_2_relative).exists());
+
+        let contents = fs::read_to_string(dest_dir.join(file_1_relative)).unwrap();
+        assert_eq!(contents, "Hello_file_1");
+    
+        let contents = fs::read_to_string(dest_dir.join(file_2_relative)).unwrap();
+        assert_eq!(contents, "Hello_file_2");
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -301,10 +301,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "adler32"
-version = "1.0.4"
+name = "adler"
+version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5d2e7343e7fc9de883d1b0341e0b13970f764c14101234857d2ddafa1cb1cac2"
+checksum = "ccc9a9dd069569f212bc4330af9f17c4afb5e8ce185e83dbb14f1349dda18b10"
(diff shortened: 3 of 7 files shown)


MarinPostma:
> Hye, great PR.
> 
> However, would it be possible to run this job in another thread and let the user choose the periodicy of it?

@Kerollmops, running meilisearch with the --snapshot argument performs the snapshot and returns right away. Isn't it preferable to rely on cron for the scheduling?

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,44 @@
...
+use std::time::{Duration};
+use tempfile::TempDir;
+use crate::error::Error;
+use crate::Data;
+
+
+pub fn create_snapshot(data: &Data, snapshot_path: &PathBuf) -> Result<(), Error> {
+    let tmp_dir = TempDir::new()?;
+
+    data.db.copy_and_compact_to_path(tmp_dir.path());
```
LegendreM, line 16: Hello @MarinPostma @Kerollmops, here we have a `meilisearch_core::error::Error`,
should I reimplement `From<meilisearch_core::error::Error>` like we made [here](https://github.com/meilisearch/MeiliSearch/pull/829/files#diff-f6b32c1d41673385c68a91309c9b2dc0R239-R244) or is there a  better way to do it?

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,44 @@
...
+pub fn schedule_snapshot(data: Data, snapshot_path: &PathBuf, time_gap_s: u64) -> Result<(), Error> {
+    if snapshot_path.file_name().is_none() { 
+        return Err(Error::Internal("invalid snapshot file path".to_string()));
+    }
+    if let Some(parent_dir) = snapshot_path.parent() { create_dir_all(parent_dir)?; }
+    let snapshot_path = snapshot_path.with_extension("tar.gz");
+    
+    thread::spawn(move || loop { 
+        thread::sleep(Duration::from_secs(time_gap_s));
+        create_snapshot(&data, &snapshot_path);
```
LegendreM, line 40: Hello @MarinPostma @Kerollmops, here we have a possible error returned by `create_snapshot`, how can I log it properly?

MarinPostma: There is a log::error macro you can use to log an error

meilisearch-http/src/snapshot.rs
MarinPostma: As we don't plan to return it with an http, and we don't have anyhow yet, I suggest that we return a box dyn error, and just log it as you mentioned after.

MarinPostma: We'll turn it in an anyhow error later

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,118 @@
...
+    ar.unpack(dest)?;
+
+    Ok(())
+}
+
+pub fn import_snapshot(db_path: &str, snapshot_path: &Path) -> Result<(), Error> {
+    let db_path = Path::new(db_path);
+
+    if db_path.exists() {
+        warn!("database already exist at {:?}, skiping snapshot importation", db_path);
```
LegendreM, line 43: here I puted a warning if a database already exist, it's not a fatal error to let user use this command in his deployment strategy (drop and restart...). It could be a log info, I don't know which is better. 😄

meilisearch-http/src/option.rs
```diff
@@ -93,6 +93,18 @@ pub struct Opt {
...
+    #[structopt(long)]
+    pub from_snapshot: Option<PathBuf>,
+
+    /// defines the file path where meilisearch will write snapshot each <snapshot_time_gap>.
+    #[structopt(long)]
+    pub snapshot_path: Option<PathBuf>,
+
+    /// defines time gap (seconds) between each snapshot creation.
+    #[structopt(long, requires = "snapshot-path", default_value = "86400")]
+    pub snapshot_time_gap_sec: u64,
```
LegendreM, line 107: should I add some ENV_VAR?

meilisearch-http/src/snapshot.rs
MarinPostma: what db should you take then? I think this should abort execution

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,118 @@
...
+        Ok(())
+    } else {
+        unpack(snapshot_path, db_path)
```
MarinPostma, line 46: this does not actually launch meili with the snapshot

curquiza:
@bidoubiwa, just a ping for the docs 🙂

bidoubiwa:
Would it be possible that @LegendreM makes the docs for it? I prefer the new features to be documented by the ones that made it.

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,134 @@
...
+pub fn recover_from_snapshot(db_path: &str, snapshot_path: &Path) -> Result<(), Error> {
+    let db_path = Path::new(db_path);
+
+    if db_path.exists() {
+        warn!("database already exist at {:?}, skiping snapshot importation", db_path);
+        Ok(())
+    } else if snapshot_path.exists() {
+        unpack(snapshot_path, db_path)
+    } else {
+        warn!("snpashot doesn't exist at {:?}, skiping snapshot importation", snapshot_path);
```
MarinPostma, line 49:
```suggestion
        warn!("snapshot doesn't exist at {:?}, skipping snapshot importation", snapshot_path);
```

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,134 @@
...
+        Ok(())
+    }
+}
+
+pub fn init_from_snapshot(db_path: &str, snapshot_path: &Path) -> Result<(), Error> {
+    let db_path = Path::new(db_path);
+
+    if db_path.exists() {
+        Err(Error::Internal(format!("database already exist at {:?}", db_path)))
```
MarinPostma, line 58:
```suggestion
        Err(Error::Internal(format!("database already exists at {:?}", db_path)))
```

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,134 @@
...
+    } else if snapshot_path.exists() {
+        unpack(snapshot_path, db_path)
+    } else {
+        Err(Error::Internal(format!("snpashot doesn't exist at {:?}", snapshot_path)))
```
MarinPostma, line 62:
```suggestion
        Err(Error::Internal(format!("snapshot doesn't exist at {:?}", snapshot_path)))
```

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+}
+
+pub fn create_snapshot(data: &Data, snapshot_path: &PathBuf) -> Result<(), Box<dyn std::error::Error>> {
```
MarinPostma, line 66: Can you make it return a concrete error please?

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,134 @@
...
+    data.db.copy_and_compact_to_path(tmp_dir.path())?;
+
+    pack(tmp_dir.path(), &snapshot_path)
+}
+
+pub fn schedule_snapshot(data: Data, snapshot_dir: &PathBuf, time_gap_s: u64) -> Result<(), Error> {
+    if snapshot_dir.file_name().is_none() { 
+        return Err(Error::Internal("invalid snapshot file path".to_string()));
+    }
+    let db_name = Path::new(&data.db_path).file_name().ok_or(Error::Internal("invalid database name".to_string()))?;
```
MarinPostma, line 78:
```suggestion
    let db_name = Path::new(&data.db_path).file_name().ok_or_else(Error::Internal("invalid database name".to_string()))?;
```

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        let mut tar_encoder = Builder::new(&mut gz_encoder);
+
+        tar_encoder.append_dir_all(".", src)?;
+        tar_encoder.finish()?;
+    }
+    gz_encoder.finish()?;
+
+    Ok(())
+}
```
Kerollmops, line 27:
```suggestion
fn pack(src: &Path, dest: &Path) -> io::Result<()> {
    let f = File::create(dest)?;
    let mut gz_encoder = GzEncoder::new(f, Compression::default());
    let mut tar_encoder = Builder::new(gz_encoder);
    tar_encoder.append_dir_all(".", src)?;
    let gz_encoder = tar_encoder.into_inner()?;
    gz_encoder.finish()
}
```

meilisearch-http/src/option.rs
```diff
@@ -93,6 +93,26 @@ pub struct Opt {
...
+    #[structopt(long, conflicts_with = "init-from-snapshot")]
+    pub recover_from_snapshot: Option<PathBuf>,
+
+    /// defines the directory path where meilisearch will create snapshot each <snapshot_time_gap>.
+    #[structopt(long)]
+    pub snapshot_path: Option<PathBuf>,
+
+    /// defines time gap (seconds) between each snapshot creation.
+    #[structopt(long, requires = "snapshot-path", default_value = "86400")]
+    pub snapshot_interval_sec: u64,
```
Kerollmops, line 115: Could you please add the related env variables please?

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,134 @@
...
+    } else {
+        warn!("snpashot doesn't exist at {:?}, skiping snapshot importation", snapshot_path);
+        Ok(())
+    }
+}
+
+pub fn init_from_snapshot(db_path: &str, snapshot_path: &Path) -> Result<(), Error> {
+    let db_path = Path::new(db_path);
+
+    if db_path.exists() {
```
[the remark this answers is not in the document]
LegendreM: `db_path` is based on a `String` and need to be converted anyway, it's probably better to convert it here instead in the `main.rs` file

meilisearch-http/src/option.rs
```diff
@@ -93,6 +93,26 @@ pub struct Opt {
...
+    /// This command will stop the process if a database already exist or if no snapshot exists in the given path.
+    /// If this command is not called, no snapshot will be imported.
+    #[structopt(long, env = "MEILI_INIT_FROM_SNAPSHOT")]
+    pub init_from_snapshot: Option<PathBuf>,
+
+    /// Defines the path of the snapshot file to import.
+    /// This command will skip importing if a database already exist or if no snapshot exists in the given path.
+    /// If this command is not called, no snapshot will be imported.
+    #[structopt(long, conflicts_with = "init-from-snapshot", env = "MEILI_RECOVER_FROM_SNAPSHOT")]
+    pub recover_from_snapshot: Option<PathBuf>,
```
Kerollmops, line 107: I would like to replace both the `init-from-snapshot` and `recover-from-snapshot` by one `load-from-snapshot` option that can be associated with an `ignore-missing-snapshot` option. This way we avoid potential confusions and keep the documentation of the snapshot loading option in one place and specify snapshot loading behaviors with simple bolean flags.

I don't know what to do when a database already exists, maybe never return an error and just thrown a warning?
I'm not sure it would be a good idea to introduce another flag? What do you think?

```suggestion
    /// Defines the path of the snapshot file to import.
    /// This option will, by default, stop the process if a database already exist or if no snapshot exists at
    /// the given path. If this option is not specified no snapshot is imported.
    #[structopt(long, env = "MEILI_LOAD_FROM_SNAPSHOT")]
    pub load_from_snapshot: Option<PathBuf>,

    /// The engine will ignore a missing snapshot and not return an error in such case.
    #[structopt(long, requires = "load-from-snapshot", env = "MEILI_IGNORE_MISSING_SNAPSHOT")]
    pub ignore_missing_snapshot: bool,
```

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,134 @@
...
+    let f = File::create(dest)?;
+    let mut gz_encoder = GzEncoder::new(f, Compression::default());
+    {
+        let mut tar_encoder = Builder::new(&mut gz_encoder);
+        tar_encoder.append_dir_all(".", src)?;
+        tar_encoder.finish()?;
+    }
+    gz_encoder.finish()?;
+
+    Ok(())
```
Kerollmops, line 26: Notice the `&mut gz_encoder` replaced by `gz_encoder` here.

```suggestion
    let f = File::create(dest)?;
    let gz_encoder = GzEncoder::new(f, Compression::default());
    let mut tar_encoder = Builder::new(gz_encoder);
    tar_encoder.append_dir_all(".", src)?;
    let gz_encoder = tar_encoder.into_inner()?;
    gz_encoder.finish()?;
    Ok(())
```

meilisearch-http/src/option.rs
Kerollmops: We concluded in using the `ignore-snapshot-if-db-exists` to make the engine throw an error if the database already exists.

MarinPostma: so the default behavior is to erase user data? Isn't it good practice to do the other way around? (`rm -f`, `git push -f` etc...)

LegendreM: @MarinPostma nope, db will never be erased, the behavior of `ignore-snapshot-if-db-exists` is:
* `false` (default): crash if db exists
* `true`: skip snapshoting process and run meilisearch with actual db

meilisearch-http/src/option.rs
```diff
@@ -93,6 +93,28 @@ pub struct Opt {
...
+
+    /// The engine will skip snapshot importation and not return an error in such case.
+    #[structopt(long, requires = "load-from-snapshot", env = "MEILI_IGNORE_SNAPSHOT_IF_DB_EXISTS")]
+    pub ignore_snapshot_if_db_exists: bool,
+
+    /// Defines the directory path where meilisearch will create snapshot each snapshot_time_gap.
+    #[structopt(long, env = "MEILI_SNAPSHOT_PATH")]
+    pub snapshot_path: Option<PathBuf>,
+
+    /// Defines time gap in seconds between each snapshot creation.
```
MarinPostma, line 115:
```suggestion
    /// Defines time interval, in seconds, between each snapshot creation.
```

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,124 @@
...
+    ignore_missing_snapshot: bool
+) -> Result<(), Error> {
+    let db_path = Path::new(db_path);
+
+    if !db_path.exists() && snapshot_path.exists() {
+        unpack(snapshot_path, db_path)
+    } else if db_path.exists() && !ignore_snapshot_if_db_exists {
+        Err(Error::Internal(format!("database already exists at {:?}", db_path)))
+    } else if !snapshot_path.exists() && !ignore_missing_snapshot {
+        Err(Error::Internal(format!("snapshot doesn't exists at {:?}", snapshot_path)))
```
MarinPostma, line 53:
```suggestion
        Err(Error::Internal(format!("snapshot doesn't exist at {:?}", snapshot_path)))
```

meilisearch-http/src/snapshot.rs
```diff
@@ -0,0 +1,124 @@
...
+        Ok(())
+    }
+}
+
+pub fn create_snapshot(data: &Data, snapshot_path: &Path) -> Result<(), Error> {
+    let tmp_dir = TempDir::new()?;
+
+    data.db.copy_and_compact_to_path(tmp_dir.path())?;
+
+    pack(tmp_dir.path(), snapshot_path).or_else(|e| Err(Error::Internal(format!("someting went wrong during snapshot compression: {}", e))))
```
MarinPostma, line 64:
```suggestion
    pack(tmp_dir.path(), snapshot_path).or_else(|e| Err(Error::Internal(format!("something went wrong during snapshot compression: {}", e))))
```

mikerogerz:
Shouldn't the --snapshot-path argument be optional when starting MeiliSearch? I'm unable to start without it.

I'm assuming not all users will want to use snapshot functionality.

MarinPostma:
@mikerogerz thanks, working on a fix rn.

PR #846, opened 2020-07-10
Change settings behavior

### LINKED ISSUE #824: Change the settings behavior
As a daily user of MS, I noticed some weird or unintuitive behaviors when manipulating the settings.
- sometimes, MS knows attributes that the users don't know (see #767)
- the concept of "new field" with the `acceptNewField` settings is not really intuitive since an attribute is not "new" because already declared in the settings (see #761). It leads to unintuitive behavior because the order (adding documents before the settings or the opposite) matters, but it should not.
- the reset of searchable and displayed attributes has a weird behavior. See the feedback of one of our users:
Screenshot 2020-07-01 at 14 54 14

### How to solve them?

What we suggest to solve these problems:

The default behavior of `searchableAttributes` and `displayedAttributes` is now `["*"]`.
`["*"]` means "all the attributes".
When resetting one of these settings, the default value is `["*"]` again.

What does it imply?
- the users use `searchableAttributes` and `displayedAttributes` **only for restriction**: I mean, if they want to restrict the display (e.g.: they have `title` and `author` in their documents but only want `title` to be displayed) you use the `displayedAttributes` settings. Otherwise, you don't, because by default all the settings are taken into account.
- the `acceptNewFields` is not needed anymore: no more unintuitive notion of "new fields". If you want to stay safe and avoid new unexpected fields to be added to searchable or displayed attributes (that's the role of the current `acceptNewFields` settings), you just have to change directly the `searchableAttributes` and `displayedAttributes`. Indeed, these two settings when no being set as `["*"]` are whitelists.
- No more unintuitive behavior when resetting because the reset will set the value `["*"]` again.

About the issue that "MeiliSearch knows attributes that the users don't", we suggest changing the behavior of the current `GET indexes/:uid/stats` route.
Currently, when calling this route, the `fieldsDistribution` field returns all the attributes found in the documents **according to the `displayedAttributes` you set**. It means that if you only set `['title']` as `displayedAttributes` you will not see the attribute `author` in `fieldsDistribution` despite MeiliSearch knows `author` and has it in its DB.
We suggest that this route now returns all the known attributes no matter the `displayedAttributes` are.

### TODO

- [x] remove `acceptNewFields` routes
- [x] change the default value of `searchableAttributes` and `displayedAttributes`. Now it's `["*"]` that means "all the attribtues found in the document". This value (`["*"]`) should be set when resetting.
- [ ] `fieldsDistribution` in the `GET indexes/:uid/stats` route should now returns all the known attributes no matter the `displayedAttributes` are.
- [ ]  and of course, add/update tests ❤️

### LINKED ISSUE #853: Setting displayed attributes to empty array cause corrupted data error on search
**Describe the bug**

Setting displayed attributes to empty array cause corrupted data error on search

**To Reproduce**
1. create an index
2. index documents
3. perform search (it should work)
4. update settings with all settings route and set displayed attrbutes to []
5. perform a search:
```
{
    "errorCode":"internal",
    "errorLink":"https://docs.meilisearch.com/error/internal",
    "errorType":"internal_error",
    "message":"Impossible to retrieve the document; Corrupted data"
}
```

**Expected behavior**

I should get a search response with a set of empty documents

## Todo
- [ ] find and fix the bug
- [ ] test behavior of settings when passed empty arrays

MarinPostma:
partially implements #824.

Returning the field distribution for all know fields is more complicated that anticipated, see https://github.com/meilisearch/MeiliSearch/issues/824#issuecomment-657656561

If we decide to to it anyway, and find a reasonable solution, I will make another PR.

fix #853 by resetting displayed and searchable attributes to wildcard when attributes are set to `[]` in the all settings route. @curquiza @bidoubiwa can you confirm me that this is the expected behavior?


MarinPostma:
bors try

PR #849, opened 2020-07-11
Update nbHits count with filtered documents

### LINKED ISSUE #764: Total result count (nbHits) not updated for filters
I want to show the number of total results (for pagination and for feedback to my users). The result field nbHits is updated by the search string but is not updated by the filters that are given. Because of that it is impossible to give a total count to the users and make a good pagination.

It would be nice if nbHits is also updated by filters or if there is a secord variable where the filtered count is shown.

### LINKED ISSUE #1039: Question: Combine facets with filter
Hi
When set a filter example price and select a facet filter the response about de hits it's correct but the count of facets no chance

Filter price > 15000 AND price < 160000 and brand:acura the response say have 1 but it's not true, not have any car with that price range
image

The response say I have 3389 hits with and without filter price
image

Without price filter it's same response
image

It's a bug?? or how to get correct counting into facets using filter

balajisivaraman:
Closes #764 
close #1039

After discussing with @MarinPostma on Slack, this is my first attempt at implementing this for the basic flow that will go through `bucket_sort_with_distinct`.

A few thoughts here: 

- For getting the count of filtered documents alone, I originally thought of using `filter_map.values().filter(|&&v| !v).count()`. In a few cases, this was the same as what I have now implemented. But I realised I couldn't do something similar for `distinct`. So for being consistent, I have implemented both in a similar fashion.
- I also needed the `contains_key` check to ensure we're not counting the same document ID twice.

@MarinPostma also mentioned that this will be an approximation since the sort is lazy. In the test example that I've updated, the actual filtered count will be just 19 (for `male` records), but due to the `limit` in play, it returns 32 (filtering out 11 records overall).

Please let me know if this is the kind of fix we are looking for, and I can implement it in the placeholder search also.

### DIFF
--- a/meilisearch-core/src/bucket_sort.rs
+++ b/meilisearch-core/src/bucket_sort.rs
@@ -212,6 +212,7 @@ where
     FD: Fn(DocumentId) -> Option<u64>,
 {
     let mut result = SortResult::default();
+    let mut filtered_count = 0;
 
     let words_set = index.main.words_fst(reader)?;
     let stop_words = index.main.stop_words_fst(reader)?;
@@ -322,19 +323,36 @@ where
                     let filter_accepted = match &filter {
                         Some(filter) => {
                             let entry = filter_map.entry(document.id);
-                            *entry.or_insert_with(|| (filter)(document.id))
+                            *entry.or_insert_with(|| {
+                                let accepted = (filter)(document.id);
+                                // we only want to count it out the first time we see it
+                                if !accepted {
+                                    filtered_count += 1;
+                                }
+                                accepted
+                            })
                         }
                         None => true,
                     };
 
                     if filter_accepted {
                         let entry = key_cache.entry(document.id);
-                        let key = entry.or_insert_with(|| (distinct)(document.id).map(Rc::new));
+                        let mut seen = true;
+                        let key = entry.or_insert_with(|| {
+                            seen = false;
+                            (distinct)(document.id).map(Rc::new)
+                        });
 
-                        match key.clone() {
+                        let distinct = match key.clone() {
                             Some(key) => buf_distinct.register(key),
                             None => buf_distinct.register_without_key(),
                         };
+
+                        // we only want to count the document if it is the first time we see it and
+                        // if it wasn't accepted by distinct
+                        if !seen && !distinct {
+                            filtered_count += 1;
+                        }
                     }
 
                     // the requested range end is reached: stop computing distinct
@@ -396,7 +414,7 @@ where
         }
     }
     result.documents = documents;
-    result.nb_hits = docids.len();
+    result.nb_hits = docids.len() - filtered_count;
 
     Ok(result)
 }
--- a/meilisearch-core/src/query_builder.rs
+++ b/meilisearch-core/src/query_builder.rs
@@ -225,10 +225,17 @@ impl<'c, 'f, 'd, 'i> QueryBuilder<'c, 'f, 'd, 'i> {
 
     fn sort_result_from_docids(&self, docids: &[DocumentId], range: Range<usize>) -> SortResult {
         let mut sort_result = SortResult::default();
+        let mut filtered_count = 0;
         let mut result = match self.filter {
             Some(ref filter) => docids
                 .iter()
-                .filter(|item| (filter)(**item))
+                .filter(|item| {
+                    let accepted = (filter)(**item);
+                    if !accepted {
+                        filtered_count += 1;
+                    }
+                    accepted
+                })
                 .skip(range.start)
                 .take(range.end - range.start)
                 .map(|&id| Document::from_highlights(id, &[]))
@@ -248,15 +255,19 @@ impl<'c, 'f, 'd, 'i> QueryBuilder<'c, 'f, 'd, 'i> {
             result.retain(|doc| {
                 let id = doc.id;
                 let key = (distinct)(id);
-                match key {
+                let distinct_accepted = match key {
                     Some(key) => distinct_map.register(key),
                     None => distinct_map.register_without_key(),
+                };
+                if !distinct_accepted {
+                    filtered_count += 1;
                 }
+                distinct_accepted
             });
         }
 
         sort_result.documents = result;
-        sort_result.nb_hits = docids.len();
+        sort_result.nb_hits = docids.len() - filtered_count;
         sort_result
     }
 
--- a/meilisearch-http/tests/placeholder_search.rs
+++ b/meilisearch-http/tests/placeholder_search.rs
@@ -588,3 +588,48 @@ async fn placeholder_search_with_empty_query() {
         assert_eq!(response["hits"].as_array().unwrap().len(), 3);
     });
 }
+
+#[actix_rt::test]
+async fn test_filter_nb_hits_search_placeholder() {
+    let mut server = common::Server::with_uid("test");
+
+    let body = json!({
+        "uid": "test",
+        "primaryKey": "id",
+    });
+
+    server.create_index(body).await;
+    let documents = json!([
+        {
+            "id": 1,
+            "content": "a",
+            "color": "green",
+            "size": 1,
+        },
+        {
+            "id": 2,
+            "content": "a",
+            "color": "green",
+            "size": 2,
+        },
+        {
+            "id": 3,
+            "content": "a",
+            "color": "blue",
+            "size": 3,
+        },
+    ]);
+
+    server.add_or_update_multiple_documents(documents).await;
+    let (response, _) = server.search_post(json!({})).await;
+    assert_eq!(response["nbHits"], 3);
+
+    server.update_distinct_attribute(json!("color")).await;
+
+    let (response, _) = server.search_post(json!({})).await;
+    assert_eq!(response["nbHits"], 2);
+
+    let (response, _) = server.search_post(json!({"filters": "size < 3"})).await;
+    println!("result: {}", response);
+    assert_eq!(response["nbHits"], 1);
+}
--- a/meilisearch-http/tests/search.rs
+++ b/meilisearch-http/tests/search.rs
@@ -1829,3 +1829,51 @@ async fn update_documents_with_facet_distribution() {
     let (response2, _) = server.search_post(search).await;
     assert_json_eq!(expected_facet_distribution, response2["facetsDistribution"].clone());
 }
+
+#[actix_rt::test]
+async fn test_filter_nb_hits_search_normal() {
+    let mut server = common::Server::with_uid("test");
+
+    let body = json!({
+        "uid": "test",
+        "primaryKey": "id",
+    });
+
+    server.create_index(body).await;
+    let documents = json!([
+        {
+            "id": 1,
+            "content": "a",
+            "color": "green",
+            "size": 1,
+        },
+        {
+            "id": 2,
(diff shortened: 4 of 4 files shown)


MarinPostma:
Hey @balajisivaraman, sorry I need time to look into this, I'll review this PR asap

balajisivaraman:
@MarinPostma, No worries! Thanks.

meilisearch-core/src/bucket_sort.rs
```diff
@@ -328,13 +330,20 @@ where
...
                         let entry = key_cache.entry(document.id);
                         let key = entry.or_insert_with(|| (distinct)(document.id).map(Rc::new));
 
-                        match key.clone() {
+                        let distinct_accepted = match key.clone() {
                             Some(key) => buf_distinct.register(key),
                             None => buf_distinct.register_without_key(),
                         };
+
+                        if !distinct_accepted && !contains_key {
```
MarinPostma, line 342: why do you need to check contain_key?

meilisearch-core/src/bucket_sort.rs
```diff
@@ -328,13 +330,20 @@ where
                     };
 
                     if filter_accepted {
+                        let contains_key = key_cache.contains_key(&document.id);
                         let entry = key_cache.entry(document.id);
```
MarinPostma, line 334:
```suggestion
                        let entry = key_cache.entry(document.id);
                        let contains_key = entry.is_some();
```

meilisearch-core/src/query_builder.rs
```diff
@@ -250,15 +257,19 @@ impl<'c, 'f, 'd, 'i> QueryBuilder<'c, 'f, 'd, 'i> {
             result.retain(|doc| {
                 let id = doc.id;
                 let key = (distinct)(id);
-                match key {
+                let distinct_accepted = match key {
                     Some(key) => distinct_map.register(key),
                     None => distinct_map.register_without_key(),
+                };
+                if distinct_accepted {
```
MarinPostma, line 264: is it not the opposite?
```suggestion
                if !distinct_accepted {
```

meilisearch-core/src/bucket_sort.rs
balajisivaraman: Ah, now I see this is a mistake. I noticed that when I did the `filter_accepted` check, I needed the `contains_key` since the document IDs were repeated in the groups. If I didn't do the `contains_key` for filter, then I got overall `nbHits` as 0 since `filtered_count` was high. I think I don't need this for `distinct_accepted` like you pointed out. I will remove it.

meilisearch-core/src/query_builder.rs
balajisivaraman: Fixed.

neevany:
Hi Guys, anything i can contribute to this issue to get it merged? Thanks

meilisearch-core/src/bucket_sort.rs
```diff
@@ -331,10 +333,16 @@ where
...
+                        let distinct_accepted = match key.clone() {
                             Some(key) => buf_distinct.register(key),
                             None => buf_distinct.register_without_key(),
                         };
+
+                        if !distinct_accepted {
+                            filtered_count += 1;
+                        }
+                    } else if !contains_key {
+                        filtered_count += 1;
```
MarinPostma, line 345:
```suggestion
                    }
```

meilisearch-core/src/bucket_sort.rs
```diff
@@ -331,10 +333,16 @@ where
                         let entry = key_cache.entry(document.id);
                         let key = entry.or_insert_with(|| (distinct)(document.id).map(Rc::new));
 
-                        match key.clone() {
+                        let distinct_accepted = match key.clone() {
```
MarinPostma, line 336:
```suggestion
                for document in group.iter() {
                    let filter_accepted = match &filter {
                        Some(filter) => {
                            let entry = filter_map.entry(document.id);
                            let accepted = *entry.or_insert_with(|| (filter)(document.id));
                            if !accepted {
                                filtered_count += 1;
                            }
                            accepted
                        }
                        None => true,
                    };

```

I would suggest this instead, this is more straightforward I think

meilisearch-core/src/query_builder.rs
```diff
@@ -227,10 +227,17 @@ impl<'c, 'f, 'd, 'i> QueryBuilder<'c, 'f, 'd, 'i> {
...
         let mut result = match self.filter {
             Some(ref filter) => docids
                 .iter()
-                .filter(|item| (filter)(**item))
+                .filter(|item| {
+                    let is_filtered = (filter)(**item);
+                    if is_filtered {
+                        filtered_count += 1;
+                    }
+                    is_filtered
```
MarinPostma, line 239:
```suggestion
                    let accepted = (filter)(**item);
                    if !accepted {
                        filtered_count += 1;
                    }
                    accepted
```

MarinPostma (changes requested):
Hello! Very sorry for being this long to review this PR. It still need some little changes, but it is a good job nonetheless. Thanks a lot for your contribution!

balajisivaraman:
@MarinPostma, Thanks for the comments. Is it okay if I get to this in the next couple of days and close it out by the weekend?

MarinPostma:
Yes take your time! ThankS for your help 🙂

LegendreM:
we should test the behavior of #1039 to know if this PR fix it.

meilisearch-core/src/bucket_sort.rs
balajisivaraman: @MarinPostma, I finally got around to this. Just for clarification, this diff should be applied in place of Lines 322 to 333 right? And I should just get rid of the `distinct_accepted` logic in the `if filter_accepted` block. However, when I tried this originally (and again now after your suggestion), for the `test search_with_filter`, I get 0 as the final count, presumably because it filters out everything in the sample set. When I tried this now, I got a `'attempt to subtract with overflow'` panic also. Am I missing something?

MarinPostma: You're right I have been a little too fast on this one, I have tried this one out and it seems to be working:

```rust
            for group in group.binary_group_by_mut(|a, b| criterion.eq(&ctx, a, b)) {
                // we must compute the real distinguished len of this sub-group
                for document in group.iter() {
                    let filter_accepted = match &filter {
                        Some(filter) => {
                            let entry = filter_map.entry(document.id);
                            *entry.or_insert_with(|| { 
                                let accepted = (filter)(document.id);
                                // we only want to count it out the first time we see it
                                if !accepted {
                                    filtered_count += 1;
                                }
                                accepted
                            })
                        }
                        None => true,
                    };

                    if filter_accepted {
                        let entry = key_cache.entry(document.id);
                        let mut seen = true;
                        let key = entry.or_insert_with(|| {
                            seen = false;
                            (distinct)(document.id).map(Rc::new)
                        });

                        let distinct = match key.clone() {
                            Some(key) => buf_distinct.register(key),
                            None => buf_distinct.register_without_key(),
                        };

                        // we only want to count the document if it is the first time we see it and
                        // if it wasn't accepted by distinct
                        if !seen && !distinct {
                            filtered_count += 1;
                        }
                    } 
                    // the requested range end is reached: stop computing distinct
                    if buf_distinct.len() >= range.end {
                        break;
                    }
                }

                documents_seen += group.len();
                groups.push(group);

                // if this sub-group does not overlap with the requested range
                // we must update the distinct map and its start index
                if buf_distinct.len() < range.start {
                    buf_distinct.transfert_to_internal();
                    distinct_raw_offset = documents_seen;
                }

                // we have sort enough documents if the last document sorted is after
                // the end of the requested range, we can continue to the next criterion
                if buf_distinct.len() >= range.end {
                    continue 'criteria;
                }
            }
```

these are the tests I have been trying it on:
```rust
#[actix_rt::test]
async fn test_filter_nb_hits_search_normal() {
    let mut server = common::Server::with_uid("test");

    let body = json!({
        "uid": "test",
        "primaryKey": "id",
    });

    server.create_index(body).await;
    let documents = json!([
        {
            "id": 1,
            "content": "a",
            "color": "green",
            "size": 1,
        },
        {
            "id": 2,
            "content": "a",
            "color": "green",
            "size": 2,
        },
        {
            "id": 3,
            "content": "a",
            "color": "blue",
            "size": 3,
        },
    ]);

    server.add_or_update_multiple_documents(documents).await;
    let (response, _) = server.search_post(json!({"q": "a"})).await;
    assert_eq!(response["nbHits"], 3);

    let (response, _) = server.search_post(json!({"q": "a", "filters": "size = 1"})).await;
    assert_eq!(response["nbHits"], 1);

    server.update_distinct_attribute(json!("color")).await;

    let (response, _) = server.search_post(json!({"q": "a"})).await;
    assert_eq!(response["nbHits"], 2);

    let (response, _) = server.search_post(json!({"q": "a", "filters": "size < 3"})).await;
    println!("result: {}", response);
    assert_eq!(response["nbHits"], 1);
}
```

MarinPostma: the reason for the subtracting with overflow is that we are counting some items more than once (as you've probably figured), we really want to count them only once we first see them :)

MarinPostma: we basically have to try the same tests on placeholder search now

balajisivaraman:
@MarinPostma, Done. Thanks for the help on this one, I had trouble figuring some things out. I tested the `placeholder search` test with the fix removed, and it failed correctly. So I think we look okay here.

MarinPostma:
bors try

MarinPostma (approved):
looks good to me, thank you :)

Kerollmops (approved):

balajisivaraman:
Thanks so much for merging this, and the support on this one from @MarinPostma.

MarinPostma:
Thank you! And sorry for taking so long :)

theo-lubert:
Testing the new v0.17.0 release as we speak (great work), but it seems the patch does not work if there is a `limit`

For exemple, with 3 documents in my index "events" (2 of them with a `date` in the past):

`limit: 25` and `date >= ${Date.now}` gives you `nbHits: 1` (correct)
but,
`limit: 1` and `date >= ${Date.now}` gives you `nbHits: 3` (**incorrect**)

MarinPostma:
hello @theo-lubert, this is perfectly normal since we lazily filter elements, since you only ask for one, and it is not filtered out, the other are still counted in the count.

theo-lubert:
Thanks @MarinPostma , so as I understand it, this is not a good fit for (filtered) pagination then. Any plans on that ? I don't find any feature related to that in the roadmap, should I create a new issue ?

Vincent56:
hi! is there any news on this front? having filters show the correct nbhits would be amazing!

celilsahin:
> Hello @Vincent56, the product team is currently working on the expected behavior we want for `nbHits`. We'll keep you informed 🙂
> It might help us if you would like to share with us your usecase of `nbHits`

My filters;
`{
    "q":"",
    "limit":10,
    "offset":0,
    "filters": "city="AND confirm!=DELETE AND authorId=T8jFa4"
}`
`{
    "q":"",
    "limit":10,
    "offset":0,
    "filters": "city="AND confirm!=DELETE"
}`

nbHits = +76, but hit 15

PR #873, opened 2020-07-21
Update CI for new workflow

MarinPostma:
This pr implements the necessary automation for our new release workflow.

## Pre-releases

whenever something is pushed to a branch `release-v*`, tests are triggered. If all test pass, the current reference is checked to see if it's a release branch. If it's a release branch, a pre-release is created for this branch and assets are automatically generated for this branch. The prerelease has the tag `vx.x.xrcn` where `x.x.x` is the version extracteds from the branch name, and n is the number of commits since the branch was forked from master. (starting from rc0).

## Releases

Whenever something is pushed to stable and tagged `vx.x.x` where `x.x.x` is the version, tests are run and a release is generated containing the assets, and binaries are published to docker, brew, apt, etc.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,8 +2,11 @@
 on:
   push:
     branches:
-      - staging
+      - release-v*
       - trying
+      - staging
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+' # this only concerns tags on stable
 
 name: Test binaries with cargo test
 
@@ -14,7 +17,6 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-
     steps:
     - uses: actions/checkout@v1
     - uses: actions-rs/toolchain@v1
@@ -32,6 +34,7 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: clippy
+
   build-image:
     name: Test the build of Docker image
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -2,8 +2,11 @@
 on:
   push:
     branches:
-      - staging
+      - release-v*
       - trying
+      - staging
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+' # this only concern tags on stable
```
curquiza, line 9:
```suggestion
      - 'v[0-9]+.[0-9]+.[0-9]+' # this only concerns tags on stable
```

.github/workflows/test.yml
```diff
@@ -32,10 +34,60 @@ jobs:
...
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.PUBLISH_TOKEN }} # PAT
+        with:
+          tag_name: ${{ steps.version-number.outputs.number }}rc${{ steps.commit-count.outputs.count }}
+          release_name: Pre-release ${{ steps.version-number.outputs.number }}-rc${{ steps.commit-count.outputs.count }}
+          prerelease: true
+
+  ## If a tag is pushed, a released is created for that tag, and assets will be generated
```
curquiza, line 73:
```suggestion
  ## If a tag is pushed, a release is created for this tag, and assets will be generated
```

.github/workflows/test.yml
```diff
@@ -32,10 +34,60 @@ jobs:
...
+
   build-image:
     name: Test the build of Docker image
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
     - run: docker build . --file Dockerfile -t meilisearch
       name: Docker build
+
+  ## A push occured on a release branch, a prerelease is created and assets are generated
```
curquiza, line 46:
```suggestion
  ## A push occurred on a release branch, a prerelease is created and assets are generated
```

.github/workflows/test.yml
```diff
@@ -32,10 +34,60 @@ jobs:
...
+        id: version-number
+        run: echo "##[set-output name=number;]$(echo ${{ github.ref }} | sed 's/.*\(v.*\)/\1/')"
+      - name: Get commit count
+        id: commit-count
+        run: echo "##[set-output name=count;]$(git rev-list remotes/origin/master..remotes/origin/release-${{ steps.version-number.outputs.number }} --count)"
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.PUBLISH_TOKEN }} # PAT
```
tpayet, line 67: what does PAT mean?

.github/workflows/test.yml
```diff
@@ -2,8 +2,11 @@
 on:
   push:
     branches:
-      - staging
+      - release-v*
       - trying
+      - staging
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+' # this only concerns tags on stable
```
LegendreM, line 9: are you sure that this regex match only stable tags ?
https://regex101.com/r/d4jF4S/1

shouldn't it at least be '^v[0-9]+.[0-9]+.[0-9]+$' ?

.github/workflows/test.yml
MarinPostma: Public Authentication Token, it's a token that allow application to do things on your behalf (@Kerollmops 's behalf in this case). This action cannot be done with the github token, otherwise, other workflows are not triggered on new releases

curquiza: Personal authentication token. That is the token of one of us if im not wrong 🙂

.github/workflows/test.yml
LegendreM: https://gist.github.com/jhorsman/62eeea161a13b80e39f5249281e17c39 😋

LegendreM: https://regex101.com/r/KuQoOu/2/

MarinPostma: this is not exactly like regex:
- it does not support `^` and `$`. 
- `.` are not wildcards
- it works as match and not find

source: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet.
I had the same concerns as you, but after experimenting a bit with it, it really only matches stables release tags

.github/workflows/test.yml
tpayet: ok thanks, so either we remove the comment because it does not bring clarity or we write the whole words ^^

.github/workflows/test.yml
LegendreM: they'll do everything differently....... 😞

.github/workflows/test.yml
MarinPostma: np, PAT is how they refer to it in all their doc

PR #874, opened 2020-07-21
Fixes default values on web interface

tpayet:


meilisearch-http/public/interface.html
```diff
@@ -221,7 +221,7 @@ <h2 class="subtitle">
                 results.innerHTML = '';
 
                 let processingTimeMs = httpResults.processingTimeMs;
-                let numberOfDocuments = httpResults.hits.length;
+                let numberOfDocuments = httpResults.nbHits;
```
MarinPostma, line 224: this is the total number of documents that matched the query, is that what you want?

tpayet: yes, I thought it makes more sense to have the total number of matched documents instead of the number of documents returned by the engine which is very often 20 because of the default limit. WDYT?

I would also like the input of @qdequele here :)

qdequele: Yes, I think also that is better.

MarinPostma:
@tpayet can you rebase please?

PR #926, opened 2020-08-24
Update genre field with genres

bidoubiwa:
Most code samples are made with the assumption that the `genres` field takes an `s`. I'm updating the dataset to match those code-samples.


MarinPostma (approved):
bors r+

PR #928, opened 2020-08-25
format codebase

MarinPostma:
More and more people are working on this codebase on a daily basis, is has become more convenient to adopt the rust style coding standard, with formatting provided by rustfmt, this ensure that we have a consistent style accross the codebase, and that contributors can  easily comply with it.

### DIFF
--- a/meilisearch-core/src/bucket_sort.rs
+++ b/meilisearch-core/src/bucket_sort.rs
@@ -470,16 +470,13 @@ fn cleanup_bare_matches<'tag, 'txn>(
     let docidslen = docids.len() as f32;
     let mut bare_matches = Vec::new();
 
-    for (
-        PostingsKey {
+    for (postings_key, matches) in queries {
+        let PostingsKey {
             query,
             input,
             distance,
             is_exact,
-        },
-        matches,
-    ) in queries
-    {
+        } = postings_key;
         let postings_list_view = PostingsListView::original(Rc::from(input), Rc::new(matches));
         let pllen = postings_list_view.len() as f32;
 
--- a/meilisearch-core/src/bucket_sort.rs
+++ b/meilisearch-core/src/bucket_sort.rs
@@ -1,28 +1,28 @@
 use std::borrow::Cow;
 use std::collections::HashMap;
+use std::fmt;
 use std::mem;
 use std::ops::Deref;
 use std::ops::Range;
 use std::rc::Rc;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::time::Instant;
-use std::fmt;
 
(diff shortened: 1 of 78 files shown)


meilisearch-core/src/bucket_sort.rs
```diff
@@ -559,15 +645,20 @@ impl<'txn> PostingsListView<'txn> {
 
     pub fn range(&self, range_offset: usize, range_len: usize) -> PostingsListView<'txn> {
         match self {
-            PostingsListView::Original { input, postings_list, offset, len } => {
+            PostingsListView::Original {
+                input,
+                postings_list,
+                offset,
+                len,
+            } => {
```
LegendreM, line 653: I prefer the previous state :(

LegendreM (approved):

meilisearch-core/src/bucket_sort.rs
MarinPostma: Yeah it feels weird sometimes, but overall I hope it will force us to write nicer code.

MarinPostma: + Ra hints the types now, it can be useful

meilisearch-core/src/bucket_sort.rs
```diff
@@ -405,12 +466,20 @@ fn cleanup_bare_matches<'tag, 'txn>(
...
+    for (
+        PostingsKey {
+            query,
+            input,
+            distance,
+            is_exact,
+        },
+        matches,
+    ) in queries
+    {
```
Kerollmops, line 482:
```suggestion
    for (postings_key, matches) in queries {
        let PostingsKey { query, input, distance, is_exact } = postings_key;
```

Kerollmops (changes requested):

qdequele:
2 things for me, I hope they have config for that. 
- Do not split macros arguments 
```rust
// keep
let path = concat!(env!("CARGO_MANIFEST_DIR"), "/../datasets/movies/movies.json");
// and not
let path = concat!(
    env!("CARGO_MANIFEST_DIR"), 
    "/../datasets/movies/movies.json"
);
```
- Do not split destructuring patern
```rust
// keep
let QueryResult { mut docids, queries } = traverse_query_tree(reader, &context, &operation)?;
// and not
let QueryResult {
    mut docids,
    queries,
} = traverse_query_tree(reader, &context, &operation)?;
```

MarinPostma:
@qdequele I went throught [rustfmt's settings](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#configuration-file-resolution), and I think we can work out solutions. I'll try things out next week.

PR #931, opened 2020-08-25
Replicated meilisearch

MarinPostma:
# wip

## Cluster formation and dynamic membership changes

Before starting, raft nodes will attempt to join other node to form a cluster.  A cluster is defined by its name, and represent a set of nodes working together thanks o the raft consensus algorithm. During it's lifetime, the cluster will go through differents states which need to be handled. Theere are three main states that need to be handled, which are: peer discovery, cluster initialization, and dynamic membership change.

### Peers discovery

Before a node can form (or be part) of a cluster, it needs to discover other peers. Meilisearch raft uses mDNS for that purpose. mDNS allows to advertize and discover services on a local network in a serverless fashion, thanks to the DNS protocol and multicast sockets.  When a node first comes online, it will start discovering and advertizing a service. This service is in the form `_meilisearch_<cluster-name>._tcp.local`.
Specifying a cluster name allows different meilisearch raft clusters to exists on the same network. The mDNS server, embedded in meilisearch, will send a request at a regular interval looking for new nodes with that service (a PTR
question). When a peer receives a request for a service it advertizes, it will respond with a DNS entry containing:
- A PTR answer with the fully qualified domain name (in the form
  `<node-id>._meilisearch_<cluster-name>._tcp.local`)
- A SRV answer with the port number on which the service is available
- A A or AAAA answer with the ip of the machine with the service.

This process is done continuously, so nodes are able to discover new nodes that could join the cluster later in the lifetime of the cluster (dynamic membership changes).

Initially, when a node goes online, it does not know about the state of the other nodes in the cluster, or if a cluster exists at all. The first thing a node does when going online is then to discover as many nodes as possible for a
certain period of time. As a raft cluster need at least 3 nodes to be albe to run, this process is repeated as long as 3 nodes are not discovered. Everytime a node discovers a new node, it sends a hanshake request. This handshakes serves two pusposes:
- exchange basic information about the nodes state
- establish a bidirectionnal connection
For the handshake to proceed, the node sends it's id and state to the node it want to connects to, and gets back the node's state if there was no error.

```rust
fn handshake(id, State) -> Result<State>;
```

For the `timeout` period, and while there is less than 3 peers connected with each other, the node tries to discover all of it's peers, and adds them to a list along with their state.

In order to ensure that all nodes discover all the other node, one should chose a `timeout` periode that is orders of maginitude larger than the tine actually required fo all the peers to discover each other.

### Cluster formation

After te discovery period has ended, a node end up with a list of `(PeerId, State)`. This list is sufficient to know if a cluster is formed (as long as sufficiently many nodes were discovered). If all nodes returned a Uninitialized state, it is safe to assume that no cluster have been formed yet, so we begin the cluster initialisation process for all the nodes. From there the raft will take over, and proceed to elections.

### Dynamic membership changes

If a cluster has already been formed, at least one node will return an Initialized state. The node that wants to join the running cluster will try to join the leader and sent a join request. It will send a request to an initialized node to join the cluster, hoping for the leader to response. If the contacted node is the leader, it will perform the node addition procedure,
returning error if any occurs. If an error occurs, the node will retry to join the cluster with another peer, and repeat this process until it succeeds. If the node the request was sent to is not a leader, it sends an error to the requesting node, with the id of the known leader, along with all information it has about the current cluster.

If 2 nodes from the same cluster discover each other and no connection existed between the two, a connection is simply initialized between the two.


jessebradley817:
Lit

Makefile
```diff
@@ -0,0 +1,6 @@
+all: image
+	docker-compose up --scale raft=3 -V
+bin: 
+	cargo build --release
+image: bin
+	docker build -t marinpostma/meilisearch .
```
LegendreM, line 6: you can use `docker build -t marinpostma/meilisearch -f Dockerfile.raft .` and avoid Dockerfile renaming tricks

MarinPostma: thanks for the tip!!

liesislukas:
Looking how to run meilisearch cross-regions, any hints?  @MarinPostma @curquiza

Issue #929, opened 2020-08-25
`facetsDistribution` counts don't take into account filters


lerouxrgd:
**Describe the bug**
The `facetsDistribution` counts don't take into account `filters`.

**To Reproduce**
Compare the `facetsDistribution` of a query with and without `filters`.

**Expected behavior**
The `facetsDistribution` counts should change based on the `filters`.

MarinPostma:
Hello @lerouxrgd, we are aware of this. In fact there is already an open pr that tries to partially address this issue (#849).

There are also two fields in the search response (`exhaustiveNbHits` and `exhaustiveFacetsCount`), whose intent is to tell  you if the returned count are approximate of not. Let me explain a bit further:
When you perform a search  a search, a collection of relevant documents is lazily created from a set of candidates documents; that means that not all documents are procesed, but only enough to return you the number of documents you asked. That means that in most cases it is impossible to return an exact count, because that would mean we would have to process all the documents.
Now there are still some things that can be done, and that are not yet implemented:
- remove the filtered documents from the count, to return a better approximation of the number of documents (this is what #849 does in part)
- when the number of candidates documents is small enough, process them all and return an exhaustive count (setting the relevant fields to `true` in the response)
- do some clever statistics on the database, and return even more accurate results (maybe in more distant future)

I hope this helps you better understand the what's going on under the hood.

hadjedjvincent:
Hi @MarinPostma I’ve a question regarding the facets distributions. It looks like that the count for each criterions is missing except for the one that are currently selected.
Is it a known bug or is there something I have not configured correctly ?
It looks like after at least one criterion is selected, the group is being calculated as "AND" rules between criterions. Whereas most of the time, we want to do "OR" between criterions of the same facet, then "AND" between each facets.

I’ve tried latest release 0.13, also compiled from master (0.14) and also compiled using PR #849, the result is the same (i'm only talking about facets, not filters ;)

MarinPostma:
Hello @hadjedjvincent, I am not completely sure to understand what you mean by "currently selected criterion", but the only thing that facet distribution does is returning an approximation of the number of documents that belong to each variant of the facet categories you used in your facet filter. If a document belongs to 2 categories, then it will be counted twice, one for each category.

hadjedjvincent:
@MarinPostma 
> If a document belongs to 2 categories, then it will be counted twice, one for each category.

Yes, i perfectly understand this :)
[...]

PR #944, opened 2020-09-01
Fix facet count

MarinPostma:
fix bug reported in: https://github.com/meilisearch/MeiliSearch/issues/929#issuecomment-683683728


LegendreM (approved):

curquiza:
Is this feature still in the milestone for the next release? Could we remove it if it's not?

Issue #945, opened 2020-09-01
facetsDistribution does not take others filters & facets in consideration


hadjedjvincent:
**Describe the bug**

As discussed into #929 , `facetsDistribution` doesn't take into consideration others applied facets.
Let's assume that we have a catalog of Dresses and we are using two facets (Color & Size).

If I select Red & Blue facets, i have all my dresses that have Red **OR** Blue color.
Now, I would want to filter theses dresses by Size, but only for Red **OR** blue colors.

If I provide the right facets while searching (red **OR** blue colors), the facets distribution is unchanged for any others facets than "Color". Meaning that I have all sizes related to the search query, and not the search query + current facets.

**Expected behavior**

`facetsDistribution` should take any previous applied filters or facets.

LegendreM:
Sorry I don't understand your need.

Lets imagine that we sell t-shirts. I have different colors (red, blue, green) and several sizes (S, M, L). the global distribution would be:
- S
  - red: 7
  - blue: 12
  - green: 5
- M
  - red: 17
  - blue: 32
  - green: 15
- L
  - red: 3
  - blue: 1
  - green: 4

now I make several request:

1) `/search?facetFilters=[["color:red", "color:blue"]]` (OR)
2) `/search?facetFilters=["color:red", "color:blue"]` (AND)
3) `/search?facetFilters=["size:S", ["color:red", "color:blue"]]`
4) `/search?facetFilters=["color:red", ["size:M", "size:L"]]`


what would be for you the facet count for each request ? 

thanks for your response :)

hadjedjvincent:
`facetsDistribution` for Size facet should be:
But, we have to take in consideration the primary key of a product, this is just an example, assuming each variant as a different ID here:

case 1:
Red = 7 + 17 + 3 = 27
Blue = 12 + 32 + 1 = 45

S = 7 + 12 = 19
M = 17 + 32 = 49
L = 3 + 1 = 4

case 2 (if we assume that a t-shirt can only have one color):
Red = 0
Blue = 0
S = 0
M = 0
L = 0

case 3:
Red = 7
Blue = 12
S = 7 + 12 = 19
M = 17 + 32 = 49
L = 3 + 1 = 4

case 4:
Red = 20
S = 7
M = 17
L = 3

The goal is to calculate how many documents would be available for each facet, if we apply facetFilters (and filters) on each of them.
Else, this would end to a possible case where you have no results when combining multiple facets (at least by user interaction)

[...]

PR #976, opened 2020-09-24
Revert 944

MarinPostma:
revert #944 
@bidoubiwa  @curquiza @eskombro, this was a misunderstanding from our side. Doing this would in fact be an error, and would prevent us to do this: https://github.com/meilisearch/MeiliSearch/issues/945#issuecomment-685526678, which is what we are really after. We are resetting this to its default behaviour before it goes in prodution. Sorry for the confusion.


Kerollmops (approved):
Thank yçu, no worries, that's on us! :)

PR #997, opened 2020-10-08
fix(core): fix benchmark in core with types

### LINKED ISSUE #994: search_benchmark is broken
**Describe the bug**
The MeiliSearch Core `search_benchmark` is broken, this is easily fixed [here](https://github.com/meilisearch/MeiliSearch/blob/b406b6ee44353812314877ebabd30dc54492a70f/meilisearch-core/benches/search_benchmark.rs#L101), by dereferencing and forcing it into an Option, 

**To Reproduce**
```
cargo bench -p meilisearch-core
```

**Expected behavior**
The benchmark to run correctly

**Additional context**
Also, the benchmarking setup doesn't currently take advantage of criterion's black_box function, which allows for better benchmarking of software, I can create these changes in another PR rather than here!

neeldug:
forces a dereference onto query and then creates an option to wrap the
query

Closes #994

### DIFF
--- a/meilisearch-core/benches/search_benchmark.rs
+++ b/meilisearch-core/benches/search_benchmark.rs
@@ -15,7 +15,7 @@ use meilisearch_core::settings::{Settings, SettingsUpdate};
 use meilisearch_schema::Schema;
 use serde_json::Value;
 
-use criterion::{criterion_group, criterion_main, Criterion, BenchmarkId};
+use criterion::{black_box, criterion_group, criterion_main, Criterion, BenchmarkId};
 
 fn prepare_database(path: &Path) -> Database {
     let database = Database::open_or_create(path, DatabaseOptions::default()).unwrap();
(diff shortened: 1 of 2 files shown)


LegendreM:
bors try

meilisearch-core/benches/search_benchmark.rs
```diff
@@ -15,7 +15,7 @@ use meilisearch_core::settings::{Settings, SettingsUpdate};
 use meilisearch_schema::Schema;
 use serde_json::Value;
 
-use criterion::{criterion_group, criterion_main, Criterion, BenchmarkId};
+use criterion::{black_box, criterion_group, criterion_main, Criterion, BenchmarkId};
```
LegendreM, line 18: hello @neeldug  could you revert this line?

LegendreM (changes requested):

meilisearch-core/benches/search_benchmark.rs
neeldug: Yep, my bad, was trying out black_box for more accurate benchmarking forgot to remove it from the PR.

LegendreM (approved):
bors r+

PR #1001, opened 2020-10-09
Fix settings bug

### LINKED ISSUE #942: Some Settings seem to be overwritten when documents are pushed after updating settings
**Describe the bug**
Some settings seem to be overwritten when documents are pushed after updating settings:
- `searchableAttributes`
- `displayedAttributes`

**To Reproduce**
I made a test on this branch:
https://github.com/LegendreM/MeiliSearch/commit/2d109cabbc798003f705a1f02d03ba0e6973f22e

**Expected behavior**
`searchableAttributes` and `displayedAttributes`should be unchanged.

**Additional context**
related to #891

MarinPostma:
fix #942, see https://github.com/meilisearch/MeiliSearch/issues/942#issuecomment-706266440

### DIFF
--- a/meilisearch-core/src/serde/deserializer.rs
+++ b/meilisearch-core/src/serde/deserializer.rs
@@ -94,6 +94,8 @@ impl<'de, 'a, 'b> de::Deserializer<'de> for &'b mut Deserializer<'a> {
                 };
 
                 let is_displayed = self.schema.is_displayed(attr);
+                // Check if displayed fields were requested, if yes, return only displayed fields,
+                // else return all fields
                 if !self.displayed || (is_displayed && self.fields.map_or(true, |f| f.contains(&attr))) {
                     if let Some(attribute_name) = self.schema.name(attr) {
                         let cursor = Cursor::new(value.to_owned());
--- a/meilisearch-core/src/serde/deserializer.rs
+++ b/meilisearch-core/src/serde/deserializer.rs
@@ -55,6 +55,7 @@ pub struct Deserializer<'a> {
     pub documents_fields: DocumentsFields,
     pub schema: &'a Schema,
     pub fields: Option<&'a HashSet<FieldId>>,
+    pub displayed: bool,
 }
 
 impl<'de, 'a, 'b> de::Deserializer<'de> for &'b mut Deserializer<'a> {
@@ -93,7 +94,7 @@ impl<'de, 'a, 'b> de::Deserializer<'de> for &'b mut Deserializer<'a> {
                 };
 
                 let is_displayed = self.schema.is_displayed(attr);
-                if is_displayed && self.fields.map_or(true, |f| f.contains(&attr)) {
+                if !self.displayed || (is_displayed && self.fields.map_or(true, |f| f.contains(&attr))) {
                     if let Some(attribute_name) = self.schema.name(attr) {
                         let cursor = Cursor::new(value.to_owned());
                         let ioread = SerdeJsonIoRead::new(cursor);
--- a/meilisearch-schema/src/schema.rs
+++ b/meilisearch-schema/src/schema.rs
@@ -16,14 +16,6 @@ impl<T> OptionAll<T> {
         std::mem::replace(self, OptionAll::None)
     }
 
-    fn map<U, F: FnOnce(T) -> U>(self, f: F) -> OptionAll<U> {
-        match self {
-            OptionAll::Some(x) => OptionAll::Some(f(x)),
-            OptionAll::All => OptionAll::All,
-            OptionAll::None => OptionAll::None,
-        }
-    }
-
     pub fn is_all(&self) -> bool {
         matches!(self, OptionAll::All)
     }
@@ -44,7 +36,7 @@ pub struct Schema {
     displayed: OptionAll<HashSet<FieldId>>,
 
     indexed: OptionAll<HashSet<FieldId>>,
-    indexed_map: HashMap<FieldId, IndexedPos>,
+    fields_position: HashMap<FieldId, IndexedPos>,
 }
 
 impl Schema {
@@ -68,7 +60,7 @@ impl Schema {
             ranked: HashSet::new(),
             displayed: OptionAll::All,
             indexed: OptionAll::All,
-            indexed_map,
+            fields_position: indexed_map,
         }
     }
 
@@ -109,15 +101,15 @@ impl Schema {
         self.fields_map.insert(name)
     }
 
-    pub fn insert_and_index(&mut self, name: &str) -> SResult<FieldId> {
+    pub fn register_field(&mut self, name: &str) -> SResult<FieldId> {
         match self.fields_map.id(name) {
             Some(id) => {
                 Ok(id)
             }
             None => {
                 let id = self.fields_map.insert(name)?;
-                let pos = self.indexed_map.len() as u16;
-                self.indexed_map.insert(id, pos.into());
+                let pos = self.fields_position.len() as u16;
+                self.fields_position.insert(id, pos.into());
                 Ok(id)
             }
         }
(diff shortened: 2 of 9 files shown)


MarinPostma:
bors try

meilisearch-core/src/serde/deserializer.rs
```diff
@@ -93,7 +94,7 @@ impl<'de, 'a, 'b> de::Deserializer<'de> for &'b mut Deserializer<'a> {
                 };
 
                 let is_displayed = self.schema.is_displayed(attr);
-                if is_displayed && self.fields.map_or(true, |f| f.contains(&attr)) {
+                if !self.displayed || (is_displayed && self.fields.map_or(true, |f| f.contains(&attr))) {
```
LegendreM, line 97: could you add a comment explaining this if statement please

meilisearch-schema/src/schema.rs
```diff
@@ -159,18 +161,18 @@ impl Schema {
...
 
-    pub fn indexed(&self) -> Cow<[FieldId]> {
+    pub fn indexed(&self) -> Vec<FieldId> {
         match self.indexed {
-            OptionAll::Some(ref v) => Cow::Borrowed(v),
+            OptionAll::Some(ref v) => v.iter().cloned().collect(),
             OptionAll::All => {
                 let fields = self
-                    .fields_map
+                    .indexed_map
```
Kerollmops, line 169: Why have you changed that? Was that a bug in the previous version?

meilisearch-schema/src/schema.rs
```diff
@@ -204,15 +206,16 @@ impl Schema {
     pub fn set_indexed(&mut self, name: &str) -> SResult<(FieldId, IndexedPos)> {
         let id = self.fields_map.insert(name)?;
 
+        self.indexed = self.indexed.take().map(|mut v| {
+            v.insert(id);
+            v
+        });
+
```
Kerollmops, line 213:
```suggestion
        if let Some(mut v) = self.indexed {
            v.insert(id);
        }

```

meilisearch-schema/src/schema.rs
```diff
@@ -260,7 +263,13 @@ impl Schema {
             self.indexed = match self.indexed.take() {
                 // valid because indexed is All and indexed() return the content of
                 // indexed_map that is already updated
-                OptionAll::All => OptionAll::Some(self.indexed().into_owned()),
+                OptionAll::All => {
+                    let indexed = self.indexed_map
+                        .iter()
+                        .filter_map(|(&i, _)| if i != id { Some(i) } else { None })
+                        .collect();
+                    OptionAll::Some(indexed)
```
Kerollmops, line 271: Are you just search for `id` in this `HashSet` ? Why not `get`ing it?

Kerollmops (changes requested):

meilisearch-schema/src/schema.rs
MarinPostma: this is strictly equivalent in this context, but it seemed easier to understand this way

meilisearch-schema/src/schema.rs
Kerollmops: Whoops, no, you are search for all others, sorry, closing!

MarinPostma: no, I am collecting all the fields in the indexed_map (`HashMap`) except the `id` one, and collect all this in a `HashSet`. I could collect all in an `HashSet` too and `HashSet::remove(id)` too:

```rust
                    let mut indexed: HashSet<_> = self.indexed_map
                        .keys()
                        .cloned()
                        .collect();
                    indexed.remove(&id);
                    OptionAll::Some(indexed)
```
do you prefer that?

Kerollmops: I would maybe prefer you just `filter` on `keys` instead of a `filter_map`.

```rust
self.indexed_map
                        .keys()
                         .filter(|i| i != id)
                        .cloned()
                        .collect();
```

Kerollmops (changes requested):
This is good, just use a filter instead and it's good for me :)

meilisearch-schema/src/schema.rs
MarinPostma: well well, turns out it isn't necessary:
```rust
            self.indexed_map.remove(&id);
```
is done just before

meilisearch-schema/src/schema.rs
```diff
@@ -220,56 +220,6 @@ impl Schema {
...
-                },
-                OptionAll::Some(mut v) => {
-                    v.retain(|x| *x != id);
-                    OptionAll::Some(v)
-                }
-                OptionAll::None => OptionAll::None,
-            }
-        }
-    }
-
```
MarinPostma: turns out it's dead code haha

Kerollmops: Really, how is it possible?

MarinPostma: those functions are too public, so rustc doesn't complain, we should use `pub(crate)` for everything that isn't exported out of the crate, I suppose

Kerollmops: Okep, are they used by meilisearch-core or meilisearch-http, so?

MarinPostma: nope it compiles and test pass

Kerollmops (changes requested):
Seems good to me! Could you rebase or will bors do it by itself?

meilisearch-schema/src/schema.rs
```diff
@@ -118,8 +110,10 @@ impl Schema {
                 Ok(id)
             }
             None => {
-                self.set_indexed(name)?;
-                self.set_displayed(name)
+                let id = self.fields_map.insert(name)?;
+                let pos = self.indexed_map.len() as u16;
+                self.indexed_map.insert(id, pos.into());
+                Ok(id)
```
LegendreM, line 116: why `self.indexed` is no more updated here ?

MarinPostma: this method is called whenever a new field is found. I want to add it to the known fields (`field_map`), and generate an indexed position for it, by inserting it in `indexed_map`. if I added it to  `indexed` too, then it would be considered an indexed field and this is what we do not want. I guess we could rename this method `add_known_field`, what do you think?

Kerollmops: So what's the purpose of `indexed_map`? Why this name if this is not here to store the indexed fields?

Kerollmops: Yeah, we can maybe find a better name than `insert_and_index`, indeed.

LegendreM: I don't understand everything.

For a user we have 3 states for an attribute:
`indexed`: indexed in meilisearch but not used 
`displayed`: displayed in a search result
`searchable`: used when I search something

Internally their are:
`fields_map`
`indexed_map`
`indexed`
`displayed_map`

what is what ? 🤔

MarinPostma:
- rename `indexed_map` into `field_positions`
- rename `insert_and_index` into `register_field`

meilisearch-schema/src/schema.rs
```diff
@@ -158,7 +158,7 @@ impl Schema {
             OptionAll::Some(ref v) => v.iter().cloned().collect(),
             OptionAll::All => {
                 let fields = self
-                    .indexed_map
+                    .fields_position
                     .iter()
                     .map(|(&f, _)| f)
```
Kerollmops: Is it just `fields_position.keys().cloned().collect()` here?

Kerollmops (changes requested):

meilisearch-schema/src/schema.rs
MarinPostma: yes that's right, fixed

Kerollmops (approved):
Thank you, seems good to me! :)

PR #1008, opened 2020-10-13
Dump info

### LINKED ISSUE #988: Rename MEILI_DUMPS_FOLDER environment variable
There has been a few discussions about the environment variable name `MEILI_DUMPS_FOLDER`. This issue's goal is to create a discussion and share everyone's point of view and arguments, so we can reach a consensus about changing it or keeping it

Please, feel free to express your ideas

:)

### LINKED ISSUE #998: Update of the dump status body
# Summary

The body of the dump status should be improved.

# Motivations

For the moment the body has two informations:
- the `dumps_uid` written as `uid`
- the status of the given `dumps_uid`

The status can have 3 values `processing`, `dump_process_failed` and `done`. There is a lack of consistency between the status itself, and between these 3 status and the status we use in the `indexes/:index/updates` route. 
Because we want generic status that we will use in any other futur sub `status` route, these are the suggested status: `in_process`, `failed`, `done`.

`failed` is also the one used in the `/indexes/:index/updates` route. Which makes it possible to make a check on the `failed` status in every asynchronous update in MeiliSearch.

In case the status returns `failed`, we should provide the full error body present in every other failing-case in MeiliSearch. Which contains the following keys: `message`, `errorCode`, `errorType`, `errorLink`

Why are we doing this? What use cases does it support? What is the expected outcome?

# Explanation

Explain the proposal like it was the final documentation of this proposal.

- What is changing for end-users.
The status key in the status body in the dump route used to have 3 possible outcomes: `processing`, `dump_process_failed` and `done`. With this change they will have these possible outcomes: `in_progress`, `failed`, `done`.

Also, in case it fails, the error_code was previously given as a `status`. With these changes, the error information will be added in the body of the status, this way: 
```json
{
    "uid": "20201008-105011897",
    "status": "failed",
    "message": "Another dump is already in progress",
    "errorCode": "dump_already_in_progress",
    "errorLink": "https://docs.meilisearch.com/errors#dump_already_in_progress"
}
```
- How it works.
`GET /dumps/:dumps_uid/status`

- What is breaking?
The status codes of the dumps status route.

- Examples.

```bash
$ curl -X GET 'http://localhost:7700/dumps/123456-456/status'
```

```json
{
    "uid": "20201008-105011897",
    "status": "failed",
    "message": "Dump process failed",
    "errorCode": "dump_process_failed",
    "errorLink": "https://docs.meilisearch.com/errors#dump_process_failed"
}
```

# Implementation

// @LegendreM 

# Related

#991 and #995 

## Todo

- [x] Change status name to `done`, `in_progress` and `failed`
- [x] Add error body in case of failure
- [x] Test added
- [ ] [Documentation](https://github.com/meilisearch/documentation/issues/#xxx) //Change xxx or remove the line

### LINKED ISSUE #1009: Snapshot redesign
# Summary

Snapshots design could have some improvements in CLI options and in error messages 

# CLI

1) Add a boolean option to activate or deactivate snapshot scheduling instead of using `--snapshot-path` as defined:
`--schedule-snapshot` 

2) rename `--load-from-snapshot` to `--import-snapshot` to be consistent with dumps

3) rename `--snapshot-path` to `--snapshot-dir` to be consistent with dumps (related to #988)

# error message

add a tips on error message `"database already exists at data.ms"`:
`"database already exists at data.ms, try to delete it or rename it"`

# Related

- [ ] Validated by the team
  - [ ]  @bidoubiwa
  - [ ] @MarinPostma 
- [ ] [Documentation](https://github.com/meilisearch/documentation/pull/552)

### LINKED ISSUE #1010: Consider changing the extension name of dump and snapshot
For the moment when creating `dumps` or `snapshots` a file is created with a `tar.gz` extension. This leads to believe that information inside can be unzipped and used. Which they are for dumps.

This is an unwanted perk. The problem with it being a perk is that we need to maintain it. Users will get used to this perk and changes to it will have to be notified. 

> we want to be free to completely change the content or the compression without impacting users

By changing the extension for `tar.gz` to respectively `dump` and `snapshot` user are not inclined to uncompress it and use it elsewhere as those will be MeiliSearch files.

Encode the content of the file.

### LINKED ISSUE #1033: dump/snapshot env variables and options
For the moment in the dump and snapshot design we have the following `env variables`/`cli options`: 
- --schedule-snapshot
- --snapshot-dir
- --snapshot-interval-sec
- --import-snapshot
- --ignore-missing-snapshot
- --ignore-snapshot-if-db-exists
- --dumps-dir
- --import-dump
- --dump-batch-size

There are two inconsistent cases within these options: 

1. If the directory of a snapshot is given but the snapshot is not activated => `--snapshot-dir` but not `--schedule-snapshot` 
    => nothing happens. **No error is thrown**.
2. If `--ignore-snapshot-if-db-exists` and `--ignore-missing-snapshot` are given but not `--import-snapshot`
    => **Error is thrown** because those options can only be added when a snapshot is imported

### 1. Both behaviors should be consistent. 

**Option A**:  In both cases, no error is thrown but a warning is given
**Option B**. In both cases, we stop MeiliSearch and throw an error.
**Option C**. In both cases, no error is thrown and no warning is given.
**Option D**. Both behaviors should not be consistent, the way it is right now is good.

### 2. Should all options have an env variable equivalent

**Option A**: yes

**Option B**: no, in this case, all options that are **only** used on launch should never have an ENV_VAR equivalent.
This means: 
- --import-snapshot
- --import-dump
- --ignore-missing-snapshot
- --ignore-snapshot-if-db-exists

Will have NO corresponding:
- MEILI_IMPORT_SNAPSHOT
- MEILI_IMPORT_DUMP
- MEILI_IGNORE_MISSING_SNAPSHOT
- MEILI_SNAPSHOT_IF_DB_EXISTS

**Option C**: No, options that generate errors should never have an ENV_VAR equivalent 
This means: 
- --ignore-missing-snapshot
- --ignore-snapshot-if-db-exists

Will have NO corresponding:
- MEILI_IGNORE_MISSING_SNAPSHOT
- MEILI_SNAPSHOT_IF_DB_EXISTS

### 3. Should options removing error exists?
 In the case of the following errors: 
- --ignore-missing-snapshot
- --ignore-snapshot-if-db-exists

The user might use then in the following scenario: 
```bash
./meilisearch --import-snapshot bad_path/snapshot --ignore-missing-snapshot=true
``` 
In which case MeiliSearch will **not** import because he could not find a snapshot. But MeiliSearch will also **not** throw an error because of the `--ignore-missing-snapshot` option. Resulting in MeiliSearch not doing anything differently than if you had NOT given any of those two options.

Should the user not create their own manual or automatic check with another script instead of using options that remove usage obligations?

**Option A**: Yes. In which case we remove the following options and their env variable.
- --ignore-missing-snapshot
- --ignore-snapshot-if-db-exists

**Option B**: No. Nothing changes.

## How to Vote

Please state the problem number and the solution letter. Example:

- 1 - B
- 2 - A
- 3 - B

Feel free to suggest add any other solution or any other problem.

LegendreM:
fix #998 
fix #988 
fix #1009
fix #1010
fix #1033


meilisearch-http/src/option.rs
```diff
@@ -109,16 +109,16 @@ pub struct Opt {
     pub ignore_snapshot_if_db_exists: bool,
 
     /// Defines the directory path where meilisearch will create snapshot each snapshot_time_gap.
-    #[structopt(long, env = "MEILI_SNAPSHOT_PATH")]
-    pub snapshot_path: Option<PathBuf>,
+    #[structopt(long, env = "MEILI_SNAPSHOT_DIR")]
+    pub snapshot_dir: Option<PathBuf>,
 
     /// Defines time interval, in seconds, between each snapshot creation.
     #[structopt(long, requires = "snapshot-path", env = "MEILI_SNAPSHOT_INTERVAL_SEC")]
```
eskombro, line 116: snapshot-path seems to have changed to snapshot-dir, not sure. Can youy check please? :)

eskombro: Otherwise, averything LGTM! 🎉

eskombro: I mean L116 in the `requires`

eskombro: Oh it is fixed, thanks! :)

meilisearch-http/tests/documents_add.rs
```diff
@@ -187,7 +187,7 @@ async fn add_document_with_long_field() {
...
...ue).\n  headers: Any additional headers to apply to static assets. This section is a mapping of header names to header values. Responses from the application aren&rsquo;t affected, to avoid overlap with the application&rsquo;s own ability to include custom headers in the response.\n  rules: Specific overrides for a specific location. The key is a PCRE (regular expression) that is matched against the full request path.\n  request_buffering: Most application servers do not support chunked requests (e.g. fpm, uwsgi), so Platform.sh enables request_buffering by default to handle them. That default configuration would look like this if it was present in .platform.app.yaml:\nweb:locations:&#39;/&#39;:passthru:truerequest_buffering:enabled:truemax_request_size:250mIf the application server can already efficiently handle chunked requests, the request_buffering subkey can be modified to disable it entirely (enabled: false). Additionally, applications that frequently deal with uploads greater than 250MB in size can update the max_request_size key to the application&rsquo;s needs. Note that modifications to request_buffering will need to be specified at each location where it is desired.\n ",
```
Kerollmops, line 190: Your sed/vscode find and replace, touched this line, and it should not had done that. sori.

Kerollmops (changes requested):

Kerollmops (approved):
bors r+

Kerollmops (approved):
bors r+

Issue #1019, opened 2020-10-15
Improve Dumps


theswerd:
**Is your feature request related to a problem? Please describe.**
When I create Meilisearch dumps, I don't want to store them on the container/volume/computer that I am dumping from, I want to export them. I would like more tools for exporting dumps.

**Describe the solution you'd like**
1. The ability to add API Hooks to the events in the process (Dump started, Dump finished, Dump in progress)
2. The ability to set dumps to run at a set interval
3. The ability to export to an external URL rather than a local folder

**Describe alternatives you've considered**
For now I am looking into build a script that gets a dump once a day and exports it.

**Other Info**
- [How Postgres Covers Dumps](https://www.postgresql.org/docs/current/backup-dump.html)

Ps. I love using MeiliSearch!

LegendreM:
related to #1023

yashiroiori:
> **Is your feature request related to a problem? Please describe.**
> When I create Meilisearch dumps, I don't want to store them on the container/volume/computer that I am dumping from, I want to export them. I would like more tools for exporting dumps.
> 
> **Describe the solution you'd like**
> 
> 1. The ability to add API Hooks to the events in the process (Dump started, Dump finished, Dump in progress)
> 2. The ability to set dumps to run at a set interval
> 3. The ability to export to an external URL rather than a local folder
> 
> **Describe alternatives you've considered**
> For now I am looking into build a script that gets a dump once a day and exports it.
> 
> **Other Info**
> 
> * [How Postgres Covers Dumps](https://www.postgresql.org/docs/current/backup-dump.html)
> 
> Ps. I love using MeiliSearch!

About sort, I think make similar to Algolia make multiples index with each sort a need

Excellent, thank you

qdequele:
Hello, Thanks for this feature proposal! 

I will close this issue in favor of our [public roadmap](https://roadmap.meilisearch.com/c/66-download-dumps?utm_medium=social&utm_source=portal_share). 

I invite everyone interested in this feature to update it on the roadmap.

PR #1022, opened 2020-10-15
Stream dump

ppamorim:
# Description

Add a route that allows the download of a dump stored in the server. This function is interesting when using the server in conjunction with the official SDKs and other tools. This allows the manager to fetch the document by simply calling the endpoint.

## Test

If you want to test this feature, you first need to start the `MeiliSearch` instance:

```
cargo run
```

Then you need to request the server to create a dump:

```
curl -X POST 'http://localhost:7700/dumps'
```

Copy the `uid`.

You can keep checking the status of the dump by calling:

```
curl -X GET 'http://localhost:7700/dumps/20201101-110357260/status'
```

Then finally when the dump is processed, download the dump:

```
curl -X GET 'http://localhost:7700/dumps/20201101-110357260' --output dump.tar.gz
```

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

## Checklist

- [X] Implement the fetch dump route
- [X] Write unit test
- [ ] Stress test (large file)
- [ ] Update documentation
- [ ] Update changelog 

## Related

#1023
#1019


ppamorim:
@LegendreM The test is supposed to work but the file `assets/dumps/v1/test/documents.jsonl` didn't exist when I created the branch.

meilisearch-http/src/routes/dump.rs
```diff
@@ -62,3 +67,54 @@ async fn get_dump_status(
...
+                if count > 0 {
+                    Poll::Ready(Some(Ok(Bytes::copy_from_slice(&self.buffer[..count]))))
+                } else {
+                    Poll::Ready(None)
+                }
+            },
+            Err(e) => Poll::Ready(Some(Err(Error::dump_read_failed(e).into())))
+        }
+    }
+}
```
MarinPostma, line 106: this is blocking, you must use `poll_read` on tokio::fs::File instead.

LegendreM:
Hello @ppamorim I created a more specific issue (#1023) related to your work, I think this feature is relevant if the user can upload dumps too.

ppamorim:
@MarinPostma I partially transformed it to `tokio::fs::File` but I got stuck in the `stream.file.read(&mut stream.buffer).await` inside the `poll_next` function. I think we will need to use the macro `async_stream::stream` but I am not sure how to do this.

meilisearch-http/src/routes/dump.rs
```diff
@@ -92,16 +96,16 @@ impl tokio::stream::Stream for StreamBody {
...
+        match stream.file.read(&mut stream.buffer).await {
+            Ok(count) => {
+                if count > 0 {
+                    Poll::Ready(Some(Ok(Bytes::copy_from_slice(&self.buffer[..count]))))
+                } else {
+                    Poll::Ready(None)
+                }
+            },
+            Err(e) => Poll::Ready(Some(Err(Error::dump_read_failed(e).into())))
+        }
```
MarinPostma, line 108: that's not right indeed. You are trying to use await out of an async context. This is a lower level of abstraction, and we are getting into the internal functioning of futures in rust. I would suggest your read https://cfsamson.github.io/books-futures-explained/ in order to get a better understanding of how it works.

In the mean time, what you want to use is [this](https://docs.rs/tokio/0.2.22/tokio/io/trait.AsyncRead.html#tymethod.poll_read)

MarinPostma: I haven't tested it, but something like this could work:
```rust
impl tokio::stream::Stream for StreamBody {
    type Item = Result<Bytes, ResponseError>;
    fn poll_next(
        mut self: Pin<&mut Self>,
        cx: &mut Context<'_>,
    ) -> Poll<Option<Self::Item>> {
        //  you can't hold to mut refs to self, need to take ownership on the buffer for some time
        let mut buffer = std::mem::replace(&mut self.buffer, vec![]);
        let file = &mut self.file;
        pin_utils::pin_mut!(file);
        match file.poll_read(cx, &mut buffer) {
            Poll::Ready(Ok(size)) => {
                if size > 0 {
                    // place it back when done with it
                    let _ = std::mem::replace(&mut self.buffer, buffer);
                    Poll::Ready(Some(Ok(Bytes::copy_from_slice(&self.buffer[..size]))))
                } else {
                    Poll::Ready(None)
                }
            }
            Poll::Ready(Err(e)) => Poll::Ready(Some(Err(Error::Internal(e.to_string()).into()))),
            Poll::Pending => Poll::Pending,
        }
    }
}
```

MarinPostma: But, just for sake of clarity, this is a feature we have thought about, but we are still not sure we want to have it, there are many ongoing discussion about that. I'm happy to give you mentorship on this, and get you going, but keep in mind that nothing is decided about the future of this feature :)

ppamorim: @MarinPostma I had to change your code a bit, it was not working because `vec![]` was causing the second poll to return an empty vec. I also replaced  `pin_utils::pin_mut!(file);` with `Pin::new(&mut self.file);` as suggested by @Koxiaet

MarinPostma:
hey @ppamorim, I'm closing this for now, thanks for the contribution. We will  probably reconsider this in another form in the future. cheers ;)

Issue #1030, opened 2020-10-21
movies.json has two different fields "genre" and "genres"


curquiza:
Following this comment https://github.com/meilisearch/MeiliSearch/commit/d937aeac0a27ef21873b1a186ce2270968621815#commitcomment-43423208

We might check that we don't have the same issue on the other datasets 🙂

bidoubiwa:
It should be changed to genres everywhere.

bidoubiwa:
You can assign me on this

curquiza:
It's fixed now, closing

PR #1031, opened 2020-10-21
Change all genre fields to genres

bidoubiwa:
Fixes: meilisearch/MeiliSearch#1030

In the `movies.json` dataset some fields where still "genre" instead of "genres". 
They are now all called `genres`.


LegendreM:
bors try

MarinPostma:
@bidoubiwa I only see one line of change, is this ok? can I merge this?

bidoubiwa:
It's because in the other PR where I changed the URL's I used this updated json.
I looked at the change files and it is weird that it only removes the first line. This is not what we want. Thanks for checking

PR #1032, opened 2020-10-21
Remove not maintained csv movies dataset

bidoubiwa:
Remove `movies.csv` from the dataset folder as it is not updated and not usable with MeiliSearch without converting it to json.


LegendreM:
bors try

MarinPostma (approved):
bors r+

PR #1036, opened 2020-10-22
fix: cors error since v0.15

### LINKED ISSUE #1021: Cors error returned instead of authentication_error on wrong API key in browser
**Describe the bug**
Previously when integrate MeiliSearch in a browser script and run it without API key or with a wrong API key I would receive a `authentication_error`.
Since v0.15 i receive a CORS error.

**To Reproduce**
1. Launch a MeiliSearch instance with a masterKey
2. Use the following code in a http file (notice it does not use a master key)
```html
<!DOCTYPE html>
<html>
<head>
  <meta charset='utf-8'>
  <meta http-equiv='X-UA-Compatible' content='IE=edge'>
</head>
<body>
    Indexes:
    <div id="indexes"></div>
</body>
</html>
<script src="../../dist/bundles/meilisearch.umd.js"></script>
<script>
  const client = new window.MeiliSearch({
    host: 'http://127.0.0.1:7700',
  })
  client.listIndexes().then(res => {
    console.log({ res });
    res.map(index => document.querySelector('#indexes').innerHTML = `<div>${index.name}</div>`)
  })
</script>

```
3. Open the html file or run a http server using `http-server yourhtmlfile.html`
4. Open it on a browser and look at the console

**Expected behavior**
Return the `authentication_error` body instead of a CORS error as it not clear what failed.

```json
{
  "message": "You must have an authorization token",
  "errorCode": "missing_authorization_header",
  "errorType": "authentication_error",
  "errorLink": "https://docs.meilisearch.com/errors#missing_authorization_header"
}
```

**Screenshots**

v.0.15:
Screenshot 2020-10-15 at 14 00 37

LegendreM:
fix #1021

How to test it simply:
run meilisearch from master, and run the HTML below on navigator:
```html
<!DOCTYPE html>
<html>
<head>
  <meta charset='utf-8'>
  <meta http-equiv='X-UA-Compatible' content='IE=edge'>
</head>
<body>
    Indexes:
    <div id="indexes"></div>
</body>
</html>
<script>
    httpRequest = new XMLHttpRequest();
    httpRequest.open('GET', 'http://127.0.0.1:7700/test/search', true);
    httpRequest.send();
</script>
```

retry with fixing branch

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,14 +40,16 @@ dependencies = [
 
 [[package]]
 name = "actix-cors"
-version = "0.4.1"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f3e5c769e4d332bfad27f11b8139b5818c4bbddb02c385b8f16344d93ff1a8eb"
+checksum = "aaf0c4345c9663a2822d42602391418fd5766f269109ec6bf1784b056a9356a7"
 dependencies = [
- "actix-service",
  "actix-web",
  "derive_more",
  "futures-util",
+ "log",
+ "once_cell",
+ "tinyvec 1.0.1",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


Kerollmops (changes requested):
I am not sure the js/html script you provide is valid, the route to search is not correct and even when I change it, the previous version of the engine (master) doesn't trigger any CORS error (firefox).

MarinPostma:
@LegendreM Any update on this?

mikerogerz:
Without this fix, I only get a 400 response "Origin is not allowed to make this request." from the search server.

bidoubiwa:
In this PR a user has a similar CORS issue I think: 
https://github.com/meilisearch/meilisearch-js/issues/652

PR #1037, opened 2020-10-22
Synonym unidecode

### LINKED ISSUE #964: Accents work strangely with synonyms
**Describe the bug**
I had some issues with some of the synonyms working and some not and it seems I managed to identify a bug.
It seems words with **accents** are not used from the synonym table (and/or the query is deaccentized before looking up synonyms (?))

**To Reproduce**
Add synonyms:

["case"]: ["holder"] --> works (searches for holder with queries both "case" and "cáse"!)
["cáse"]: ["holder"] --> does not work (for either)

**Expected behavior**
In a previous issue (#949) it was mentioned there is no typo tolerance with synonyms, however it now seems when considering synonyms, there is some sort of deaccenting the query before looking up in the synonyms table (?)

MarinPostma:
fix #964 

- unidecodes all synonyms before adding them to the synonyms fst
- stores a copy of the original synonyms (unicoded) for later retrieve

### DIFF
--- a/meilisearch-core/src/store/synonyms.rs
+++ b/meilisearch-core/src/store/synonyms.rs
@@ -19,10 +19,6 @@ impl Synonyms {
         self.synonyms.put(writer, word, bytes)
     }
 
-    pub fn del_synonyms(self, writer: &mut heed::RwTxn<MainT>, word: &[u8]) -> ZResult<bool> {
-        self.synonyms.delete(writer, word)
-    }
-
     pub fn clear(self, writer: &mut heed::RwTxn<MainT>) -> ZResult<()> {
         self.synonyms.clear(writer)
     }
--- a/meilisearch-core/src/store/synonyms.rs
+++ b/meilisearch-core/src/store/synonyms.rs
@@ -12,14 +12,14 @@ pub struct Synonyms {
 }
 
 impl Synonyms {
-    pub fn put_synonyms<A>(self, writer: &mut heed::RwTxn<MainT>, word: &[u8], synonyms: &fst::Set<A>) -> ZResult<()>
+    pub(crate) fn put_synonyms<A>(self, writer: &mut heed::RwTxn<MainT>, word: &[u8], synonyms: &fst::Set<A>) -> ZResult<()>
     where A: AsRef<[u8]>,
     {
         let bytes = synonyms.as_fst().as_bytes();
         self.synonyms.put(writer, word, bytes)
     }
 
-    pub fn clear(self, writer: &mut heed::RwTxn<MainT>) -> ZResult<()> {
+    pub(crate) fn clear(self, writer: &mut heed::RwTxn<MainT>) -> ZResult<()> {
         self.synonyms.clear(writer)
     }
 
--- a/meilisearch-core/src/update/settings_update.rs
+++ b/meilisearch-core/src/update/settings_update.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet, btree_map::Entry};
 
 use heed::Result as ZResult;
 use fst::{set::OpBuilder, SetBuilder};
@@ -126,7 +126,7 @@ pub fn apply_settings_update(
     }
 
     match settings.synonyms {
-        UpdateState::Update(synonyms) => apply_synonyms_update(writer, index, synonyms)?,
+        UpdateState::Update(synonyms) => apply_synonyms_update(writer, index, transform_synonyms(synonyms))? ,
         UpdateState::Clear => apply_synonyms_update(writer, index, BTreeMap::new())?,
         UpdateState::Nothing => (),
     }
@@ -136,6 +136,23 @@ pub fn apply_settings_update(
     }
 
     Ok(())
+}
+
+fn transform_synonyms(synonyms: BTreeMap<String, Vec<String>>) -> BTreeMap<String, Vec<String>> {
+    synonyms
+        .into_iter()
+        .fold(BTreeMap::new(), |mut map, (key, values)| {
+            let deunicoded = deunicode::deunicode(&key);
+            match map.entry(deunicoded) {
+                Entry::Vacant(entry) => {
+                    entry.insert(values);
+                }
+                Entry::Occupied(mut entry) => {
+                    entry.get_mut().extend_from_slice(&values);
+                }
+            }
+            map
+        })
 }
 
 fn apply_attributes_for_faceting_update(
--- a/meilisearch-core/src/update/settings_update.rs
+++ b/meilisearch-core/src/update/settings_update.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, BTreeSet, btree_map::Entry};
+use std::collections::{BTreeMap, BTreeSet};
 
 use heed::Result as ZResult;
 use fst::{set::OpBuilder, SetBuilder};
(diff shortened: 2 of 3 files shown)


meilisearch-core/src/store/main.rs
```diff
@@ -28,6 +28,7 @@ const SCHEMA_KEY: &str = "schema";
 const SORTED_DOCUMENT_IDS_CACHE_KEY: &str = "sorted-document-ids-cache";
 const STOP_WORDS_KEY: &str = "stop-words";
 const SYNONYMS_KEY: &str = "synonyms";
+const SYNONYMS_UNICASED_KEY: &str = "synonyms-unicased";
```
qdequele, line 31: Why do you keep both?

MarinPostma: so the synonyms retrieved are the same as the one that were put in

LegendreM: I think we should transparently assume that we transform synonyms, no ?

lets say I add,
```
"synonyms": {
            "wolverine": ["xmen", "logan"],
            "wolvérine": ["filin", "toto"]
        }
```

It will be displayed as:

```
"synonyms": {
            "wolverine": ["xmen", "logan"],
            "wolvérine": ["filin", "toto"]
        }
```

but behave as:

```
"synonyms": {
            "wolverine": ["xmen", "logan", "filin", "toto"]
        }
```

am I right ?

meilisearch-core/src/store/synonyms.rs
```diff
@@ -1,40 +1,55 @@
...
+        self.synonyms_unicased.put(writer, word, &synonyms_vec)?;
+        Ok(())
     }
 
-    pub fn del_synonyms(self, writer: &mut heed::RwTxn<MainT>, word: &[u8]) -> ZResult<bool> {
-        self.synonyms.delete(writer, word)
+    pub fn del_synonyms(self, writer: &mut heed::RwTxn<MainT>, word: &str) -> ZResult<bool> {
+        let deunicoded = deunicode(word);
+        self.synonyms_unicased.delete(writer, word)?;
+        self.synonyms.delete(writer, deunicoded.as_bytes())
```
LegendreM, line 31: this function is never used

LegendreM (changes requested):

meilisearch-core/src/store/main.rs
MarinPostma: uhmm, I hadn't thought of that. It actually replace the old one, but we would expect it to behave as:
```
"synonyms": {
            "wolverine": ["xmen", "logan", "filin", "toto"]
        }
```
indeed. This does not answer your question though... What are your opinions @curquiza @qdequele @Kerollmops ?

Kerollmops: I am more on the many side here, I prefer when behaviors are explicit, and here we will 1. remove some code and 2. make it explicit to the user that we always transform syninyms.

MarinPostma: and what should be the behaviour of this?
```
"synonyms": {
            "wolverine": ["xmen", "logan"],
            "wolvérine": ["filin", "toto"]
        }
```

Kerollmops: I think we should just merge them as wolverine is the same as wolvérine for the engine, for now at least, maybe in the future we will change that (e.g. for chinese, thaï).

curquiza: Hello sorry for the delay!
I agree with Kero and Many for the moment, maybe it's better to assume this behaviorm for a first try.

MarinPostma:
final implemetation:
- the synonyms are deunicoded and merged if their deunicoded representation are equal
- the merged synonyms arrays are not sorted (this could be part of another PR, if we decide that we need it)

LegendreM (approved):

meilisearch-core/src/update/settings_update.rs
```diff
@@ -138,6 +138,23 @@ pub fn apply_settings_update(
     Ok(())
 }
 
+fn transform_synonyms(synonyms: BTreeMap<String, Vec<String>>) -> BTreeMap<String, Vec<String>> {
```
Kerollmops, line 141:
```suggestion
fn canonicalize_synonyms(synonyms: BTreeMap<String, Vec<String>>) -> BTreeMap<String, Vec<String>> {
```

meilisearch-core/src/update/settings_update.rs
```diff
@@ -138,6 +138,23 @@ pub fn apply_settings_update(
...
+        .fold(BTreeMap::new(), |mut map, (key, values)| {
+            let deunicoded = deunicode::deunicode(&key);
+            match map.entry(deunicoded) {
+                Entry::Vacant(entry) => {
+                    entry.insert(values);
+                }
+                Entry::Occupied(mut entry) => {
+                    entry.get_mut().extend_from_slice(&values);
+                }
+            }
```
Kerollmops, line 153:
```suggestion
            map.entry(deunicoded).or_default().extend_from_slice(&values);
```

meilisearch-core/src/update/settings_update.rs
```diff
@@ -138,6 +138,23 @@ pub fn apply_settings_update(
     Ok(())
 }
 
+fn transform_synonyms(synonyms: BTreeMap<String, Vec<String>>) -> BTreeMap<String, Vec<String>> {
+    synonyms
+        .into_iter()
+        .fold(BTreeMap::new(), |mut map, (key, values)| {
```
Kerollmops, line 144: I am not a big fan 💨 of this `fold`, I would highly prefer a for loop for readability but if you think I am wrong, keep it :)

Kerollmops (changes requested):

meilisearch-core/src/update/settings_update.rs
Kerollmops: And you can remove the `btree_map::Entry` import.

meilisearch-core/src/update/settings_update.rs
MarinPostma: I don't mind, I like fold, but I prefer readability

Kerollmops (approved):

meilisearch-core/src/update/settings_update.rs
```diff
@@ -138,6 +138,15 @@ pub fn apply_settings_update(
     Ok(())
 }
 
+fn canonicalize_synonyms(synonyms: BTreeMap<String, Vec<String>>) -> BTreeMap<String, Vec<String>> {
+    let mut canonicalized: BTreeMap<String, Vec<String>> = BTreeMap::new();
```
Kerollmops, line 142:
```suggestion
    let mut canonicalized = BTreeMap::new();
```

Kerollmops (changes requested):

meilisearch-core/src/update/settings_update.rs
```diff
@@ -138,6 +138,18 @@ pub fn apply_settings_update(
...
 }
 
+fn canonicalize_synonyms(synonyms: BTreeMap<String, Vec<String>>) -> BTreeMap<String, Vec<String>> {
+    let mut canonicalized = BTreeMap::new();
+    for (key, values) in synonyms {
+        let deunicoded = deunicode::deunicode(&key);
+        canonicalized
+            .entry(deunicoded)
+            .or_insert_with(Vec::new)
+            .extend_from_slice(&values);
```
Kerollmops, line 148: Wow!

bors r+

Kerollmops (approved):

PR #1040, opened 2020-10-22
Update movie posters

bidoubiwa:
This PR resolves 3 issues: 

1. update posters URLs that changed
2. All posters point to a smaller image ( +- 20kb instead of 500kb+-) this was done by changing the width size from 1280 px to 500 px. 
3. Remove films that are not in the tmdb database


Kerollmops (approved):
bors r+

PR #1044, opened 2020-10-27
Synonyms case

MarinPostma:


LegendreM:
Ok I think it's not the good way to do it, the goal of theses changes seem to have an iso state between synonyms and what is searched, we should probably have a common function called by both to `normalize` them. 🤔

PR #1053, opened 2020-10-30
Filter 0 in facetDistribution

hah:
adding the posibility to ignore/discard facet counts if the value is 0
the param is called **ignoreIfZero** it's false if not set

I haven't updated the tests, in case you are fine with
this feature, I'll update them and refactor the code abit
default example:
image

parameter set:
image

thanks


LegendreM:
Hello @hah,
you are too fast 🏃‍♂️, your feature change user documentation which needs to be discussed,
you should create an issue explaining what is the feature and why you want it in meilisearch.

thanks for your interest in our project and contribute to enhance it! 🚀

hah:
closing per #1064

PR #1054, opened 2020-10-31
Make small improvements

whoan:
Thanks for this great tool!


download-latest.sh
```diff
@@ -168,16 +168,17 @@ failure_usage() {
 
 # MAIN
 latest="$(get_latest)"
-get_os
-if [ "$?" -eq 1 ]; then
+
+if ! get_os; then
```
Kerollmops, line 172: I suppose this is fair, we were only testing that the `get_os` out was `1` instead of checking that it is simply different from `0`, so...

Kerollmops (approved):
Thank you very much for your help, this is very appreciated 👍 😃

PR #1100, opened 2020-11-23
[fix] Remove some clippy warnings

### LINKED ISSUE #1099: Setup more targets for clippy continuous integration
`cargo clippy` doesn't address lint problems in all possible Rust targets (examples, tests, etc.). We can widen the scope of `clippy`.

- Fix all current clippy problems in current other targets
- Setup in the CI the additional targets 

- I'm wondering also if we should also add a `--deny warnings`, this would be stricter but is generally sound and we can deactivate locally (with `#[allow(the_lint)]`) when one particular warning doesn't make sense.

**Additional context**
I'll propose a PR associated to this ticket.

woshilapin:
fix #1099 

I'm also wondering if I should add `-- --deny warnings` to the modified line in `test.yml`.

### DIFF
--- a/meilisearch-core/examples/from_file.rs
+++ b/meilisearch-core/examples/from_file.rs
@@ -349,8 +349,8 @@ fn search_command(command: SearchCommand, database: Database) -> Result<(), Box<
 
                 if let Some(ref filter) = command.filter {
                     let filter = filter.as_str();
-                    let (positive, filter) = if filter.chars().next() == Some('!') {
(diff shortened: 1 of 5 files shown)


LegendreM (commented):
bors try

MarinPostma (approved):
looks good to me! Thanks!!

MarinPostma:
> I'm also wondering if I should add -- --deny warnings to the modified line in test.yml.

I agree that this could be useful

bors r+

PR #1131, opened 2020-12-10
fix attributes to retrieve bug

MarinPostma:
fix bug when using empty `attributeToRetrieve`


Kerollmops (approved):
Seems good, I hope the tests passes! :)

bors r+

PR #1143, opened 2020-12-27
Add facet filtering to the development dashboard

aidanhs:
GIF preview of what this looks like when filtering on genres with the movies dataset:
ezgif-3-6ae2e49a5006

It also should degrade gracefully and essentially disappear if there are no facets (there's a small amount of dead space, but it doesn't really get in the way):
image


curquiza:
Hello @aidanhs!

This is really kind of you to suggest this kind of improvement. However, we've just hired someone who will take care of the dashboard and will improve the current one.

Thanks for your involvement anyway 🙂

PR #1164, opened 2021-01-04
change health route

### LINKED ISSUE #1144: Set the authorization token in the header
It's not always possible to set the token in the headers, can we have it also as a query param?

LegendreM:
change response code from 204 to 200
to make it work with GCP healthchecks
https://cloud.google.com/load-balancing/docs/health-checks#optional-flags-hc-protocol-http

The state of this PR is not a final state, and should be improved:
- we could create a new ENV var to let the user choose the HTTP CODE (`200`, `204` ....)
- we could detect the origin/referer of the request and if it's `GCP` return `200`

fix #1144

### DIFF
--- a/meilisearch-http/src/routes/health.rs
+++ b/meilisearch-http/src/routes/health.rs
@@ -9,5 +9,5 @@ pub fn services(cfg: &mut web::ServiceConfig) {
 
 #[get("/health")]
 async fn get_health() -> Result<HttpResponse, ResponseError> {
-    Ok(HttpResponse::NoContent().finish())
+    Ok(HttpResponse::Ok().finish())
 }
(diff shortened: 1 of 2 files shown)


bidoubiwa:
The documentation should be updated accordingly @react-learner or @LegendreM (I'm not sure who is supposed to do it @fharper ?). 
This update will not be added in this release (2021-w1) but in the next one.

fharper:
We encourage anyone to create a PR, especially when it doesn't take much time, but your priority is the code, so please at least create an issue for it if you don't do the PR and @react-learner (or anyone else in the community) will make the change.

MightySCollins:
I am not a rust developer but I would really like this implemented. What are you guys saying you would need in addition to this change? For it to be toggled based on referrer or env? I am happy to look into those if it would help

curquiza:
Closing it according to https://github.com/meilisearch/MeiliSearch/issues/1282#issuecomment-799424875

PR #1167, opened 2021-01-05
Update dumps ci

MarinPostma:
Now that the dump test are re-entrant, they can be run from a multithreaded context, whereas they used to be ran from a single threaded context, in a separate CI task.


LegendreM (approved):
bors r+

PR #1172, opened 2021-01-07
Fix atomic snapshot creation

raszi:
Compress gzip files to a temporary file first and then do an atomic rename.

In our setup we have an indexer which does snapshoting for the instances serving the requests. Since currently the snapshoting mechanism is replacing the file in place therefore the indexer could not share the snapshot with a live instance. 

With this small patch we first create a new temporary file in the same directory as the snapshot dir and then we do an atomic rename therefore the snapshot path would always contain a valid snapshot.
After applying this change it would be enough to simply restart the serving instances to pick up the new snapshot from a shared storage without worrying them to die because of an incomplete snapshot.

### DIFF
--- a/meilisearch-http/src/helpers/compression.rs
+++ b/meilisearch-http/src/helpers/compression.rs
@@ -1,19 +1,27 @@
 use flate2::Compression;
 use flate2::read::GzDecoder;
 use flate2::write::GzEncoder;
-use std::fs::{create_dir_all, File};
+use std::fs::{create_dir_all, rename, File};
 use std::path::Path;
 use tar::{Builder, Archive};
+use uuid::Uuid;
 
 use crate::error::Error;
 
 pub fn to_tar_gz(src: &Path, dest: &Path) -> Result<(), Error> {
-    let f = File::create(dest)?;
+    let file_name = format!(".{}", Uuid::new_v4().to_urn());
+    let p = dest.with_file_name(file_name);
+    let tmp_dest = p.as_path();
+
+    let f = File::create(tmp_dest)?;
     let gz_encoder = GzEncoder::new(f, Compression::default());
     let mut tar_encoder = Builder::new(gz_encoder);
     tar_encoder.append_dir_all(".", src)?;
     let gz_encoder = tar_encoder.into_inner()?;
     gz_encoder.finish()?;
+
+    rename(tmp_dest, dest)?;
+
     Ok(())
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1734,6 +1734,7 @@ dependencies = [
  "tempfile",
  "tokio",
  "ureq",
+ "uuid",
  "vergen",
  "walkdir",
  "whoami",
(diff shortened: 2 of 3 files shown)


meilisearch-http/src/helpers/compression.rs
```diff
@@ -1,19 +1,27 @@
...
+use std::fs::{create_dir_all, rename, File};
 use std::path::Path;
 use tar::{Builder, Archive};
+use uuid::Uuid;
 
 use crate::error::Error;
 
 pub fn to_tar_gz(src: &Path, dest: &Path) -> Result<(), Error> {
-    let f = File::create(dest)?;
+    let file_name = format!(".{}", Uuid::new_v4().to_urn());
```
MarinPostma, line 12: I think adding Uuid as a dependency for this purpose is a bit overkill, a timestamp should be largely enough, and chrono is already a dependency
```suggestion
    let file_name = format!(".{}", Utc::now().timestamp_nanos());
```

MarinPostma (changes requested):
Thanks for the contribution, I like the idea! there is something I would like to change though

cheers :)

meilisearch-http/src/helpers/compression.rs
raszi: I believe that is a valid concern that this introduces a new dependency (although `uuid` is already a dependency for the overall project) and of course it can be changed to the `chrono` usage.

The reason why I went with the `uuid` solution is security. The UUIDv4s cannot be guessed and an attack vector could be eliminated with this solution. If we change the implementation to the timestamp way then an attacker could create a file (or rather multiple files) and get ownership of the final snapshot file after the rename and overwrite with arbitrary content.

If you feel this is not a real concern since the snapshot directory should be protected by correct ownership then I will change the code to your suggestion of course.

MarinPostma: Well if an attacker can create an arbitrary file on your filesystem, he can also create one that is named like your database don't you think? 🙂 Imo even "temp" would be enough for this file name, but since uuid is already a dep, I don't have any strong opinion

raszi: If the snapshot file does not exist yet on the system then you are absolutely right. So what should I do? Accept the change or is it good as it is now? 😄

MarinPostma: @LegendreM what do you think?

LegendreM: Hello @raszi, thanks for your PR, I will try it locally before, but I think we will merge the current state 👍

raszi: Any update on this?

LegendreM (approved):

LegendreM:
bors try

MarinPostma:
All good! we merge that. Thanks for your contribution @raszi :)

MarinPostma (approved):

MarinPostma:
bors merge

PR #1174, opened 2021-01-08
Limit query words number

### LINKED ISSUE #941: Allow user to limit the max number of words in a query
# Summary

Be able to limit the number of word accepted in a query.

# Motivations

meilisearch is getting slower the longer the query is (#925), which induce several issues:
- a common user may not be interested by a request which the response time is long
- a malicious user could flood long request to make meilisearch hang

# Explanation

add a `max_query_word` settings to choose it

# Related

- [ ] Validated by the team
- [ ] Test added
- [ ] [Documentation](https://github.com/meilisearch/documentation/issues/#xxx) TBD
- [ ] [SDK/Integrations](https://github.com/meilisearch/integration-guides/issues/#xxx) TBD

MarinPostma:
This pr adds a limit to the number of words taken into account in a search query. Using query string that are too long leads to huge performance hits and ressources consumtion, that occasionally crashes the machine. The limit has been hard set to 10, and tests have been added to make sure that it is taken into account.

close #941

### DIFF
--- a/meilisearch-core/src/lib.rs
+++ b/meilisearch-core/src/lib.rs
@@ -39,6 +39,7 @@ pub use self::update::{EnqueuedUpdateResult, ProcessedUpdateResult, UpdateStatus
 pub use meilisearch_types::{DocIndex, DocumentId, Highlight};
 pub use meilisearch_schema::Schema;
 pub use query_words_mapper::QueryWordsMapper;
+pub use query_tree::MAX_QUERY_LEN;
 
 use compact_arena::SmallArena;
 use log::{error, trace};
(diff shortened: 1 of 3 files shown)


MarinPostma:
@react-learner what are the relevant part of the documentation I should update?

nubpro:
I understand 10 is a safe bet. Can we anticipate this hard limit be raised in the future? I worried that 10 might be too short for my use cases.

MarinPostma:
Yes why not, we could make it an option in the future, this would require a bit more thinking though.

Can you tell me more about your use case? How many words would you need?

nubpro:
> Yes why not, we could make it an option in the future, this would require a bit more thinking though.
> 
> Can you tell me more about your use case? How many words would you need?

I initially thought u meant 10 letters. (I need sleeeeep)
Oppsie, I'll come back to you when I have more to share.

Kerollmops (approved):
Great work!

MarinPostma:
bors merge

dsikes:
Thank you so much! Was hitting this very issue. After updating to the latest version, the issue was resolved. Great work!

PR #1176, opened 2021-01-09
fix race condition in  document addition

### LINKED ISSUE #1160: Updating settings on a new document can be overwritten when loading the first document
(possibly related to #1159)

**Describe the bug**
Creating a new index, setting some settings and then (without waiting for that update to complete) loading a single initial document has been observed to lose the settings updates.

**To Reproduce**
Run the test script (listed at the bottom) against a fresh meilisearch instance (I'm using `RUST_BACKTRACE=1 ./MeiliSearch/target/release/meilisearch --db-path data.ms` (I'm on a05aef5c14d445139addb5e8709acaf427cec0cd).

The script logs will look like:
```
$ python3 repro2.py
Deleting index
Making index
Response:  b'{"name":"songs","uid":"songs","createdAt":"2021-01-01T15:29:15.263684857Z","updatedAt":"2021-01-01T15:29:15.263756264Z","primaryKey":null}'
Configuring index
Response:  b'{"updateId":0}'
Beginning song load
Songs loading with update id 1
Checking settings
Response:  b'{"rankingRules":["typo","words","proximity","attribute","wordsPosition","exactness"],"distinctAttribute":null,"searchableAttributes":["*"],"displayedAttributes":["*"],"stopWords":[],"synonyms":{},"attributesForFaceting":[]}'
Song indexing complete
```

This is wrong - the resulting settings are not the ones that I set in the test script (e.g. `searchableAttributes` is `["*"]`).

**Expected behavior**
I expect updates to be commutative - the end result should be the same no matter what order I apply things in (as long as all updates are compatible with each other, and relative ordering of updates that overwrite each other are the same).

In this case, there are two settings (S1, the initial settings, and S2, my custom settings) and a doc. As long as S1 happens before S2, I expect all end results to be the same *even if* I don't wait for updates of the settings to complete.

What I think is happening here is
 - S2 begins being applied
 - the document comes in, is treated as the first document, and figures it should register the fields into the settings
 - S2 finishes being applied
 - the document finishes being added, and settings are applied based on the fields of the document

 - OS: Linux
 - Browser N/A
 - Version Ubuntu 18.04

**Additional context**
Test script (this differs from the one in #1159 by not doing anything with facets, and printing settings out afterwards):
```
#!/usr/bin/env python3

import datetime
import json
from urllib import request
import time
import sys

assert sys.version_info.major == 3

def updatesearch():
    print("Deleting index")
    r = request.Request('http://localhost:7700/indexes/songs', method='DELETE')
    try:
        request.urlopen(r)
    except Exception as e:
        if e.getcode() != 404:
            raise

    songs = [{
        'song_key': '1', 'title': 'succducc - me & u', 'song_name': 'me & u', 'song_sub_name': '', 'level_author_name': 'datkami', 'description': 'Hard Only / ~330 notes / Event Lighting! / https://soundcloud.com/succducc/me-n-u', 'total_votes': 640, 'pct_upvoted': 82.65625, 'uploaded_at_tstamp': 1525786136, 'post_id': 7809, 'uploader_name': 'datkami', 'genres': ['Anime', 'EDM'], 'modes': ['Standard']
    }]

    print("Making index")
    r = request.Request('http://localhost:7700/indexes', method='POST',
            data=b'{"uid":"songs"}', headers={'Content-Type': 'application/json'})
    res = request.urlopen(r)
    print('Response: ', res.read())

    print("Configuring index")
    settings = {
        "searchableAttributes": ["title", "song_name", "song_sub_name", "level_author_name", "description", "uploader_name", "genres", "modes"],
        "displayedAttributes": ["*"]
    }
    settings = json.dumps(settings).encode('utf8')
    r = request.Request('http://localhost:7700/indexes/songs/settings', method='POST',
            data=settings, headers={'Content-Type': 'application/json'})
    res = request.urlopen(r)
    print('Response: ', res.read())

    print("Beginning song load")
    r = request.Request('http://localhost:7700/indexes/songs/documents?primaryKey=song_key', method='POST',
            data=json.dumps(songs[:1]).encode('utf8'), headers={'Content-Type': 'application/json'})
    res = request.urlopen(r)
    update_id = json.loads(res.read())['updateId']
    print('Songs loading with update id', update_id)
    while True:
        r = request.Request('http://localhost:7700/indexes/songs/updates/'+str(update_id))
        res = request.urlopen(r)
        data = json.loads(res.read())
        if data['status'] == 'enqueued':
            time.sleep(1)
        elif data['status'] == 'processed':
            break
        else:
            print(data)
            raise Exception('failed to process song load')

    time.sleep(5)
    print('Checking settings')
    r = request.Request('http://localhost:7700/indexes/songs/settings', method='GET',
            headers={'Content-Type': 'application/json'})
    res = request.urlopen(r)
    print('Response: ', res.read())

    print("Song indexing complete")

updatesearch()
```

MarinPostma:
As described in #1160, there was a race condition when updating settings and adding documents simultaneously. This was due to the schema being updated and document addition being processed in two different transactions. This PR moves the schema update logic for the primary key in the same transaction as the document addition, while maintaining the input checks for the validity of the primary key in the http route, in order not to break the error reporting for the document addition route.

close #1160.

### DIFF
--- a/meilisearch-core/src/update/documents_addition.rs
+++ b/meilisearch-core/src/update/documents_addition.rs
@@ -23,6 +23,8 @@ pub struct DocumentsAddition<D> {
     updates_store: store::Updates,
     updates_results_store: store::UpdatesResults,
     updates_notifier: UpdateEventsEmitter,
+    // Whether the user explicitely set the primary key in the update
+    primary_key: Option<String>,
     documents: Vec<D>,
     is_partial: bool,
 }
@@ -39,6 +41,7 @@ impl<D> DocumentsAddition<D> {
             updates_notifier,
             documents: Vec::new(),
             is_partial: false,
+            primary_key: None,
         }
     }
 
@@ -53,7 +56,12 @@ impl<D> DocumentsAddition<D> {
             updates_notifier,
             documents: Vec::new(),
             is_partial: true,
+            primary_key: None,
         }
+    }
+
+    pub fn set_primary_key(&mut self, primary_key: String) {
+        self.primary_key = Some(primary_key);
     }
 
     pub fn update_document(&mut self, document: D) {
@@ -71,6 +79,7 @@ impl<D> DocumentsAddition<D> {
             self.updates_results_store,
             self.documents,
             self.is_partial,
+            self.primary_key,
         )?;
         Ok(update_id)
     }
@@ -88,6 +97,7 @@ pub fn push_documents_addition<D: serde::Serialize>(
     updates_results_store: store::UpdatesResults,
     addition: Vec<D>,
     is_partial: bool,
+    primary_key: Option<String>,
 ) -> MResult<u64> {
     let mut values = Vec::with_capacity(addition.len());
     for add in addition {
@@ -99,9 +109,9 @@ pub fn push_documents_addition<D: serde::Serialize>(
     let last_update_id = next_update_id(writer, updates_store, updates_results_store)?;
 
     let update = if is_partial {
-        Update::documents_partial(values)
+        Update::documents_partial(primary_key, values)
     } else {
-        Update::documents_addition(values)
+        Update::documents_addition(primary_key, values)
     };
 
     updates_store.put_update(writer, last_update_id, &update)?;
@@ -149,7 +159,8 @@ pub fn apply_addition(
     writer: &mut heed::RwTxn<MainT>,
     index: &store::Index,
     new_documents: Vec<IndexMap<String, Value>>,
-    partial: bool
+    partial: bool,
+    primary_key: Option<String>,
 ) -> MResult<()>
 {
     let mut schema = match index.main.schema(writer)? {
@@ -162,7 +173,14 @@ pub fn apply_addition(
     let internal_docids = index.main.internal_docids(writer)?;
     let mut available_ids = DiscoverIds::new(&internal_docids);
 
-    let primary_key = schema.primary_key().ok_or(Error::MissingPrimaryKey)?;
+    let primary_key = match schema.primary_key() {
+        Some(primary_key) => primary_key.to_string(),
+        None => {
+            let name = primary_key.ok_or(Error::MissingPrimaryKey)?;
+            schema.set_primary_key(&name)?;
+            name
+        }
+    };
 
     // 1. store documents ids for future deletion
     let mut documents_additions = HashMap::new();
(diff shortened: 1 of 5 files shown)


meilisearch-http/src/routes/document.rs
```diff
@@ -168,33 +159,34 @@ async fn update_multiple_documents(
...
-                    .ok_or(meilisearch_core::Error::MissingPrimaryKey)?,
-            };
-
-            schema.set_primary_key(&id).map_err(Error::bad_request)?;
-
-            data.db.main_write(|w| index.main.put_schema(w, &schema))?;
+                    .ok_or(meilisearch_core::Error::MissingPrimaryKey)?;
+                document_addition.set_primary_key(key);
+            }
+            _ => ()
```
LegendreM, line 187: Wouldn't it better to have the explicit last case `(None, Some(_))` instead of `_`?
```suggestion
            (None, Some(_)) => ()
```

meilisearch-core/src/update/documents_addition.rs
```diff
@@ -23,6 +23,8 @@ pub struct DocumentsAddition<D> {
     updates_store: store::Updates,
     updates_results_store: store::UpdatesResults,
     updates_notifier: UpdateEventsEmitter,
+    // Whether the user explicitely set the primary key in the update
```
Kerollmops, line 26:
```suggestion
    // Whether the user explicitly set the primary key in the update
```

Kerollmops (approved):

MarinPostma:
bors merge

MarinPostma:
bors merge

MarinPostma:
@Kerollmops, there are flaky tests on macos, can you check on your end?

MarinPostma:
bors try

MarinPostma:
I have found the bug causing the CI to fail.

The test `return_update_status_of_pushed_documents` should not have been working in the first place, since it is trying to push documents multiple times while trying to set the primary_key. Since this was done almost simultaneously, the test passed since it always read a schema without a primary key.

This raises an issue with the `PrimaryKey` being passed along the with the document addition, since the primary key is written **asynchronously**, update made in short intervals will lead to this kind on inconsistencies. There are multiple things we can do:

1) Do nothing and accept that thoses inconsistencies will be resolved when the update is processed. This means that the update may be accepted, and the update will fail with a primary key error later (this is the current state). If the schema is up to date at the time the update is processed, then an error will be thrown synchronously (returned with the update query).

2) Stop checking for primary key synchronously, and only report these errors in the updates status.

@qdequele @Kerollmops What do you think?

MarinPostma:
We decided with @Kerollmops to go with 1, and fix the test.

After discussing with @qdequele, we need to think about removing this options from the document update route, and maybe move it to the settings.

MarinPostma:
bors try

Kerollmops (approved):
Seems good to me, now that the test is right :)

PR #1178, opened 2021-01-11
Fix blocking search

### LINKED ISSUE #1177: Use of blocking functions inside async functions
Hi,
While auditing meilliseach's source code, I've noticed that you use blocking functions inside actix web handlers which are `async` functions.
As we all know, blocking the event loop is a bad idea and may introduce hard to debug performance issues, especially in high-concurrency environments. The recommended time between each `.await` is 10 to 100 microseconds (https://ryhl.io/blog/async-what-is-blocking).

Here is an example of the incriminated code: https://github.com/meilisearch/MeiliSearch/blob/fa40c6e3d42ec9406828df3aa48e20f7c306bb68/meilisearch-http/src/routes/search.rs#L88

Issues that may be related: https://github.com/meilisearch/MeiliSearch/issues/900  & https://github.com/meilisearch/MeiliSearch/issues/522

The solution is to use a threadpool, either tokio's one or actix web's one.

Feel free to close if you have strong reasons to believe that it's not a problem 🌻

MarinPostma:
As pointed out by #1177, the search query were processed in the main tokio loop. This pr fixes that by offloadingn this work in a tokio blocking task.

close #1177

### DIFF
--- a/meilisearch-http/src/error.rs
+++ b/meilisearch-http/src/error.rs
@@ -11,7 +11,7 @@ use meilisearch_error::{ErrorCode, Code};
 
 #[derive(Debug)]
 pub struct ResponseError {
-    inner: Box<dyn ErrorCode>,
+    inner: Box<dyn ErrorCode + Sync + Send>,
 }
 
 impl error::Error for ResponseError {}
--- a/meilisearch-http/src/routes/search.rs
+++ b/meilisearch-http/src/routes/search.rs
@@ -40,7 +40,9 @@ async fn search_with_url_query(
     path: web::Path<IndexParam>,
     params: web::Query<SearchQuery>,
 ) -> Result<HttpResponse, ResponseError> {
-    let search_result = params.search(&path.index_uid, data)?;
+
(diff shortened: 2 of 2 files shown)


skerkour:
Hi,
Please note that the issues was more about describing the underlying issues rather than a specific case.

From a quick scan I can also find blocking code at least here
https://github.com/meilisearch/MeiliSearch/blob/fa40c6e3d42ec9406828df3aa48e20f7c306bb68/meilisearch-http/src/routes/dump.rs#L23
https://github.com/meilisearch/MeiliSearch/blob/fa40c6e3d42ec9406828df3aa48e20f7c306bb68/meilisearch-http/src/routes/dump.rs#L54
https://github.com/meilisearch/MeiliSearch/blob/fa40c6e3d42ec9406828df3aa48e20f7c306bb68/meilisearch-http/src/routes/setting.rs#L56

skerkour:
@LegendreM 👋

curquiza:
is this PR can be merged?

MarinPostma:
I'm closing this in favour of transplant

PR #1185, opened 2021-01-12
fix cors issue

MarinPostma:
This PR fixes a bug where foreign origin were not accepted.
This was due to an update to actix-cors

It also fixes the cors bug when authentication failed, with the caveat that request that are denied for permissions reason are not logged. 

it introduces a bug described in  #1186


LegendreM:
bors try

LegendreM (approved):

MarinPostma:
bors merge

PR #1204, opened 2021-01-22
Indexes name are now sorted by name

Cronos87:


MarinPostma:
I think this is something we want to do directly in the core engine.

Cronos87:
Hello @MarinPostma! Thank you for the reply!

I would like to do it in the core, but sadly I don't know yet Rust, so I prefer avoid doing a mistake in the core 😅

But thank you for the proposition.

meilisearch-http/public/interface.html
```diff
@@ -203,6 +203,10 @@ <h1 class="title is-1 is-spaced">
 
         if (!Array.isArray(result)) { return }
 
+        result.sort(function(a, b) {
+            return a.name.localeCompare(b.name);
```
fflorent, line 207: Beware of performance issues with `String.prototype.localeCompare`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#performance

You'd may rather use Intl.Collator: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator

Hope to help :)

curquiza:
Hello @Cronos87!

Thanks for your suggestion of improvement! However, we are currently working hard on a new search engine that would change the way the indexes are managed. That's why we don't accept PR about this for the moment since it will be outdated in few months, but thanks for your involvement in MeiliSearch 🙂

PR #1206, opened 2021-01-26
fix running URL display

fharper:
by doing that you can just click on it in the terminal if you want


fharper:
@MarinPostma I cannot compile MeiliSearch on my computer, so I wasn't able to test, but I assume it will work since I just updated a string...

Also, let me know if I needed to tag someone else.

MarinPostma:
Uhm would that work with tls?

fharper:
You may have a warning as I changed only the display of the running URL.

curquiza (approved):
checked with @qdequele!
bors merge

PR #1222, opened 2021-02-03
Ignore existing primary key

MarinPostma:
fixing bug in #1176 made it an hard error to try to re-set the primary key on a document addition. This PR makes Meilisearch ignore a primary key passed as an argument to a document addition. This has been decided after a discussion with @curquiza, in order to make the bug fix non breaking.

Turns out it was a good catch too, since contrary to what I thought the error was not caught asynchronously, thank you @curquiza


curquiza (approved):
As a non rust developer, I approve 😄

Kerollmops (approved):
Thank you, seems good to me!

bors r+

PR #1224, opened 2021-02-03
fix synonyms normalization

LegendreM:
Synonyms needs to be indexed in ascendant order,
and the new normalization step for synonyms potentially changes this order
which break the indexation process
because "Harry Potter" > "HP"  but "harry potter" < "hp"


Kerollmops (approved):
This seems good to me!

MarinPostma (approved):
bors r+

PR #1264, opened 2021-02-23
1138 payload size unit

carlosb1:


meilisearch-http/src/option.rs
```diff
@@ -49,15 +65,21 @@ pub struct Opt {
...
+    // 100GB
     pub max_udb_size: usize,
 
     /// The maximum size, in bytes, of accepted JSON payloads
-    #[structopt(long, env = "MEILI_HTTP_PAYLOAD_SIZE_LIMIT", default_value = "104857600")] // 100MB
+    #[structopt(
+        long,
+        env = "MEILI_HTTP_PAYLOAD_SIZE_LIMIT",
+        default_value = "104857600"
+    )] // 100MB
```
qdequele, line 82:
```suggestion
    #[structopt(
        long,
        env = "MEILI_HTTP_PAYLOAD_SIZE_LIMIT",
        parse(try_from_str=parse_size),
        default_value = "104857600"
    )] // 100MB
```

You don't use your method for the payload size?

carlosb1:
From my side, i think all is correct... If you prefer I can include more tests but they don't seem mandatory.

curquiza:
Hello @carlosb1!
I've checked with the team, we are currently working on a new search engine, and this functionality will be already improved in it. Sorry you spent time on it, but thanks for your PR anyway :)
Closing this

fharper:
Hey @carlosb1, I just want to add to what @curquiza wrote! We are so sorry, it's totally our fault.

We are not doing a good job right now since our roadmap is not public. That would have prevented this situation. Be assure that as soon as we release our new engine (version 0.21), which is soon, we will be more transparent. Just so you know, we are actually hiring a PM to help us with that.

Can you send me an email at fred@meilisearch.com, I would like to send you some stickers, a small gesture to apologize :D

carlosb1:
Don't worry, I work in another startup, I understand the situation. I only hope to contribute more for future roadmaps-

fharper:
Still want to send you some stickers if you want :D

Issue #1261, opened 2021-02-22
Web Interface Design


dichotommy:
# Web Interface Design Feedback (for eventual re-design)

It’s been a while since the web interface was designed, but it’s very important and visible—ships with MeiliSearch and is the first entry-point for most newcomers (especially those without a lot of API/command line experience). Particularly with v1.0 coming soon, we should think about redesigning it in the near future. Here are some small points of feedback for whomever ends up working on this project: 

- ~~**Problem**: The API key text box is described as “optional”, which is confusing. An API key is optional for a MeiliSearch instance; however, if you do have an API key set then the web interface will not make any search queries without the API key (in other words, it’s not optional to input the key into the web interface).~~

- ~~**Potential solution**: It would be better for the web interface to infer whether your instance has an API key and show or hide the text input box accordingly. If not possible, then there should at least be a toggle switch: API Key? YES | NO, with the text input box appearing only if “YES” is selected.~~

**FIXED** by #1265 
---
- **Problem**: Missing information: total number of documents in the index.

- **Potential solution**: Instead of only displaying “Documents: {number of returned documents}”, display “Hits: {number of returned documents}” and “Documents: {total number of documents}”
---
- **Problem**: No support for faceted search

- **Potential solution**: Left-hand facet area with text truncation (to handle overly long facet values)
---
- **Problem**: ~~Before performing any searches, the index list appears but no documents are displayed (document counter says “0 documents”). This makes it hard to tell if the interface is functioning or if documents have successfully been added to the index.~~ If your instance is protected by an API key and you open the web interface, you will not see any indexes or documents until after entering your API key. This is normal; however, it's strange that even after the API key has been verified, no documents are displayed until the first search is performed. Accessing a placeholder search requires typing, then deleting a query.

- **Potential solution**: Entering (and verifying) the API key should automatically perform a Placeholder Search on the current index; this would be more consistent with online search interfaces.

curquiza:
Fixed by the integration of the new search engine in https://github.com/meilisearch/MeiliSearch/pull/1421

PR #1265, opened 2021-02-25
Inferring whether to show or Hide API Key box

sanders41:
Relates to #1261

This is one potential solution for inferring whether an instance has an API key and show or hide the text input box accordingly. When the page first loads a request is sent to the server with no API key. If that request was successful then no API key is need so the box is hidden. If the request returns with a 401 status then the API Key was needed and it is shown.

### DIFF
--- a/meilisearch-http/public/interface.html
+++ b/meilisearch-http/public/interface.html
@@ -174,7 +174,7 @@
         
         xmlHttp.onload = function () {
             let apiKeyContainer = document.getElementById('apiKeyContainer');
-            if (xmlHttp.status !== 200) {
+            if (xmlHttp.status === 401) {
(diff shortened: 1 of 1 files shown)


curquiza:
Without any master key defined:
Capture d’écran 2021-02-25 à 18 57 55

With a master key defined:
Capture d’écran 2021-02-25 à 18 58 43

It's a really good idea from my POV 🚀  @react-learner is it ok for you?

curquiza:
Thanks @sanders41, it's nice to see you everywhere 😄

react-learner (approved):
Awesome work! You spoil us @sanders41 ! 😁🔥

curquiza (approved):
bors merge

Issue #1136, opened 2020-12-14
Cannot compile on macOS 11.0.1


fharper:
**Describe the bug**
I cannot compile MeiliSearch. Any help would be appreciated :D

**To Reproduce**
I followed the steps at https://github.com/meilisearch/MeiliSearch#compile-and-run-it-from-sources

**Expected behavior**
It should compile without error.

**Desktop (please complete the following information):**
 - OS: macOS BigSur 11.0.1
- rustc 1.48.0 (7eac88abb 2020-11-16)
 Tested on a MacBook Pro with the new M1 chip (running in the terminal with Rosetta2), but I have the same on a MacBook Pro Intel.

**Additional context**
here is the output when I run `cargo run` on the repo

```
Compiling jemalloc-sys v0.3.2
error: failed to run custom build command for `jemalloc-sys v0.3.2`
```

full output in [output.txt](https://github.com/meilisearch/MeiliSearch/files/5691656/output.txt)

fharper:
I tried some things with @MarinPostma, like re-adding the toolchain

```
rustc 1.48.0 (7eac88abb 2020-11-16)
rustup self update
```

we also try to install `jemalloc` before compiling using homebrew.

tpayet:
I can compile (`cargo clean && cargo build`) MeiliSearch last commit (#861c6fec06367fe) on my MacBook Pro 2019 with Intel processor. I am running Mac OS 11.0.1 aka Big Sur.

```bash
$ rustc --version
rustc 1.48.0 (7eac88abb 2020-11-16)
```

```bash
$ rustup --version
rustup 1.22.1 (b01adbbc3 2020-07-08)
```

let me know if you need more information ^^

qdequele:
@fharper Is this issue could be closed?

fharper:
No, I need to spend more time to try it again. It would be nice if someone on the core team can help as I have no idea what is going on as I'm even installing the "faulty" packages manually...

qdequele:
does the issue come from the Apple M1 processor and not from MacOS?

fharper:
I have the same issue on my intel.

tpayet:
On a fresh new M1 MacBook Pro running macOS Big Sur 11.2.1
I did the following:
- install Rust toolchain using [rustup.rs installer](https://rustup.rs) (`rustc --version` returned  1.50.0)
- in the terminal, I run `git clone git@github.com:meilisearch/MeiliSearch.git`
[...]

PR #1266, opened 2021-02-25
Simplify compile and run from sources

tpayet:
Related to #1136, I just saw that compile & run instructions from sources were not up to date


curquiza (approved):
bors merge

PR #1270, opened 2021-02-27
Remove regex

Mubelotix:
This is an optimisation proposal.
Regex are slow and can often be replaced with simple Rust code.

### DIFF
--- a/meilisearch-core/src/settings.rs
+++ b/meilisearch-core/src/settings.rs
@@ -210,3 +210,30 @@ impl Default for SettingsUpdate {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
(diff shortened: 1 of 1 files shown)


MarinPostma:
Hey @Mubelotix, first of all, thank you for contributing :)

It turns out regex are pretty fast, once they are compiled, and the difference between a simple regex like this and tweaked code is not very significant, and even less so that it's not on the hot path. However, a small regex like this is a lot more readable than a handmade parser.

For this reason I won't merge this pr

I hope you understand :)

Mubelotix:
Yeah actually the primary advantage of this PR was to make the removal of the `regex` dependency easier. I usually like optimizating things and removing dependencies but I understand that MeiliSearch's focus might be readability.

MarinPostma:
It is very probably a transitive dependant anyway 🙂

PR #1279, opened 2021-03-03
fix Docker volume path

fharper:
essential if `$(pwd)` returns a path with spaces


fharper:
FYI, I forced push to change my commit message since pwd returned my path in the message 🤦‍♂️

MarinPostma (approved):
Lgtm

bors r+

PR #1291, opened 2021-03-16
Use 200 status code for healthcheck endpoint 

### LINKED ISSUE #1282: Use 200 status code for healthcheck endpoint
My issue is related to https://github.com/meilisearch/meilisearch-kubernetes/issues/56. If you want to deploy this to GKE using there ingress you need a health check which returns a 200 status code

Either a new health check endpoint or the existing one to return a 200 status code. I understand why 204 is used so I guess it could return 200 with the words success or similar.

The solution in the other ticket uses / as the endpoint for the health check but seems to not return 200 in production when using a master key.
The other solution I can think of is using Nginx to proxy the 204 and change it to a 200 code but this is another dependency which should not be required.

irevoire:
closes  #1282

### DIFF
--- a/meilisearch-http/src/routes/health.rs
+++ b/meilisearch-http/src/routes/health.rs
@@ -9,5 +9,6 @@ pub fn services(cfg: &mut web::ServiceConfig) {
 
 #[get("/health")]
 async fn get_health() -> Result<HttpResponse, ResponseError> {
-    Ok(HttpResponse::Ok().finish())
+    let payload = serde_json::json!({ "status": "ok" });
+    Ok(HttpResponse::Ok().body(payload.to_string()))
 }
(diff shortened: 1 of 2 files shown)


curquiza (changes requested):
Thanks for your PR!

Could you please replace "status: ok" by "status: available"? We chose to follow the IBM convention -> https://github.com/meilisearch/MeiliSearch/issues/1282#issuecomment-799424875

curquiza (approved):

meilisearch-http/src/routes/health.rs
```diff
@@ -9,5 +9,6 @@ pub fn services(cfg: &mut web::ServiceConfig) {
 
 #[get("/health")]
 async fn get_health() -> Result<HttpResponse, ResponseError> {
-    Ok(HttpResponse::NoContent().finish())
+    let payload = serde_json::json!({ "status": "available" });
+    Ok(HttpResponse::Ok().body(payload.to_string()))
```
MarinPostma, line 13: Can you change to use the json method please?

MarinPostma (changes requested):

MarinPostma (approved):
Lgtm

MarinPostma:
bors merge

PR #1292, opened 2021-03-16
return a 200 on / when meilisearch is running in production

### LINKED ISSUE #1235: MeiliSearch shouldn't return 404 on `/` path when PRODUCTION environment enabled
When MeiliSearch is run in `development` environment, a GET request to `/` path will return the MeiliSearch test dashboard. When `production` environment is enabled, it returns a 404 error. Even if at first sight this makes sense, since the dashboard shouldn't be enabled in production, it is confusing for some users. I have had several situations and feedback of users moving their MeiliSearch from `dev` to `prod` and reading this 404 error as a problem in MeiliSearch configuration. (I.E. recent example: https://github.com/meilisearch/meilisearch-digitalocean/issues/115)

MeiliSearch server could send any other type of answer when doing an HTTP request on `/` while the server is in production: `No content`, `Unathorized`, or a message to the user are a few examples of what could be returned. II would like to debate what is the best option since I don't have the perfect answer, but this behavior clearly creates confusion for our users.

irevoire:
close #1235

### DIFF
--- a/meilisearch-http/src/routes/mod.rs
+++ b/meilisearch-http/src/routes/mod.rs
@@ -29,11 +29,24 @@ impl IndexUpdateResponse {
     }
 }
 
+/// Return the dashboard, should not be used in production. See [running]
 #[get("/")]
 pub async fn load_html() -> HttpResponse {
     HttpResponse::Ok()
         .content_type("text/html; charset=utf-8")
         .body(include_str!("../../public/interface.html").to_string())
+}
+
+/// Always return a 200 with:
+/// ```json
+/// {
+///     "status": "Meilisearch is running"
+/// }
+/// ```
+#[get("/")]
+pub async fn running() -> HttpResponse {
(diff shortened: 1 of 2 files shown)


meilisearch-http/src/routes/mod.rs
```diff
@@ -29,13 +29,26 @@ impl IndexUpdateResponse {
...
+/// Always return a 200 with:
+/// ```json
+/// {
+///     "status": "Meilisearch is running"
+/// }
+/// ```
+#[get("/")]
+pub async fn running() -> HttpResponse {
+    let payload = serde_json::json!({ "status": "MeiliSearch is running" }).to_string();
+    HttpResponse::Ok().body(payload)
```
MarinPostma, line 49:
```suggestion
    HttpResponse::Ok().json(serde_json::json!({ "status": "MeiliSearch is running" }))
```

MarinPostma (approved):
Lgtm, thanks

bors merge

PR #1326, opened 2021-04-01
improve the performances

irevoire:
I removed as much useless characters as possible in the source code:
- This improve the compilation time by at least 0.2% since the parser now doesn't have to skip these characters
- I'm pretty sure it also reduce the size of the final binary
- It also improve the readability of the code. Personally I can see way more code on my screen

image

For  further improvement we could probably merge every files in one single big file.


meilisearch-core/src/criterion/attribute.rs
```diff
@@ -1,37 +1 @@
...
-            sum_of_attribute
-        }
-
-        let lhs = sum_of_attribute(&lhs.processed_matches);
-        let rhs = sum_of_attribute(&rhs.processed_matches);
-
-        lhs.cmp(&rhs)
-    }
-}
+use std::cmp::Ordering;use slice_group_by::GroupBy;use crate::{RawDocument, MResult};use crate::bucket_sort::SimpleMatch;use super::{Criterion, Context, ContextMut, prepare_bare_matches};pub struct Attribute;impl Criterion for Attribute {fn name(&self) -> &str { "attribute" }fn prepare<'h, 'p, 'tag, 'txn, 'q, 'r>(&self,ctx: ContextMut<'h, 'p, 'tag, 'txn, 'q>,documents: &mut [RawDocument<'r, 'tag>],) -> MResult<()>{prepare_bare_matches(documents, ctx.postings_lists, ctx.query_mapping);Ok(())}fn evaluate(&self, _ctx: &Context, lhs: &RawDocument, rhs: &RawDocument) -> Ordering {#[inline]fn sum_of_attribute(matches: &[SimpleMatch]) -> usize {let mut sum_of_attribute = 0;for group in matches.linear_group_by_key(|bm| bm.query_index) {sum_of_attribute += group[0].attribute as usize;}sum_of_attribute}let lhs = sum_of_attribute(&lhs.processed_matches);let rhs = sum_of_attribute(&rhs.processed_matches);lhs.cmp(&rhs)}}
```
MarinPostma, line 1:
```suggestion
use std::cmp::Ordering;use slice_group_by::GroupBy;use crate::{RawDocument, MResult};use crate::bucket_sort::SimpleMatch;use super::{Criterion, Context, ContextMut, prepare_bare_matches};pub struct Attribute;impl Criterion for Attribute {fn name(&self) -> &str { "attribute" }fn prepare<'h, 'p, 'tag, 'txn, 'q, 'r>(&self,ctx: ContextMut<'h, 'p, 'tag, 'txn, 'q>,documents: &mut [RawDocument<'r, 'tag>],) -> MResult<()>{prepare_bare_matches(documents, ctx.postings_lists, ctx.query_mapping);Ok(())}fn evaluate(&self, _ctx: &Context, lhs: &RawDocument, rhs: &RawDocument) -> Ordering {#[inline]fn sum_of_attribute(matches: &[SimpleMatch]) -> usize {let mut sum_of_attribute = 0;for group in matches.linear_group_by_key(|bm| bm.query_index) {sum_of_attribute += group[0].attribute as usize;}sum_of_attribute}let lhs = sum_of_attribute(&lhs.processed_matches);let rhs =sum_of_attribute(&rhs.processed_matches);lhs.cmp(&rhs)}}
```

this spaca is unecessary, please remove it.

MarinPostma (changes requested):
still many useless spaces. Needs more work.

meilisearch-core/src/criterion/attribute.rs
Kerollmops: Spaces between lifetimes too!

irevoire:
Easter is over, closing.

PR #1340, opened 2021-04-15
Fix web interface when using api key

### LINKED ISSUE #1339: Web interface not working if a master key is set
**Describe the bug**

The web interface of meilisearch (`meilisearch-http/public/interface.html`) is not working as expected if a `MEILI_MASTER_KEY` has been set for the meilisearch instance.
The user fill the apiKey but it doesn't do anything and he can't see the indexes or the results of the search.
In the interface, when entering the `apiKey` in the `API Key input`, the value of the `X-Meili-API-Key` for the queries will never be updated and the user will never be able to see anything in the interface. (All the requests get a 401 response)
At each change of the `API Key input`, the variable `apiKey` in the localstorage should be updated.

**To Reproduce**

Run `cargo run --release -- --master-key "masterkey"`
Add some data to your meilisearch
Go to `localhost:7700` in your favourite browser
Make sure that you don't have a variable `apiKey` set in your localstorage, if so delete it.
Try to do a search or check indexes -> should not work
Enter your apiKey `masterkey`
Try to do a search or check indexes -> **should work** but is not working at the moment

**Expected behavior**

The value of the variable `apiKey` in the localstorage should be updated each time the user change the `API Key input`.

 - OS: macos
 - Browser: chrome/safari

**Additional context**

Since Meilisearch v0.20.0

nicolasvienot:
closes #1339 

The `apiKey` variable in the localstorage is not updated each time the input with id `apiKey` changes as it should be [because the element gets removed](https://github.com/meilisearch/MeiliSearch/blob/94c0858c276743e8b2f4c8c97880070a198341bf/meilisearch-http/public/interface.html#L178) when it gets the response of the query in `setApiKeyField` checking whether there is an apiKey needed or not.

Since the element with id `apiKey` has been removed, [the event listener](https://github.com/meilisearch/MeiliSearch/blob/94c0858c276743e8b2f4c8c97880070a198341bf/meilisearch-http/public/interface.html#L348) is removed too and the apiKey won't be updated anymore.

Adding the eventlistener and checking for the variable in the localstorage when re-creating the input fixes this.

### DIFF
--- a/meilisearch-http/public/interface.html
+++ b/meilisearch-http/public/interface.html
@@ -203,7 +203,6 @@
                 apiKeyContainer.appendChild(columnNode);
 
                 apiKey.addEventListener('input', function(e) {
-                    console.log("addEventListener");
                     localStorage.setItem('apiKey', apiKey.value);
                     refreshIndexList();
                 }, false);
--- a/meilisearch-http/public/interface.html
+++ b/meilisearch-http/public/interface.html
@@ -201,6 +201,16 @@
                 columnNode.classList.add('column', 'is-4');
                 columnNode.appendChild(fieldNode);
                 apiKeyContainer.appendChild(columnNode);
+
+                apiKey.addEventListener('input', function(e) {
+                    console.log("addEventListener");
+                    localStorage.setItem('apiKey', apiKey.value);
+                    refreshIndexList();
+                }, false);
+
+                if (!apiKey.value) {
+                  apiKey.value = localStorage.getItem('apiKey');
+                }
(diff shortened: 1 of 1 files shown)


curquiza:
Hello @nicolasvienot!
Thanks for your fix, however, MeiliSearch does not accept any contribution before the release of the v0.21.0, sorry I will close your PR! https://github.com/meilisearch/MeiliSearch#contributing
A new dashboard will be released in the v0.21.0 so this bug will be removed 😉

PR #1390, opened 2021-05-26
Create da.stopwords.txt

kerns:
This is a merge / dup removal created from two popular lists of Danish stop words used in other OS projects.


curquiza:
Hello @kerns!
Thanks for the Danish stop word addition!
However this repository is in a transition step, and we would rather not use it to store datasets and dictionaries. We plan to move all of these to this repo -> https://github.com/meilisearch/datasets/
It's not done for the moment, but when we will migrate them, be sure we will include the Danish stop words 🙂

I close this issue, but thank you so much for this addition, it will be useful for Danish users!

PR #1403, opened 2021-06-10
fix amount of time

TheTechRobo:
The new MeiliSearch sandboix website says "48 hours" rather than 72, so I updated the readme to reflect that


curquiza (approved):
thanks!!
bors merge

Issue #1368, opened 2021-05-16
Highlight on emojis


JacobLey:
**Describe the bug**
Using Meilisearch's `_matches` and `_formattedInfo` of the response should show the searcher the exact text that was found in the search.

I have found that _mostly_ works, but appears to break when an emoji is in the middle of a word. Note to the best of my knowledge the actual search results are still correct, it is just over-highlighting the results.

**To Reproduce**

1) Create an index with a searchable attribute `text`.
2) Upload a document with the text `Go💼od luck.`
3) Perform the following search on the index:
```
{
  "q": "Go💼",
  "attributesToHighlight": ["text"],
  "matches": true
}
```
4) View results:
```
{
  "hits": [
    {
      "text": "Go💼od luck.",
      "_formatted": {
        "text": "<em>Go💼od lu</em>ck."
      },
      "_matchesInfo": {
        "text": [
          {
            "start": 0,
            "length": 11
          }
        ]
      }
    }
  ],
  "offset": 0,
  "limit": 20,
  "nbHits": 1,
  "exhaustiveNbHits": false,
  "processingTimeMs": 1,
  "query": "Go💼"
}
```

**Expected behavior**

I would expect the same document to be returned, but the `<em>` highlighting should only extend to the briefcase emoji (e.g. `"text": "<em>Go💼</em>od luck."`, and the matchesInfo length should be 6 (2 bytes for plain text + 4 for emoji in UTF-8)

**Additional context**
Running 0.19.0
I know that is a version behind, but I don't see anything notable in the changelog that I would expect to resolve this: https://github.com/meilisearch/MeiliSearch/releases

Note I am _not_ able to recreate the issue when the text is something like `💼Good💼 luck.` and I search for part or all of that word (e.g. `💼Go` or `💼Good💼`).

curquiza:
Hello, thanks for the report, we will see if it's possible to fix it in the next release.

curquiza:
@JacobLey, it's impossible for the search engine to highlight correctly emojis. The search engine de-unicodes the string and cannot work with emoji in the string when applying the highlight.
The next version will just not highlight any emoji then. Sorry for the inconvenience.

@CaroFG and @meilisearch/docs-team, it might interest you.

JacobLey:
> will just not highlight any emoji then
That sounds fair to me.

[...]

Issue #1173, opened 2021-01-08
Highlighting is broken with non ascii characters


burtgulash:
There is a field in my index **category** with value **Vývoj**, that I am searching for.
- In the first case I cut the last letter and send **vývo** as a query. This case actually highlights correctly.
- In the second case I try to search directly by sending the full word **vývoj** as a query. No highlight in the result. Matches object contains two overlapping match ranges.

```
curl -X POST http://localhost:7700/indexes/myindex/search --data '{"q":"vývo","attributesToHighlight": ["category"],"matches":true}'

"_formatted": {
  "category": "<em>Vývoj</em>"
},
"_matchesInfo": {
  "category": [
    {
      "start": 0,
      "length": 5
    }
  ]       
}
```

This one gets messed up.
```
curl -X POST http://localhost:7700/indexes/myindex/search --data '{"q":"vývoj","attributesToHighlight": ["category"],"matches":true}'

"_formatted": {
  "category": "Vývoj"  
},      
"_matchesInfo": {      
  "category": [      
    {      
      "start": 0,      
      "length": 5      
    },      
    {      
      "start": 0,      
      "length": 6      
    }      
  ]       
}
```

```
% meilisearch --version
meilisearch-http 0.17.0
```

timoschwarzer:
### Reproduction

#### When searching documents without special characters:

```json
[
  {
    "id": "123",
    "title": "This is a test with special characters blah blub"
  }
]
```

Search term is "characters":

```json
title: "This is a test with special <em>characters</em> blah blub"
```

#### And **with** special characters:

```json
[
  {
    "id": "123",
    "title": "This is a test with special chäräcters blah blub"
  }
]
```

Search term is "chäräcters":

```json
title: "This is a test with special <em>chäräcte</em>rs blah blub"
```

One thing I noticed is that it ignores a word entirely for highlighting if the last character of a query word is a special one, e.g.

<details>
	<summary>Examples</summary>
	
![image](https://user-images.githubusercontent.com/8849554/108190857-c138b080-7112-11eb-9363-bec255ca3b9b.png)

![image](https://user-images.githubusercontent.com/8849554/108190916-d01f6300-7112-11eb-9069-267a37193006.png)

![image](https://user-images.githubusercontent.com/8849554/108190945-d7df0780-7112-11eb-9d05-1c5352b67523.png)


</details>

timoschwarzer:
Quoting from @curquiza 

[...]

PR #1437, opened 2021-06-30
Fix highlighting panic for special characters

irevoire:
Helps #1173 and helps #1368.

### DIFF
--- a/meilisearch-http/src/index/search.rs
+++ b/meilisearch-http/src/index/search.rs
@@ -738,7 +738,7 @@ mod test {
         )
         .unwrap();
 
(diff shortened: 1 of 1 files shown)


Kerollmops (approved):
bors merge

curquiza:
⚠️ Can we add tests for highlight
- with emoji (no matter is the final behavior, highlight or not, we just want to check it does no panic)
- with special caracters (no matter is the final behavior, highlight or not, we just want to check it does no panic)
- when doing a prefix search, cf https://github.com/meilisearch/MeiliSearch/issues/1441

curquiza:
Replaced by https://github.com/meilisearch/MeiliSearch/pull/1457

PR #1254, opened 2021-02-17
Fix highlighting with special characters

timoschwarzer:
Closes #1173

This PR uses the `String::chars()` iterator for splitting strings. Before it was possible that the `highlighted` String got split in between an UTF-8 multibyte sequence, which lead to errors.

:warning: Please be warned that I have practically no experience with Rust and this PR might be total :hankey: But *maybe* it's a solution, it worked locally at least. Feedback appreciated :)

PR #1457, opened 2021-07-01
Hotfix highlight on emojis panic

### LINKED ISSUE #1368: Highlight on emojis

ManyTheFish:
When the highlight bound is in the middle of a character
or if we are out of bounds, we highlight the complete matching word.

note: we should enhance the tokenizer and the Highlighter to match char indices.

Fix #1368

### DIFF
--- a/meilisearch-http/src/index/search.rs
+++ b/meilisearch-http/src/index/search.rs
@@ -580,13 +580,23 @@ impl<'a, A: AsRef<[u8]>> Formatter<'a, A> {
             // Matcher::match since the call is expensive.
             if format_options.highlight && token.is_word() {
                 if let Some(length) = matcher.matches(token.text()) {
-                    if format_options.highlight {
-                        out.push_str(&self.marks.0);
-                        out.push_str(&word[..length]);
-                        out.push_str(&self.marks.1);
-                        out.push_str(&word[length..]);
-                        return out;
+                    match word.get(..length).zip(word.get(length..)) {
+                        Some((head, tail)) => {
+                            out.push_str(&self.marks.0);
+                            out.push_str(head);
+                            out.push_str(&self.marks.1);
+                            out.push_str(tail);
+                        }
+                        // if we are in the middle of a character
+                        // or if all the word should be highlighted,
+                        // we highlight the complete word.
+                        None => {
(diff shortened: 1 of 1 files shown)


curquiza:
Since this PR replaces #1254, can you apply my request here: https://github.com/meilisearch/MeiliSearch/pull/1437#issuecomment-871613873

ManyTheFish:
@curquiza I have added the tests for the 2 first cases, unfortunately, I think your third case is not related to this fix.

Kerollmops (approved):
bors merge

irevoire (approved):

PR #1468, opened 2021-07-05
fix update loop bug

### LINKED ISSUE #1465: v0.21.0rc1 Fatal error while processing an update
**Describe the bug**
I have been doing some testing with v0.21.0rc1 and have been getting an error about `Fatal error while processing an update` and then `Update store loop exited.` This seems to be related to sending multiple requests in succession, and seems to happen after deleting an index and then reusing it. This scenario is one I have been able to pin down, but I'm not sure if there are others that would also cause it.  I first noticed it while running automated tests so this issue may not be "real world" relevant, but v0.20.0 doesn't throw this error in the same situation so I wanted to report it in case there is something more here than I have noticed so far.

**To Reproduce**
I was able to come up with an example using the Python client that has reproduced the error for me every time I run it. In this example I used the small movies data set here https://github.com/meilisearch/meilisearch-python/blob/main/datasets/small_movies.json.

```py
import json

from meilisearch import Client


def main():
    with open("/path/to/small_movies.json") as f:  # Replace this path with your path to small_movies.json
        data = json.load(f)
        client = Client("http://localhost:7700", "masterKey")
        index = client.index("test")
        index.add_documents(data)
        index.delete()
        index.add_documents(data)


if __name__ == "__main__":
    main()
```

**Expected behavior**
Program should run without error.

**Actual behavior**
MeiliSearch returns an error

```
[2021-07-02T19:22:49Z ERROR meilisearch_http::index_controller::update_actor::store] Fatal error while processing an update that requires the update store to shutdown: Internal error: No such file or directory (os error 2)
[2021-07-02T19:22:49Z ERROR meilisearch_http::index_controller::update_actor::store] Update store loop exited.
```

Then any subsequent request results in an error until the server is restarted (the "deleting" in this message would correspond to the request type).

```
[2021-07-02T19:39:44Z ERROR meilisearch_http::index_controller] Error while deleting index: update store was shut down due to a fatal error, please check your logs for more info.
```

 - `lsb-release` info
```
DISTRIB_ID=ManjaroLinux
DISTRIB_RELEASE=21.0.7
DISTRIB_CODENAME=Ornara
DISTRIB_DESCRIPTION="Manjaro Linux"
```
 - MeiliSearch v0.21.0rc1 running in Docker version 20.10.7

**Additional context**
If I modify the program to slow it down the issue goes away.

```py
import json
from time import sleep

from meilisearch import Client


def main():
    with open("/path/to/small_movies.json") as f:  # Replace this path with your path to small_movies.json
        data = json.load(f)
        client = Client("http://localhost:7700", "masterKey")
        index = client.index("test")
        index.add_documents(data)
        sleep(5)
        index.delete()
        sleep(5)
        index.add_documents(data)


if __name__ == "__main__":
    main()
```

MarinPostma:
fix #1465

When deleted an index, the update files were deleted before the currently update for the index to delete was processed. If the delete operation occured before the currently processing update occured, then it triggered a file not found error and crashed the update loop.

tested with:

```python
import json

from meilisearch import Client


def main():
    with open("small.json") as f:  # Replace this path with your path to small_movies.json
        data = json.load(f)
        client = Client("http://localhost:7700", "masterKey")
        for i in range(0, 100):
            index = client.index("test")
            index.add_documents(data)
            index.delete()
            index.add_documents(data)


if __name__ == "__main__":
    main()
```

### DIFF
--- a/meilisearch-http/src/index_controller/update_actor/store/mod.rs
+++ b/meilisearch-http/src/index_controller/update_actor/store/mod.rs
@@ -471,13 +471,6 @@ impl UpdateStore {
 
         txn.commit()?;
 
-        uuids_to_remove
-            .iter()
-            .map(|uuid| update_uuid_to_file_path(&self.path, *uuid))
-            .for_each(|path| {
-                let _ = remove_file(path);
-            });
-
         // If the currently processing update is from our index, we wait until it is
         // finished before returning. This ensure that no write to the index occurs after we delete it.
         if let State::Processing(uuid, _) = *self.state.read() {
@@ -486,6 +479,17 @@ impl UpdateStore {
                 self.state.write();
             }
         }
+
+        // Finally, remove any outstanding update files. This must be done after waiting for the
+        // last update to ensure that the update files are not deleted before the update needs
+        // them.
+        uuids_to_remove
+            .iter()
+            .map(|uuid| update_uuid_to_file_path(&self.path, *uuid))
+            .for_each(|path| {
+                let _ = remove_file(path);
+            });
+
 
         Ok(())
     }
--- a/meilisearch-http/src/index_controller/update_actor/store/mod.rs
+++ b/meilisearch-http/src/index_controller/update_actor/store/mod.rs
@@ -490,7 +490,6 @@ impl UpdateStore {
                 let _ = remove_file(path);
             });
 
-
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


Kerollmops (approved):
Thank you very much! I'll wait for the @irevoire review.

irevoire (approved):
The fix looks good to me and I was not able to reproduce the bug :slightly_smiling_face: 
Thanks @MarinPostma 
bors merge

PR #1478, opened 2021-07-05
refactor routes

MarinPostma:
refactor the route directory, so the module tree follows the route structure


irevoire (approved):
lgtm, I tried a bunch of routes and ran the tests.
It should be easier to maintain this way, thanks @MarinPostma 
bors merge

PR #1496, opened 2021-07-08
fix arrays in GET search

### LINKED ISSUE #1489: Breaking changes in GET /search facetsDistribution usage
**Describe the bug**

Previously `facetsDistribution` was expecting an array of attributes: 
`/search?facetsDistribution=["genres", "title"]`

It is now expected a string of attributes:

`search?facetsDistribution=genres,title`

**To Reproduce**
1. Create index with movies containing `title` and `genres` field
2. try to get facetsDistribution the way it was done in v0.20
3. Error thrown
```json
{
    "message": "invalid facet distribution, the fields [\"genres'] are not set as filterable",
    "errorCode": "bad_request",
    "errorLink": "https://docs.meilisearch.com/errors#bad_request"
}
```

**Expected behavior**
Same behavior 

**MeiliSearch version**:
v0.21.0

MarinPostma:
fix #1489

### DIFF
--- a/meilisearch-http/src/error.rs
+++ b/meilisearch-http/src/error.rs
@@ -165,3 +165,13 @@ where
 {
     err.into().into()
 }
+
+#[derive(thiserror::Error, Debug)]
+#[error(transparent)]
+pub struct SerdeJsonError(#[from] serde_json::Error);
+
+impl ErrorCode for SerdeJsonError {
+    fn error_code(&self) -> Code {
+        Code::BadRequest
+    }
+}
--- a/meilisearch-http/src/routes/indexes/search.rs
+++ b/meilisearch-http/src/routes/indexes/search.rs
@@ -1,11 +1,11 @@
-use std::collections::{BTreeSet, HashSet};
+use std::convert::{TryFrom, TryInto};
 
 use actix_web::{web, HttpResponse};
 use log::debug;
 use serde::Deserialize;
 use serde_json::Value;
 
-use crate::error::ResponseError;
+use crate::error::{ResponseError, SerdeJsonError};
 use crate::extractors::authentication::{policies::*, GuardedData};
 use crate::index::{default_crop_length, SearchQuery, DEFAULT_SEARCH_LIMIT};
 use crate::routes::IndexParam;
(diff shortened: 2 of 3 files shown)


bidoubiwa:
`attributesToCrop` could previously have the value `*`.
I tried with your branch and the following GET request: `/search?q=&attributesToCrop=*` raises the following error: 
```json
{
    "message": "expected value at line 1 column 1",
    "errorCode": "bad_request",
    "errorLink": "https://docs.meilisearch.com/errors#bad_request"
}
```

bidoubiwa:
Ah your PR broke formating the other way around: 

`GET /search?attributesToRetrieve=title,author`
this is now throws an error

```
{
    "message": "expected ident at line 1 column 2",
    "errorCode": "bad_request",
    "errorLink": "https://docs.meilisearch.com/errors#bad_request"
}
```

MarinPostma:
Wait you mean that we're supposed to support the two?

bidoubiwa:
Its the behavior of 0.20 😕

MarinPostma:
@gmourier, only the array syntax is documented, this is at best undefined behaviour, at wort a bug. Do we reintroduce it?

bidoubiwa:
It will introduce a new breaking that we might consider as users will have to update their query parameters either way because of facetFilters and filters becoming filter. 
So it may be a good time to go with best practices of GET parameters. And thus not implement this PR.

gmourier:
For the moment I would say to be ISO at the maximum with 0.20. I think we can afford to introduce breaking changes later. The API convention will probably introduce a lot of breaking changes in Q3. We might as well give ourselves time to decide on the syntax of the common parameters between GET / POST. What do you think?

gmourier:
Edit:

Since we may reintroduce an unspecified/undocumented behavior if we want to be ISO with 0.20. It makes sense to push this pull request to move in the right direction.

We'll fix this during Q3 to globalize the parameter syntax.

The most important point here is that people using the old syntax can quickly fix and understand how to make the code work by reading the changelog or other support.

bidoubiwa:
Tests should be made to guarantee that both GET and POST works with the same query parameters types as in meilisearch-js I have to parse specifically for GET requests: 

```js
const getParams: Types.GetSearchRequest = {
        ...params,
        filter: parseFilter(options?.filter),
        facetsDistribution: options?.facetsDistribution
          ? JSON.stringify(options.facetsDistribution)
          : undefined,
        attributesToRetrieve: options?.attributesToRetrieve
          ? options.attributesToRetrieve.join(',')
          : undefined,
        attributesToCrop: options?.attributesToCrop
          ? options.attributesToCrop.join(',')
          : undefined,
        attributesToHighlight: options?.attributesToHighlight
          ? options.attributesToHighlight.join(',')
          : undefined,
      }
```

These might be outdated and could be removed. I think it could be interesting to make tests to compare both usages

gmourier:
Do you think that an API specification can help to keep your test up-to-date?

bidoubiwa:
> Do you think that an API specification changelog can help to keep your test up-to-date for all our future evolutions?

I'm not sure I understood, but an explanatory table with formats compatibilities in GET and POST would be very useful

curquiza:
Closing this PR. Explanation here: https://github.com/meilisearch/MeiliSearch/issues/1489#issuecomment-886767116

PR #1521, opened 2021-07-21
Sentry was never sending anything

irevoire:
@Kerollmops noticed that we had no log of this release in sentry, and it look like I badly tested my code after ignoring the “No space left on device” errors.

Now it should be fixed.


Kerollmops (approved):
Thank you :)
bors merge

bb:
Curious: If I set no-analytics but do _not_ no-sentry - will it send exception messages or not?

irevoire:
No, for the 0.21, we removed the `no-sentry` flag to make it easier to understand to our users.
Now, if you disable the analytics, it'll disable sentry and amplitude at the same time.
You can read a little bit more in this PR where I implemented the analytics as they are currently: https://github.com/meilisearch/transplant/issues/180.
And if you have more questions do not hesitate :smile:

PR #1537, opened 2021-07-28
Import `.git` to the docker build image to fix vergen

irevoire:
I observed a small difference in the size of the build image, but I think we can allow it:
image

I was not able to see any difference in build time, though.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,4 @@
 target
 Dockerfile
 .dockerignore
-.git
 .gitignore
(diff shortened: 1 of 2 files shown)


MarinPostma:
Are those difference for the same builds? Does that means that we can't have reproducible builds?

Kerollmops:
Wouldn't it be better to only import the file used by vergen? like for example the HEAD file or files related to tags maybe?

irevoire:
@MarinPostma it's the same build. One time with the `.git` in the filesystem and one time without it.

@Kerollmops I looked at the `.git` directory and it look like that:
image

If I remove the `.git/objects/pack` then vergen does not works. So it look like we wont win anything by cherry-picking the right file from our `.git`.

curquiza (approved):

MarinPostma (approved):

curquiza:
wunderbar

bors merge

PR #1577, opened 2021-08-12
Update milli dependency: fix facet values bugs

### LINKED ISSUE #1576: FacetDistribution values are incomplete
**Describe the bug**

When doing a search including facetsDistribution in 0.21.0rc4, I don't receive all values for a field, even when I'm not filtering. In earlier versions (0.20 and before) the facetDistribution was complete and correct for the same dataset.

I made sure it's not missing data by checking the values in the hits from the very same response.

My dataset contains a field `region_ids` which is an array of strings (which look like ints, but they're strings because of <v0.21 limitations).

Tiny example (most likely not big enough for reproduction, just to see the structure see [this slack message](https://meilicommunity.slack.com/archives/CP9DVS1RQ/p1628699235091200?thread_ts=1628696972.089500&cid=CP9DVS1RQ) for more):

```json
  [
  { "region_ids": ["-2", "-2202162", "-8649", "-71525"]  },
  { "region_ids": ["-2", "-49715", "-224459", "-2431739"] },
  { "region_ids": ["-5", "-60199", "-71248", "-11900985"] }
]
```

**To Reproduce**
1. Do a search including facets: `res=await index.search("", {facetsDistribution: ["*"], limit: 1000 })`
2. Observe facet results: `Object.keys(res.facetsDistribution.region_ids).length` (18 in my case)
3. Compare with the values in the hits: `new Set(res.hits.map (h => h.region_ids).flat()).size` (44 in my case)
4. facet distribution values (18) should be bigger than number of values in hits (44)

**Expected behavior**
Each field in the facetsDistribution should contain *all* values and their complete counts. Thus the number keys of a field in facetsDistribution should always be greater or equal than the actual values in the hits. If the default behavior is to limit that to a reasonable number, there should be a setting to enable returning all. 

Of course, the facetDistribution should adhere to the filters. It did so for facetFilters in <0.21 but according to #929 not for filters.

**MeiliSearch version:** [e.g. v0.21.0rc4]

curquiza:
Fixes #1576

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1756,8 +1756,8 @@ dependencies = [
 
 [[package]]
 name = "milli"
-version = "0.9.0"
-source = "git+https://github.com/meilisearch/milli.git?tag=v0.9.0#2102e0da6b046f252b75b2bb5a9fd34a5954ac8b"
+version = "0.10.0"
+source = "git+https://github.com/meilisearch/milli.git?tag=v0.10.0#51581d14f859b01411b586475f48b54eb81b7b1d"
 dependencies = [
  "bstr",
  "byteorder",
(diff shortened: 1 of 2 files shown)


Kerollmops (approved):
Looks good to me!
bors merge

PR #1586, opened 2021-08-13
Update grenad

bb:
See also https://github.com/meilisearch/milli/pull/306

### DIFF
--- a/meilisearch-http/Cargo.toml
+++ b/meilisearch-http/Cargo.toml
@@ -41,7 +41,7 @@ flate2 = "1.0.19"
 fst = "0.4.5"
 futures = "0.3.7"
 futures-util = "0.3.8"
-grenad = { git = "https://github.com/Kerollmops/grenad.git", rev = "3adcb26" }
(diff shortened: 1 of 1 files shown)


MarinPostma:
Wait I'm not sure why we even need grenade in MeiliSearch http. @irevoire can you check that out if you come back before I do?

bb:
It's only here for the CompressionType Option - very minor compared to the usage in Milli.
However, I think it's better to close both PRs because I later noticed they're basically useless because of API changes in Grenad: Introduction of ChunkCreator, Removal of FileFuse. I mentioned that in Slack to @Kerollmops already.

The upgrade here shouldn't be an issue but the upgrade in Milli requires quite a few changes regarding ChunkCreator, so I think it's better if someone from the Core team does it.

Kerollmops:
Thanks @bb for this issue, we removed the grenad dependency in MeiliSearch and made milli reexport the `grenad::CompressionType` this way we can directly use the `milli::CompressionType` in MeiliSearch now. 🚀

Issue #1606, opened 2021-08-23
Remove Sentry


curquiza:
Checked with @gmourier and the approval of @qdequele 

## Motivation

We want to remove Sentry since we don't use it.

## TODO

Remove the Sentry calls from the code base.
The `--no-analytics` flag only concerns Amplitude anymore.

Impacts the @meilisearch/docs-team.

PR #1656, opened 2021-08-31
Remove unused Arc import

Kerollmops:
This PR removes a warning introduced by #1606 which removed Sentry that was using an `Arc` but forgot to remove the scope import, we remove it here.


MarinPostma (approved):
ty!

bors merge

Issue #1273, opened 2021-03-01
Cannot build MeiliSearch v0.19.0 as NixOS package


michcioperz:
**Describe the bug**
I tried to use MeiliSearch on my NixOS server today, but noticed the official repos only have v0.9, so I tried to update it.
I took the build script from the official repository and upped the version and fixed the hash of the source code.
The build process stops with an error as seen below.

**To Reproduce**
1. Run NixOS 20.09
2. Download this Nix expression: https://gist.github.com/michcioperz/4ecdeaf7921d63784eedf7197eafcb71
3. Run `nix-build meilisearch.nix` (the filename you saved to)
4. An error like this should show up during the vendoring phase:

```
error: failed to sync

Caused by:
  found duplicate version of package `pest v2.1.3` vendored from two sources:

        source 1: https://github.com/pest-parser/pest.git?rev=51fd1d49f1041f7839975664ef71fe15c7dcaf67#51fd1d49
        source 2: registry `https://github.com/rust-lang/crates.io-index`
Traceback (most recent call last):
  File "/nix/store/m6x1vskdlgsray23j2m2k5j7xzgcjhry-cargo-vendor-normalise/bin/.cargo-vendor-normalise-wrapped", line 42, in <module>
    main()
  File "/nix/store/m6x1vskdlgsray23j2m2k5j7xzgcjhry-cargo-vendor-normalise/bin/.cargo-vendor-normalise-wrapped", line 17, in main
    assert list(data.keys()) == ["source"]
AssertionError
builder for '/nix/store/vm7f4lngx8ar6qbh0n1snrwrscldwab8-meilisearch-0.19.0-vendor.tar.gz.drv' failed with exit code 1
cannot build derivation '/nix/store/cjxkk5frkqdikx6rzvyjlbviz2mlqd99-meilisearch-0.19.0.drv': 1 dependencies couldn't be built
error: build of '/nix/store/cjxkk5frkqdikx6rzvyjlbviz2mlqd99-meilisearch-0.19.0.drv' failed
```

**Expected behavior**
The vendoring process should finish and Nix should complain that the hash of the vendor directory is different from what I told it to expect.

**Desktop:**
 - OS: NixOS 20.09

**Additional context**
NixOS is really into reproducible builds and pinning hashes of all things that are downloaded, so the first part of the process of building a package of something in Rust is vendoring the dependencies and checking that the hash of vendor directory matches what was given in the build script.
[...]

PR #1659, opened 2021-09-01
deps: unify pest dependency

happysalada:
meilisearch dependends on two different versions of pest.
This can be problematic for some build systems (e.g. NixOS).
Since the repo hasn't received an update in a while, in the meantime, use the later version of the two pest dependencies.

Context: this has been discussed previously https://github.com/meilisearch/MeiliSearch/issues/1273
meilisearch has been selected by ngi to be packaged for nixos. A patch can be applied to make the changes proposed in this PR. This PR intends to see how the maintainers of meilisearch would feel about the patch.

What was done.
- Add an override for the pest dependency in Cargo.toml.
- recreate the Cargo.lock with `cargo update`. This has had the side effect of updating some dependencies.

I ran the tests on darwin. My machine is quite old so I had 8 failures due to a timeout. None of the failures look like they are due to the new dependencies.

Checking the pest repo, it seems there are some recent commits, however no sure date of when there could be a new release.

If this gets accepted, there is no need to do a new release, nixos can just target the new commit.

If you feel it's too much pain for not enough gain, no worries at all!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,10 @@ members = [
     "meilisearch-http",
     "meilisearch-error",
 ]
+resolver = "2"
 
 [profile.release]
 debug = true
+
+[patch.crates-io]
+pest = { git = "https://github.com/pest-parser/pest.git", rev = "51fd1d49f1041f7839975664ef71fe15c7dcaf67" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1674,7 +1674,7 @@ dependencies = [
  "obkv",
  "once_cell",
  "ordered-float",
- "pest 2.1.3 (git+https://github.com/pest-parser/pest.git?rev=51fd1d49f1041f7839975664ef71fe15c7dcaf67)",
+ "pest",
  "pest_derive",
  "rayon",
  "roaring",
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [workspace]
-members = [
-    "meilisearch-http",
-    "meilisearch-error",
-]
+members = ["meilisearch-http", "meilisearch-error"]
```
happysalada, line 2: This change is unrelated, I'm using an automatic toml formatter and this changed was made. Happy to revert if this proves to be a problem.

Cargo.toml
```diff
@@ -1,8 +1,8 @@
...
 
 [profile.release]
 debug = true
+
+[patch.crates-io]
+pest = { git = "https://github.com/pest-parser/pest.git", rev = "51fd1d49f1041f7839975664ef71fe15c7dcaf67" }
```
happysalada, line 8: this is the main change of the PR.
The Cargo.toml has just been recreated to take this change into account.

happysalada:
I've reverted the dependency updates so as to not make it confusing to review.

This includes setting the "resolver" version of cargo to 2. With this using `--no-default-features` can also remove default features of workspace crates. Without this, it's not possible to remove default features for `meilisearch-http`

MarinPostma:
Sorry I need a bit of time to look into this. It looks ok, i just want to take a bit of time to understand the fix

happysalada:
No problem at all, thanks a lot for taking a look at this!
I've find a way around setting the resolver version to 2, so the only thing needed is the pest dependency override.
I've applied the patch to the version being packaged in NixOS and updated it to 0.21.1.
No rush.

MarinPostma (dismissed):
Ok this is a good fix. Ultimately, I think we should get rid of the pest dependency, I am currently looking into nom, which I think will be a better long term solution.

thanks!

bors merge

happysalada:
Thanks a lot for merging this, it will make the next release of meilisearch easier to package with!

curquiza:
Needs to have the changes on `release-v0.22.0` to be merge, will be merge on v0.22.0 is out :)

happysalada:
@curquiza let me know if that can be merged now that 0.22 is out.

curquiza:
can be merged!
sorry for the delay

bors merge

curquiza:
bors try

curquiza:
@happysalada looks like there is an issue with your PR with the cargo.lock, cf https://github.com/meilisearch/MeiliSearch/runs/3600376194

MarinPostma:
hey @happysalada can you rebase and then run `cargo update`, I think it should fix the error. Thanks :)

happysalada:
Just rebased off main. Let me know if anything.

MarinPostma:
We'll try

Bors merge

MarinPostma (approved):

PR #1669, opened 2021-09-06
Fix windows integration tests

ManyTheFish:
Set max_memory value to unlimited during tests:
because tests run several meilisearch in parallel,
we overestimate the value for max_memory making the tests on Windows crash

### DIFF
--- a/meilisearch-http/src/index_controller/index_actor/actor.rs
+++ b/meilisearch-http/src/index_controller/index_actor/actor.rs
@@ -31,9 +31,12 @@ pub struct IndexActor<S> {
 }
 
 impl<S: IndexStore + Sync + Send> IndexActor<S> {
-    pub fn new(receiver: mpsc::Receiver<IndexMsg>, store: S) -> anyhow::Result<Self> {
-        let options = IndexerOpts::default();
-        let update_handler = UpdateHandler::new(&options)?;
+    pub fn new(
+        receiver: mpsc::Receiver<IndexMsg>,
+        store: S,
+        options: &IndexerOpts,
+    ) -> anyhow::Result<Self> {
+        let update_handler = UpdateHandler::new(options)?;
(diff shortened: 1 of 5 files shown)


Kerollmops:
We should [open an issue](https://github.com/GuillaumeGomez/sysinfo/issues) to inform @GuillaumeGomez of that issue. It seems to be related to [the `total_memory_bytes` function and on Windows only](https://github.com/meilisearch/MeiliSearch/blob/33514b28be4fcace8dfd77f75abec3aca9a9d5c9/meilisearch-http/src/option.rs#L289-L299).

We could use [a `cfg` for windows only](https://doc.rust-lang.org/rust-by-example/attribute/cfg.html) that use `None` instead of trying to get the total memory.

GuillaumeGomez:
What's the issue exactly?

Kerollmops:
Hey @GuillaumeGomez,

Sorry for pinging you, maybe it's a little bit early but we [have a `STATUS_ACCESS_VIOLATION`](https://github.com/meilisearch/MeiliSearch/runs/3524069020?check_suite_focus=true#step:5:285) on windows when executing our tests on windows. But, I would like to ping you back when we are sure it is not just related to our memory consumption. I didn't found much information about this error and when it is raised on the internet.

Sorry for that ❤️

GuillaumeGomez:
I see, no problem. Well, feel free to send a PR or a more complete explanation on how to reproduce it locally. ;)

curquiza (changes requested):
⚠️ The branch should point to `release-v0.22.0`

ManyTheFish:
Hey @GuillaumeGomez, no issue on your side, we had excessive memory consumption.
Thanks for your quick response!

MarinPostma (approved):
thanks you!!

bors merge

PR #1688, opened 2021-09-09
Geosearch

irevoire:


irevoire:
Hello,
To test this PR easily, you can find a well formatted dataset here: irevoire.ovh/smol-all-countries.json.xz
Download and uncompress it and then:
```
echo '{ "filterableAttributes": ["_geo", "population", "dem"], "sortableAttributes": ["_geo", "population", "dem"], "searchableAttributes": ["name", "asciiname", "alternatenames", "timezone"] }' | mieli settings
mieli add smol-all-countries.json
echo '{ "q": "Lille", "sort": ["_geoPoint(50.63857209529606, 3.064491389569509):asc"] }' | mieli search
```

meilisearch-http/src/index/search.rs
```diff
@@ -247,6 +252,25 @@ impl Index {
     }
 }
 
+fn insert_geo_distance(sorts: &[String], document: &mut Document) {
+    lazy_static::lazy_static! {
+        static ref GEO_REGEX: Regex =
+            Regex::new(r"_geoPoint\(\s*([[:digit:].\-]+)\s*,\s*([[:digit:].\-]+)\s*\)").unwrap();
+    };
+    if let Some(capture_group) = sorts.iter().find_map(|sort| GEO_REGEX.captures(sort)) {
+        // TODO: TAMO: milli encountered an internal error, what do we want to do?
```
curquiza, line 261: Should this stay? Can we open an issue instead? :)

meilisearch-http/src/routes/indexes/search.rs
```diff
@@ -81,6 +79,30 @@ impl From<SearchQueryGet> for SearchQuery {
     }
 }
 
+/// Transform the sort query parameter into something that matches the post expected format.
```
curquiza, line 82: I'm not sure to understand why we need to fix the sort parameter

meilisearch-http/src/routes/indexes/search.rs
```diff
@@ -115,3 +137,42 @@ pub async fn search_with_post(
...
+        assert_eq!(
+            sort,
+            vec![
+                "doggo:asc".to_string(),
+                "_geoPoint(12.45,13.56,2590352):desc".to_string(),
+                "catto:desc".to_string(),
+            ]
+        );
+        let sort = fix_sort_query_parameters("doggo:asc , _geoPoint(1, 2), catto:desc");
+        // This is ugly but eh, I don't want to write a full parser just for this unused route
```
curquiza, line 169: Which unused route do you mean?

curquiza (changes requested):
Sending a search with
```
{ "q": "amer", "sort": ["lol"] }
```
Returns this error:
```json
{
  "errorCode": "bad_request",
  "errorLink": "https://docs.meilisearch.com/errors#bad_request",
  "message": "invalid criterion lol"
}
```

but should return this (as v0.22.0 does)
```json
{
  "errorCode": "invalid_sort",
  "errorLink": "https://docs.meilisearch.com/errors#invalid_sort",
  "message": "Invalid syntax for the sort parameter: lol"
}
```

meilisearch-http/src/routes/indexes/search.rs
irevoire: The search route with the verb get!

meilisearch-http/src/index/search.rs
irevoire: I wanted to know what @MarinPostma thought of this; I think we can take a decision right away and remove this TODO entirely.

meilisearch-http/src/routes/indexes/search.rs
irevoire: Because when we get the sort parameters on the `post` route, we get something looking like that:
```
["price:asc", "date:desc"]
```

While on the `get` route, we get something like that:

```
"price:asc,date:desc"
```

What we were doing until now was a split on the `,` and the parse each thing independently.
Sadly, since the `_geoPoint(x, y)` contains a `,` we can't just do that and hope it works.
So one solution would be to write a full parser, but that would be a lot of code.
The solution I choose is to still split on the `,` but when I encounter `_geoPoint(...` I join all parameters until I find something ending by `):asc` or `):desc`.
It's a bit hacky, but it should work well for our use case.

curquiza:
Also:
Search
```
{ "q": "", "limit": 2, "sort": ["_geoPoint:asc"] }
```
Returns
```
{
  "errorCode": "bad_request",
  "errorLink": "https://docs.meilisearch.com/errors#bad_request",
  "message": "_geoPoint is a reserved keyword and thus can't be used as a ranking rule"
}
```

According to the spec ([link to the related section](https://github.com/meilisearch/specifications/blob/develop/text/0059-geo-search.md#post-search-indexesindexuidsearch-1)), shouldn't it be the first error rather than the second one? At first, when reading the error, I thought I could not use `_geoPoint` with `sort`, but I just miss used it.

curquiza:
And the final one: if I do a search with `sort` using `_geoPoint` but `_geo` is NOT set as `sortableAttributes`, I don't get any error, but I don't get any result.
Doing the same with `filter` (i.e. `_geo` is not set as `filterableAttributes`) returns an error as expected.

curquiza:
Sorry I realize I should have checked this before the milli release

gmourier:
Hi! 👋 

According to the [specification](https://github.com/meilisearch/specifications/blob/develop/text/0059-geo-search.md#as-an-end-user-i-want-to-know-the-document-distance-when-i-am-sorting-around-a-geo-point) `_geoDistance` should be an int. Currently, it's displaying a float value in the search results response.

```json
{
    "hits": [
        {
            "id": 1,
            "_geo": {
                "lat": 45.7597786530353,
                "lng": 4.843585698860502
            },
            "tags": [
                "user"
            ],
            "_geoDistance": 492.0
        },
        {
            "id": 2,
            "_geo": {
                "lat": 45.76185949082058,
                "lng": 4.843446223990196
            },
            "_geoDistance": 717.0
        },
        {
            "id": 3,
            "title": "Hotel de ville",
            "_geo": {
                "lat": 45.76826583227509,
                "lng": 4.835775088782751
            },
            "_geoDistance": 1506.0
        },
        {
            "id": 4,
            "_geo": {
                "lat": 45.76347584820389,
                "lng": 4.79710836356847
            },
            "_geoDistance": 3606.0
        }
    ],
    "nbHits": 4,
    "exhaustiveNbHits": false,
    "query": "",
    "limit": 20,
    "offset": 0,
    "processingTimeMs": 0
}
```

irevoire:
Closing in favour of #1725. 
Sadly you can't try the geosearch right now because the `obkv-json` branch is not working yet.
I'll add you as reviewers once it works.

irevoire:
@curquiza I'll delete this branch only after we merge #1725 in case we change our mind and want to restore this one

PR #1738, opened 2021-09-29
Update Dockerfile following the refactor

curquiza:


MarinPostma:
is there a way to glob that?

curquiza:
I investigated with @tpayet yersterday evening but we did not find any trick to avoid adding this line when a new workspace is created ☹️

MarinPostma (approved):
thanks! 

bors merge

PR #1764, opened 2021-10-04
Bump Milli to improve geosearch errors

### LINKED ISSUE #1734: Expecting error when using `_geoPoint` with `filter`
According to [the spec](https://github.com/meilisearch/specifications/blob/develop/text/0059-geo-search.md#post-search-indexesindexuidsearch):

> 🔴 Using `_geo`, `_geoDistance`, `_geoPoint` in a filter expression cause the API to return an invalid_filter error. `message` should be `:reservedKeyword is a reserved keyword and thus can't be used as a filter expression`.

When I send

```json
'{ "q": "Lille", "filter": "_geoDistance"  }'
```

or

```json
{ "q": "Lille", "filter": "_geoPoint(12.4, 14.6, 1)"  }
```

I don't get this error.

irevoire:
closes #1734 

@curquiza, your two examples still don't work because a filter must be composed of multiples operations; look at my screenshot to see what works and what doesn't.
Is this ok? :thinking: 

@gmourier,
What do you think?

image

And here is a screenshot with all the new errors that have been implemented

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1757,8 +1757,8 @@ dependencies = [
 
 [[package]]
 name = "milli"
-version = "0.17.0"
-source = "git+https://github.com/meilisearch/milli.git?tag=v0.17.0#22551d0941bee1a9cdcf7d5bfc4ca46517dd25f3"
+version = "0.17.1"
+source = "git+https://github.com/meilisearch/milli.git?tag=v0.17.1#31c18f095396d45e67ace32038dda7d0fbc9a4e4"
 dependencies = [
  "bimap",
  "bincode",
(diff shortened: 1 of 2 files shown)


curquiza:
It's ok for me. I used `geoPoint` as `geoRadius` is described in the spec, I mean without any `=`.
https://github.com/meilisearch/specifications/blob/develop/text/0059-geo-search.md#post-search-indexesindexuidsearch
But the behavior is ok for me

irevoire:
Maybe we could define some `reserved_function` errors that are used alone without any operators and followed by parenthesis (`_geoRadius(...)` and `_geo(...)`).
And some `reserved_keywords` that are used as fields names (`_geoDistance` and `_geo`).
But personally, I would say this is enough for a first version :slightly_smiling_face:

curquiza:
I agree, I think it's ok for a first version (if @gmourier is also ok with this of course)
Let's wait for the user feedback :)

curquiza (approved):
Check with @gmourier, it's ok!! 🎉 

bors merge

PR #1769, opened 2021-10-05
Content type

### LINKED ISSUE #1665: Enforce the `Content-Type` header for routes with payload
Checked with @gmourier: we will not open any specification for this since the changes are minimal and make the current behavior more consistent instead of integrating a new feature.

### Goal

Make the routes accepting a payload check if a `Content-Type` header has been defined. If it's not, MeiliSearch should return a `415 Unsupported Media Type` error:

```
{
    "message": "A Content-Type header is missing. Accepted values for Content-Type are: :contentTypeList",
    "code": "missing_content_type",
    "type": "invalid_request",
    "link": "https://docs.meilisearch.com/errors#missing_content_type"
}
```

### Why?

- The current behavior of MeiliSearch is inconsistent: the `POST /indexes/:uid/settings` and `POST /indexes/:uid/search` routes returns a `415` error when not `Content-Type` header are set, but the `POST /indexes/:uid/documents` route does not and infer the type as JSON.
- This enforces more precision and the user to avoid mistakes: he's aware he's sending a JSON
- This is not really impactful for most of the users: the users using SDKs will not see the change since the SDKs send a `Content-Type: application/json` by default.

### Impact for the docs

@meilisearch/docs-team 

The documentation needs to be updated:
- the `POST`/`PUT` examples (for the `cURL` tab) needs to add the following line: `-H 'Content-Type: application/json'`
- we also recommend updating all the docs examples using `--data-binary` instead of `--data`: we ensure the data are not interpreted by `cURL`, which is not the case with `--data`.
Here is the description of `--data-binary` in `man curl`: "(HTTP) This posts data exactly as specified with no extra processing whatsoever."

----

### TODO

The following routes should return `415 Unsupported Media Type` when no `Content-Type` header is defined or if it is defined with an invalid media type:
  - `POST indexes/
  - `POST indexes/:uid/documents`
  - `POST indexes/:uid/documents/delete-batch`
  - `PUT indexes/:uid/documents`
  - `POST indexes/:uid/search`
  - `POST indexes/:uid/settings`
  - all the sub-settings routes like `POST indexes/:uid/settings/XXX`
  - `POST /dumps`

irevoire:
closes #1665

### DIFF
--- a/meilisearch-http/src/lib.rs
+++ b/meilisearch-http/src/lib.rs
@@ -11,7 +11,10 @@ pub mod routes;
 use std::path::Path;
 use std::time::Duration;
 
+use crate::error::{MeilisearchHttpError, ResponseError};
 use crate::extractors::authentication::AuthConfig;
+use actix_web::error::JsonPayloadError;
+use http::header::CONTENT_TYPE;
 pub use option::Opt;
 
 use actix_web::web;
@@ -98,9 +101,28 @@ pub fn configure_data(config: &mut web::ServiceConfig, data: MeiliSearch, opt: &
         .app_data(data)
         .app_data(
             web::JsonConfig::default()
-                .limit(http_payload_size_limit)
-                .content_type(|_mime| true) // Accept all mime types
-                .error_handler(|err, _req| error::payload_error_handler(err).into()),
+                .content_type(|mime| mime == mime::APPLICATION_JSON)
+                .error_handler(|err, req| match err {
+                    JsonPayloadError::ContentType if req.headers().get(CONTENT_TYPE).is_none() => {
+                        ResponseError::from(MeilisearchHttpError::MissingContentType(vec![
+                            mime::APPLICATION_JSON.to_string(),
+                        ]))
+                        .into()
+                    }
+                    JsonPayloadError::ContentType => {
+                        ResponseError::from(MeilisearchHttpError::InvalidContentType(
+                            req.headers()
+                                .get(CONTENT_TYPE)
+                                .unwrap()
+                                .to_str()
+                                .unwrap_or("unknown")
+                                .to_string(),
+                            vec![mime::APPLICATION_JSON.to_string()],
+                        ))
+                        .into()
+                    }
+                    err => error::payload_error_handler(err).into(),
+                }),
         )
         .app_data(PayloadConfig::new(http_payload_size_limit))
         .app_data(
@@ -180,6 +202,7 @@ macro_rules! create_app {
         use actix_web::middleware::TrailingSlash;
         use actix_web::App;
         use actix_web::{middleware, web};
+        use meilisearch_http::error::{MeilisearchHttpError, ResponseError};
         use meilisearch_http::routes;
         use meilisearch_http::{configure_auth, configure_data, dashboard};
 
--- a/meilisearch-http/src/routes/indexes/documents.rs
+++ b/meilisearch-http/src/routes/indexes/documents.rs
@@ -181,9 +181,24 @@ async fn document_addition(
         Some("application/x-ndjson") => DocumentAdditionFormat::Ndjson,
         Some("text/csv") => DocumentAdditionFormat::Csv,
         Some(other) => {
-            return Err(MeilisearchHttpError::InvalidContentType(other.to_string()).into())
+            return Err(MeilisearchHttpError::InvalidContentType(
+                other.to_string(),
+                vec![
+                    "application/json".to_string(),
+                    "application/x-ndjson".to_string(),
+                    "application/csv".to_string(),
+                ],
+            )
+            .into())
         }
-        None => return Err(MeilisearchHttpError::MissingContentType.into()),
+        None => {
+            return Err(MeilisearchHttpError::MissingContentType(vec![
+                "application/json".to_string(),
+                "application/x-ndjson".to_string(),
+                "application/csv".to_string(),
+            ])
+            .into())
+        }
     };
 
     let update = Update::DocumentAddition {
--- a/meilisearch-http/tests/content_type.rs
+++ b/meilisearch-http/tests/content_type.rs
@@ -1,0 +1,109 @@
+mod common;
+
+use crate::common::{GetAllDocumentsOptions, Server};
+use actix_web::test;
+use meilisearch_http::create_app;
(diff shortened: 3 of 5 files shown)


meilisearch-http/src/routes/indexes/documents.rs
```diff
@@ -181,9 +181,24 @@ async fn document_addition(
...
+            )
+            .into())
+        }
+        None => {
+            return Err(MeilisearchHttpError::MissingContentType(vec![
+                "application/json".to_string(),
+                "application/x-ndjson".to_string(),
+                "application/csv".to_string(),
+            ])
+            .into())
```
MarinPostma, line 200: make it a const instead:
```const aACCEPTED_CONTENT_TYPE = [
                "application/json".to_string(),
                "application/x-ndjson".to_string(),
                "application/csv".to_string(),
            ];

// ... later
 return Err(MeilisearchHttpError::MissingContentType(ACCEPTED_CONTENT_TYPE.clone()))

File: meilisearch-http/tests/content_type.rs [hunk 2]
```diff
@@ -0,0 +1,109 @@
+mod common;
+
+use crate::common::{GetAllDocumentsOptions, Server};
+use actix_web::test;
+use meilisearch_http::create_app;
+use serde_json::{json, Value};
+
+#[actix_rt::test]
+async fn strict_json_bad_content_type() {
```
MarinPostma, line 9: :+1:

meilisearch-http/tests/content_type.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    let document = "{}";
+    let server = Server::new().await;
+    let app = test::init_service(create_app!(
+        &server.service.meilisearch,
+        true,
+        &server.service.options
+    ))
+    .await;
+    for route in routes {
```
MarinPostma, line 44: bonus: the partition can be made even more explicit with https://docs.rs/itertools/0.10.1/itertools/trait.Itertools.html#method.cartesian_product :) (i know you like iterators)

MarinPostma (changes requested):

meilisearch-http/tests/content_type.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use crate::common::{GetAllDocumentsOptions, Server};
+use actix_web::test;
+use meilisearch_http::create_app;
+use serde_json::{json, Value};
+
+#[actix_rt::test]
+async fn strict_json_bad_content_type() {
+    let routes = [
+        // all the POST route except the dumps that can be created without any body or content-type
```
curquiza, line 11:
```suggestion
        // all the POST routes except the dumps that cannot be created without any body or content-type
```

curquiza: "Cannot" right? Or maybe I missed something

curquiza (commented):
It works 🦅

meilisearch-http/tests/content_type.rs
curquiza: Also the PUT methods, I don't know if you can add them in this test
- update documents 
- update index (I forgot this one in the issue, I've just updated it)

gmourier:
Thanks for your work guys! ❤️

meilisearch-http/src/lib.rs
```diff
@@ -98,9 +101,28 @@ pub fn configure_data(config: &mut web::ServiceConfig, data: MeiliSearch, opt: &
...
+                                .get(CONTENT_TYPE)
+                                .unwrap()
+                                .to_str()
+                                .unwrap_or("unknown")
+                                .to_string(),
+                            vec![mime::APPLICATION_JSON.to_string()],
+                        ))
+                        .into()
+                    }
+                    err => error::payload_error_handler(err).into(),
```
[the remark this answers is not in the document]
irevoire: Here is the fix:
https://github.com/meilisearch/MeiliSearch/pull/1769/commits/5bc47cdb99ccf24bd83b2eb025ceb81459b88a5f

meilisearch-http/src/lib.rs
```diff
@@ -98,9 +101,28 @@ pub fn configure_data(config: &mut web::ServiceConfig, data: MeiliSearch, opt: &
...
+                        ResponseError::from(MeilisearchHttpError::InvalidContentType(
+                            req.headers()
+                                .get(CONTENT_TYPE)
+                                .unwrap()
+                                .to_str()
+                                .unwrap_or("unknown")
+                                .to_string(),
+                            vec![mime::APPLICATION_JSON.to_string()],
+                        ))
+                        .into()
```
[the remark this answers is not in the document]
irevoire: Here is the fix: https://github.com/meilisearch/MeiliSearch/pull/1769/commits/64e2dd41a887e0f7f6ff9c2b9fda7d948c0a8989

meilisearch-http/src/routes/indexes/documents.rs
irevoire: Here is the fix: https://github.com/meilisearch/MeiliSearch/pull/1769/commits/2f0768dcb03648a40aa32a5613041cd1caa80e32

meilisearch-http/tests/content_type.rs
irevoire: I could, but the problem is that I make two tests independently of the content-type (the no content-type and the good content-type tests).

If I understood correctly, the cartesian_product would not help me here, or I would need to redo these two tests for every content_type instead of doing it only once?

MarinPostma (approved):
Thanks!

bors merge

irevoire:
@MarinPostma I squashed and fixuped some commits together and fixed the clippy warning, we're good to merge now :grin:

meilisearch-http/tests/content_type.rs
irevoire: I duplicated all the tests for the index document route with the put verb :+1:

MarinPostma (dismissed):
:+1: 

bors merge

MarinPostma:
bors cancel

curquiza (approved):
Still a Clippy error, otherwise it's still ok :)

MarinPostma (dismissed):
Bors merge

MarinPostma (approved):

PR #1787, opened 2021-10-07
Handle empty dump

### LINKED ISSUE #1701: Dump import crashed when dump file is created out of empty MeiliSearch
**Describe the bug**
When creating a dump on an empty MeiliSearch instance, the dump is created correctly but when launching it using `--import-dump` the launch fails with the following error: 

```
Error: No such file or directory (os error 2)
```

But the file does exists.

**To Reproduce**
1. Launch empty MeiliSearch instance
2. Create dump
3. Wait for dump completion
4. Stop and restart meilisearch with `--import-dump`
5. Error: `Error: No such file or directory (os error 2)`

**Expected behavior**
No crash.

**Screenshots**

**MeiliSearch version:** 
v0.22.0

irevoire:
Fixes #1701

### DIFF
--- a/meilisearch-lib/src/index_controller/dump_actor/mod.rs
+++ b/meilisearch-lib/src/index_controller/dump_actor/mod.rs
@@ -248,6 +248,7 @@ where
         let mut meta_file = File::create(&meta_path)?;
         serde_json::to_writer(&mut meta_file, &meta)?;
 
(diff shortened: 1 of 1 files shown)


curquiza:
I would rather this PR being merged into `main` to avoid bringing any problem just before the release :)

PR #1806, opened 2021-10-12
Fix csv content-type error message

### LINKED ISSUE #1805: Incorrect error message for content-type error
**Describe the bug**
When an invalid content-type header is sent the the error message that is returned says: `Error message: The Content-Type "applicaiton/csv" is invalid. Accepted values for the Content-Type header are: "application/json", "application/x-ndjson", "application/csv".`, however for `csv` files it should be `text/csv` instead of `applicaiton/csv`.

**To Reproduce**
Send documents to either the add documents or update documents routes with and invalid content-type, such as `Content-Type: application/csv`

**Expected behavior**
The error message should state ``Error message: The Content-Type "applicaiton/csv" is invalid. Accepted values for the Content-Type header are: "application/json", "application/x-ndjson", "text/csv".``

**MeiliSearch version:**
v0.23.1

sanders41:
Fixes #1805

I was not sure if the `application/csv` [here](https://github.com/meilisearch/MeiliSearch/blob/23f11e355d300a572030117ed7860ffbcf2d027c/meilisearch-http/tests/content_type.rs#L29) should also be changed? I'm thinking yes, but `applicaiton/csv` is a bad type.

### DIFF
--- a/meilisearch-http/src/routes/indexes/documents.rs
+++ b/meilisearch-http/src/routes/indexes/documents.rs
@@ -181,7 +181,7 @@ async fn document_addition(
         vec![
             "application/json".to_string(),
             "application/x-ndjson".to_string(),
-            "application/csv".to_string(),
+            "text/csv".to_string(),
         ]
     });
     let format = match content_type {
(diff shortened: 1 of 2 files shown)


curquiza:
bors merge

PR #1817, opened 2021-10-15
Improve RUNs in Dockerfile

nfsec:
Align apk add commands between images.


irevoire:
I think we could also merge the apk update with the apk add what do you think?
Something like that:
```
RUN apk update --quiet && apk add -q --no-cache curl build-base
```

curquiza:
Oh since the user branch does not exist anymore, I have to manually merge.

PR #1824, opened 2021-10-18
Fix indexation perfomances on mounted disk

ManyTheFish:
We were creating all of our tempfiles in data.ms directory, but when the database directory is stored in a mounted disk, tempfiles I/O throughput decreases, impacting the indexation time.

Now, only the persisting tempfiles will be created in the database directory. Other tempfiles will stay in the default tmpdir.


curquiza:
Many, the hero of the mounted disk 🦸

PR #1873, opened 2021-11-03
Change lacking errors

ManyTheFish:


irevoire:
Here is a list of little things I found missing in this PR.
But overall everything seems to be working and I'm not sure these errors are in the scope of this PR?

- [ ] When a primary key inference fails the message is wrong: `Missing primary key.` vs [`The primary key inference process failed because...`](https://github.com/meilisearch/specifications/blob/develop/text/0061-error-format-and-definitions.md#primary_key_inference_failed)
- [ ] When the maximum number of fields is reached the message is wrong: `Maximum number of attributes reached.` vs [`A document cannot contain more than 65,535 fields.`](https://github.com/meilisearch/specifications/blob/develop/text/0061-error-format-and-definitions.md#document_fields_limit_reached)
- [ ] When we send an invalid ranking rule the message is wrong: `invalid ranking rule truc` vs [`:rankingRule` ranking rule is invalid. Valid ranking rules are Words, Typo, Sort, Proximity, Attribute, Exactness and custom ranking rules.`](https://github.com/meilisearch/specifications/blob/develop/text/0061-error-format-and-definitions.md#invalid_ranking_rule)

irevoire (approved):
It's good to me, I think we can merge :+1:

ManyTheFish:
bors merge

PR #1875, opened 2021-11-04
Fix search post event and disk size analytics

gmourier:
- Branch POST search on the post_search aggregator
- Use largest disk `total_space` instead of `available_space`


irevoire (approved):
Good catch thanks a lot!
bors merge

PR #1889, opened 2021-11-10
return 404 when getting a task on an unexisting index

MarinPostma:
When getting a task from `/indexex/uid/tasks`, a 404 is now returned to the user if the index hasn't been created.

When the index is created, the task id that lead to it's creation is ascociated to it, so only tasks that have a higher `task_id` than the creation `task_id` are returned to the the user. This solves the issue when a user deletes an index and then re-creates one with the same name. Now older tasks are not returned.

The older tasks are still available from the `/tasks` route.


irevoire (approved):
bors merge

irevoire:
poor bors

PR #1891, opened 2021-11-10
List task by indexes until it was created

irevoire:


meilisearch-lib/src/tasks/task_store/store.rs
```diff
@@ -161,18 +162,25 @@ impl Store {
...
+                        let (_, id) =
+                            IndexUidTaskIdCodec::bytes_decode(key).ok_or(heed::Error::Decoding)?;
+                        Ok(id)
+                    })
+                    .take_while(|entry| {
+                        entry
+                            .as_ref()
+                            .ok()
+                            // as soon as we are out of the range we exit
+                            .map(|key| !range.contains(key))
```
MarinPostma, line 177: uhmm, what if the first element is out of the range? I think you mean to filter here?

MarinPostma (changes requested):

MarinPostma (approved):
lgtm

Issue #1883, opened 2021-11-09
Make `matches` work with numerical value


curquiza:
Since v0.24.0, and since this issue is done (https://github.com/meilisearch/MeiliSearch/issues/1758) MeiliSearch can apply highlight on numerical fields.

But this does not work when using `matches`.

Example

- With this document:

```json
[
  { "id": 1, "title": "Pride and Prejudice", "price": 3.5 }
]
```

- And this search

```json
{
    "q": "3",
    "attributesToHighlight": ["*"],
    "matches": true
}
```

Current behavior:

```json
{
    "hits": [
        {
            "id": 1,
            "title": "Pride and Prejudice",
            "price": 3.5,
            "_formatted": {
                "id": "1",
                "title": "Pride and Prejudice",
                "price": "<em>3</em>.5"
            },
            "_matchesInfo": {}
        }
    ],
    "nbHits": 1,
    "exhaustiveNbHits": false,
    "query": "3",
    "limit": 20,
    "offset": 0,
    "processingTimeMs": 0
}
```

Expected behavior:

```json
{
    "hits": [
        {
            "id": 1,
            "title": "Pride and Prejudice",
            "price": 3.5,
            "_formatted": {
                "id": "1",
                "title": "Pride and Prejudice",
                "price": "<em>3</em>.5"
            },
            "_matchesInfo": {
                "price": [
                    {
                        "start": 0,
                        "length": 1
                    }
                ]
            }
        }
    ],
    "nbHits": 1,
    "exhaustiveNbHits": false,
    "query": "3",
    "limit": 20,
    "offset": 0,
    "processingTimeMs": 0
}
```

This change should be done in this file: https://github.com/meilisearch/MeiliSearch/blob/main/meilisearch-lib/src/index/search.rs

🔥 Tests should be added 🔥

Thearas:
Hi @curquiza, I‘’m interested in Meilisearch and I'd like to take this issue if no one else is currently working on it.

curquiza:
Hello @Thearas 

Thanks for your interest in this project :)

FYI, we prefer not assigning people to our issues because sometimes people ask to be assigned and never come back, which discourages the volunteer contributors to open a PR to fix this issue.
We will accept and merge the first PR that fixes correctly and well implements the issue following our [contributing guidelines](https://github.com/meilisearch/meilisearch/blob/main/CONTRIBUTING.md).   

We are looking forward to reviewing your PR 🙂

[...]

PR #1893, opened 2021-11-10
Make matches work with numerical value

Thearas:
# Pull Request

## What does this PR do?

Implement #1883.

I have test this PR with unit test. It appears to be working properly:
image

PTAL @curquiza 

## PR checklist

### DIFF
--- a/meilisearch-lib/src/index/search.rs
+++ b/meilisearch-lib/src/index/search.rs
@@ -310,6 +310,9 @@ fn compute_value_matches<'a, A: AsRef<[u8]>>(
         Value::Object(vals) => vals
             .values()
             .for_each(|val| compute_value_matches(infos, val, matcher, analyzer)),
+        Value::Number(number) => {
+            compute_value_matches(infos, &Value::String(number.to_string()), matcher, analyzer)
+        }
         _ => (),
     }
 }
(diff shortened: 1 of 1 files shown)


MarinPostma:
I find it weird that we partially highlight a number. I think a number should be considered as one word @gmourier @curquiza

gmourier:
> I find it weird that we partially highlight a number. I think a number should be considered as one word @gmourier @curquiza

We should keep the same behavior as on a string where MeiliSearch does not match the whole word as well. The highlighting feature consists of having markers to guide the user's eye into the details. I'm not convinced by having different highlighting behaviors between "types" as the default behavior, I think it could create more confusion in the end than being really helpful.

On the other hand, we could give the possibility in the future to define the behavior of the highlighting feature e.g. `fully highlighted`, or even how each "type" could behave in case of partial highlighting.

curquiza (approved):
It works!

Thanks for the PR @Thearas 

⚠️ We need the @MarinPostma's review before any merge :)

ManyTheFish:
Hello @Thearas @MarinPostma!

This PR should fix the issue, thanks!
However, we know that will not highlight numbers perfectly, mainly because of the behavior of our current tokenizer.
We don't have the time to enhance the tokenizer  before the beginning of 2022 and [an issue with the label `good first issue`](https://github.com/meilisearch/tokenizer/issues/62) has been opened for this.

thanks again for your contribution!

MarinPostma (approved):
thank you for your contribution :)

bors merge

PR #1896, opened 2021-11-11
Remove email address from the message at the launch

curquiza:
I suggest removing this email address from the message at the launch since it can encourage people to think this is an email address for support. Is it something we want @meilisearch/devrel-team since we mostly redirect them to the forum or the slack?


ferdi05 (approved):
Totally agree, this address is not made for getting support.
btw we may need to rediscuss the _Support_ topic to avoid some side effects that the team experienced lately

ferdi05 (approved):

ferdi05:
@curquiza I approved the changes, but it looks like there was an issue...

PR #1315, opened 2021-03-29
fix armv7

MarinPostma:
fix armv7 build

this was caused by usize being 32 bit on armv7 and 64bits on all other targeted architectures.

PR #1897, opened 2021-11-11
Add ARM image for Docker to CI

curquiza:
Fixes #1315 

- [x] Publish MeiliSearch's docker image for `arm64`
- [x] Add `workflow_dispatch` event in case we need to re-trigger it after a failure without creating a new release
- [x] Use our own server to run the github runner since this CI is really slow (1h instead of 4h)
- [x] Open an issue for a refactor by merging both files in one file (https://github.com/meilisearch/MeiliSearch/issues/1901)

### DIFF
--- a/.github/workflows/publish-docker-latest.yml
+++ b/.github/workflows/publish-docker-latest.yml
@@ -2,7 +2,6 @@
 on:
   release:
     types: [released]
-  workflow_dispatch:
 
 name: Publish latest image to Docker Hub
 
--- a/.github/workflows/publish-docker-latest.yml
+++ b/.github/workflows/publish-docker-latest.yml
@@ -6,17 +6,24 @@ on:
 name: Publish latest image to Docker Hub
 
 jobs:
-  build:
-    runs-on: ubuntu-18.04
+  docker-latest:
+    runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v2
-      - name: Check if current release is latest
-        run: echo "##[set-output name=is_latest;]$(sh .github/is-latest-release.sh)"
-        id: release
-      - name: Publish to Registry
-        if: steps.release.outputs.is_latest == 'true'
-        uses: elgohr/Publish-Docker-Github-Action@master
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
         with:
-          name: getmeili/meilisearch
-          username: ${{ secrets.DOCKER_USERNAME }}
-          password: ${{ secrets.DOCKER_PASSWORD }}
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          tags: getmeili/meilisearch:latest
--- a/.github/workflows/publish-docker-latest.yml
+++ b/.github/workflows/publish-docker-latest.yml
@@ -2,6 +2,7 @@
 on:
   release:
     types: [released]
+  workflow_dispatch:
 
 name: Publish latest image to Docker Hub
 
--- a/.github/workflows/publish-docker-latest.yml
+++ b/.github/workflows/publish-docker-latest.yml
@@ -10,9 +10,6 @@ jobs:
   docker-latest:
     runs-on: self-hosted
     steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-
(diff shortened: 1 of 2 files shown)


curquiza:
If someone with a MacOs m1 could check it works with the temporary `arm-test` tag I created, it would be really helpful 😄 

```bash
docker run --rm -p 7700:7700 getmeili/meilisearch:test-arm ./meilisearch
```

Maybe @robsontenorio if you have the time :)

iksaku:
Working pretty good on my side. Great work!

robsontenorio:
@curquiza please note, that was the way I have done few months ago. Maybe there is something new on action versions.

curquiza:
> @curquiza please note, that was the way I have done few months ago. Maybe there is something new on action versions.

Thanks! Of course, I did this PR regarding the docs of docker for GH actions, not only your file 😉

> Consider nowdays there is a new Ubuntu ARM64 version. So the builds can be faster than a while ago.

Do you mean GH actions provides Ubuntu ARM64 runners?

.github/workflows/publish-docker-latest.yml
```diff
@@ -2,21 +2,30 @@
 on:
   release:
     types: [released]
+  workflow_dispatch:
```
irevoire, line 5:
```suggestion
```

Are we sure we want to keep this in production?
A new docker image should only be made when a release is made I guess 🤔

.github/workflows/publish-docker-tag.yml
```diff
@@ -3,20 +3,38 @@ on:
...
-      - uses: actions/checkout@v2
-      - name: Publish to Registry
-        uses: elgohr/Publish-Docker-Github-Action@master
-        env:
-          COMMIT_SHA: ${{ github.sha }}
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
```
irevoire, line 18: Here if I understand correctly we setup qemu so we can virtualize “any” platform and build our binary for it?
Is this better than [`cargo-cross`](https://github.com/rust-embedded/cross) which is made specifically to cross compile rust to other platform?

irevoire (dismissed):
I have to little question but overall I think we can merge as-is.
If `cargo-cross` has not been tested and we don’t have time to check if it works I think we should open an issue.

Logically it should reduce the compile-time _drastically_ and simplify the CI.

.github/workflows/publish-docker-latest.yml
curquiza: I added in the description of the PR

> Add `workflow_dispatch` event in case we need to re-trigger it after a failure without creating a new release

In the past, sometimes, we had issues with the docker CI that failed, not related to MeiliSearch, but to the Dockerfile or the CI itself. Because we cannot delete the MeiliSearch release and re-create the same one, I built the docker image on my machine and pushed it to DockerHub manually.
Since this new build of Docker image, I'm not sure I'm able to reproduce it locally since I don't have all the required docker tools on my machine. So I need a way to re-trigger the CI without recreating a release

Why can't I delete and re-create a release for MeiliSearch?
Once a release (not for a RC) is created, it triggers a lot of CIs, and among them, the CI to update Brew and APT package. Once these CIs are triggered, it's not really possible to roll back.

curquiza: But I realized, in any case, I'm stuck... 
Let's say the CI to push v0.24.0 docker image fails. If I want to re-trigger the same CI, I need to provide a `tag`, and the v0.24.0 is already taken. So, either I create a new release, or I find a way to build it manually on my machine...
So you're right, there is no point to let the `workflow_dispatch`

.github/workflows/publish-docker-tag.yml
curquiza: I don't know if it's better, but if I understand correctly,  using `cargo-cross` would involve creating another dockerfile or modifying the current one, right?
The CI creates both docker images, and according to the machine of the user, will pull the right one

robsontenorio: Buildx concurrently builds same image for multiples architecture and that includes a kind of manifest. You are right, when user pulls the image it gets the right one , based on machine architecture. 

I don’t know about cargo-cross. But buildx is the official tool from Docker.

irevoire: Ah ok, from what I see we would need to keep buildx and there is no good integration with cargo-cross so lets call it a day and roll with it 😁

irevoire (approved):
Thank you 🏅 
bors merge

mayel:
👋 it seems that in docker hub the arm image is only available on test tags, will multiarch be made available on 'latest' tag?

brysonreece:
@mayel I would assume it's waiting on #1901

curquiza:
Hello! It will be once v0.25 is out. For the moment, the latest tag does not work for arm

PR #1906, opened 2021-11-16
index uid validator

MarinPostma:
Introduce the `IndexUid` type to validate and represent a valid index uid.

This also allows to return a synchronous error when trying to push an update to a malformed index uid.

should be merged after #1902


meilisearch-http/src/task.rs
```diff
@@ -90,7 +90,7 @@ impl From<Task> for TaskResponse {
             index_uid,
             content,
             events,
-        } = task;
+        } = dbg!(task);
```
irevoire, line 93:
```suggestion
        } = task;
```

irevoire (changes requested):
Oops, maybe you mixed this PR with the rename of `index.wait_update_id` into `index.wait_task` but overall I’m ok with both changes 👍

meilisearch-http/src/task.rs
MarinPostma: yes sorry this must be merged after #1902, since it is partially based on it

irevoire (approved):

PR #1911, opened 2021-11-17
Implement the import of unfinished updates

irevoire:


meilisearch-lib/src/tasks/task_store/store.rs
```diff
@@ -68,11 +68,42 @@ impl Store {
...
+            unfinished_tasks.push(id.get());
+
+            // the id stayed unchanged so this is a safe operation.
+            unsafe {
+                iter.put_current(&id, &task)?;
+            }
+        }
+
+        drop(iter);
+        wtxn.commit()?;
```
MarinPostma, line 97: i think this could be its own function

MarinPostma: uhm i even think that rather than returning the pending tasks when created, we could have `TaskStore` call that method and initialize it's queue itself.

meilisearch-lib/src/tasks/task_store/store.rs
```diff
@@ -98,6 +129,18 @@ impl Store {
...
+        &self,
+        txn: &'a RoTxn,
+    ) -> Result<impl Iterator<Item = Result<Task>> + 'a> {
+        Ok(self
+            .tasks
+            .rev_iter(txn)?
+            .map(|res| res.map(|(_, task)| task))
+            .map(|res| res.map_err(|e| e.into())))
+    }
+
```
MarinPostma, line 143: I don't see where this is called?

MarinPostma (changes requested):

meilisearch-lib/src/tasks/task_store/store.rs
```diff
@@ -68,11 +68,42 @@ impl Store {
...
+            // Since all tasks are ordered, we can stop iterating on all tasks when we encounter our first non-finished task.
+            if task.is_finished() {
+                break;
+            }
+
+            // we only keep the first state. It’s supposed to be a `Created` state.
+            task.events.drain(1..);
+            unfinished_tasks.push(id.get());
+
+            // the id stayed unchanged so this is a safe operation.
```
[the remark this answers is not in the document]
irevoire: [The invariant was](https://docs.rs/heed/0.11.0/heed/struct.RwRevIter.html#method.put_current):
>>> The given key must be equal to the one this cursor is pointing at.

So I guess we are good?

meilisearch-lib/src/tasks/task_store/store.rs
irevoire: I removed it.

meilisearch-lib/src/tasks/task.rs
```diff
@@ -55,6 +54,16 @@ pub struct Task {
     pub events: Vec<TaskEvent>,
 }
 
+impl Task {
+    /// Return true will no longer change its state. We can consider the task as `finished`.
+    /// The only two state that makes a task finished are the `Succeeded` and `Failed` states.
```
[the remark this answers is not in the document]
irevoire: Yes you're right!
```
    /// Return true when a task is finished.
    /// A task is finished when its last state is either `Succeeded` or `Failed`.
```

Here is my new comment

meilisearch-lib/src/tasks/task_store/store.rs
```diff
@@ -59,6 +60,11 @@ pub struct Store {
 }
 
 impl Store {
+    /// Create a new store from the specified `Path`.
+    /// Be really cautious when calling this function, the returned `Store` is
+    /// not altered and is a direct representation of what was laying on the disk.
+    /// You probably want to put back all un-finished update in your queue with the
+    /// `reset_and_return_unfinished_update` method.
```
MarinPostma, line 67:
```suggestion
    /// Be really cautious when calling this function, the returned `Store` may be in an invalid state, with dangling processing tasks.
    /// You will want to patch  all un-finished tasksm and put them in your pending queue with the
    /// `reset_and_return_unfinished_update` method.
```

meilisearch-lib/src/tasks/task_store/store.rs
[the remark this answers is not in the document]
MarinPostma: this is an outdated version or the doc, here is the updated safety note:
```
    /// # Safety
    ///
    /// It is _[undefined behavior]_ to keep a reference of a value from this database while
    /// modifying it, so you can't use the key/value that comes from the cursor to feed
    /// this function.
    ///
    /// In other words: Tranform the key and value that you borrow from this database into an owned
    /// version of them i.e. `&str` into `String`.
    ///
    /// > [Values returned from the database are valid only until a subsequent update operation,
    /// or the end of the transaction.](http://www.lmdb.tech/doc/group__mdb.html#structMDB__val).
```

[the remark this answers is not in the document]
MarinPostma: this seems to be fine since we own both `Task` and `id` :)

MarinPostma (changes requested):

MarinPostma (approved):
great, thanks!

PR #1914, opened 2021-11-18
Replace quickeck with proptest

MarinPostma:
replace quickcheck with proptest, and write a few tests


MarinPostma:
@irevoire, I have fixed the rebase on the #1916 branch instead, I will close this one, since #1916 is based on this.

PR #1915, opened 2021-11-18
Fix task not written back to store after processing

MarinPostma:
tasks were not put back to the store after being updated


irevoire (approved):
👍

PR #1916, opened 2021-11-18
fix tests

MarinPostma:
fix all remaining test but snapshot tests

also fix the `too many fd` issue when running tests.


irevoire (approved):
I think we can merge as-is but we should not forget to do a release of nelson once the tests for the snapshots are implemented :+1:

PR #1928, opened 2021-11-22
Expose indexer opts

ManyTheFish:
# Pull Request

Expose indexer options in order to manage the consumed memory and threads during indexing.

`--max-memory` (env: `MEILI_MAX_MEMORY`): Humanized maximum memory used by the indexer.
`--indexing-jobs` (env: `MEILI_INDEXING_JOBS`): number of threads used by the indexer.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1713,7 +1713,7 @@ checksum = "a3e378b66a060d48947b590737b30a1be76706c8dd7b8ba0f2fe3989c68a853f"
 
 [[package]]
 name = "meilisearch-auth"
-version = "0.26.0"
+version = "0.26.1"
 dependencies = [
  "enum-iterator",
  "heed",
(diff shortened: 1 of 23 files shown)


gmourier:
Thanks, @ManyTheFish 💪, we might decide to explain that in the documentation, it will go through a specification for each newly exposed env var and CLI option anyway.

We may want to keep by default the values that we have chosen at the engine level and to replace it with the user values as soon as they are filled in. What do you think about it?

ManyTheFish:
@gmourier, this draft PR is not meant to be merged for now; I want to know if it fits the need of the  @meilisearch/saas-team before. 🤓
The default behavior is not changed, these new options will override it. 🙂

curquiza:
bors try

curquiza:
Closing this since v0.27.0 is out!

PR #1933, opened 2021-11-24
Extract actual Content-Type from the header

### LINKED ISSUE #1866: Accept `charset` in `Content-Type` header
Replaces https://github.com/meilisearch/MeiliSearch/issues/1791

After discussion with the team, we want to permit `charset` in the `Content-Type` header for the routes requiring a body, so a `Content-Type` header.

### Reminder

The following routes should only accept `application/json` in this header and not `text/csv` or `application/x-ndjson`
- `POST indexes/`
- `PUT indexes/:uid`
- `POST indexes/:uid/documents/delete-batch`
- `POST indexes/:uid/search`
- `POST indexes/:uid/settings`
- all the sub-settings routes like `POST indexes/:uid/settings/XXX`

The following routes can accept `application/json`, `text/csv` or `application/x-ndjson`
- `PUT indexes/:uid/documents`
- `POST indexes/:uid/documents`

Kerollmops:
This PR should fix #1866 by extracting the actual Content-Type from the header and ignoring everything that follows the first semicolon (`;`) which, most of the time, means ignoring the `charset=utf-8`. We use [the `mime_type` method](https://docs.rs/actix-web/3.3.2/actix_web/trait.HttpMessage.html#method.mime_type) to do so.

### DIFF
--- a/meilisearch-http/src/routes/indexes/documents.rs
+++ b/meilisearch-http/src/routes/indexes/documents.rs
@@ -1,6 +1,9 @@
 use actix_web::error::PayloadError;
+use actix_web::http::header::CONTENT_TYPE;
+use actix_web::http::HeaderMap;
 use actix_web::web::Bytes;
 use actix_web::{web, HttpRequest, HttpResponse};
+use bstr::ByteSlice;
 use futures::{Stream, StreamExt};
 use log::debug;
 use meilisearch_error::ResponseError;
@@ -144,9 +147,7 @@ pub async fn add_documents(
     );
 
     let task = document_addition(
-        req.headers()
-            .get("Content-type")
-            .map(|s| s.to_str().unwrap_or("unkown")),
+        extract_content_type(req.headers()),
         meilisearch,
         index_uid,
         params.primary_key,
@@ -176,9 +177,7 @@ pub async fn update_documents(
     );
 
     let task = document_addition(
-        req.headers()
-            .get("Content-type")
-            .map(|s| s.to_str().unwrap_or("unkown")),
+        extract_content_type(req.headers()),
         meilisearch,
         index_uid,
         params.into_inner().primary_key,
@@ -191,7 +190,7 @@ pub async fn update_documents(
 }
 
 async fn document_addition(
-    content_type: Option<&str>,
+    content_type: Option<String>,
     meilisearch: GuardedData<Private, MeiliSearch>,
     index_uid: String,
     primary_key: Option<String>,
@@ -205,7 +204,8 @@ async fn document_addition(
             "text/csv".to_string(),
         ]
     });
-    let format = match content_type {
+
+    let format = match content_type.as_ref().map(AsRef::as_ref) {
         Some("application/json") => DocumentAdditionFormat::Json,
         Some("application/x-ndjson") => DocumentAdditionFormat::Ndjson,
         Some("text/csv") => DocumentAdditionFormat::Csv,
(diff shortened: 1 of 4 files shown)


meilisearch-http/src/routes/indexes/documents.rs
```diff
@@ -274,3 +274,20 @@ pub async fn clear_all_documents(
     debug!("returns: {:?}", task);
     Ok(HttpResponse::Accepted().json(task))
 }
+
+/// Returns the Content-Type header value without the charset=utf-8 or anything,
+/// only the value that is before the first semicolon (;).
```
irevoire, line 279: Is it forbidden to put the charset as the first value?
```
charset=utf-8;type=application/json
```

Kerollmops: Do you mean `charset=utf-8; application/json`? It is very uncommon to see this so I decide to throw them out in this case.

MarinPostma: I would prefer that we use [this method](https://docs.rs/actix-web/3.3.2/actix_web/trait.HttpMessage.html#method.mime_type) that is implemented on `HttpRequest`, and leverage the `Mine` library, instead of doing this ourselves

Kerollmops: It seems, [according to RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.1), that the media type is always before anything else (`charset`).

irevoire (approved):
@MarinPostma I let you merge :+1:

meilisearch-http/src/routes/indexes/documents.rs
```diff
@@ -274,3 +274,23 @@ pub async fn clear_all_documents(
...
+        Some(value) => match value.to_str() {
+            Ok(content) => Some(
+                content
+                    .split_once(";")
+                    .map_or(content, |(ct, _)| ct.trim())
+                    .to_string(),
+            ),
+            Err(_) => Some(format!("{}", value.as_bytes().as_bstr())), // convenient bytes display
+        },
+        None => None,
```
MarinPostma, line 294: I would prefer that we use [this method](https://docs.rs/actix-web/3.3.2/actix_web/trait.HttpMessage.html#method.mime_type) that is implemented on `HttpRequest`, and leverage the Mine library, instead of doing this ourselves

MarinPostma (changes requested):
I would like to use methods provided by actix to handle that :)

meilisearch-http/src/routes/indexes/documents.rs
Kerollmops: Wow, didn't see this, that's awesome. The only difference that I see with my function is that it will not try to return a malformed content-type, one that is not utf-8 encoded for example, and we will not be able to return it in the error messages. But, not sure that's very important. Changing my PR to use this method, thank you!

irevoire (approved):

meilisearch-http/tests/content_type.rs
```diff
@@ -110,3 +110,39 @@ async fn error_json_bad_content_type() {
         }
     }
 }
+
+#[actix_rt::test]
+async fn extract_actual_content_type() {
+    let route = "/indexes/doggo/documents";
+    let documents = "[{}]";
+    let server = Server::new().await;
+    let app = test::init_service(create_app!(
```
MarinPostma, line 119: You can get access to a service from the server itself: https://github.com/meilisearch/MeiliSearch/blob/main/meilisearch-http/tests/common/service.rs,

You are actually creating two instances here

MarinPostma (changes requested):
Can you update the tests to make use of the server

MarinPostma (approved):
look good to me, thank you :)

PR #1936, opened 2021-11-28
Merge Task env and Index env

MarinPostma:
Make the `IndexStore` and the `TaskStore` share the same environment.

One must now be careful not to hold a lock on the task store while trying to write to the index store.


meilisearch-lib/src/tasks/task_store/store.rs
```diff
@@ -206,7 +200,7 @@ impl Store {
 
                 self.uids_task_ids
                     .remap_key_type::<ByteSlice>()
-                    .rev_prefix_iter(txn, &index_uid)?
+                    .rev_prefix_iter(txn, dbg!(&index_uid))?
```
irevoire, line 203:
```suggestion
                    .rev_prefix_iter(txn, &index_uid)?
```

meilisearch-lib/src/tasks/task_store/store.rs
```diff
@@ -238,17 +232,20 @@ impl Store {
             }
         }
 
-        Ok(candidates)
+        Ok(dbg!(candidates))
```
irevoire, line 235:
```suggestion
        Ok(candidates)
```

irevoire (changes requested):

irevoire (approved):

PR #1937, opened 2021-11-28
Restore snapshots for TaskStore

MarinPostma:
restore snapshots after they were disabled by the TaskStore.

must be merged after #1930


meilisearch-lib/src/snapshot.rs
```diff
@@ -0,0 +1,178 @@
...
+            };
+            let job = Job::Snapshot(snapshot_job);
+            self.task_store.register_job(job).await;
+
+            sleep(self.snapshot_period).await;
+        }
+    }
+}
+
+#[allow(dead_code)]
```
irevoire, line 43:
```suggestion
```

irevoire (approved):

PR #1942, opened 2021-11-30
feat(analytics): Send the maximum number of terms number

irevoire:
We realized that the average number of terms number was always going to be ±1 while what actually interests us is to see if 10 cover most of our case. We think we'll be able to validate this theory by looking at the maximum number of terms number.

Implements https://github.com/meilisearch/specifications/pull/95


MarinPostma (approved):
I'm good with the code, but I would like that we find a way to send sufficiently generic data that can be analyzed without having to make changes to the analytics code all the time...

PR #1945, opened 2021-11-30
return tasks in order specified by spec

MarinPostma:
collect task filter candidates in a max binary heap so they are returned in the right order to the user, and add a regression test.


irevoire (approved):
:+1:

PR #1950, opened 2021-11-30
Implement the export of the dump

irevoire:


meilisearch-lib/src/tasks/task_store/mod.rs
```diff
@@ -231,6 +233,19 @@ impl TaskStore {
...
+
+    pub async fn dump(&self, dir_path: &Path) -> Result<()> {
+        let update_dir = dir_path.join("updates");
+        tokio::fs::create_dir(&update_dir).await?;
+        let updates_file = update_dir.join("data.jsonl");
+        dbg!(&updates_file);
+        let mut updates_file = tokio::fs::File::create(updates_file).await?;
+        for task in self.list_tasks(None, None, None).await? {
+            updates_file.write_all(&serde_json::to_vec(&task)?).await?;
+        }
```
MarinPostma, line 245: can we make this function synchronous, and write to the file using `serde_json::to_writer`?
also I am pretty sure you need a line break between each json

MarinPostma: an also use BufferedIo

MarinPostma (changes requested):

MarinPostma (approved):

PR #1951, opened 2021-11-30
compat dump v3

MarinPostma:
compat layer between task v3 and v4

@gmourier, some things are somewhat broken:

- We don't have the information of whether the number of documents added, so we se it to 0. The number of indexed documents should work alright.
- We don't know a setting update comes from a DELETE route, so we assume that it is not, and this may lead to index creation on setting reset on an unexisting index.

I think the remainign should be working correctly


gmourier:
Thanks, @MarinPostma!

irevoire (approved):

PR #1960, opened 2021-12-01
fix snapshot permission

### LINKED ISSUE #1956: Snapshot file permissions changed in v0.24.0
**Describe the bug**
The created snapshots have 600 unix file permissions, whereas previously, they had 644.

With an older Meilisearch version (v0.22.0, I believe), the created snapshots had different unix file permissions than with v0.24.0.

```
-rw-------   1 user  group   2.4M  1 Dec 13:48 data.ms.snapshot
-rw-r--r--   1 user  group   1.2M 22 Nov 11:59 data.ms.snapshot_1141
```

This change broke our workflow. On CI, we use a docker container to create a snapshot, which we deploy at a later stage. We save the snapshot to a volume attached to the docker container.
When deploying the snapshot, we get a permission issue.  
In the newer Meilisearch docker images, the default user was changed, so on the host image filesystem, the owner of the file will be different, and as the permissions also changed, the current user won't have read permission to the file.

Is there a reason to change the 644 permissions to 600?

If there is, then we'll find another solution.

Sorry for any potentially missing information.

**To Reproduce**
Create a snapshot on a unix system with Meilisearch v0.24.0.

**Expected behavior**
The snapshot should have 644 permission.

**MeiliSearch version:** v0.24.0.

**Additional context**
Environment:
Ubuntu 18.04.6 LTS

MarinPostma:
close #1956

tempfiles are created with 600 permission, I changed that to 644 for unix OSes

### DIFF
--- a/meilisearch-lib/src/snapshot.rs
+++ b/meilisearch-lib/src/snapshot.rs
@@ -123,7 +123,6 @@ impl SnapshotJob {
             use std::fs::Permissions;
             use std::os::unix::fs::PermissionsExt;
 
-            dbg!();
             let perm = Permissions::from_mode(0o644);
             _file.set_permissions(perm)?;
         }
--- a/meilisearch-lib/src/snapshot.rs
+++ b/meilisearch-lib/src/snapshot.rs
@@ -117,7 +117,16 @@ impl SnapshotJob {
         let temp_snapshot_file = tempfile::NamedTempFile::new_in(&snapshot_dir)?;
         let temp_snapshot_file_path = temp_snapshot_file.path().to_owned();
         crate::compression::to_tar_gz(temp_snapshot_path, temp_snapshot_file_path)?;
-        temp_snapshot_file.persist(&snapshot_path)?;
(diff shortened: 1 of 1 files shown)


meilisearch-lib/src/snapshot.rs
```diff
@@ -117,7 +117,16 @@ impl SnapshotJob {
...
         let temp_snapshot_file_path = temp_snapshot_file.path().to_owned();
         crate::compression::to_tar_gz(temp_snapshot_path, temp_snapshot_file_path)?;
-        temp_snapshot_file.persist(&snapshot_path)?;
+        let _file = temp_snapshot_file.persist(&snapshot_path)?;
+
+        if cfg!(unix) {
+            use std::fs::Permissions;
+            use std::os::unix::fs::PermissionsExt;
+
+            dbg!();
```
irevoire, line 126:
```suggestion
```

irevoire (changes requested):

irevoire (approved):

PR #1796, opened 2021-10-11
Feature branch: Task store

MarinPostma:
# Feature branch: Task Store

## Spec todo
https://github.com/meilisearch/specifications/blob/develop/text/0060-refashion-updates-apis.md

- [x] The update resource is renamed task. The names of existing API routes are also changed to reflect this change.
- [x]    Tasks are now also accessible as an independent resource of an index. GET - /tasks; GET - /tasks/:taskUid
- [x] The task uid is not incremented by index anymore. The sequence is generated globally.
- [x] A task_not_found error is introduced.
- [x] The format of the task object is updated.
- [x] updateId becomes uid.
- [x] Attributes of an error appearing in a failed task are now contained in a dedicated error object.
- [x] type is no longer an object. It now becomes a string containing the values of its name field previously defined in the type object.
- [x] The possible values for the type field are reworked to be more clear and consistent with our naming rules.
- [x] A details object is added to contain specific information related to a task payload that was previously displayed in the type nested object. Previous number key is renamed numberOfDocuments.
- [x] An indexUid field is added to give information about the related index on which the task is performed.
- [x] duration format has been updated to express an ISO 8601 duration.
- [x] processed status changes to succeeded.
- [x] startedProcessingAt is updated to startedAt.
- [x] processedAt is updated to finishedAt.
- [x] 202 Accepted requests previously returning an updateId are now returning a summarized task object.
- [x] MEILI_MAX_UDB_SIZE env var is updated MEILI_MAX_TASK_DB_SIZE.
- [x] --max-udb-size cli option is updated to --max-task-db-size.
- [x] task object lists are now returned under a results array.
- [x] Each operation on an index (creation, update, deletion) is now asynchronous and represented by a task.

## Todo tech

- [x] Restore Snapshots
- [x] Restore dumps of documents
- [x] Implements the dump of updates
- [x] Error  handling
- [x] Fix stats
- [x] Restore the Analytics
- [x] [Add the new analytics](https://github.com/meilisearch/specifications/pull/92/files)
- [x] Fix tests
- [x] ~Deleting tasks when index is deleted (see bellow)~ see #1891 instead
- [x] Improve details for documents addition and deletion tasks
- [ ] Add integration test
- [ ] Test task store filtering
- [x] Rename `UuidStore` to `IndexMetaStore`, and simplify the trait.
- [x] Fix task store initialization: fill pending queue from hard state
- [x] Synchronously return error when creating an index with an invalid index_uid and add test
[...]

PR #1965, opened 2021-12-02
Reintroduce engine version file

### LINKED ISSUE #1840: MeiliSearch removes indexes when `data.ms` is not compatible
Steps to reproduce
- Start MeiliSearch v0.22.0
- Add documents
- Stop MeiliSearch v0.22.0 and run MeiliSearch v0.23.0 using the same `data.ms`

-> MeiliSearch starts without any error, but should not, and remove the indexes in `data.ms`

MeiliSearch should display the following error
```
Cannot open database, expected MeiliSearch engine version: X.X.X, current engine version Y.Y.Y
```
at it was before, and should not start if data.ms is not compatible without removing the indexes

irevoire:
Right now if you boot up MeiliSearch and point it to a DB directory created with a previous version of MeiliSearch the existing indexes will be deleted. This [used to be](https://github.com/meilisearch/MeiliSearch/commit/51d7c84e7327b14f84522a1f5521f6d98c398a6f) prevented by a startup check which would compare the current engine version vs what was stored in the DB directory's version file, but this functionality seems to have been lost after a few refactorings of the code.

In order to go back to the old behavior we'll need to reintroduce the `VERSION` file that used to be present; I considered reusing the `metadata.json` file used in the dumps feature, but this seemed like the simpler and more approach. As the intent is just to restore functionality, the implementation is quite basic. I imagine that in the future we could build on this and do things like compatibility across major/minor versions and even migrating between formats.

This PR was made thanks to @mbStavola and is basically a port of his PR #1860 after a big refacto of the code #1796.

Closes #1840

### DIFF
--- a/meilisearch-lib/src/index_controller/mod.rs
+++ b/meilisearch-lib/src/index_controller/mod.rs
@@ -39,6 +39,7 @@ use crate::update_file_store::UpdateFileStore;
 
 mod dump_actor;
 pub mod error;
+pub mod versioning;
 
 /// Concrete implementation of the IndexController, exposed by meilisearch-lib
 pub type MeiliSearch = IndexController<HeedMetaStore, MapIndexStore>;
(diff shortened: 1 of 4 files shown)


MarinPostma (approved):

MarinPostma (approved):
bors merge

PR #2005, opened 2021-12-15
auto batching

MarinPostma:
This pr implements auto batching. The basic functioning of this is that all updates that can be batched together are batched together while the previous batch is being processed.

For now, the only updates that can be batched together are the document addition updates (both update and replace), for a single index.

The batching is disabled by default for multiple reasons:
- We need more experimentation with the scheduling techniques
- Right now, if one task fails in a batch, the whole batch fails. We need more permissive error handling when processing document indexation.

There are four CLI options, for now, to interact with how the batch is scheduled:
- `enable-autobatching`: enable the autobatching feature.
- `debounce-duration-sec`: When an update is received, wait that number of seconds before batching and performing the updates. Defaults to 0s.
- `max-batch-size`: the maximum number of tasks per batch, defaults to unlimited.
- `max-documents-per-batch`: the maximum number of documents in a batch, defaults to unlimited. The batch will always contain a least 1 task, no matter the number of documents in that task.

# Implementation

The current implementation is made of 3 major components:

## TaskStore
The `TaskStore` contains all the tasks. When a task is pushed, it is directly registered to the task store.

## Scheduler
The scheduler is in charge of making the batches. At its core, there is a `TaskQueue` and a job queue. `Job`s are always processed first. They are *volatile* tasks, that is, they don't have a TaskId and are not persisted to disk. Snapshots and dumps are examples of Jobs.

If no `Job` is available for processing, then the scheduler attempts to make a `Task` batch from the `TaskQueue`. The first step is to gather new tasks from the `TaskStore` to populate the `TaskQueue`. When this is done, we can prepare our batch. The `TaskQueue` is itself a `BinaryHeap` of `Tasklist`. Each `index_uid` is associated with a `TaskList` that contains all the updates associated with that index uid. Each `TaskList` in the `TaskQueue` is ordered by the id of its first task.

When preparing a batch, the `TaskList` at the top of the `TaskQueue` is popped, and the tasks are popped from the list to make the next batch. If there are remaining tasks in the list, the list is inserted back in the `TaskQueue`.

## UpdateLoop
The `UpdateLoop` role is to perform batch sequentially. Each time updates are pushed to the update store, the scheduler is notified, and will in turn notify the update loop that work can be performed. When notified, the update loop waits some time to wait for more incoming update and then asks the scheduler for the next batch to perform and perform it. When it is done, the status of the task is put back into the store, and the next batch is processed.


Kerollmops (changes requested):
Why do we see the changes between structopt and clap even if there is only one commit?

MarinPostma:
@Kerollmops I needed this change for the `enable-autobatching`, there was a bug with structopt. This is the reason I made the change initally. I have cherry-picked the commit and made another PR with it, as you know.

meilisearch-lib/proptest-regressions/index_resolver/mod.txt
```diff
@@ -17,4 +17,3 @@ cc 3a01c78db082434b8a4f8914abf0d1059d39f4426d16df20d72e1bd7ebb94a6a # shrinks to
...
 cc fb6b98947cbdbdee05ed3c0bf2923aad2c311edc276253642eb43a0c0ec4888a # shrinks to task = Task { id: 0, index_uid: IndexUid("A"), content: CreateIndex { primary_key: Some("") }, events: [] }, index_exists = false, index_op_fails = true, any_int = 0
 cc 1aa59d8e22484e9915efbb5818e1e1ab684aa61b166dc82130d6221663ba00bf # shrinks to task = Task { id: 0, index_uid: IndexUid("a"), content: DocumentDeletion(Clear), events: [] }, index_exists = true, index_op_fails = false, any_int = 0
-cc 2e8644e6397b5f76e0b79f961fa125e2f45f42f26e03c453c9a174dfb427500d # shrinks to task = Task { id: 0, index_uid: IndexUid("0"), content: SettingsUpdate { settings: Settings { displayed_attributes: NotSet, searchable_attributes: NotSet, filterable_attributes: NotSet, sortable_attributes: NotSet, ranking_rules: NotSet, stop_words: NotSet, synonyms: NotSet, distinct_attribute: NotSet, _kind: PhantomData }, is_deletion: false, allow_index_creation: false }, events: [] }, index_exists = false, index_op_fails = false, any_int = 0
```
Kerollmops, line 20: What that's change? Why is this file commited?

meilisearch-lib/src/index_resolver/mod.rs
```diff
@@ -180,33 +188,99 @@ where
...
+                let index = match index {
+                    Ok(index) => index,
+                    Err(e) => {
+                        let error = ResponseError::from(e);
+                        for task in batch.tasks.iter_mut() {
+                            task.events.push(TaskEvent::Failed {
+                                error: error.clone(),
+                                timestamp: Utc::now(),
+                            });
+                        }
```
Kerollmops, line 237:
```suggestion
                        let now = Utc::now();
                        for task in batch.tasks.iter_mut() {
                            task.events.push(TaskEvent::Failed {
                                error: error.clone(),
                                timestamp: now.clone(),
                            });
                        }
```

meilisearch-lib/src/options.rs
```diff
@@ -24,25 +24,69 @@ pub struct IndexerOpts {
...
+    // The maximum number of documents in a document batch. Since batch must contain at least one
+    // batch for the schedulet to be able to make progress, this is ignored if a single update contains
+    // more updates that what is specified.
+    #[clap(long, requires = "enable-autobatching", hide = true)]
+    pub max_documents_per_batch: Option<usize>,
+
+    /// Debounce duration in seconds
+    ///
+    /// When a new task in enqueud, the scheduler waits `debounce_duration_sec` for new updates before
+    /// starting to process the batch.
```
Kerollmops, line 64:
```suggestion
    /// When a new task is enqueued, the scheduler waits for `debounce_duration_sec` for
    /// new updates of the same type before starting to process the batch.
```

meilisearch-lib/src/options.rs
```diff
@@ -24,25 +24,69 @@ pub struct IndexerOpts {
...
+    /// enable the autobatching experimental feature
+    #[clap(long, hide = true)]
+    pub enable_autobatching: bool,
+    // The maximum number of updates that can be batched together. If None, this is unlimited. A
+    // value of 0 is interpreted as 1.
+    #[clap(long, requires = "enable-autobatching", hide = true)]
+    pub max_batch_size: Option<usize>,
+    // The maximum number of documents in a document batch. Since batch must contain at least one
+    // batch for the schedulet to be able to make progress, this is ignored if a single update contains
+    // more updates that what is specified.
```
Kerollmops, line 57:
```suggestion
    // The maximum number of documents in a document batch. Since the scheduler needs
    // at least one update to make progress, the number of documents in a batch will always
    // be at least the size of the first update to batch.
```

Kerollmops (changes requested):
I have done half of the review, will continue later.

meilisearch-lib/proptest-regressions/index_resolver/mod.txt
MarinPostma: Those are protest regression cache, things have changed, and this test wasn't relevant anymore apparently :)

meilisearch-lib/src/index_resolver/mod.rs
```diff
@@ -180,33 +188,99 @@ where
...
+                batch
             }
+            _ => panic!("invalid batch!"),
+        }
+    }
+
+    async fn process_task(&self, task: &Task) -> Result<TaskResult> {
+        let index_uid = task.index_uid.clone();
+        match &task.content {
+            TaskContent::DocumentAddition { .. } => panic!("updates should be handled by batch"),
```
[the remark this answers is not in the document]
MarinPostma: Yes, but it can be a batch of 1

meilisearch-http/src/lib.rs
```diff
@@ -25,8 +26,14 @@ use extractors::payload::PayloadConfig;
 use meilisearch_auth::AuthController;
 use meilisearch_lib::MeiliSearch;
 
+pub static AUTOBATCHING_ENABLED: OnceCell<bool> = OnceCell::new();
```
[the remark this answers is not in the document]
MarinPostma: Yeah i though about it and didn't want to think about memory ordering... But I think relaxed should be plenty enough, noting critical

meilisearch-lib/src/options.rs
MarinPostma: > new updates of the same type before

this implies that we stop waiting if we receive an update of a different type, but it is not the case, we simply wait

Kerollmops: We wait for how long? `debounce_duration_sec`?

MarinPostma: yes, I have updated the docstring, tell me if it makes more sense?

meilisearch-lib/src/options.rs
```diff
@@ -24,25 +24,69 @@ pub struct IndexerOpts {
...
     /// Number of parallel jobs for indexing, defaults to # of CPUs.
-    #[structopt(long)]
+    #[clap(long)]
     pub indexing_jobs: Option<usize>,
 }
 
+#[derive(Debug, Clone, Parser, Default)]
+pub struct SchedulerConfig {
+    /// enable the autobatching experimental feature
+    #[clap(long, hide = true)]
```
Kerollmops, line 49: What's that `hide = true`? Do we want to hide it while this feature is experimental?

meilisearch-lib/src/options.rs
```diff
@@ -24,25 +24,69 @@ pub struct IndexerOpts {
...
+    pub enable_autobatching: bool,
+    // The maximum number of updates of the same type that can be batched together.
+    // If unspecified, this is unlimited. A value of 0 is interpreted as 1.
+    #[clap(long, requires = "enable-autobatching", hide = true)]
+    pub max_batch_size: Option<usize>,
+    // The maximum number of documents in a document batch. Since batches must contain at least one
```
Kerollmops, line 55:
```suggestion
    pub enable_autobatching: bool,

    // The maximum number of updates of the same type that can be batched together.
    // If unspecified, this is unlimited. A value of 0 is interpreted as 1.
    #[clap(long, requires = "enable-autobatching", hide = true)]
    pub max_batch_size: Option<usize>,

    // The maximum number of documents in a document batch. Since batches must contain at least one
```

meilisearch-lib/src/options.rs
```diff
@@ -24,25 +24,69 @@ pub struct IndexerOpts {
...
+    // update for the scheduler to make progress, the number of documents in a batch will be at
+    // least the number of documents of its first update.
+    #[clap(long, requires = "enable-autobatching", hide = true)]
+    pub max_documents_per_batch: Option<usize>,
+
+    /// Debounce duration in seconds
+    ///
+    /// When a new task is enqueued, the scheduler waits for `debounce_duration_sec` seconds for new updates before
+    /// starting to process the batch.
```
Kerollmops, line 64:
```suggestion
    /// When a new task is enqueued, the scheduler waits for `debounce_duration_sec` seconds
    /// for new updates before starting to process a batch of updates of the same type.
```

Kerollmops (changes requested):

Kerollmops:
I'll wait for #2068 to be merged and that this PR to be rebased on it before continuing the review. That'll be easier to review without all of this noise.

meilisearch-lib/src/options.rs
MarinPostma: > Updates of the same time is confusing 
That implies that two settings updates would be processed together, whereas for now only document additions are processed in batches. Furthermore, batching rules will expand in the future, and this may not hold true.

Kerollmops: Ok, so we can keep it like that for now!

meilisearch-lib/src/index_controller/dump_actor/mod.rs
```diff
@@ -291,21 +292,28 @@ impl DumpJob {
...
 
         AuthController::dump(&self.db_path, &temp_dump_path)?;
 
+        //TODO(marin): this is not right, the scheduler should dump iteself, not do it here...
+        self.scheduler
+            .read()
+            .await
+            .dump(&temp_dump_path, self.update_file_store.clone())
+            .await?;
+
```
irevoire, line 315:
```suggestion
        //TODO(marin): this is not right, the scheduler should dump itself, not do it here...
        self.scheduler
            .read()
            .await
            .dump(&temp_dump_path, self.update_file_store.clone())
            .await?;

```

So, should we keep this?

irevoire (changes requested):
There is an issue when I try to see the status of a dump.
Meilisearch wait for the dump to be finished before answering.

Kerollmops (changes requested):
LOOK good to me!

However, I am just not sure about the time we will take when we boot the server, it needs to read all of the tasks in the queue to find the batches, if there is a lot of tasks (which can highly be the case) it could take a lot of time that's `O(N)` on the number of tasks.

What do we do when we list all of the tasks? Do we deserialize the content of it? I don't think so.

meilisearch-lib/src/index/search.rs
```diff
@@ -579,7 +579,7 @@ impl<'a, A: AsRef<[u8]>> Formatter<'a, A> {
                 let mut tokens = analyzed.reconstruct().peekable();
 
                 while let Some((word, token)) =
-                    tokens.next_if(|(_, token)| matcher.matches(token.text()).is_none())
+                    tokens.next_if(|(_, token)| matcher.matches(token).is_none())
```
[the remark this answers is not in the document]
MarinPostma: what do you mean?

MarinPostma:
@Kerollmops I agree that it may be an issue. This only happens when you restart meilisearch AND have a lot of pending updates. I am running some test without autobatching enabled to simulate a lot of pending updates and restart meilisearch to see how it impact startup time, and I'll report.

meilisearch-lib/src/index/search.rs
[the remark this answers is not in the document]
Kerollmops: A `Token` struct doesn't return the same data from the `Deref::deref` and `text` functions, IIRC, be carreful.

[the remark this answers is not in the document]
irevoire: @Kerollmops This `Token` doesn't come from the filters, it comes from the tokenizer and from what I see `Deref` is not implemented on it `Token` :grin:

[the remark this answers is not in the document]
MarinPostma: but wait, `matches` expects a `Token` now, not a `&str` anymore, I am just passing it what it expects, I don't understand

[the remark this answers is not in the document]
Kerollmops: Ooops, I wasn't sure about which `Token` was behaving like that, OK thank you!

MarinPostma:
followup @irevoire @Kerollmops: first of all there was a bug, the scheduler was not loading the pending updates on startup, this is fixed now, and I notify the update loop that there is work to do on startup. The good new is that this does not block startup, and is done asynchronously when the update loop asks for the first batch. The other good new it that with over 1 million pending updates, it took no more that 2-3s to load it all :)

there is no bad new

meilisearch-lib/Cargo.toml
```diff
@@ -55,12 +55,13 @@ reqwest = { version = "0.11.4", features = ["json", "rustls-tls"], default-featu
...
 derivative = "2.2.0"
 fs_extra = "1.2.0"
+atomic_refcell = "0.1.8"
 
 [dev-dependencies]
 actix-rt = "2.2.0"
 mockall = "0.10.2"
 paste = "1.0.5"
-nelson = { git = "https://github.com/MarinPostma/nelson.git", rev = "e5f4ff046c21e7e986c7cb31550d1c9e7f0b693b"}
+nelson = { git = "https://github.com/MarinPostma/nelson.git", rev = "675f13885548fb415ead8fbb447e9e6d9314000a"}
```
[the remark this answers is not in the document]
MarinPostma: done

curquiza:
bors try

curquiza:
bors try

MarinPostma:
bors try

MarinPostma:
@curquiza it is fixed, a bad merge messed up with the tests

meilisearch-lib/src/tasks/scheduler.rs
```diff
@@ -1,253 +1,526 @@
...
+
+    pub async fn schedule_job(&mut self, job: Job) {
+        self.jobs.push_back(job);
+        self.notify();
+    }
+
+    async fn fetch_pending_tasks(&mut self) -> Result<()> {
+        // We must NEVER re-enqueue an already processed task! It's content uuid would point to an unexisting file.
+        //
+        // TODO(marin): This may lead to slower startup times, and should be optimized in the future.
```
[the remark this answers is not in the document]
MarinPostma: cf https://github.com/meilisearch/meilisearch/pull/2005#issuecomment-1023097535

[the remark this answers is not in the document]
MarinPostma: leaving that here because it could be done at some point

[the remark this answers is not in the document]
MarinPostma: actually I will edit the comment

Kerollmops (approved):
That's better, thank you!

MarinPostma:
bors merge

irevoire (approved):
👍

PR #2007, opened 2021-12-15
bug(lib): ignore primary if already set on document addition

### LINKED ISSUE #2002: v0.25.0 RC: add documents with primary key should not fail
Checked with @gmourier

---

Using v0.25.0rc0, I noticed a change of behavior that is not expected regarding the document addition/update with a primary key

Steps to reproduce
- add a batch of documents with a primary key as query parameter
- re-add the same batch of documents with the same primary key as query parameter

The second batch fails because a primary already exists. It should not fail and just ignore the primary key, no matter is the value of the primary key.

TODO:
- [ ] Fix the bug
- [ ] 🔥 Add tests 🔥

MarinPostma:
Ignore the primary key if it is already set on documents updates. Add a test for verify behaviour.

close #2002

### DIFF
--- a/meilisearch-http/tests/documents/add_documents.rs
+++ b/meilisearch-http/tests/documents/add_documents.rs
@@ -1032,3 +1032,26 @@ async fn error_primary_key_inference() {
 
     assert_eq!(response["error"], expected_error);
 }
+
+#[actix_rt::test]
+async fn add_documents_with_primary_key_twice() {
+    let server = Server::new().await;
+    let index = server.index("test");
+
+    let documents = json!([
+        {
+            "title": "11",
+            "desc": "foobar"
+        }
+    ]);
+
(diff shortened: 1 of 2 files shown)


curquiza (changes requested):
The commits related to the CONTRIBUTING.md file should not be in this PR.

curquiza (approved):
💪

Kerollmops (commented):
Not sure about what happens if I send two batches of documents but with two different primary keys, shouldn't we raise an error as the primary key already exists **and** is different from the previous one?

curquiza:
At the moment, we want to roll back to the behavior of v0.24.0 -> ignore the primary key if it's already defined, no matter is the value of the new primary key sent.
We indeed might want to improve this behavior later, but not in v0.25.0 :)

curquiza:
bors merge

PR #2025, opened 2021-12-20
Fix security index creation

### LINKED ISSUE #2024: v0.25.0 RC: `documents.add` right should not allow to create indexes
We identified a potential security issue regarding the access rights when using API keys.

Only available in v0.25.0 release candidates at the moment, so no issue for our users in production since we strongly do not recommend using RCs in production. 

When creating this kind of API key

```json
{
    "description": "Key to add documents for all the indexes.",
    "actions": [
        "documents.add"
    ],
    "indexes": [
        "*"
    ],
    "expiresAt": null
}
```

Currently, with this key, I can add documents in every index, but also, I can create as many indexes as I want, without the `indexes.create` right.
This is due to the implicit index creation: when adding documents, MeiliSearch automatically creates an index if it does not exist.

### Solution

We chose to make the API return an asynchronous error: when getting the task information using the `GET /tasks/:task_uid` route, the user will see an `Index not found error`.

The solution we decided to implement is not perfect since it can bring some user experience frustration, but we would rather a user frustration than a security issue for this first implementation of the API keys.

ManyTheFish:
Forbid index creation on alternates routes when the action `index.create` is not given

fix #2024


MarinPostma (approved):
look good to me, thanks @ManyTheFish

ManyTheFish:
bors merge

PR #2026, opened 2021-12-20
Bug(auth): Parse YMD date

### LINKED ISSUE #2017: v0.25.0 RC Post key by sending only the date part `YYYY-MM-DD` failed
**Describe the bug**

Post a new key by sending only the date part for `expiresAt` field failed.
According to the [API Keys](https://github.com/meilisearch/specifications/blob/develop/text/0085-api-keys.md#155-post-patch---keys---api-key-object-payload-definition) specification and the API response, the `expiresAt` value can be defined only with the date such as: `YYYY-MM-DD`.

| field       | type    | required |description                     |
|-------------|---------|----------|--------------------------------|
| expiresAt   | string  | Required | The expiration date and time as `ISO-8601` format. `null` equals to no expiration time. Sending only the date part e.g `2021-12-01` leads to having an `expiresAt` value set to `2021-12-01T00:00:00`. **Default**: `No Default` |

**To Reproduce**

*Post on `/keys` route:*
```json
{
    "description": "Indexing Products API key",
    "actions": [
        "documents.add"
    ],
    "indexes": ["products"],
    "expiresAt": "2022-01-01"
}
```
*Response:*
```json
{
	"message": "expiresAt field value `\"2022-01-01\"` is invalid. It should be in ISO-8601 format to represents a date or datetime in the future or specified as a null value. e.g. 'YYYY-MM-DD' or 'YYYY-MM-DDTHH:MM:SS'.",
	"code": "invalid_api_key_expires_at",
	"type": "invalid_request",
	"link": "https://docs.meilisearch.com/errors#invalid_api_key_expires_at"
}
```

**Expected behavior**

The `expiredAt` must be able to be defined by a date format or the API message must be modified.

**MeiliSearch version:** v0.25.0rc1

### LINKED ISSUE #2024: v0.25.0 RC: `documents.add` right should not allow to create indexes

ManyTheFish:
Use NaiveDate to parse YMD date instead of NaiveDatetime

fix #2017

### DIFF
--- a/meilisearch-auth/src/key.rs
+++ b/meilisearch-auth/src/key.rs
@@ -1,7 +1,7 @@
 use crate::action::Action;
 use crate::error::{AuthControllerError, Result};
 use crate::store::{KeyId, KEY_ID_LENGTH};
-use chrono::{DateTime, NaiveDateTime, Utc};
+use chrono::{DateTime, NaiveDate, NaiveDateTime, Utc};
 use rand::Rng;
 use serde::{Deserialize, Serialize};
 use serde_json::{from_value, Value};
@@ -142,8 +142,8 @@ fn parse_expiration_date(value: &Value) -> Result<Option<DateTime<Utc>>> {
                     .map(|naive| DateTime::from_utc(naive, Utc))
             })
             .or_else(|_| {
-                NaiveDateTime::parse_from_str(string, "%Y-%m-%d")
-                    .map(|naive| DateTime::from_utc(naive, Utc))
+                NaiveDate::parse_from_str(string, "%Y-%m-%d")
+                    .map(|naive| DateTime::from_utc(naive.and_hms(0, 0, 0), Utc))
             })
             .map_err(|_| AuthControllerError::InvalidApiKeyExpiresAt(value.clone()))
             // check if the key is already expired.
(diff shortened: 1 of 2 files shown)


MarinPostma (approved):

ManyTheFish:
bors merge

curquiza:
bors merge

ManyTheFish:
bors try

curquiza:
bors merge

PR #2056, opened 2022-01-05
Allow any header for CORS

curquiza:
Bug fix: trigger a CORS error when trying to send the `User-Agent` header via the browser

@bidoubiwa thanks for the bug report!


MarinPostma (approved):
thanks!

curquiza:
checked with @gmourier and the whole @meilisearch/core-team: we would rather accept every request and use only the needed ones in the codebase

MarinPostma (approved):

curquiza:
bors merge

PR #2057, opened 2022-01-05
fix(dump): Uncompress the dump IN the data.ms 

### LINKED ISSUE #1833: Import dumps in docker instance got os error 16
**Describe the bug**
when use MeiliSearch docker instance with data volumes to import dumps get os error 16:
```bash
docker run -d --name meilisearch -p 7700:7700 -v /root/meilisearch/data:/data -v /root/meilisearch/dumps:/dumps getmeili/meilisearch:v0.23.1 ./meilisearch --db-path /data --import-dump /dumps/20211020-04504144.dump
```

and if without data volumes is works well:
```
docker run -d --name meilisearch -p 7700:7700 -v /root/meilisearch/dumps:/dumps getmeili/meilisearch:v0.23.1 ./meilisearch --import-dump /dumps/20211020-04504144.dump
```
but it is not convenient to direct import dump to db path while db path is a volume, needs docker cp data to local directory.

is this just as designed?

**MeiliSearch version:** [v0.23.1]

irevoire:
When loading a dump with docker, we had two problems.
After creating a tempdirectory, uncompressing and re-indexing the dump:
1. We try to `move` the new “data.ms” onto the currently present
   one. The problem is that if the `data.ms` is a mount point because
   that's what peoples do with docker usually. We can't override
   a mount point, and thus we were throwing an error.
2. The tempdir is created in `/tmp`, which is usually quite small AND may not
   be on the same partition as the `data.ms`. This means when we tried to move
   the dump over the `data.ms`, it was also failing because we can't move data
   between two partitions.
------------------
1 was fixed by deleting the *content* of the `data.ms` and moving the *content*
of the tempdir *inside* the `data.ms`. If someone tries to create volumes inside
the `data.ms` that's his problem, not ours.
2 was fixed by creating the tempdir *inside* of the `data.ms`. If a user mounted
its `data.ms` on a large partition, there is no reason he could not load a big
dump because his `/tmp` was too small. This solves the issue; now the dump is
extracted and indexed on the same partition the `data.ms` will lay.

fix #1833

### DIFF
--- a/meilisearch-lib/src/index_controller/dump_actor/mod.rs
+++ b/meilisearch-lib/src/index_controller/dump_actor/mod.rs
@@ -185,7 +185,7 @@ pub fn load_dump(
     let mut meta_file = File::open(&meta_path)?;
     let meta: MetadataVersion = serde_json::from_reader(&mut meta_file)?;
 
-    let tmp_dst = tempfile::tempdir()?;
+    let tmp_dst = tempfile::tempdir_in(dst_path.as_ref())?;
 
     info!(
         "Loading dump {}, dump database version: {}, dump version: {}",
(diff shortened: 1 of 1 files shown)


MarinPostma (approved):

irevoire:
bors merge

PR #2059, opened 2022-01-06
change indexed doc count on error

### LINKED ISSUE #2053: V0.25.0rc3 Task details contains a null value when it can contains a finite value
**Describe the bug**

The value of `indexedDocuments` could be `0` rather than `null` when a task fails.

```json
{
            "uid": 1,
            "indexUid": "test",
            "status": "failed",
            "type": "documentAddition",
            "details": {
                "receivedDocuments": 2500,
                "indexedDocuments": null
            },
            "error": {
                "message": "Index `test` not found.",
                "code": "index_not_found",
                "type": "invalid_request",
                "link": "https://docs.meilisearch.com/errors#index_not_found"
            },
            "duration": "PT0.000578S",
            "enqueuedAt": "2022-01-04T12:09:36.709163Z",
            "startedAt": "2022-01-04T12:09:36.709532Z",
            "finishedAt": "2022-01-04T12:09:36.709741Z"
}
```

**To Reproduce**
1. Add documents on an inexistent index with an API key that do not have the right to create an index. (It also happens if the user tries to push documents missing a document identifier so it's probably not related to that particular case)
2. Fetch the related task

**Expected behavior**

When a real value can be given, it should be preferred over `null` in the details of a task for the field `indexedDocuments` and `deletedDocuments`. In this particular case, we know that 0 documents have been indexed.

**MeiliSearch version:** [e.g. v0.25.0rc3]

MarinPostma:
change `indexed_documents` and `deleted_documents` to return 0 instead of null when empty when the task has failed.

close #2053

### DIFF
--- a/meilisearch-http/src/task.rs
+++ b/meilisearch-http/src/task.rs
@@ -215,6 +215,27 @@ impl From<Task> for TaskView {
                 (TaskStatus::Succeeded, None, Some(*timestamp))
             }
             TaskEvent::Failed { timestamp, error } => {
+                match details {
+                    Some(TaskDetails::DocumentDeletion {
+                        ref mut deleted_documents,
+                        ..
+                    }) => {
+                        deleted_documents.replace(0);
+                    }
+                    Some(TaskDetails::ClearAll {
+                        ref mut deleted_documents,
+                        ..
+                    }) => {
+                        deleted_documents.replace(0);
+                    }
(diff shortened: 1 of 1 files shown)


irevoire (dismissed):
:+1:

irevoire (approved):

irevoire:
bors merge

PR #2066, opened 2022-01-11
bug(http): fix task duration

MarinPostma:
@gmourier found that the duration in the task view was not computed correctly, this pr fixes it.

@curquiza, I let you decide if we need to make a hotfix out of this or wait for the next release. This is not breaking.

### DIFF
--- a/meilisearch-http/src/task.rs
+++ b/meilisearch-http/src/task.rs
@@ -224,12 +224,12 @@ impl From<Task> for TaskView {
             _ => unreachable!("A task must always have a creation event."),
         };
 
-        let duration = finished_at.map(|ts| (ts - enqueued_at));
-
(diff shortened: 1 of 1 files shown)


Kerollmops (changes requested):
Be careful, your commit seems anonymous or something. Maybe a missing email?

irevoire (approved):
:+1:

curquiza:
Not something that deserves a hotfix. If a bigger bug is reported, I will put this bug fix in the patch version as well :)
Let's merge this on `main` only for the moment

curquiza (approved):

MarinPostma:
Bors merge

MarinPostma:
bors merge

PR #2068, opened 2022-01-12
chore(http): migrate from structopt to clap3

MarinPostma:
migrate from structopt to clap3

This fix the long lasting issue with flags require a value, such as `--no-analytics` or `--schedule-snapshot`.

All flag arguments now take NO argument, i.e:
`meilisearch --schedule-snapshot true` becomes `meilisearch --schedule-snapshot`

as per https://docs.rs/clap/latest/clap/struct.Arg.html#method.env, the env variable is defines as:
> A false literal is n, no, f, false, off or 0. An absent environment variable will also be considered as false. Anything else will considered as true.

@gmourier 
@curquiza 
@meilisearch/docs-team

### DIFF
--- a/meilisearch-http/src/option.rs
+++ b/meilisearch-http/src/option.rs
@@ -129,16 +129,6 @@ pub struct Opt {
 }
 
 impl Opt {
-    /// Wether analytics should be enabled or not.
-    #[cfg(all(not(debug_assertions), feature = "analytics"))]
-    pub fn analytics(&self) -> bool {
-        match self.no_analytics {
-            None => true,
-            Some(None) => false,
-            Some(Some(disabled)) => !disabled,
-        }
-    }
-
     pub fn get_ssl_config(&self) -> anyhow::Result<Option<rustls::ServerConfig>> {
         if let (Some(cert_path), Some(key_path)) = (&self.ssl_cert_path, &self.ssl_key_path) {
             let client_auth = match &self.ssl_auth_path {
--- a/meilisearch-http/src/option.rs
+++ b/meilisearch-http/src/option.rs
@@ -4,127 +4,127 @@ use std::path::PathBuf;
 use std::sync::Arc;
 
 use byte_unit::Byte;
+use clap::Parser;
 use meilisearch_lib::options::IndexerOpts;
 use rustls::internal::pemfile::{certs, pkcs8_private_keys, rsa_private_keys};
 use rustls::{
     AllowAnyAnonymousOrAuthenticatedClient, AllowAnyAuthenticatedClient, NoClientAuth,
     RootCertStore,
 };
-use structopt::StructOpt;
 
 const POSSIBLE_ENV: [&str; 2] = ["development", "production"];
 
-#[derive(Debug, Clone, StructOpt)]
+#[derive(Debug, Clone, Parser)]
 pub struct Opt {
     /// The destination where the database must be created.
-    #[structopt(long, env = "MEILI_DB_PATH", default_value = "./data.ms")]
+    #[clap(long, env = "MEILI_DB_PATH", default_value = "./data.ms")]
     pub db_path: PathBuf,
 
     /// The address on which the http server will listen.
-    #[structopt(long, env = "MEILI_HTTP_ADDR", default_value = "127.0.0.1:7700")]
+    #[clap(long, env = "MEILI_HTTP_ADDR", default_value = "127.0.0.1:7700")]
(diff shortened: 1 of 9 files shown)


MarinPostma:
ping @Kerollmops just to make sure that the `no-analytics` behavior is correct

irevoire:
:clap: :clap: :clap:

meilisearch-http/src/option.rs
```diff
@@ -4,127 +4,127 @@ use std::path::PathBuf;
...
-    #[structopt(long, env = "MEILI_ENV", default_value = "development", possible_values = &POSSIBLE_ENV)]
+    #[clap(long, env = "MEILI_ENV", default_value = "development", possible_values = &POSSIBLE_ENV)]
     pub env: String,
 
     /// Do not send analytics to Meili.
     #[cfg(all(not(debug_assertions), feature = "analytics"))]
-    #[structopt(long, env = "MEILI_NO_ANALYTICS")]
-    pub no_analytics: Option<Option<bool>>,
+    #[clap(long, env = "MEILI_NO_ANALYTICS")]
+    pub no_analytics: bool,
```
Kerollmops, line 41: I am at the regret to say that you must keep it like before, use an `Option<Option<bool>>` to keep the correct behavior: being able to specify `true` or `false` or nothing to enable the analitycs.

Kerollmops (changes requested):
I am quite happy with this PR, thank you!

meilisearch-http/src/option.rs
MarinPostma: The goal is not to have to specify true or false, i.e is no analytics is not passed, then analytics are enabled, and to disable them you need to pass the flag

gmourier: @nicolasvienot I would like to have your opinion on this 🤓

> All flag arguments now take NO argument, i.e:
> meilisearch --schedule-snapshot true becomes meilisearch --schedule-snapshot

What do you think of this? Do you see a limit in your use of the CLI, I'm thinking mostly of the manipulation of containers dynamically within kubernetes.

Kerollmops: Yup, by using an `Option<Option<bool>>` you do both: specify `true` or `false` and also not specify anything. It is the behavior explained in the documentation and the behavior that @nicolasvienot wanted for the SaaS. Waiting for Guigui about what we should do, keep or change.

curquiza:
@meilisearch/docs-team, I will create a detailed issue regarding this :)

meilisearch-http/src/option.rs
nicolasvienot: Hey @gmourier,

I don't think this is an issue for the manipulation of containers dynamically within Kubernetes or MeiliSearch Cloud.
We should be able to add the flag conditionally, even if this was a bit easier with `true` / `false`.

As long as this seems better for the core team and it's the same behavior everywhere, I see no objection. 😃 

@Kerollmops I made the issue about the `MEILI_NO_ANALYTICS` because I was following the doc and it seemed wrong, but we actually handle both for now because we still support < `0.25.0` versions.

MarinPostma: @nicolasvienot, there is a misanderstanding. For the environment variable, the behaviour is the following:
> A false literal is n, no, f, false, off or 0. An absent environment variable will also be considered as false. Anything else will considered as true.
that means that a set empty env variable is considered a true as well

for the cli flag, the flag commands do not accept an argument anymore.

If you plan on only using the env variables, this should not change anything for you.

nicolasvienot: @MarinPostma Thanks for the clarification. I just checked and as you said, since we are using env variables for now, this will still work fine. :seal:

gmourier: Thanks, @nicolasvienot, and @MarinPostma!

@curquiza Do you think we can plan it for v0.26.0?

MarinPostma:
@curquiza @Kerollmops, can we merge that?

Kerollmops (changes requested):
I am OK to merge this, just make sure to fix Clippy.

curquiza (approved):
Thanks for this, sorry for the delay! I've created the related issue https://github.com/meilisearch/MeiliSearch/issues/2097

We can merge this when the clippy error is fixed!

Kerollmops (approved):
I approve and merge this PR, thank you!
bors merge

MarinPostma:
bors merge

PR #2073, opened 2022-01-13
fix(dump): Fix the import of dump from the v24 and before

irevoire:
When trying to import [this dump](https://files.slack.com/files-pri/TD4JJU04D-F02UK3ZP2KA/download/dumps-steam.dump) we were getting this error:
```
Error: Internal error: No such file or directory (os error 2)
```

It’s because in any dump made from before the v25 we had no `key_file_store`.

PR #2076, opened 2022-01-13
fix(dump): Fix the import of dump from the v24 and before

irevoire:
Same as https://github.com/meilisearch/MeiliSearch/pull/2073 but on main this time


ManyTheFish (approved):
bors merge

PR #2086, opened 2022-01-18
feat(analytics): send the whole set of cli options instead of only the snapshot

### LINKED ISSUE #2088: Add telemetry for v0.26.0
To keep the transparency with our users, here are the added telemetries to our product. We need to add them to the docs, @meilisearch/docs-team.

- [ ] add telemetry for the CLI flags and options, see [this spec changes](https://github.com/meilisearch/specifications/pull/107).

⚠️ User data are still anonymized and you can easily disable the analytics by passing the `--no-analytics=true` flag when launching MeiliSearch.

More about the [MeiliSearch telemetry usage](https://docs.meilisearch.com/learn/what_is_meilisearch/telemetry.html#telemetry).

irevoire:
Fixes #2088

### DIFF
--- a/meilisearch-http/src/analytics/segment_analytics.rs
+++ b/meilisearch-http/src/analytics/segment_analytics.rs
@@ -1,6 +1,6 @@
 use std::collections::{BinaryHeap, HashMap, HashSet};
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
@@ -211,10 +211,30 @@ impl Segment {
                     "server_provider": std::env::var("MEILI_SERVER_PROVIDER").ok(),
             })
         });
-        let infos = json!({
-            "env": opt.env.clone(),
-            "has_snapshot": opt.schedule_snapshot,
-        });
+        // The infos are all cli option except every option containing sensitive information.
+        // We consider an information as sensible if it contains a path, an address or a key.
+        let infos = {
+            // First we see if any sensitive fields were used.
+            let db_path = opt.db_path != PathBuf::from("./data.ms");
+            let import_dump = opt.import_dump.is_some();
+            let dumps_dir = opt.dumps_dir != PathBuf::from("dumps/");
+            let import_snapshot = opt.import_snapshot.is_some();
+            let snapshots_dir = opt.snapshot_dir != PathBuf::from("snapshots/");
+            let http_addr = opt.http_addr != "127.0.0.1:7700";
+
+            let mut infos = serde_json::to_value(opt).unwrap();
+
+            // Then we overwrite all sensitive field with a boolean representing if
+            // the feature was used or not.
+            infos["db_path"] = json!(db_path);
+            infos["import_dump"] = json!(import_dump);
+            infos["dumps_dir"] = json!(dumps_dir);
+            infos["import_snapshot"] = json!(import_snapshot);
+            infos["snapshot_dir"] = json!(snapshots_dir);
+            infos["http_addr"] = json!(http_addr);
+
+            infos
+        };
 
         let number_of_documents = stats
             .indexes
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -494,6 +494,7 @@ version = "4.0.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "956ffc5b0ec7d7a6949e3f21fd63ba5af4cffdc2ba1e0b7bf62b481458c4ae7f"
 dependencies = [
+ "serde",
  "utf8-width",
 ]
 
--- a/meilisearch-http/Cargo.toml
+++ b/meilisearch-http/Cargo.toml
@@ -30,7 +30,7 @@ arc-swap = "1.3.2"
 async-stream = "0.3.2"
 async-trait = "0.1.51"
 bstr = "0.2.17"
-byte-unit = { version = "4.0.12", default-features = false, features = ["std"] }
+byte-unit = { version = "4.0.12", default-features = false, features = ["std", "serde"] }
 bytes = "1.1.0"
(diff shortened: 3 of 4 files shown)


meilisearch-http/src/analytics/segment_analytics.rs
```diff
@@ -210,6 +210,21 @@ impl Segment {
                     "server_provider": std::env::var("MEILI_SERVER_PROVIDER").ok(),
             })
         });
+        let info = {
+            let db_path = opt.db_path != PathBuf::from("./data.ms");
+            let import_dump = opt.import_dump.is_some();
+            let dumps_dir = opt.dumps_dir != PathBuf::from("dumps/");
+            let import_snapshot = opt.import_snapshot.is_some();
+            let http_addr = opt.http_addr != "127.0.0.1:7700";
```
Kerollmops, line 218: We should maybe also check that it is different from `localhost:7700` as this is equivalent even if it is not the default value.

meilisearch-http/src/analytics/segment_analytics.rs
```diff
@@ -210,10 +210,20 @@ impl Segment {
...
-        let infos = json!({
-            "env": opt.env.clone(),
-            "has_snapshot": opt.schedule_snapshot,
-        });
+        let infos = {
+            let db_path = opt.db_path != PathBuf::from("./data.ms");
+            let import_dump = opt.import_dump.is_some();
+            let dumps_dir = opt.dumps_dir != PathBuf::from("dumps/");
+            let import_snapshot = opt.import_snapshot.is_some();
+            let http_addr = opt.http_addr != "127.0.0.1:7700";
```
Kerollmops, line 218: Can you add a comment explaining what you are really doing here, please?

meilisearch-http/src/analytics/segment_analytics.rs
```diff
@@ -210,10 +210,20 @@ impl Segment {
                     "server_provider": std::env::var("MEILI_SERVER_PROVIDER").ok(),
             })
         });
-        let infos = json!({
-            "env": opt.env.clone(),
-            "has_snapshot": opt.schedule_snapshot,
-        });
+        let infos = {
```
Kerollmops, line 213: Also, what I don't like here is that we serialize all of the Argument struct and pick the field that we don't want to for privacy reasons instead of doing it the other, safer, way: picking the struct fields that we care about and are safe. Doing it the other way can work with privacy if we forgot about modifying this block of code: we would lose information instead of getting too much from the user i.e. privacy argument.

Or maybe I am wrong and the #[serde(skip)] thing is enough.

Kerollmops (changes requested):

meilisearch-http/src/analytics/segment_analytics.rs
Kerollmops: A great example is the `snapshot_dir` field that leaks information about the user filesystem that we don't want to know about, and we forgot to serde skip it and to change the value to a boolean.

Kerollmops (changes requested):

meilisearch-http/src/analytics/segment_analytics.rs
irevoire: I can but I think the idea was just to see if the field has been used, not really how people use it :thinking: 
What do you think @gmourier?

meilisearch-http/src/analytics/segment_analytics.rs
irevoire: Yes, basically we need to choose if we want to maintain a big chunk of code importing every field one by one or if we want to import everything by default and *try to remember* to erase the thing containing user data.
I thought doing it like that was what @MarinPostma proposed Monday but I can change it.

meilisearch-http/src/analytics/segment_analytics.rs
irevoire: Done

meilisearch-http/src/analytics/segment_analytics.rs
MarinPostma: I am for skipping field as @Kerollmops said

meilisearch-http/src/analytics/segment_analytics.rs
gmourier: Absolutely, it's all about whether this option is used.

meilisearch-http/src/analytics/segment_analytics.rs
```diff
@@ -259,6 +279,7 @@ impl Segment {
     }
 
     async fn tick(&mut self, meilisearch: MeiliSearch) {
+        println!("sending analytics");
```
Kerollmops, line 282: Can you remove this?
```suggestion
```

meilisearch-http/src/analytics/segment_analytics.rs
```diff
@@ -269,7 +290,7 @@ impl Segment {
                         },
                     })),
                     user: self.user.clone(),
-                    traits: Self::compute_traits(&self.opt, stats),
+                    traits: dbg!(Self::compute_traits(&self.opt, stats)),
```
Kerollmops, line 293: Can you remove this `dbg!` too?
```suggestion
                    traits: Self::compute_traits(&self.opt, stats),
```

Kerollmops (dismissed):

meilisearch-http/src/analytics/segment_analytics.rs
irevoire: Yes sorry I left this PR a bit unfinished while waiting for the cli options to be merged :grin:

MarinPostma (dismissed):

MarinPostma:
@irevoire  this pr needs a rebase

MarinPostma:
bors merge

MarinPostma (approved):
bors merge

PR #2120, opened 2022-01-26
Bring `stable` into `main`

curquiza:
I forgot to do it, tell me @Kerollmops or @irevoire if it's useful or not. I would say yes, otherwise I will have conflict when I will try to bring `main` into `stable` for the next release. Maybe I'm wrong


curquiza:
Oh the conflict was easy to fix. Maybe it's not useful then. Tell me.

MarinPostma (approved):
we need to automate that process.

curquiza:
bors merge

PR #2130, opened 2022-02-01
feat(formatting): Implement a smarter crop

### LINKED ISSUE #1912: Phrase search does not return the correct highlight
**Describe the bug**
When performing a phrase search, "the seeker seeks the one" as one example on my data. The phrase search will return the first highlight with the word "the" in it... instead of returning the highlight with the entire phrase.

Screenshot 2021-11-18 at 12 59 29

If you take a look at the text itself, you can see the phrase "the seeker seeks the one" exists in the third paragraph but the highlight in the search hits returns the first sentence.

Screenshot 2021-11-18 at 13 01 28

**Expected behavior**
I would expect the phrase search to only return results which includes the exact phrase and only highlight the exact phrase.

**MeiliSearch version:** [e.g. v0.20.0]
```json
{
    "commitSha":"2e2eeb0a424408e27f4223e58afec303c53e1419",
    "commitDate":"2021-10-12T14:47:06+00:00",
    "pkgVersion":"0.23.1"
}
```

**Additional context**
I understand stop words might be used to exclude "the" and "of" from the search, which might help the result... but in our case we want to search dialogue exactly for "the seeker seeks the one" for example, and not "seeker of the one seeks" or another combination of the phrase, for example.

ManyTheFish:
This implementation will now crop around the highest match density instead of cropping around the first match.

This can be enhanced in a future PR by modifying milli:
- taking into account match quality (is it an ngram or a split)
- taking into account match diversity (matching several times same word)

Fix #1912

### DIFF
--- a/meilisearch-lib/src/index/search.rs
+++ b/meilisearch-lib/src/index/search.rs
@@ -575,45 +575,78 @@ impl<'a, A: AsRef<[u8]>> Formatter<'a, A> {
 
         let tokens: Box<dyn Iterator<Item = (&str, Token)>> = match format_options.crop {
             Some(crop_len) => {
-                let mut buffer = Vec::new();
-                let mut tokens = analyzed.reconstruct().peekable();
+                let crop_len = crop_len * 2;
+                let mut tokens = Vec::new();
+                let mut matches = Vec::new();
+                for (pos, (word, token)) in analyzed.reconstruct().enumerate() {
+                    if matcher.matches(token.text()).is_some() {
+                        matches.push(pos);
+                    }
 
-                while let Some((word, token)) =
-                    tokens.next_if(|(_, token)| matcher.matches(token.text()).is_none())
-                {
-                    buffer.push((word, token));
+                    tokens.push((word, token));
                 }
 
-                match tokens.next() {
-                    Some(token) => {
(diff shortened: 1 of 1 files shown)


curquiza:
What does it mean exactly "This implementation will now crop around the highest match density instead of cropping around the first match."? :)
For the docs-team, since they will have to change the [docs here](https://docs.meilisearch.com/reference/features/search_parameters.html#attributes-to-crop) if I'm correct.

ManyTheFish:
> What does it mean exactly "This implementation will now crop around the highest match density instead of cropping around the first match."? :) For the docs-team, since they will have to change the [docs here](https://docs.meilisearch.com/reference/features/search_parameters.html#attributes-to-crop) if I'm correct.

If you speak about this sentence in the doc:
> Cropping starts at the first occurrence of the search query. It only keeps cropLength characters on each side of the first match, rounded to match word boundaries.

kind of yes, the current behavior is stopping at the first occurrence of any match, for example, the query `a fun football game` has a good chance to be cropped around an isolated `a` (because it's the first one in the attribute), and that was the issue. 🙂

ManyTheFish:
I close this PR, this should be specified as a feature, and probably, the main part of the implementation should be done in milli sides to benefits from more match information.

Issue #1909, opened 2021-11-17
[ARMv8] About removing the old ARMv8 workflow and naming for the binary


patrickdung:
So, [the new GH workflow](https://github.com/meilisearch/MeiliSearch/blob/main/.github/workflows/publish-crossbuild.yml) could build the aarch64 linux binary in shorter time for the new 0.24 release.
Currently the binary is meilisearch-linux-aarch64.
[Here's Original PR](https://github.com/meilisearch/MeiliSearch/pull/1811)

Here's some discussion:
1) Removing [the old workflow](https://github.com/meilisearch/MeiliSearch/blob/main/.github/workflows/publish-binaries.yml#L40-L64) for aarch64 linux binary (meilisearch-linux-armv8)

2) Should we update the new workflow to generate binary name as meilisearch-linux-armv8?
If not we would need to update the script 'download-latest.sh' for the correct binary name.

Thanks.

curquiza:
Hello @patrickdung! Thanks for this issue.

- [ ] I would have a definitive confirmation from our raspbian users: is the `meilisearch-linux-aarch64` works on their raspberry for example? @kappa-wingman if you are around 😇 
- [ ] If the first point is validated, we also need to update the `download-latest.sh` script

I'll update your issue description :)

kappa-wingman:
@curquiza

I use MeiliSearch with ARM processor with a cloud computing provider, it runs RHEL 8. The v0.24.0 binary works for my case. So you need to find someone using the Raspberry Pi 🍓 to confirm.

curquiza:
Thanks for your answer @kappa-wingman and sorry for the miss-understanding

However, does the [`meilisearch-linux-aarch64`](https://github.com/meilisearch/MeiliSearch/releases/download/v0.24.0/meilisearch-linux-aarch64) work on this machine? Not the `meilisearch-linux-armv8` binary which is the one you download when using the [`download-latest.sh script`](https://github.com/meilisearch/MeiliSearch/blob/main/download-latest.sh) via this command

```bash
curl -L https://install.meilisearch.com | sh
```

Thanks a lot for your reactivity.

kappa-wingman:
@curquiza , I am using the meilisearch-linux-aarch64 (0.24.0 from the release page)and works on the ARM machine.
The first testing about the new binary is in [last month] (https://github.com/meilisearch/MeiliSearch/discussions/1790).

curquiza:
[...]

PR #2136, opened 2022-02-02
Refactoring CI regarding ARM binary publish

curquiza:
Fixes https://github.com/meilisearch/meilisearch/issues/1909

- Remove CI file to publish aarch64 binary and put the logic into `publish-binary.yml`
- Remove the job to publish armv8 binary
- Fix download-latest script accordingly
- Adapt dowload-latest with the specific case of the MacOS m1


.github/workflows/publish-binaries.yml
```diff
@@ -38,28 +38,69 @@ jobs:
...
+        if: matrix.use-cross
+        run: |
+          sudo apt-get install -y ${{ matrix.linker }}
+
+      - name: Configure target aarch64 GNU
+        if: matrix.target == 'aarch64-unknown-linux-gnu'
+        ## Environment variable is not passed using env:
+        ## LD gold won't work with MUSL
+        # env:
+        #   JEMALLOC_SYS_WITH_LG_PAGE: 16
```
irevoire, line 82: Removing that line for the ARM build won’t be a problem?

Kerollmops (changes requested):
Whoops, it doesn't work on my Macbook Pro M1.
Capture d’écran 2022-02-07 à 13 23 20

download-latest.sh
```diff
@@ -148,11 +148,18 @@ get_os() {
 get_archi() {
     architecture=$(uname -m)
     case "$architecture" in
-    'x86_64' | 'amd64' | 'arm64')
+    'x86_64' | 'amd64' )
         archi='amd64'
         ;;
+    'arm64')
+        if [ $os -eq 'macos' ]; then # MacOS M1
```
Kerollmops, line 155:
```suggestion
        if [ $os = 'macos' ]; then # MacOS M1
```

I just checked and it seems like `-eq` is here to compare numbers (?) and `=` is used to compare strings.

Kerollmops (changes requested):
I am not sure but I would like that first command that returns something different than zero should make the script abort or at least we should check that curl doesn't return something different than 0. It would help in not having `Not Found` in the final `meilisearch` file as it can be dangerous, we `chmod u+x` it and a simple bash script can be returned by a middleware or something... We should abort before `chmod`ing the file.

.github/workflows/publish-binaries.yml
curquiza: Which lines? These lines were already commented in the other file

Kerollmops (changes requested):
Ho and BTW the download binary is not the aarch64 one when I use the script from my MacBook Pro M1, it is the `x86_64` one and uses rosetta to run the binary.
```
$ file meilisearch
meilisearch: Mach-O 64-bit executable x86_64
```

curquiza:
> Ho and BTW the download binary is not the aarch64 one when I use the script from my MacBook Pro M1, it is the x86_64 one and uses rosetta to run the binary.

This is expected, when using a M1, the downloaded binary should be x86_64. Maybe I'm wrong, but @tamo confirmed the aarch64 binary did not work on M1. Because, if aarch64 works on M1, I can make the script simpler.

Kerollmops:
Indeed, aarch64 can't work on macOS as it is an ELF binary file, not a Mach-O one. If running the `x86_64` binary on the M1 is expected then the only thing that I don't like is that we do not [stop the script on the first encountered error](https://stackoverflow.com/a/3474556).

Kerollmops (approved):
Thank you. I'm validating this!

.github/workflows/publish-binaries.yml
```diff
@@ -38,28 +38,69 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Installing Rust toolchain
+        uses: actions-rs/toolchain@v1
         with:
-          arch: aarch64 # aka ARMv8
-          distro: ubuntu18.04
-          env: |
-            JEMALLOC_SYS_WITH_LG_PAGE: 16
```
irevoire, line 52: @curquiza 
> Which lines? These lines were already commented in the other file

It was not commented here, for the arm binary

irevoire: Ah yes sorry I missed it my bad!

irevoire (approved):

curquiza:
bors merge

PR #2157, opened 2022-02-09
fix(auth): fix env being closed when dumping

MarinPostma:
When creating a dump, the auth store environment would be closed on drop, so subsequent dumps couldn't reopen the environment. I have added a flag in the environment to prevent the closing of the environment on drop when dumping.


curquiza:
bors try

Kerollmops (approved):
Looks good to me! Is this fixing an issue we had with the CI?
bors merge

Issue #1673, opened 2021-09-06
Docker image should have curl binary


curquiza:
### Discussed in https://github.com/meilisearch/MeiliSearch/discussions/1672

<div type='discussions-op-text'>

<sup>Originally posted by **ook** September  6, 2021</sup>
Thank you for maintaining a Docker image.
For day to day investigations, this image should keep curl: binary is only around 400kb and will ease querying the container for status from the inside, removing eventual networking issue.
Shall I open a PR for it?</div>

PR #2177, opened 2022-02-17
chore(docker): remove curl as a dependency

irevoire:
I noticed we were installing curl in the final image while meilisearch don't actually use it.
This does not make a big size difference of the final image, 64.7MB before versus 64.6MB after removing it (-0.1%) but we have no reason to keep it.


curquiza:
hum I have to check but I think a contributor asked to add it

curquiza:
Found it!
https://github.com/meilisearch/meilisearch/issues/1673

This was added in v0.23.0

irevoire:
Ah ok I was wondering why it was here but yes since it's so small we can keep it 👍

Issue #2085, opened 2022-01-18
Error while persisting update to disk: No such file or directory


hahmad748:
**Getting INTERNAL error while accessing MeiliSearch**
{
    "message": "An internal error has occurred. `Error while persisting update to disk: No such file or directory (os error 2)`.",
    "code": "internal",
    "type": "internal",
    "link": "https://docs.meilisearch.com/errors#internal"
}

**To Reproduce**
1. Go to POSTMAN
2. Add Authorization Key
3. Add IP and Port
4. Hit any API, Like See stats, or get indexes list
5. Got above mentioned error in every case

**Expected behavior**
Should display the desired result, stats or indexes

**MeiliSearch version:**
{
    "commitSha": "0a1d2ce2318c26f36ddcb82d3159b8003120b84f",
    "commitDate": "2021-11-15T18:02:07+00:00",
    "pkgVersion": "0.24.0"
}

curquiza:
Hello @hahmad748 

Sorry for the inconvenience :(

Can you share the logs?
Can you say which actions (adding documents, adding settings) you did before having this error?

Also, if it's possible, can you try using [MeiliSearch v0.25.2](https://github.com/meilisearch/MeiliSearch/releases/tag/v0.25.2), which is the latest one, and report us if you have the same error?

Thanks in advance!

hahmad748:
Hello @curquiza 

- I can easily get the indexes that are listed before the mentioned error. Now am getting error for almost every request for settings, like fetching global stats, getting list of indexes, fetching filterable attributes etc.

I'll update the MeiliSearch version to 0.25.2 and verify if the issue persist or not

curquiza:
Thank you @hahmad748, let us know!

geoioanid:
_Meilisearch Version_: v0.25.2
_Dockerized_: Yes

I'm facing a similar issue, when trying to create a dump. 
The dump creation process runs for some minutes, and then fails with the following error:
```
{
    "uid": "20220130-072536070",
    "status": "failed",
    "error": "An internal error has occurred. `Internal error: Error while persisting update to disk: No such file or directory (os error 2)`.",
    "startedAt": "2022-01-30T07:25:36.070952400Z",
    "finishedAt": "2022-01-30T07:45:21.146419389Z"
}
```
This error doesn't create any side effects for me as @hahmad748 has mentioned (like fetching global stats, getting list of indexes, fetching filterable attributes etc.) Refer to the following quote
> Hello @curquiza
[...]

PR #2178, opened 2022-02-17
Refacto docker

### LINKED ISSUE #2166: Clean persistent volume on Kubernetes cases an error
**Describe the bug**
Running the latest version of Meilisearch (0.25) on Kubernetes with a clean PVC mounted on `/data.ms` throws the error: _VERSION file not found or an older version was running. Use dump...._

However, the volume is new and empty. The only thing in the folder is `lost+found` folder added by default in Ubuntu. Removing `lost+found` folder solves the problem, but requires either a manual start of the container pod (to shell and remove the folder for the first time) or use of an init script to delete it.

**To Reproduce**
1. Deploy Meilisearch to a k8s, using a PVC mounted on `/data.ms`.

**Expected behavior**
Search daemon to start without errors.

**Meilisearch version:**
v0.25

irevoire:
closes #2166 and #2085

-----------

I noticed many people had issues with the default configuration of our Dockerfile.
Some examples:
- #2166: If you use ubuntu and mount your `data.ms` in a volume (as shown in the [doc](https://docs.meilisearch.com/learn/getting_started/installation.html#download-and-launch)), you can't run meilisearch
- #2085: Here, meilisearch was not able to erase the `data.ms` when loading a dump because it's the mount point
Currently, we don't show how to use the snapshot and dumps with docker in the documentation. And it's quite hard to do:
  - You either send a big command to meilisearch to change the dump-path, snapshot-path and db-path a single directory and then mount that one
  - Or you mount three volumes
- And there were other issues on the slack community

I think this PR solve the problem.
Now the image contains the `meilisearch` binary in the `/bin` directory, so it's easy to find and always in the `PATH`.
It creates a `data` directory and moves the working-dir in it.
So now you can find the `dumps`, `snapshots` and `data.ms` directory in `/data`.

Here is the new command to run meilisearch with a volume:
```
docker run -it --rm -v $PWD/meili_data:/data -p 7700:7700 getmeili/meilisearch:latest
```

And if you need to import a dump or a snapshot, you don't need to restart your container and mount another volume. You can directly hit the `POST /dumps` route and then run:
```
docker run -it --rm -v $PWD/meili_data:/data -p 7700:7700 getmeili/meilisearch:latest meilisearch --import-dump dumps/20220217-152115159.dump
```

-------

You can already try this PR with the following docker image:
```
getmeili/meilisearch:test-docker-v0.26.0
```

If you want to use the v0.25.2 I created another image;
```
getmeili/meilisearch:test-docker-v0.25.2
```

------

If you're using helm I created a branch [here](https://github.com/meilisearch/meilisearch-kubernetes/tree/test-docker-v0.26.0) that use the v0.26.0 image with the good volume :+1: 
If you use this conf with the v0.25.2, it should also work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,9 +27,20 @@ ENV     MEILI_SERVER_PROVIDER docker
 RUN     apk update --quiet \
         && apk add -q --no-cache libgcc tini curl
 
-COPY    --from=compiler /meilisearch/target/release/meilisearch .
+# add meilisearch to the `/bin` so you can run it from anywhere and it's easy
+# to find.
+COPY    --from=compiler /meilisearch/target/release/meilisearch /bin/meilisearch
+# To stay compatible with the older version of the container (pre v0.27.0) we're
+# going to symlink the meilisearch binary in the path to `/meilisearch`
+RUN     ln -s /bin/meilisearch /meilisearch
+
+# This directory should hold all the data related to meilisearch so we're going
+# to move our PWD in there.
+# We don't want to put the meilisearch binary
+WORKDIR /meili_data
+
 
 EXPOSE  7700/tcp
 
 ENTRYPOINT ["tini", "--"]
-CMD     ./meilisearch
+CMD     /bin/meilisearch


fr3fou:
Can I use this Dockerfile with 0.25.2 safely? I'm not sure what changes are in the 0.26.0 release of MeiliSearch but I need to use this and so far it seems to have solved https://github.com/meilisearch/meilisearch-kubernetes/issues/95 for (the Pod is running but I'm not sure of any internal changes that might have ocurred)

irevoire:
Yes but you'll need to rebuild the image yourself.
Something like 
```bash
git checkout v0.25.2
git checkout refacto-docker -- Dockerfile
docker build . # name your image and everything 
```

fr3fou:
Yup, that's what I did exactly!

Dockerfile
```diff
@@ -45,11 +45,21 @@ ENV     MEILI_HTTP_ADDR 0.0.0.0:7700
...
-        && apk add -q --no-cache libgcc tini curl
+        && apk add -q --no-cache libgcc tini
 
-COPY    --from=compiler /meilisearch/target/release/meilisearch .
+# add meilisearch to the `/bin` so you can run it from anywhere and it's easy
+# to find.
+COPY    --from=compiler /meilisearch/target/release/meilisearch /bin/meilisearch
+
+# We want user to mount this directory as a volume
+RUN     mkdir /data
```
ChillFish8, line 55: So I'm not really sure what's going on here, we make a directory first but then also tag it as our working directory.
If we're marking this as our WD then we don't need to make the directory first, it's just an implicit step iirc.

irevoire:
> Can I use this Dockerfile with 0.25.2 safely? I'm not sure what changes are in the 0.26.0 release of MeiliSearch but I need to use this and so far it seems to have solved [meilisearch/meilisearch-kubernetes#95](https://github.com/meilisearch/meilisearch-kubernetes/issues/95) for (the Pod is running but I'm not sure of any internal changes that might have ocurred)

Hey, I created a new image for you and everyone who wants to test this PR with the v0.25.2.
You can find it here; `getmeili/meilisearch:test-docker-v0.25.2`

Dockerfile
irevoire: Oh cool, I didn't know that!
Thanks :+1:

irevoire:
bors merge

curquiza (approved):

PR #2204, opened 2022-03-02
Fix blocking auth

### LINKED ISSUE #2201: Blocking calls in authentication extractor
Reviewing the auth implementation, I found that blocking calls to heed were performed in the implementation of the extractor.

- [blocking calls](https://github.com/meilisearch/meilisearch/blob/main/meilisearch-auth/src/lib.rs#L54-L58) are made in async contexts.
- [`FromRequest`](https://github.com/meilisearch/meilisearch/blob/main/meilisearch-http/src/extractors/authentication/mod.rs#L37) should not return a `Ready` future, but a custom future, encapsulating the authentication action. a boxed future should do the trick, but it should be possible to do it more elegantly.

MarinPostma:
Fix auth blocking runtime

I have decided to remove async code from `meilisearch-auth` and let `meilisearch-http` handle that.

Because Actix polls the extractor futures concurrently, I have made a wrapper extractor that forces the errors from the futures to be returned sequentially (though is still polls them sequentially).
close #2201

### DIFF
--- a/meilisearch-http/src/extractors/authentication/mod.rs
+++ b/meilisearch-http/src/extractors/authentication/mod.rs
@@ -41,10 +41,7 @@ impl<P, D> GuardedData<P, D> {
                 }),
                 None => Err(AuthenticationError::IrretrievableState.into()),
             },
-            (token, None) => {
-                let token = token.to_string();
-                Err(AuthenticationError::InvalidToken(token).into())
-            }
+            (token, None) => Err(AuthenticationError::InvalidToken(token).into()),
         }
     }
 
--- a/meilisearch-http/src/extractors/authentication/mod.rs
+++ b/meilisearch-http/src/extractors/authentication/mod.rs
@@ -33,7 +33,7 @@ impl<P, D> GuardedData<P, D> {
         P: Policy + 'static,
     {
         match Self::authenticate(auth, token, index).await? {
-            (_, Some(filters)) => match data {
+            Some(filters) => match data {
                 Some(data) => Ok(Self {
                     data,
                     filters,
@@ -41,7 +41,7 @@ impl<P, D> GuardedData<P, D> {
                 }),
                 None => Err(AuthenticationError::IrretrievableState.into()),
             },
-            (token, None) => Err(AuthenticationError::InvalidToken(token).into()),
+            None => Err(AuthenticationError::InvalidToken.into()),
         }
     }
 
@@ -49,7 +49,7 @@ impl<P, D> GuardedData<P, D> {
     where
         P: Policy + 'static,
     {
-        match Self::authenticate(auth, "", None).await?.1 {
+        match Self::authenticate(auth, String::new(), None).await? {
             Some(filters) => match data {
                 Some(data) => Ok(Self {
                     data,
@@ -62,18 +62,16 @@ impl<P, D> GuardedData<P, D> {
         }
     }
 
-    async fn authenticate<S>(
+    async fn authenticate(
         auth: AuthController,
-        token: S,
+        token: String,
         index: Option<String>,
-    ) -> Result<(S, Option<AuthFilter>), ResponseError>
+    ) -> Result<Option<AuthFilter>, ResponseError>
     where
         P: Policy + 'static,
-        S: AsRef<str> + 'static + Send,
     {
         Ok(tokio::task::spawn_blocking(move || {
-            let res = P::authenticate(auth, token.as_ref(), index.as_deref());
-            (token, res)
+            P::authenticate(auth, token.as_ref(), index.as_deref())
         })
         .await
         .map_err(|e| ResponseError::from_msg(e.to_string(), Code::Internal))?)
@@ -114,10 +112,7 @@ impl<P: Policy + 'static, D: 'static + Clone> FromRequest for GuardedData<P, D> 
                                 index.map(String::from),
                                 req.app_data::<D>().cloned(),
                             )),
-                            None => Box::pin(err(AuthenticationError::InvalidToken(
-                                "unknown".to_string(),
-                            )
-                            .into())),
+                            None => Box::pin(err(AuthenticationError::InvalidToken.into())),
                         }
                     }
                     _otherwise => {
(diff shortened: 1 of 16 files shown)


meilisearch-http/src/extractors/authentication/mod.rs
```diff
@@ -2,28 +2,85 @@ mod error;
...
+            },
+            None => Err(AuthenticationError::MissingAuthorizationHeader.into()),
+        }
+    }
+
+    async fn authenticate<S>(
+        auth: AuthController,
+        token: S,
+        index: Option<String>,
+    ) -> Result<(S, Option<AuthFilter>), ResponseError>
```
ManyTheFish, line 69: We may return only an `Option<AuthFilter>`, this would simplify the upper functions.

ManyTheFish (changes requested):
We may not release this in the v0.26, we may instead merge it in `main`

meilisearch-http/src/extractors/sequential_extractor.rs
```diff
@@ -0,0 +1,148 @@
...
+        },
+        Done {
+            output: Res,
+        },
+        Error {
+            error: Err,
+        },
+        Empty,
+    }
+}
```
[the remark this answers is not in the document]
MarinPostma: it's not, i took inspiration from actix implementation, bu this is my code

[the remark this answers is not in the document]
ManyTheFish: Why do you need `#![allow(non_snake_case)]` at the top of the file?

irevoire (commented):
The test of the rust-sdk all pass :+1:

meilisearch-http/src/extractors/sequential_extractor.rs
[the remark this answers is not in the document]
MarinPostma: this is because the fields of the generated `SeqFromRequestFutX` are named after the generic type of the field (`A`, `B`,...), and rust complains because they are not snake case

meilisearch-http/src/routes/api_key.rs
```diff
@@ -52,8 +65,13 @@ pub async fn get_api_key(
...
     path: web::Path<AuthParam>,
 ) -> Result<HttpResponse, ResponseError> {
-    let key = auth_controller.get_key(&path.api_key).await?;
-    let res = KeyView::from_key(key, &auth_controller);
+    let api_key = path.into_inner().api_key;
+    let res = tokio::task::spawn_blocking(move || -> Result<_, AuthControllerError> {
+        let key = auth_controller.get_key(&api_key)?;
+        Ok(KeyView::from_key(key, &auth_controller))
+    })
+    .await
```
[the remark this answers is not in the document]
MarinPostma: async auth does not actually do anything related to async, so it made more sense to move te boundary over to meilisearch-http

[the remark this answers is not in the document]
ManyTheFish: OOOOOKKK... My bad, I misunderstood the behavior of `async`, thanks!

meilisearch-http/src/extractors/authentication/mod.rs
ManyTheFish: no sorry, I was saying:
```suggestion
    ) -> Result<Option<AuthFilter>, ResponseError>
```

Is it possible ?

MarinPostma: no, we need S to be static here, so it could force a clone. This is the only way to report errors without having to clone each time on line 44.

curquiza (commented):
Works with the Rust, PHP, Python, Ruby and JS SDK tests 💪 🎉

meilisearch-http/src/extractors/authentication/mod.rs
Kerollmops: On the lines 113, 114 and 115 you clone a lot, maybe you can simplify this method and also clone this token?

meilisearch-http/src/extractors/sequential_extractor.rs
```diff
@@ -0,0 +1,148 @@
...
+pub struct SeqHandler<H>(pub H);
+
+pub struct SeqFromRequest<T>(T);
+
+/// This macro implements `FromRequest` for arbitrary arity handler, except for one, which is
+/// useless anyway.
+macro_rules! gen_seq {
+    ($ty:ident; $($T:ident)+) => {
+        pin_project! {
+            pub struct $ty<$($T: FromRequest), +> {
```
Kerollmops, line 21: Maybe you can put this allow non_snake_case on the struct instead?

meilisearch-http/src/extractors/sequential_extractor.rs
```diff
@@ -0,0 +1,148 @@
+#![allow(non_snake_case)]
+use std::{future::Future, pin::Pin, task::Poll};
+
+use actix_web::{dev::Payload, FromRequest, Handler, HttpRequest};
+use pin_project_lite::pin_project;
+
+/// `SeqHandler` is an actix `Handler` that enforces that extractors errors are returned in the
+/// same order as they are defined in the wrapped handler. This is needed because, by default, actix
+/// to resolves the extractors concurrently, whereas we always need the authentication extractor to
```
Kerollmops, line 9:
```suggestion
/// resolves the extractors concurrently, whereas we always need the authentication extractor to
```

Kerollmops (changes requested):

meilisearch-http/src/extractors/sequential_extractor.rs
MarinPostma: not possible, you also get the lint when referring to those fields, it needs to be a module level lint here

meilisearch-http/src/extractors/authentication/mod.rs
MarinPostma: The issue is not about cloning when it's needed, as it is on the lines you mention, but cloning when it is not, for error handling....

But ultimately, passing the token seems useless, I think it's better to just stop passing I it to the `InvalidToken`

Kerollmops (approved):
Looks good to me!

ManyTheFish (approved):

PR #2207, opened 2022-03-04
Fix: avoid embedding the user input into the error response.

### LINKED ISSUE #2107: Error message contains whole payload when the type is not correct
**Describe the bug**
I tried to send (wrongly) a string using `POST /indexes/x/documents`. The returned error message from MeiliSearch is as follows: 
`MeiliSearchApiError: The json payload provided is malformed. "whole payload"`.

In my case the `whole payload` is 3million character long. So my error message looks like this:

Screenshot 2022-01-24 at 13 39 54

(ignore the popup from vscode)

**To Reproduce**
1. Add a string instead of a document with content-type json
2. It throws an error containing the whole string

**Expected behavior**
Maybe a more explicit error `The json payload provided is malformed, expected an array but go a string instead`

**Screenshots**
see above

**MeiliSearch version:** v0.25.1

CNLHC:
# Pull Request

## What does this PR do?
Fix #2107. 

The problem is meilisearch embeds the user input to the error message. 

The reason for this problem is `milli` throws a `serde_json: Error` whose `Display` implementation will do this embedding.  

I tried to solve this problem in this PR by manually implementing the `Display` trait for `DocumentFormatError` instead of deriving automatically.

## PR checklist

### DIFF
--- a/meilisearch-http/tests/documents/add_documents.rs
+++ b/meilisearch-http/tests/documents/add_documents.rs
@@ -312,16 +312,7 @@ async fn error_add_malformed_json_documents() {
     // truncate
 
     // length = 100
-    let long = String::from_utf8(
-        "0123456789"
-            .as_bytes()
-            .iter()
-            .cycle()
-            .cloned()
-            .take(100)
-            .collect_vec(),
-    )
-    .unwrap();
+    let long = "0123456789".repeat(10);
 
     let document = format!("\"{}\"", long);
     let req = test::TestRequest::put()
--- a/meilisearch-http/tests/documents/add_documents.rs
+++ b/meilisearch-http/tests/documents/add_documents.rs
@@ -321,9 +321,7 @@ async fn error_add_malformed_json_documents() {
         .to_request();
     let res = test::call_service(&app, req).await;
     let body = test::read_body(res).await;
-    dbg!(&body);
     let response: Value = serde_json::from_slice(&body).unwrap_or_default();
-    dbg!(&response);
     assert_eq!(status_code, 400);
     assert_eq!(
         response["message"],
--- a/meilisearch-http/tests/documents/add_documents.rs
+++ b/meilisearch-http/tests/documents/add_documents.rs
@@ -1,6 +1,5 @@
 use crate::common::{GetAllDocumentsOptions, Server};
 use actix_web::test;
-use itertools::Itertools;
 use meilisearch_http::{analytics, create_app};
 use serde_json::{json, Value};
 use time::{format_description::well_known::Rfc3339, OffsetDateTime};
@@ -329,7 +328,7 @@ async fn error_add_malformed_json_documents() {
     assert_eq!(
         response["message"],
         json!(
-            r#"The `json` payload provided is malformed. `Couldn't serialize document value: invalid type: string "0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789", expected a documents, or a sequence of documents. at line 1 column 102`."#
+            r#"The `json` payload provided is malformed. `Couldn't serialize document value: invalid type: string "0123456789012345678901234567...890123456789", expected a documents, or a sequence of documents. at line 1 column 102`."#
         )
     );
     assert_eq!(response["code"], json!("malformed_payload"));
@@ -355,7 +354,7 @@ async fn error_add_malformed_json_documents() {
     assert_eq!(
         response["message"],
         json!(
-            r#"The `json` payload provided is malformed. `Couldn't serialize document value: invalid type: string "01234567890123456789012345678901234567890123456789 ... 1234567890123456789012345678901234567890123456789m", expected a documents, or a sequence of documents. at line 1 column 103`."#
+            r#"The `json` payload provided is malformed. `Couldn't serialize document value: invalid type: string "0123456789012345678901234567...90123456789m", expected a documents, or a sequence of documents. at line 1 column 103`."#
         )
     );
     assert_eq!(response["code"], json!("malformed_payload"));
--- a/meilisearch-http/tests/documents/add_documents.rs
+++ b/meilisearch-http/tests/documents/add_documents.rs
@@ -345,9 +345,7 @@ async fn error_add_malformed_json_documents() {
         .to_request();
     let res = test::call_service(&app, req).await;
     let body = test::read_body(res).await;
-    dbg!(&body);
     let response: Value = serde_json::from_slice(&body).unwrap_or_default();
-    dbg!(&response);
     assert_eq!(status_code, 400);
     assert_eq!(
         response["message"],
--- a/meilisearch-http/tests/documents/add_documents.rs
+++ b/meilisearch-http/tests/documents/add_documents.rs
@@ -348,8 +348,7 @@ async fn error_add_malformed_json_documents() {
         json!("https://docs.meilisearch.com/errors#malformed_payload")
     );
 
-
-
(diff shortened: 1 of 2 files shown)


meilisearch-lib/src/document_formats.rs
```diff
@@ -23,16 +24,27 @@ impl fmt::Display for PayloadType {
...
+        match self {
+            Self::Internal(e) => write!(f, "An internal error has occurred. `{}`.", e),
+            Self::MalformedPayload(me, b) => match me.borrow() {
+                milli::documents::Error::JsonError(se) => write!(
+                    f,
+                    "The `{}` payload provided is malformed. `Couldn't serialize document value. at line {} column {}`",
+                    b,se.line(),se.column()
+                ),
+                _ => write!(f, "The `{}` payload provided is malformed. `{}`.", b, me),
+            },
```
Kerollmops, line 43:
```suggestion
            Self::Internal(e) => write!(f, "An internal error has occurred: `{}`.", e),
            Self::MalformedPayload(me, b) => match me.borrow() {
                milli::documents::Error::JsonError(se) => write!(
                    f,
                    "The `{}` payload provided is malformed. `Couldn't serialize document value at line {} column {}`",
                    b, se.line(), se.column()
                ),
                _ => write!(f, "The `{}` payload provided is malformed: `{}`.", b, me),
            },
```

Kerollmops (changes requested):
Thank you very much for your PR!
Apart from some esthetic changes, it's good for me.

meilisearch-lib/src/document_formats.rs
```diff
@@ -23,16 +24,27 @@ impl fmt::Display for PayloadType {
...
 pub enum DocumentFormatError {
-    #[error("An internal error has occurred. `{0}`.")]
     Internal(Box<dyn std::error::Error + Send + Sync + 'static>),
-    #[error("The `{1}` payload provided is malformed. `{0}`.")]
-    MalformedPayload(
-        Box<dyn std::error::Error + Send + Sync + 'static>,
-        PayloadType,
-    ),
+    MalformedPayload(Box<milli::documents::Error>, PayloadType),
+}
```
Kerollmops, line 31:
```suggestion
}

```

meilisearch-lib/src/document_formats.rs
```diff
@@ -23,16 +24,27 @@ impl fmt::Display for PayloadType {
...
+                    f,
+                    "The `{}` payload provided is malformed. `Couldn't serialize document value at line {} column {}`",
+                    b, se.line(), se.column()
+                ),
+                _ => write!(f, "The `{}` payload provided is malformed: `{}`.", b, me),
+            },
+        }
+    }
 }
+impl std::error::Error for DocumentFormatError {}
```
Kerollmops, line 47:
```suggestion

impl std::error::Error for DocumentFormatError {}
```

Kerollmops (changes requested):
I had some thoughts about your PR and think it removes too much information about the error, users will be lost if an error occurs and only the line and column are specified.

What do you think about ellipsing the message? By using something like 50 chars at the start and 50 chars at the end:
```
The `json` payload provided is malformed. `Couldn't serialize document value: invalid type: string \"my very ver...very very long body here\", expected a documents, or a sequence of documents. at line 1 column 7`.
```

CNLHC:
Sorry for the late response. I will submit a patch to  follow the suggestion of @Kerollmops today.

curquiza:
@CNLHC thanks a lot for these changes 🙏 
Looks like there is a tiny issue to fix with our clippy CI before Kero reviews it :)

meilisearch-http/tests/documents/add_documents.rs
```diff
@@ -307,6 +308,71 @@ async fn error_add_malformed_json_documents() {
...
+    let long = String::from_utf8(
+        "0123456789"
+            .as_bytes()
+            .iter()
+            .cycle()
+            .cloned()
+            .take(100)
+            .collect_vec(),
+    )
+    .unwrap();
```
Kerollmops, line 324:
```suggestion
    let long = "0123456789".repeat(10);
```

Kerollmops (changes requested):
Thank you very much for your contribution ❤️
However, could you modify the tests according to the changes I proposed with the ellipsing algorithm? And could you stash everything into one final commit when we are done, please?

meilisearch-lib/src/document_formats.rs
```diff
@@ -23,17 +24,47 @@ impl fmt::Display for PayloadType {
...
+                        } else {
+                            serde_msg = String::from("");
+                        }
+                    }
+
+                    write!(
+                    f,
+                    "The `{}` payload provided is malformed. `Couldn't serialize document value: {}`.",
+                    b,serde_msg
+                )
```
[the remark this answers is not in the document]
CNLHC: In some cases, this method will lead to weird error messages. For example, if the raw message is 
```
The `json` payload provided is malformed. `Couldn't serialize document value: invalid type: string "0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789", expected a documents, or a sequence of documents. at line 1 column 102`.
```

Using your method, ellipsing the string without parsing the context will  generate

```
The `json` payload provided is malformed. `Couldn't serialize document value: invalid type: string \"0123456789012345678901234567..., or a sequence of documents. at line 1 column 102`.
```

Some meaningful information(The `expected a documents`)  is truncated.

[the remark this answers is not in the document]
CNLHC: The meaningful suffix length cannot be determined. But in most cases, its length is around `75` which is 
`", expected a documents, or a sequence of documents. at line <LINE_NUMBER> column <COLUMN_NUMBER>`

So I suggest to change  `50..serde_msg.len() - 50` to 50..serde_msg.len() - 85`

meilisearch-http/tests/documents/add_documents.rs
```diff
@@ -307,6 +307,62 @@ async fn error_add_malformed_json_documents() {
...
+    let req = test::TestRequest::put()
+        .uri("/indexes/dog/documents")
+        .set_payload(document)
+        .insert_header(("content-type", "application/json"))
+        .to_request();
+    let res = test::call_service(&app, req).await;
+    let body = test::read_body(res).await;
+    dbg!(&body);
+    let response: Value = serde_json::from_slice(&body).unwrap_or_default();
+    dbg!(&response);
```
Kerollmops, line 326:
```suggestion
    let response: Value = serde_json::from_slice(&body).unwrap_or_default();
```

meilisearch-http/tests/documents/add_documents.rs
```diff
@@ -307,6 +307,62 @@ async fn error_add_malformed_json_documents() {
...
```
Kerollmops, line 352:
```suggestion
    let response: Value = serde_json::from_slice(&body).unwrap_or_default();
```

Kerollmops (changes requested):
Thank you, that's better and clever. Could you remove the `dbg!` in the test and I'll merge after that, please?
Once you have done so, could you squash everything in one commit and rebase on main, please?

meilisearch-http/tests/documents/add_documents.rs
CNLHC: Oh, It's my fault. I should remove it.

Kerollmops (approved):
Thank you very much for your work @CNLHC, merging now!
bors merge

PR #2237, opened 2022-03-14
Update dependencies

Kerollmops:
This PR upgrade and updates the dependencies of meilisearch, but first I removed three unused dependencies. I used [cargo udeps](https://github.com/est31/cargo-udeps) to detect those and [cargo upgrade](https://github.com/killercup/cargo-edit/blob/master/README.md#available-subcommands) to upgrade ⬆️

~This PR **must** be merged when https://github.com/meilisearch/milli/pull/465 is merged and then must be updated accordingly i.e. using the latest version of milli.~

### DIFF
--- a/meilisearch-http/src/extractors/authentication/mod.rs
+++ b/meilisearch-http/src/extractors/authentication/mod.rs
@@ -144,6 +144,7 @@ pub mod policies {
     pub static TENANT_TOKEN_VALIDATION: Lazy<Validation> = Lazy::new(|| {
         let mut validation = Validation::default();
         validation.validate_exp = false;
+        validation.required_spec_claims.remove("exp");
         validation.algorithms = vec![Algorithm::HS256, Algorithm::HS384, Algorithm::HS512];
         validation
     });
@@ -205,9 +206,7 @@ pub mod policies {
                 return None;
             }
 
-            let mut validation = Validation::default();
-            validation.validate_exp = false;
-            validation.validate_nbf = false;
+            let mut validation = TENANT_TOKEN_VALIDATION.clone();
             validation.insecure_disable_signature_validation();
             let dummy_key = DecodingKey::from_secret(b"secret");
             let token_data = decode::<Claims>(token, &dummy_key, &validation).ok()?;
--- a/meilisearch-http/src/extractors/authentication/mod.rs
+++ b/meilisearch-http/src/extractors/authentication/mod.rs
@@ -131,7 +131,7 @@ pub trait Policy {
 }
 
 pub mod policies {
-    use jsonwebtoken::{dangerous_insecure_decode, decode, Algorithm, DecodingKey, Validation};
+    use jsonwebtoken::{decode, Algorithm, DecodingKey, Validation};
     use once_cell::sync::Lazy;
     use serde::{Deserialize, Serialize};
     use time::OffsetDateTime;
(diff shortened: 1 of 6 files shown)


MarinPostma (dismissed):

Kerollmops:
Hey @ManyTheFish,

I am not sure of how I should patch the `jsonwebtoken` library bump, they removed the `dangerous_insecure_decode` function. I followed [this PR](https://github.com/mozilla/sccache/pull/1137/files/351484fe899270c68b9497b9d4e3f661356e2f4f#diff-16425d15ca815137ec49d96fd0a07bd1e920bf54c13eb1eccb2db5c1d5094ae9) but it doesn't seem to pass the tests.

meilisearch-http/src/extractors/authentication/mod.rs
```diff
@@ -204,12 +206,17 @@ pub mod policies {
                 return None;
             }
 
+            let mut validation = TENANT_TOKEN_VALIDATION.clone();
+            validation.insecure_disable_signature_validation();
+            let dummy_key = DecodingKey::from_secret(b"secret");
```
MarinPostma, line 211: uh, you sure about that?

meilisearch-http/src/extractors/authentication/mod.rs
```diff
@@ -204,12 +206,17 @@ pub mod policies {
                 return None;
             }
 
+            let mut validation = TENANT_TOKEN_VALIDATION.clone();
+            validation.insecure_disable_signature_validation();
```
MarinPostma, line 210: are we sure about that too? The doc says that we should only do that if know what we are doing. If we really do, may be we should add a comment explaining why this is ok.

MarinPostma (changes requested):

meilisearch-http/src/extractors/authentication/mod.rs
Kerollmops: Many changed that in the latest commit of this PR [4aef7c5](https://github.com/meilisearch/meilisearch/pull/2237/commits/4aef7c5ac5a8d9856472ed1621ed74685e4c5cf9) as the tests were not passing. You can ask him if you want, maybe I made a mistake when rebasing.

meilisearch-http/src/extractors/authentication/mod.rs
Kerollmops: Same as above. @ManyTheFish can you review this PR and especially the two parts @MarinPostma quoted here, please?

meilisearch-http/src/extractors/authentication/mod.rs
ManyTheFish: DecodingKey doesn't have default implementation, we could change it to:
```suggestion
            let dummy_key = DecodingKey::from_secret(b"dummy");
```
or:
```suggestion
            let dummy_key = DecodingKey::from_secret(b"");
```

meilisearch-http/src/extractors/authentication/mod.rs
ManyTheFish: Yes, here we don't already know the API key to decode properly the token, to do it, we have to get the key prefix stored in the token.
So:
1) We decode the token to get the prefix
2) We retrieve the API key from the prefix
3) We verify the token with the retrieved API key

ManyTheFish (approved):

MarinPostma (changes requested):

meilisearch-http/src/extractors/authentication/mod.rs
```diff
@@ -141,10 +141,12 @@ pub mod policies {
...
-        validate_exp: false,
-        algorithms: vec![Algorithm::HS256, Algorithm::HS384, Algorithm::HS512],
-        ..Default::default()
+    pub static TENANT_TOKEN_VALIDATION: Lazy<Validation> = Lazy::new(|| {
+        let mut validation = Validation::default();
+        validation.validate_exp = false;
+        validation.required_spec_claims.remove("exp");
+        validation.algorithms = vec![Algorithm::HS256, Algorithm::HS384, Algorithm::HS512];
+        validation
     });
```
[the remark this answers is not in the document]
Kerollmops: > the implementation for `Clone` here comes from a `derive`

What do you mean?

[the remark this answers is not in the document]
MarinPostma: Sorry, I mean that this is not a cheap reference clone, so we are actually using a static to do the work of a function.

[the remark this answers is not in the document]
MarinPostma: cf: https://docs.rs/jsonwebtoken/latest/src/jsonwebtoken/validation.rs.html#29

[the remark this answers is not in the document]
Kerollmops: Ok, I will create a function then 👍

MarinPostma (approved):
lgtm, thank @Kerollmops  and @ManyTheFish

PR #2238, opened 2022-03-15
cargo: use resolver 2

happysalada:
# Pull Request

## What does this PR do?
use resolver 2 from cargo.
This enables mainly to propagate the `--no-default-features` flag to workspace crates.
I mistakenly thought before that it was enough to have edition 2021 enabled. However it turns out that for virtual workspaces, this needs to be explicitely defined.
https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html

This will also change a little how your dependencies are compiled. See https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html#cargos-new-feature-resolver for more details.

Just to give a bit more context, this is for usage in nixos. I have tried to do the upgrade today with the latest version, and the no default features flag is just ignored.

Let me know if you need more details of course.

## PR checklist
- [ ] Does this PR fix an existing issue?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 members = [
     "meilisearch-http",
     "meilisearch-error",


Kerollmops:
Hey @happysalada,

Thank you for your PR. I just reread the article summarizing the new behavior and it seems that it will not impact the time it takes to compile meilisearch, as it should compile multiple times the dependencies based on the features we use, so it look good to me. However, I would like to let @MarinPostma review this PR too.

If I summarize the change: You added the _resolve 2_ to the workspace this way you are able to pass the `--no-default-feature ` flag at the root of the repo.

Kerollmops (approved):

happysalada:
Thanks a lot for the review. That is exactly the idea!
To give a little more context, the nix builder builds without network access to make sure builds are reproducible. So any assets has to be passed to it beforehands (with the sha-256 checked). Removing that network access call for the dashboard during the build is main idea. In order to enable it, assets will be downloaded and checked before hand passed to it.

PR #2243, opened 2022-03-16
bug(http): fix panic on startup

### LINKED ISSUE #2242: Cargo run does not work
**Describe the bug**
When I execute `cargo run` meilisearch panic with the following message;
```
thread 'main' panicked at 'Command meilisearch-http: Argument or group 'enable-autobatching' specified in 'requires*' for 'max-batch-size' does not exist', /home/irevoire/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-3.1.2/src/build/debug_asserts.rs:128:13
```

**To Reproduce**
1. cargo run
2. wait
3. wait

**Expected behavior**
It doesn't panic.

**Screenshots**
image

**Meilisearch version:** running on main

**Additional context**
Additional information that may be relevant to the issue.
I use firefox.

MarinPostma:
this seems to fix #2242

I am not sure why this doesn't reproduce on v0.26.0, so we should remain vigilant.

### DIFF
--- a/meilisearch-lib/src/options.rs
+++ b/meilisearch-lib/src/options.rs
@@ -52,20 +52,20 @@ pub struct SchedulerConfig {
 
     // The maximum number of updates of the same type that can be batched together.
     // If unspecified, this is unlimited. A value of 0 is interpreted as 1.
-    #[clap(long, requires = "enable-autobatching", hide = true)]
+    #[clap(long, requires = "enable-auto-batching", hide = true)]
     pub max_batch_size: Option<usize>,
 
     // The maximum number of documents in a document batch. Since batches must contain at least one
     // update for the scheduler to make progress, the number of documents in a batch will be at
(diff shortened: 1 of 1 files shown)


sanders41:
This fixed the error for me.

@MarinPostma when I initially ran it worked, but didn't rebuild. I did a `cargo clean` to get it to rebuild and verify it worked after that. I have no idea if it is related, but could this be why it didn't reproduce?

MarinPostma:
@sanders41 IDK, @curquiza found that the bug was not present when building in release, and I don't like that at all.

irevoire (approved):
I think we can merge here and then cherry-pick this commit to make a v0.26.1 wdyt @curquiza ?

MarinPostma:
bors merge

Corn68 (commented):
G

PR #2249, opened 2022-03-17
feat(all): introduce disable typos

MarinPostma:
Introduce the disable typo setting, that allows disabling typos for an index.

waiting on https://github.com/meilisearch/milli/pull/469


irevoire (approved):
Looks good to me. I don't think there will be any bug with this PR, and if there is a bug with the feature, it'll be from the milli side thus, I think we can merge :grin:

irevoire:
bors merge

PR #2253, opened 2022-03-21
refactor authentication key extraction

MarinPostma:
I am concerned that the part of the code that performs the key prefix extraction from the jwt token migh be misused in the future. Since this is a critical part of the code, I moved it into it's own function. Since we deserialized the payload twice anyway, I reordered the verifications, and we now use the data from the validated token.


MarinPostma:
bors try

ManyTheFish (approved):
It's way clearer, thanks @MarinPostma!

MarinPostma:
bors merge

ManyTheFish:
bors merge

PR #2264, opened 2022-03-24
Import milli from meilisearch-lib

Kerollmops:
This PR directly imports the milli dependency used in _meilisearch-http_ from _meilisearch-lib_. I can't import _meilisearch-lib_ in _meiliserach-auth_ and that _meilisearch-auth_ can't use the milli exported by _meilisearch-lib_ 😞

### DIFF
--- a/meilisearch-http/src/helpers/env.rs
+++ b/meilisearch-http/src/helpers/env.rs
@@ -1,10 +1,11 @@
+use meilisearch_lib::milli::heed::Env;
 use walkdir::WalkDir;
 
 pub trait EnvSizer {
     fn size(&self) -> u64;
 }
 
-impl EnvSizer for milli::heed::Env {
+impl EnvSizer for Env {
     fn size(&self) -> u64 {
         WalkDir::new(self.path())
             .into_iter()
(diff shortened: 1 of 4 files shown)


meilisearch-http/src/helpers/env.rs
```diff
@@ -1,10 +1,11 @@
+use meilisearch_lib::milli::heed::Env;
```
MarinPostma, line 1: what about re-exporting `heed` directly, since this is the only thing we need from milli on `meilisearch-http` side
```suggestion
use meilisearch_lib::heed::Env;
```

MarinPostma (changes requested):
I agree that the `meilisearch-auth` situation is not ideal, I'm not sure yet what to do about it...

meilisearch-http/src/helpers/env.rs
Kerollmops: Yup! Done!

MarinPostma (approved):
👍 

bors merge

Kerollmops:
This is lindera which is not compiling due to Google Drive...

bors merge

PR #2277, opened 2022-03-29
fix(http): fix panic when sending document update without content type header

MarinPostma:
I found a panic when pushing documents without a content-type. This fixes is by returning unknown instead of crashing.


irevoire (approved):
👍

curquiza:
Is it something that would deserve v0.26.2?

MarinPostma:
bors merge

PR #2281, opened 2022-03-29
Hard limit the number of results returned by a search

### LINKED ISSUE #2133: Hard limit when searching documents
To avoid issues like this one: https://github.com/meilisearch/meilisearch/issues/1708, we need to fix an hard limit during the search.

WIP on the product side.

Kerollmops:
This PR fixes #2133 by hard-limiting the number of results that a search request can return at any time. I would like the guidance of @MarinPostma to test that, should I use a mocking test here? Or should I do anything else?

I talked about touching the _nb_hits_ value with @qdequele and we concluded that it was not correct to do so.

Could you please confirm that it is the right place to change that?

### DIFF
--- a/meilisearch-lib/src/index/search.rs
+++ b/meilisearch-lib/src/index/search.rs
@@ -1,3 +1,4 @@
+use std::cmp::min;
 use std::collections::{BTreeMap, BTreeSet, HashSet};
 use std::str::FromStr;
 use std::time::Instant;
@@ -101,8 +102,13 @@ impl Index {
             search.query(query);
         }
 
-        search.limit(query.limit);
-        search.offset(query.offset.unwrap_or_default());
+        // Make sure that a user can't get more documents than the hard limit,
+        // we align that on the offset too.
+        let offset = min(query.offset.unwrap_or(0), HARD_RESULT_LIMIT);
+        let limit = min(query.limit, HARD_RESULT_LIMIT.saturating_sub(offset));
+
+        search.offset(offset);
+        search.limit(limit);
 
         if let Some(ref filter) = query.filter {
             if let Some(facets) = parse_filter(filter)? {
@@ -127,8 +133,6 @@ impl Index {
             candidates,
             ..
         } = search.execute()?;
-
-        let documents_ids: Vec<_> = documents_ids.into_iter().take(HARD_RESULT_LIMIT).collect();
 
         let fields_ids_map = self.fields_ids_map(&rtxn).unwrap();
 
--- a/meilisearch-lib/src/index/search.rs
+++ b/meilisearch-lib/src/index/search.rs
@@ -34,6 +34,10 @@ pub const fn default_crop_length() -> usize {
     DEFAULT_CROP_LENGTH
 }
 
+/// The maximimum number of results that the engine
+/// will be able to return in one search call.
+pub const HARD_RESULT_LIMIT: usize = 1000;
+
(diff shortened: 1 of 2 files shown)


gmourier:
> I talked about touching the nb_hits value with @qdequele and we concluded that it was not correct to do so.

Curious to hear your thoughts on this! What were the reasons for making this choice?

MarinPostma:
I don't think that this is how it should be done.

what if I ask for 1 document at offset 10000?

I think we need to pass the following params to milli:

```
offset = min(offset, HARD_RESULT_LIMIT);
limit = min(limit, HARD_RESULT_LIMIT - OFFSET);
```

wdyt?

Kerollmops:
Ahha yeah, I'm dumb, you are absolutely right. That's probably due to jet lag. I will update my PR accordingly. Thanks!

meilisearch-lib/src/index/search.rs
```diff
@@ -97,8 +102,13 @@ impl Index {
             search.query(query);
         }
 
-        search.limit(query.limit);
-        search.offset(query.offset.unwrap_or_default());
+        // Make sure that a user can't get more documents than the hard limit,
+        // we align that on the offset too.
+        let offset = min(query.offset.unwrap_or(0), HARD_RESULT_LIMIT);
```
MarinPostma, line 107:
```suggestion
        let offset = query.offset.unwrap_or(0).min(HARD_RESULT_LIMIT));
```

meilisearch-lib/src/index/search.rs
```diff
@@ -97,8 +102,13 @@ impl Index {
...
+        let limit = min(query.limit, HARD_RESULT_LIMIT.saturating_sub(offset));
```
MarinPostma, line 108:
```suggestion
        let limit = query.limit.min(HARD_RESULT_LIMIT.saturating_sub(offset));
```

the saturating sub is not strictly necessary since we already clamped `offset` to `HARD_RESULT_LIMIT`, but I don't see any downsides, so cool

meilisearch-lib/src/index/search.rs
```diff
@@ -1,3 +1,4 @@
+use std::cmp::min;
```
MarinPostma, line 1:
```suggestion
```

MarinPostma (changes requested):

MarinPostma:
@Kerollmops, for testability, you could in fact do a mock here, but maybe the simple option would be to make a function like `compute_offset_limit(offset: usize, limit: usize) -> (usize, uszie)`, and unit test that function instead, wdyt?

If we wanted to make integration tests, having a way to lower the value of `MAX_RESULT_LIMIT` would be useful so we don't have to create search results with 1000 hits, but who should own this setting is another question...

meilisearch-lib/src/index/search.rs
Kerollmops: I find it way easier to use the `cmp::min` function like I do than using the `usize::min` method.

meilisearch-lib/src/index/search.rs
Kerollmops: Yeah you are right about the `saturating_sub` uselessness and I'll keep it :)

curquiza (commented):
Ok for me ✅

Wait for the final Marin's approval for the merge

MarinPostma (approved):

PR #2258, opened 2022-03-22
min word size typo

MarinPostma:
Introduce the `MinWordLenghForTypo` according to https://github.com/meilisearch/specifications/pull/117

awaiting https://github.com/meilisearch/milli/pull/473

PR #2294, opened 2022-04-06
chore(lib): bump milli to 0.25.0

MarinPostma:
bump milli

### DIFF
--- a/meilisearch-lib/src/error.rs
+++ b/meilisearch-lib/src/error.rs
@@ -41,6 +41,7 @@ impl ErrorCode for MilliError<'_> {
                     UserError::CriterionError(_) => Code::InvalidRankingRule,
                     UserError::InvalidGeoField { .. } => Code::InvalidGeoField,
                     UserError::SortError(_) => Code::Sort,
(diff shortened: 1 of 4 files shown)


meilisearch-lib/src/error.rs
```diff
@@ -41,6 +41,7 @@ impl ErrorCode for MilliError<'_> {
                     UserError::CriterionError(_) => Code::InvalidRankingRule,
                     UserError::InvalidGeoField { .. } => Code::InvalidGeoField,
                     UserError::SortError(_) => Code::Sort,
+                    UserError::InvalidMinTypoWordLenSetting(_, _) => unreachable!(),
```
ManyTheFish, line 44: Not sure but, it should be better to return at least an internal error, no? 🤔

MarinPostma: this error cannot be reached now, but will in two PR :) (#2258)

ManyTheFish: Ah ok! Thanks!

MarinPostma:
bors merge

PR #2298, opened 2022-04-07
Nested fields

irevoire:
There are a few things that I want to fix _AFTER_ merging this PR.
For the following RCs.

## Stop the useless conversion
In the `search.rs` I convert a `Document` to a `Value`, and then the `Value` to a `Document` and then back to a `Value` etc. I should stop doing all these conversion and stick to one format.
Probably by merging my `permissive-json-pointer` crate into meilisearch.
That would also give me the opportunity to work directly with obkvs and stops deserializing fields I don't need.

## Add more test specific to the nested
Everything seems to works but I should write tests to double check that the nested works well with the `formatted` field.

## See how I could stop iterating on hashmap and instead fill them correctly
This is related to milli. I really often needs to iterate over hashmap to see if a field is a subset of another field. I could probably generate a structure containing all the possible key values.
ie. the user say `doggo` is an attribute to retrieve. Instead of iterating on all the attributes to retrieve to check if `doggo.name` is a subset of `doggo`. I should insert `doggo.name` in the attributes to retrieve map.

### DIFF
--- a/meilisearch-auth/Cargo.toml
+++ b/meilisearch-auth/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2021"
 [dependencies]
 enum-iterator = "0.7.0"
 meilisearch-error = { path = "../meilisearch-error" }
-milli = { git = "https://github.com/meilisearch/milli.git", tag = "v0.25.0" }
+milli = { git = "https://github.com/meilisearch/milli.git", tag = "v0.26.0" }
 rand = "0.8.4"
 serde = { version = "1.0.136", features = ["derive"] }
 serde_json = { version = "1.0.79", features = ["preserve_order"] }
--- a/meilisearch-lib/src/index/search.rs
+++ b/meilisearch-lib/src/index/search.rs
@@ -106,10 +106,19 @@ pub struct SearchResult {
     pub exhaustive_facets_count: Option<bool>,
 }
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Default)]
 struct FormatOptions {
     highlight: bool,
     crop: Option<usize>,
+}
+
+impl FormatOptions {
+    pub fn merge(self, other: Self) -> Self {
+        Self {
+            highlight: self.highlight || other.highlight,
+            crop: self.crop.or(other.crop),
+        }
+    }
 }
 
 impl Index {
@@ -231,8 +240,8 @@ impl Index {
                 .then(|| compute_matches(&matching_words, &document, &analyzer));
 
             let formatted = format_fields(
+                &document,
                 &fields_ids_map,
-                obkv,
                 &formatter,
                 &matching_words,
                 &formatted_options,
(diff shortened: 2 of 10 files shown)


meilisearch-lib/src/index/search.rs
```diff
@@ -701,28 +731,33 @@ mod test {
...
         let value = format_fields(
+            &document,
             &fields,
-            obkv,
             &formatter,
             &matching_words,
             &formatted_options,
         )
         .unwrap();
+        dbg!(&value);
```
MarinPostma, line 760:
```suggestion
```

meilisearch-lib/src/index/search.rs
```diff
@@ -436,50 +445,71 @@ fn make_document(
...
+    permissive_json_pointer::map_leaf_values(&mut document, selectors, |key, value| {
+        let format = formatted_options
+            .iter()
+            .filter(|(field, _option)| {
+                milli::is_faceted_by(field_ids_map.name(**field).unwrap(), key)
+                    || milli::is_faceted_by(key, field_ids_map.name(**field).unwrap())
+            })
+            .fold(FormatOptions::default(), |acc, (_, option)| {
+                acc.merge(*option)
+            });
```
MarinPostma, line 505: what exactly is going on here? why do we need to merge the format options? can you add some explanations?

meilisearch-lib/src/index/search.rs
```diff
@@ -436,50 +445,71 @@ fn make_document(
...
-                .expect("Missing field name")
-                .to_string();
-
-            document.insert(key, value);
-        }
-    }
+    // Convert the `IndexMap` into a `serde_json::Map`.
+    let document = document
+        .iter()
+        .map(|(k, v)| (k.clone(), v.clone()))
```
MarinPostma, line 484: maybe this works
```suggestion
        .cloned()
```

meilisearch-lib/src/index/search.rs
```diff
@@ -436,50 +445,71 @@ fn make_document(
...
+        .collect();
+
+    let mut document = permissive_json_pointer::select_values(&document, selectors.iter().copied());
+
+    permissive_json_pointer::map_leaf_values(&mut document, selectors, |key, value| {
+        let format = formatted_options
+            .iter()
+            .filter(|(field, _option)| {
+                milli::is_faceted_by(field_ids_map.name(**field).unwrap(), key)
+                    || milli::is_faceted_by(key, field_ids_map.name(**field).unwrap())
```
MarinPostma, line 501:
```suggestion
               let name = field_ids_map.name(**field).unwrap();
                milli::is_faceted_by(name, key)
                    || milli::is_faceted_by(key, name)
```

irevoire: That's better thanks :grin:

meilisearch-lib/src/index/search.rs
irevoire: Sadly no, it doesn't work :pensive:

meilisearch-lib/src/index/search.rs
```diff
@@ -436,50 +445,71 @@ fn make_document(
...
+
+    // select the attributes to retrieve
+    let attributes_to_retrieve = attributes_to_retrieve
+        .iter()
+        .map(|&fid| field_ids_map.name(fid).expect("Missing field name"));
+
+    let document = permissive_json_pointer::select_values(&document.into(), attributes_to_retrieve);
+
+    // then we need to convert the `serde_json::Map` into an `IndexMap`.
+    let document = document.into_iter().collect();
```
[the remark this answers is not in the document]
irevoire: Yes, I was concerned about that too. From what I see it works but it's not specified it should so I guess I'm just lucky.
I would like to move the `permissive-json-pointer` crate into meilisearch so we could make it work directly with the `IndexMap`.
Do you think this needs to be fixed before the RC0 or can it wait until the RC1?

@gmourier @curquiza 
Basically, the jsons in `hits` could have their fields in a different order each time.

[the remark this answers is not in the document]
MarinPostma: actually, when the `preserve_order` feature from serde_json is enables, a `serde_json::Map` is nothing but an `IndexMap` under the hood: https://docs.serde.rs/src/serde_json/map.rs.html#23-26

and it turns out that we're lucky: https://github.com/meilisearch/meilisearch/blob/main/meilisearch-lib/Cargo.toml#L45

This now begs the question: Do we need to do the back and forth conversion?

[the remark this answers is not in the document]
irevoire: Ah! No we don't need and since that's the case I guess we should just remove entirely the `IndexMap` and define a `Document` as a `serde_json::Map<String, Value>`!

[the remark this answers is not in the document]
MarinPostma: I think we can indeed 🤔

meilisearch-auth/Cargo.toml
```diff
@@ -6,7 +6,8 @@ edition = "2021"
 [dependencies]
 enum-iterator = "0.7.0"
 meilisearch-error = { path = "../meilisearch-error" }
-milli = { git = "https://github.com/meilisearch/milli.git", tag = "v0.25.0" }
+# milli = { path = "../../milli/milli" }
+milli = { git = "https://github.com/meilisearch/milli.git", branch = "nested_fields" }
```
Kerollmops, line 10: I conclude that you are awaiting the `nested_fields` PR to be merged on milli?

Kerollmops: I just reviewed https://github.com/meilisearch/milli/pull/458 for you and bors merged it.

Kerollmops (approved):
Looks good to me. We can now wait for the users to try to break this feature! I am also awaiting the benchmarks too.

meilisearch-lib/src/index/search.rs
```diff
@@ -798,23 +831,27 @@ mod test {
...
+            "test": "hello",
+        });
 
-        let obkv = obkv::KvReader::new(&buf);
+        // we need to convert the `serde_json::Map` into an `IndexMap`.
+        let document = document
+            .as_object()
+            .unwrap()
+            .into_iter()
+            .map(|(k, v)| (k.clone(), v.clone()))
```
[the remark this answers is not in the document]
irevoire: we can't do that sadly :pleading_face:

Kerollmops (dismissed):
Looks good to me. Are we waiting for https://github.com/meilisearch/milli/pull/458 to be merged and released?

Kerollmops (approved):
Thank you for the clippy fix!

irevoire:
Yep we need to wait for the release because we need to publish the new `Cargo.lock`

Kerollmops (dismissed):
Looks good to me, thank you @irevoire!

meilisearch-lib/src/index/search.rs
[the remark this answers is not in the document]
curquiza: > Basically, the jsons in hits could have their fields in a different order each time.

Ok to wait for rc1 to fix this!

MarinPostma (approved):

irevoire:
bors merge

Kerollmops:
Is it a bug in Clippy?

irevoire:
yeah look like, it says it's an internal compiler error :unamused: 

Also the benchmarks just finished running and it's worse than I thought :weary: 
```
group                                                             indexing_main_4ae7aea3                 indexing_nested_fields_0438ab7d
-----                                                             ----------------------                 -------------------------------
indexing/Indexing geo_point                                       1.00       8.7±0.19s        ? ?/sec    2.81      24.5±0.17s        ? ?/sec
indexing/Indexing movies in three batches                         1.00      18.2±0.26s        ? ?/sec    1.03      18.7±0.27s        ? ?/sec
indexing/Indexing movies with default settings                    1.00      17.6±0.11s        ? ?/sec    1.01      17.9±0.08s        ? ?/sec
indexing/Indexing songs in three batches with default settings    1.00      63.4±0.41s        ? ?/sec    1.08      68.3±0.93s        ? ?/sec
indexing/Indexing songs with default settings                     1.00      53.2±0.95s        ? ?/sec    1.21      64.2±0.93s        ? ?/sec
indexing/Indexing songs without any facets                        1.00      49.2±0.96s        ? ?/sec    1.20      59.2±1.22s        ? ?/sec
indexing/Indexing songs without faceted numbers                   1.00      52.0±0.23s        ? ?/sec    1.21      62.8±0.28s        ? ?/sec
indexing/Indexing wiki                                            1.00   1007.1±10.25s        ? ?/sec    1.01   1014.7±10.74s        ? ?/sec
indexing/Indexing wiki in three batches                           1.00    1136.9±7.83s        ? ?/sec    1.01    1152.7±9.64s        ? ?/sec
```

I don't really understand how the geosearch can be so slow

irevoire:
bors merge

PR #2336, opened 2022-04-20
Move permissive-json-pointer in the meilisearch repository

irevoire:
Move the permissive-json-pointer crate in the meilisearch repository.

### DIFF
--- a/permissive-json-pointer/Cargo.toml
+++ b/permissive-json-pointer/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "permissive-json-pointer"
+version = "0.2.0"
+edition = "2021"
+description = "A permissive json pointer"
+readme = "README.md"
+
+[dependencies]
+serde_json = "1.0"
+
+[dev-dependencies]
+big_s = "1.0"
--- a/permissive-json-pointer/src/lib.rs
+++ b/permissive-json-pointer/src/lib.rs
@@ -1,0 +1,786 @@
+#![doc = include_str!("../README.md")]
+
+use std::collections::HashSet;
+
+use serde_json::*;
+
+type Document = Map<String, Value>;
+
+const SPLIT_SYMBOL: char = '.';
+
+/// Returns `true` if the `selector` match the `key`.
+///
+/// ```text
+/// Example:
+/// `animaux`           match `animaux`
+/// `animaux.chien`     match `animaux`
+/// `animaux.chien`     match `animaux`
+/// `animaux.chien.nom` match `animaux`
+/// `animaux.chien.nom` match `animaux.chien`
+/// -----------------------------------------
+/// `animaux`    doesn't match `animaux.chien`
+/// `animaux.`   doesn't match `animaux`
+/// `animaux.ch` doesn't match `animaux.chien`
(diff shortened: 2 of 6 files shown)


permissive-json-pointer/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.2.0"
+edition = "2021"
+description = "A permissive json pointer"
+readme = "README.md"
+
+[dependencies]
+serde_json = "1.0"
+
+[dev-dependencies]
+big_s = "1.0"
```
Kerollmops, line 12: Could you add a newline, here, please?

Kerollmops (changes requested):

permissive-json-pointer/src/lib.rs
```diff
@@ -0,0 +1,772 @@
...
+
+use std::collections::HashSet;
+
+use serde_json::*;
+
+type Document = Map<String, Value>;
+
+const SPLIT_SYMBOL: char = '.';
+
+/// Returns `true` if the `subset` is contained in the `main` string.
```
Kerollmops, line 11: Could you change this description, improve it and use the names of the parameters?

Kerollmops (changes requested):
There are a lot of tests and that's good!

permissive-json-pointer/src/lib.rs
irevoire: I improved the message :+1:

permissive-json-pointer/src/lib.rs
```diff
@@ -0,0 +1,783 @@
...
+/// `animaux`           match `animaux`
+/// `animaux.chien`     match `animaux`
+/// `animaux.chien`     match `animaux`
+/// `animaux.chien.nom` match `animaux`
+/// `animaux.chien.nom` match `animaux.chien`
+/// -----------------------------------------
+/// `animaux`          doesn't match `animaux.chien`
+/// `animaux.`    doesn't match `animaux`
+/// `animaux.ch`          doesn't match `animaux.chien`
+/// `animau` doesn't match `animaux`
```
Kerollmops, line 22:
```suggestion
/// Returns `true` if the `selector` match the `key`.
///
/// ```
/// Example:
/// `animaux`           match `animaux`
/// `animaux.chien`     match `animaux`
/// `animaux.chien`     match `animaux`
/// `animaux.chien.nom` match `animaux`
/// `animaux.chien.nom` match `animaux.chien`
/// -----------------------------------------
/// `animaux`          doesn't match `animaux.chien`
/// `animaux.`    doesn't match `animaux`
/// `animaux.ch`          doesn't match `animaux.chien`
/// `animau` doesn't match `animaux`
/// ```
```

Kerollmops (changes requested):

Kerollmops (approved):
bors merge

permissive-json-pointer/src/lib.rs
irevoire: You can't do that because then it's considered a test.
And you can't write a real test with `assert_eq` either because the function isn't public.
Thus I made the block a text block instead of a code block.

PR #2355, opened 2022-04-26
feat(http): add analytics on typo tolerance setting

MarinPostma:
Add analytics for the typo settings.

Also make settings analytics return null for settings that are not set, instead of a default value, as seen with @gmourier

### DIFF
--- a/meilisearch-http/src/routes/indexes/settings.rs
+++ b/meilisearch-http/src/routes/indexes/settings.rs
@@ -171,7 +171,7 @@ make_setting_route!(
         use serde_json::json;
 
         analytics.publish(
-            "TypoToleranceUpdated Updated".to_string(),
+            "TypoTolerance Updated".to_string(),
             json!({
                 "typo_tolerance": {
                     "enabled": setting.as_ref().map(|s| !matches!(s.enabled, Setting::Set(false))).unwrap_or(true),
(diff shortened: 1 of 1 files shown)


meilisearch-http/src/routes/indexes/settings.rs
```diff
@@ -165,7 +165,49 @@ make_setting_route!(
...
+                        .unwrap_or(5),
+                    "min_word_size_for_two_typos": setting
+                        .as_ref()
+                        .map(|s| s.min_word_size_for_typos
+                            .as_ref()
+                            .set()
+                            .map(|s| s.two_typos.set()))
+                        .flatten()
+                        .flatten()
+                        .unwrap_or(9),
```
gmourier, line 205: Just out of curiosity, is it a good thing to specify constants here too? e.g. `.unwrap_or(9)`. If one day these values change, we should not forget them here I guess?

irevoire: Yeah, it could be a good idea to create two constants to hold these values so everything always stays synchronized.

irevoire (dismissed):
Good job :+1:

curquiza:
bors merge

irevoire (approved):
:+1:

curquiza:
bors merge

PR #2374, opened 2022-05-04
feat(analytics): handle the `X-Meilisearch-Client` header

### LINKED ISSUE #2367: Telemetry: Add `X-Meilisearch-Client` header
Why?
Following this [comment](https://github.com/meilisearch/integration-guides/issues/150#issuecomment-966360488) pointing out that some browsers forbid the customization of the User-Agent header, the analytics stack can now take the value of a `x-meilisearch-client` query parameter if found instead of the user-agent header to track the usage of our SDKs/Integrations.

Spec: https://github.com/meilisearch/specifications/pull/145

Reminder: the telemetry can be disable at any moment using the `--no-analytics` flag when launching Meilisearch
```bash
meilisearch --no-analytics
```
----

⚠️ EDIT ⚠️

After discussion with @brunoocasali, @Kerollmops and @irevoire, it appears adding a query parameter is not a good solution (cf the [related specification PR description](https://github.com/meilisearch/specifications/pull/150#issue-1241650655))
Instead, we will use the `X-Meilisearch-Client` header for front-end integration analytics

Please refer to the specification: https://github.com/meilisearch/specifications/pull/150

irevoire:
Fix #2367

### DIFF
--- a/meilisearch-http/src/analytics/segment_analytics.rs
+++ b/meilisearch-http/src/analytics/segment_analytics.rs
@@ -31,6 +31,8 @@ use crate::Opt;
 
 use super::{config_user_id_path, MEILISEARCH_CONFIG_PATH};
 
+const ANALYTICS_HEADER: &str = "X-Meilisearch-Client";
+
 /// Write the instance-uid in the `data.ms` and in `~/.config/MeiliSearch/path-to-db-instance-uid`. Ignore the errors.
 fn write_user_id(db_path: &Path, user_id: &str) {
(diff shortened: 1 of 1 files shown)


irevoire:
Converting to draft until I have the time to test it

irevoire:
A query parameter! But the first version I implemented used a custom header 😂

Kerollmops (approved):
Thank you very much, I just checked and headers are not case sensitive and they are not!
https://stackoverflow.com/a/5259004/1941280

bors merge

PR #2410, opened 2022-05-19
Make dump a task

MarinPostma:
This PR transforms the dump task into a proper task.
The `GET /dumps/:dump_uid` is removed.

Some changes were made to make this work, and a bit a refactoring was necessary.
- The `dump_actor` module has been renamed do `dumps` and moved to the root
- There isn't a `DumpActor` anymore, and the dump process is handled by the `DumpHandler`.
- The `TaskPerformer` is renamed to `BatchHandler`
- The `BatchHandler` trait no longer has a `perform_job` method, but instead has a `accept` method returning whether a handler can proccess a batch
- The scheduler now accept a list of `BatchHandler`, and iterates trhough them until it finds one to accept the current batch.
- `Job` doesn't exist anymore, and everything in now inside of the `BatchContent` enum.
- The `Vec<TaskId>` from `Batch` is replaced with a `BatchContent` enum which hints at the content.
- The Scheduler is slightly modified to accept batch, and prioritize them before regular tasks.
- The `TaskList` are not identified by a `String` representing the index uid anymore, but by a `TaskListIdentifier` which also works for dumps which are not targeting any specific indexes.
- The `GET /dump/:dump_id` no longer exists
- `DumpActorError` is renamed to `DumpError`

close #2410


meilisearch-http/src/task.rs
```diff
@@ -342,7 +350,7 @@ impl From<Vec<TaskView>> for TaskListView {
 #[serde(rename_all = "camelCase")]
 pub struct SummarizedTaskView {
     uid: TaskId,
-    index_uid: String,
+    index_uid: Option<String>,
```
Kerollmops, line 353: Do we want to display a `null` here or do we prefer to remove the field from the payload by not serializing it? What does the spec says?

meilisearch-lib/src/index_resolver/mod.rs
```diff
@@ -493,17 +411,23 @@ mod test {
...
+            error::{IndexError, Result as IndexResult},
+            Checked, IndexMeta, IndexStats, Settings,
+        },
+        tasks::{batch::Batch, BatchHandler},
     };
     use index_store::MockIndexStore;
     use meta_store::MockIndexMetaStore;
 
+    // TODO: ignoring this test, it has become too complex to maintain, and rather implement
+    // handler logic test.
```
Kerollmops, line 425: Could you create a PR to help us keep track of that? With a permanent link to the associated code part?

meilisearch-lib/src/tasks/scheduler.rs
```diff
@@ -82,29 +83,38 @@ impl DerefMut for TaskList {
...
+                    (None, None) => Ordering::Equal,
+                    (None, Some(_)) => Ordering::Less,
+                    (Some(_), None) => Ordering::Greater,
+                    (Some(lhs), Some(rhs)) => lhs.cmp(rhs),
+                }
+            }
+            (TaskListIdentifier::Index(_), TaskListIdentifier::Dump) => Ordering::Less,
+            (TaskListIdentifier::Dump, TaskListIdentifier::Index(_)) => Ordering::Greater,
+            (TaskListIdentifier::Dump, TaskListIdentifier::Dump) => {
+                unreachable!("There should be only one Dump task list")
```
Kerollmops, line 116: I am not sure that I understand, does it mean that there can only be one dump task in the task list? That a user cannot ask for two dumps one after the other?

meilisearch-lib/src/tasks/task.rs
```diff
@@ -74,7 +86,17 @@ pub enum TaskEvent {
 #[cfg_attr(test, derive(proptest_derive::Arbitrary))]
 pub struct Task {
     pub id: TaskId,
-    pub index_uid: IndexUid,
+    /// The name of the index the task is targeting. If it isn't targeting any idex (i.e Dump task)
+    /// then this is None
+    // TODO: when next forward breaking dumps, it would be a good idea to move this field inside of
+    // the TaskContent.
```
Kerollmops, line 92: I agree with you, could you create an issue to help us track this? By adding a persistent link to this part of the code.

meilisearch-lib/src/tasks/task_store/mod.rs
```diff
@@ -30,10 +32,14 @@ pub struct TaskFilter {
...
-        self.indexes
-            .as_ref()
-            .map(|indexes| indexes.contains(&*task.index_uid))
-            .unwrap_or(true)
+        match task.index_uid {
+            Some(ref index_uid) => self
+                .indexes
+                .as_ref()
+                .map(|indexes| indexes.contains(index_uid.as_str()))
+                .unwrap_or(true),
```
Kerollmops, line 40:
```suggestion
                .map_or(true, |indexes| indexes.contains(index_uid.as_str())),
```

Kerollmops (changes requested):

meilisearch-http/src/task.rs
MarinPostma: the specs says `null`:
> The 202 Accepted (Summarized task response) and the related task object define the indexUid to null since dump are related to all indexes.

meilisearch-lib/src/index_resolver/mod.rs
MarinPostma: This is part of my next PR, I need to create a mocker for IndexResolver and work from there. I will remove this test in it :)

meilisearch-lib/src/tasks/scheduler.rs
MarinPostma: no, this is not what is happening here. We create a specific task list for dump, that is identified with the `TaskListIdentifier::Dump` identifier. If the user asks for a new dump, it will be added to the existing dump list if it exists, or it will be created, but I can't have 2 task list with the `Dump` identifier

meilisearch-lib/src/tasks/task.rs
MarinPostma: this is already done here https://github.com/meilisearch/meilisearch/issues/2430, will be done in this release

meilisearch-lib/src/tasks/scheduler.rs
irevoire: > That a user cannot ask for two dumps one after the other?
Also, yes, that's a limitation we always had in meilisearch;
https://docs.meilisearch.com/reference/api/error_codes.html#error-list

meilisearch-lib/src/tasks/handlers/index_resolver_handler.rs
```diff
@@ -0,0 +1,146 @@
...
+            }
+            BatchContent::IndexUpdate(ref mut task) => match self.process_task(task).await {
+                Ok(success) => task.events.push(TaskEvent::succeeded(success)),
+                Err(err) => task.events.push(TaskEvent::failed(err.into())),
+            },
+            _ => unreachable!(),
+        }
+
+        batch
+    }
```
irevoire, line 35: I find it quite strange to have an `accept` method + an `unreachable` in the `process_batch`.
Couldn't we implement a `process(self, Context)` on the tasks directly?

It might be strange too to have a `Context` which contains everything though, but I think you have access to everything when you call the accept to process_batch methods.

I don't think we should change that right now, but I would like to hear if you considered doing it the other way around and what was the issue.

irevoire (approved):

meilisearch-lib/src/tasks/handlers/index_resolver_handler.rs
MarinPostma: On the task or on the batch? What would Context be?

irevoire: > On the task or on the batch? What would Context be?

On the task. The `Context` would be something that contains;
- The `IndexMetaStore`
- The `IndexStore`
- The `SnapshotHandler`
- The `DumpHandler`

Basically, everything that implements `BatchHandler` currently.

MarinPostma: I'm not convinced about this approach, it doesn't feel right that tasks process themselves to me, they're just data after all. we would be shifting a lot of logic from the handlers to the tasks, and that would not be right for IndexResolver. It seems natural that IndexResolver has the logic necessary to apply an update to an index, hence the logic to do so is in its implementation of `BatchHandler`. While this seems a bit less true for `BatchHandler` because I've just introduced it, for me it contains the logic of performing the batch into a struct whose job is explicitly to do that.

If the goal is to remove the accepted method, I can agree with that and let do what  I proposed, otherwise I am interested in stronger arguments that would justify a change (even though we're not doing it know as you said:) )

Kerollmops (approved):
Thank you very much @MarinPostma for the changes!
bors merge

PR #2414, opened 2022-05-21
Improve index uid validation upon API key creation

### LINKED ISSUE #2158: API key creation should returns an error
**Describe the bug**
While creating an API key we can specify the indexes it applies to.
And if we specify a [bad formatted index name](https://github.com/meilisearch/specifications/blob/main/text/0061-error-format-and-definitions.md#invalid_index_uid) we’re supposed to get an [`invalid_api_key_indexes`](https://github.com/meilisearch/specifications/blob/develop/text/0061-error-format-and-definitions.md#invalid_api_key_indexes) error from what I see.

That’s not the case.
The same happens for the update route.

**To Reproduce**
1. Send the following request: `echo '{ "actions": [], "description": "test_error_create_key", "indexes": ["invalid index # / \\name with spaces"], "expiresAt": null }' | http :7700/keys 'Authorization: Bearer masterKey'`
2. Get the result:
```
{
    "actions": [],
    "createdAt": "2022-02-09T14:47:49Z",
    "description": "test_error_create_key",
    "expiresAt": null,
    "indexes": [
        "invalid index # / \name with spaces"
    ],
    "key": "AG5o4s740edcfd04ce86f2a92259ca6f886485caf22ebb156b91211ac4e10119482f6e67",
    "updatedAt": "2022-02-09T14:47:49Z"
}
```

**Expected behavior**
Error

**Screenshots**
image

**Meilisearch version:** v0.25.2

**Additional context**
I gave you my private key please don’t hack me.

--- 

Edit by @curquiza 

TODO:

The spec specifies:
> Sending a value that is not a valid `index` (e.g. regarding the index uid format) in the `indexes` field returns an [invalid_index_uid](0061-error-format-and-definitions.md#invalid_index_uid) error.

See https://github.com/meilisearch/specifications/pull/112/files

We want to make the search engine returns an error when the index uid is invalid

Definitely something that can be fixed by an external contributor! Feel free to ask for help

### LINKED ISSUE #2168: facetsDistribution returning thousands of Facets, instead of just the few calculated ones
facetsDistribution seems to be getting all my facets, which causes the page to load really slow. Here's an example. Why isn't facetsDistribution filtering to only. the top 10, for instance. 

[``]([url](url)
Screen Shot 2022-02-14 at 1 00 42 AM
Screen Shot 2022-02-14 at 1 00 19 AM
)

What seems to be happening is that its' not just grabbing the top 10 keywords, it's grabbing the whole database of keywords. Any idea what might be causing this?

pierre-l:
- ~Use an IndexUid newtype to enforce stronger constraints~
- ~`cargo update -p vergen`~ (`rustup update` was the proper fix for this)
- Add a new `meilisearch_types` crate
- Move `meilisearch_error` to `meilisearch_types::error`
- Move `meilisearch_lib::index_resolver::IndexUid` to `meilisearch_types::index_uid`
- Add a new `InvalidIndexUid` error in `meilisearch_types::index_uid`
- Move `meilisearch_http::routes::StarOr` to `meilisearch_types::star_or`
- Use the `IndexUid` and `StarOr` in `meilisearch_auth::Key`

Fixes #2158


meilisearch-auth/src/index_uid.rs
```diff
@@ -0,0 +1,157 @@
...
+pub struct InvalidIndexUid(pub String);
+
+impl TryFrom<String> for IndexUid {
+    type Error = InvalidIndexUid;
+
+    fn try_from(uid: String) -> Result<Self, Self::Error> {
+        if uid
+            .chars()
+            .all(|x| x.is_ascii_alphanumeric() || x == '-' || x == '_')
+            || !(1..=400).contains(&uid.len())
```
pierre-l, line 37: This is duplicated from `meilisearch-lib`, inaccessible from `meilisearch-auth`.
The former already has a `IndexUid` struct representing the same concept, with similar implementation.

I would personally advocate for the creation of a `meilisearch-types` crate that would regroup these implementations, and include all that's in `meilisearch-error`. This new crate could be used to implement Rust clients for Meilisearch. It would be a breaking change though, and it doesn't feel right to me to do this in this PR.

Thoughts?

permissive-json-pointer/src/lib.rs
```diff
@@ -206,7 +206,7 @@ fn create_value(value: &Document, mut selectors: HashSet<&str>) -> Document {
     new_value
 }
 
-fn create_array(array: &Vec<Value>, selectors: &HashSet<&str>) -> Vec<Value> {
+fn create_array(array: &[Value], selectors: &HashSet<&str>) -> Vec<Value> {
```
pierre-l, line 209: Unrelated to the issue this PR fixes, but Clippy warns about this:

```
warning: writing `&Vec<_>` instead of `&[_]` involves one more reference and cannot be used with non-Vec-based slices
   --> permissive-json-pointer/src/lib.rs:209:24
    |
209 | fn create_array(array: &Vec<Value>, selectors: &HashSet<&str>) -> Vec<Value> {
    |                        ^^^^^^^^^^^ help: change this to: `&[Value]`
    |
    = note: `#[warn(clippy::ptr_arg)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg
```

Let me know if I should remove this from the PR.

curquiza:
Hello @pierre-l thanks for your PR, we will review it asap, the team is focusing on the v0.28.0 sprint at the moment 😄

meilisearch-auth/src/index_uid.rs
MarinPostma: agreed, I don't like that this is duplicated here. Having a `meilisearch-type` crate would solve this issue, and maybe some others.

permissive-json-pointer/src/lib.rs
MarinPostma: that ok, thanks

MarinPostma (dismissed):
I really don't like that we need to repeat code here. I think your idea of creating a `meilisearch-type` is a good idea

Kerollmops (approved):
Thank you very much for your PR. The codebase will be much better and easier to contribute now!
bors merge

curquiza:
@pierre-l sorry for asking this, but we cannot merge the PR because of the merge conflict 😅 could you please rebase and fix the conflicts?

pierre-l:
@curquiza Done!

Kerollmops (approved):
Thank you @pierre-l, great job!
bors merge

PR #2429, opened 2022-05-25
Send the analytics to `telemetry.meilisearch.com` instead of segment

### LINKED ISSUE #2425: Telemetry: specify `telemetry.meilisearch.com` as a transit domain
Related spectification: https://github.com/meilisearch/specifications/pull/153

The telemetry data previously sent to Segment shoud now be forwarded to the domain `telemetry.meilisearch.com`, which redirects them to Segment.

Reminder: the telemetry can be disabled at any moment using the `--no-analytics` flag when launching Meilisearch
```bash
meilisearch --no-analytics
```

irevoire:
Fix #2425

### DIFF
--- a/meilisearch-http/src/analytics/segment_analytics.rs
+++ b/meilisearch-http/src/analytics/segment_analytics.rs
@@ -81,7 +81,19 @@ impl SegmentAnalytics {
         let user_id = user_id.unwrap_or_else(|| Uuid::new_v4().to_string());
         write_user_id(&opt.db_path, &user_id);
 
-        let client = HttpClient::default();
+        let client = reqwest::Client::builder()
+            .connect_timeout(Duration::from_secs(10))
+            .build();
+
+        // if reqwest throws an error we won't be able to send analytics
+        if client.is_err() {
+            return super::MockAnalytics::new(opt);
+        }
+
+        let client = HttpClient::new(
+            client.unwrap(),
+            "https://telemetry.meilisearch.com".to_string(),
+        );
         let user = User::UserId { user_id };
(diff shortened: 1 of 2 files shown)


meilisearch-http/src/analytics/segment_analytics.rs
```diff
@@ -81,7 +81,11 @@ impl SegmentAnalytics {
         let user_id = user_id.unwrap_or_else(|| Uuid::new_v4().to_string());
         write_user_id(&opt.db_path, &user_id);
 
-        let client = HttpClient::default();
+        let client = reqwest::Client::builder()
+            .connect_timeout(Duration::new(10, 0))
```
MarinPostma, line 85: I find this more readable, don't you think?
```suggestion
            .connect_timeout(Duration::from_secs(10))
```

meilisearch-http/src/analytics/segment_analytics.rs
```diff
@@ -81,7 +81,11 @@ impl SegmentAnalytics {
...
+            .build()
+            .unwrap();
```
MarinPostma, line 87: are you sure this can't fail?
https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.build

MarinPostma (changes requested):

meilisearch-http/src/analytics/segment_analytics.rs
irevoire: Yeah that's true :+1:

MarinPostma:
bors merge

PR #2434, opened 2022-05-26
Update docker volume path

0x0x1:
Currently, the args `$(pwd)/data.ms:/data.ms` cannot share data from the container, makes docker volume same as the [Dockerfile](https://github.com/0x0x1/meilisearch/blob/67b6f4340aaea005471a0ad41c700bde1d88afe9/Dockerfile#L40) to fixing it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ meilisearch
 #### Docker
 
 ```bash
-docker run -p 7700:7700 -v "$(pwd)/data.ms:/data.ms" getmeili/meilisearch
+docker run -p 7700:7700 -v "$(pwd)/data.ms:/meili_data/data.ms" getmeili/meilisearch
 ```
 
 #### Announcing a cloud-hosted Meilisearch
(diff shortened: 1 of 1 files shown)


irevoire (approved):
Ah yes, good catch!
@curquiza it looks good to me but I let you merge

curquiza:
@irevoire isn't it better with `-v $(pwd)/meili_data:/meili_data` instead?

irevoire:
Ah yes obviously :facepalm: 
That's why we changed the path initially!

README.md
```diff
@@ -58,7 +58,7 @@ meilisearch
 #### Docker
 
 ```bash
-docker run -p 7700:7700 -v "$(pwd)/data.ms:/data.ms" getmeili/meilisearch
+docker run -p 7700:7700 -v "$(pwd)/data.ms:/meili_data/data.ms" getmeili/meilisearch
```
irevoire, line 61:
```suggestion
docker run -p 7700:7700 -v "$(pwd)/meili_data:/meili_data" getmeili/meilisearch
```

irevoire (changes requested):

curquiza (approved):

PR #2444, opened 2022-05-30
add boilerplate for dump v5

MarinPostma:
add the boilerplate files for dump v5


meilisearch-lib/src/dump/loaders/v4.rs
```diff
@@ -19,31 +12,15 @@ pub fn load_dump(
     index_db_size: usize,
     meta_env_size: usize,
     indexing_options: &IndexerOpts,
-    version: &str,
 ) -> anyhow::Result<()> {
-    info!(
-        "Loading dump from {}, dump database version: {}, dump version: {}",
-        meta.dump_date, meta.db_version, version
-    );
+    info!("Patching dump V4 to dump V5...");
```
ManyTheFish, line 16: I know this is useless here, but I think it will ease future merges:

```suggestion
    info!("Patching dump V4 to dump V5...");
    
    let patched_dir = tempdir()?;
    let options = CopyOptions::default();
    
    // Indexes
    dir::copy(src.as_ref().join("indexes"), patched_dir.path(), &options)?;
    
    // Index uuids
    dir::copy(
        src.as_ref().join("index_uuids"),
        patched_dir.path(),
        &options,
    )?;
    
    // Metadata
    fs::copy(
        src.as_ref().join("metadata.json"),
        patched_dir.path().join("metadata.json"),
    )?;
    
    // Updates
    dir::copy(src.as_ref().join("updates"), patched_dir.path(), &options)?;
    
    // Keys
    fs::copy(
        src.as_ref().join("keys"),
        patched_dir.path().join("keys"),
    )?;
```

ManyTheFish (changes requested):

meilisearch-lib/src/dump/loaders/v4.rs
ManyTheFish: obviously, we may return `patched_dir` instead of `src`

ManyTheFish:
poke @MarinPostma (I can't set you as a reviewer)

ManyTheFish (approved):

PR #2447, opened 2022-05-31
move index uid in task content

MarinPostma:
this pr moves the index_uid from the `Task` to the `TaskContent`. This is because the task can now have content that do not target a particular index.

### DIFF
--- a/meilisearch-lib/src/dump/compat/v3.rs
+++ b/meilisearch-lib/src/dump/compat/v3.rs
@@ -4,10 +4,10 @@ use serde::{Deserialize, Serialize};
 use time::OffsetDateTime;
 use uuid::Uuid;
 
-use super::v4::{Task, TaskEvent};
+use super::v4::{Task, TaskContent, TaskEvent};
 use crate::index::{Settings, Unchecked};
 use crate::index_resolver::IndexUid;
-use crate::tasks::task::{DocumentDeletion, TaskContent, TaskId, TaskResult};
+use crate::tasks::task::{DocumentDeletion, TaskId, TaskResult};
 
 use super::v2;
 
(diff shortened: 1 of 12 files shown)


meilisearch-lib/src/dump/compat/v3.rs
```diff
@@ -186,10 +186,10 @@ impl Failed {
 impl From<(UpdateStatus, String, TaskId)> for Task {
     fn from((update, uid, task_id): (UpdateStatus, String, TaskId)) -> Self {
         // Dummy task
-        let mut task = Task {
+        let mut task = super::v4::Task {
             id: task_id,
-            index_uid: IndexUid::new(uid).unwrap(),
-            content: TaskContent::IndexDeletion,
+            index_uid: IndexUid::new_unchecked(uid),
```
Kerollmops, line 191: Why not unwrapping here? it is incorrect to not check that.

meilisearch-lib/src/index_resolver/mod.rs
```diff
@@ -411,193 +421,193 @@ where
...
-    use crate::{
-        index::{
-            error::{IndexError, Result as IndexResult},
-            Checked, IndexMeta, IndexStats, Settings,
-        },
-        tasks::{batch::Batch, BatchHandler},
-    };
-    use index_store::MockIndexStore;
-    use meta_store::MockIndexMetaStore;
+    // use std::{collections::BTreeMap, vec::IntoIter};
```
Kerollmops, line 424: Could you remove this test and not comment it, or maybe skip it, instead?

Kerollmops (changes requested):

meilisearch-lib/src/dump/compat/v3.rs
Kerollmops: Ok I understand, we shouldn't be more restrictive than before, but I understand the point. Thank you for the explanation.

meilisearch-lib/src/index_resolver/mod.rs
Kerollmops: Nice, thanks!

PR #2449, opened 2022-06-01
Use bcrypt instead of sha256

ManyTheFish:


ManyTheFish:
We encounter a lot of performance issues with bcrypt like algorithm. We will use an HMAC instead.
closing in favor of #2487

PR #2460, opened 2022-06-05
Create custom error types for `TaskType`, `TaskStatus`, and `IndexUid`

### LINKED ISSUE #2443: Create custom error types for parsing issues
We should create custom error types for when we parse the `TaskType`, and `TypeStatus` in our `FromStr` implementations. The same applies to the `IndexUid::new` method that returns an `IndexResolverError`, we should implement the right `IndexResolverError::from` for a custom error type too.

It would be much more Rust idiomatic and easier to work with.

---

We have [a `String` error for the `TaskType` and `TypeStatus` structs](https://github.com/meilisearch/meilisearch/pull/2399#discussion_r884592697) and `FromStr` implementation for `IndexUid` returns an `IndexResolverError`, this is not super consistent, maybe we want a custom error type for this error?

_Originally posted by @MarinPostma in https://github.com/meilisearch/meilisearch/pull/2399#discussion_r884579948_

walterbm:
# Pull Request

## What does this PR do?
Fixes #2443 by making the following changes:

- Add custom `TaskTypeError` for `TaskType::from_str` 
- Add custom `TaskStatusError` for `TaskStatus::from_str`
- Add custom `IndexUidFormatError` for `IndexUid::from_str`
- Implement `From<IndexUidFormatError> for IndexResolverError` to convert between errors
- Replace all usages of `IndexUid::new` with `IndexUid::from_str`
    - **NOTE** I am relatively new to Rust and I struggled a lot with this final part. This PR ended up with a messy error conversion which does not seem ideal. Please let me know if you have any suggestions for how to make this better and I'll be happy to make any updates!

## PR checklist

### DIFF
--- a/meilisearch-http/src/task.rs
+++ b/meilisearch-http/src/task.rs
@@ -1,4 +1,5 @@
-use std::fmt::Write;
+use std::error::Error;
+use std::fmt::{self, Write};
 use std::str::FromStr;
 use std::write;
 
@@ -39,10 +40,29 @@ impl From<TaskContent> for TaskType {
     }
 }
 
-impl FromStr for TaskType {
-    type Err = String;
+#[derive(Debug)]
+pub struct TaskTypeError {
+    invalid_type: String,
+}
 
-    fn from_str(status: &str) -> Result<Self, String> {
+impl fmt::Display for TaskTypeError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "invalid task type `{}`, expecting one of: \
+            indexCreation, indexUpdate, indexDeletion, documentAdditionOrUpdate, \
+            documentDeletion, settingsUpdate, dumpCreation",
+            self.invalid_type
+        )
+    }
+}
+
+impl Error for TaskTypeError {}
+
+impl FromStr for TaskType {
+    type Err = TaskTypeError;
+
+    fn from_str(status: &str) -> Result<Self, TaskTypeError> {
         if status.eq_ignore_ascii_case("indexCreation") {
             Ok(TaskType::IndexCreation)
         } else if status.eq_ignore_ascii_case("indexUpdate") {
@@ -58,12 +78,9 @@ impl FromStr for TaskType {
         } else if status.eq_ignore_ascii_case("dumpCreation") {
             Ok(TaskType::DumpCreation)
         } else {
-            Err(format!(
-                "invalid task type `{}`, expecting one of: \
-                indexCreation, indexUpdate, indexDeletion, documentAdditionOrUpdate, \
-                documentDeletion, settingsUpdate, dumpCreation",
-                status
-            ))
+            Err(TaskTypeError {
+                invalid_type: status.to_string(),
+            })
         }
     }
 }
--- a/meilisearch-http/src/task.rs
+++ b/meilisearch-http/src/task.rs
@@ -94,10 +94,28 @@ pub enum TaskStatus {
     Failed,
 }
 
-impl FromStr for TaskStatus {
-    type Err = String;
+#[derive(Debug)]
+pub struct TaskStatusError {
+    invalid_status: String,
+}
 
-    fn from_str(status: &str) -> Result<Self, String> {
+impl fmt::Display for TaskStatusError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "invalid task status `{}`, expecting one of: \
+            enqueued, processing, succeeded, or failed",
+            self.invalid_status,
+        )
+    }
+}
+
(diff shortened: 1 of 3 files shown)


meilisearch-http/src/task.rs
```diff
@@ -39,10 +40,29 @@ impl From<TaskContent> for TaskType {
...
+}
+
+impl Error for TaskTypeError {}
+
 impl FromStr for TaskType {
-    type Err = String;
+    type Err = TaskTypeError;
 
-    fn from_str(status: &str) -> Result<Self, String> {
+    fn from_str(status: &str) -> Result<Self, TaskTypeError> {
```
Kerollmops, line 65: Could you change the name of this parameter to be named `type_`, please? It has been copy/pasted from the `TaskStatus` impl by myself and I forgot to change the name :)

meilisearch-lib/src/index_resolver/mod.rs
```diff
@@ -81,18 +71,53 @@ impl std::ops::Deref for IndexUid {
...
-    type Err = IndexResolverError;
+    type Err = IndexUidFormatError;
 
-    fn from_str(s: &str) -> Result<IndexUid> {
-        IndexUid::new(s.to_string())
+    fn from_str(uid: &str) -> std::result::Result<IndexUid, IndexUidFormatError> {
+        if !uid
+            .chars()
+            .all(|x| x.is_ascii_alphanumeric() || x == '-' || x == '_')
+            || !(1..=400).contains(&uid.len())
```
Kerollmops, line 113: I understand that this was not the scope of this PR, but I find that this line is not easy to understand, could you change it to something like:

```suggestion
            || uid.len() < 1 || uid.len() > 400
```

meilisearch-lib/src/index_controller/mod.rs
```diff
@@ -356,7 +358,18 @@ where
...
+        //      let index_uid = IndexUid::from_str(&uid)?;
+        //      let index_uid: std::result::Result<_, IndexResolverError> = IndexUid::from_str(&uid).into()?;
+        // bad implementation but compiles
+        let index_uid = match IndexUid::from_str(&uid) {
+            Ok(uid) => uid,
+            Err(e) => {
+                let e: IndexResolverError = e.into();
+                Err(e)?
+            }
+        };
```
Kerollmops, line 371: Not sure of the reason why it is struggling to infer the error type but this Should do the trick:

```suggestion
let index_uid = IndexUid::from_str(&uid).map_err(IndexResolverError::from)?;
```

Kerollmops: It looks like the reason it is struggling to infer the conversion function (i.e. `Into::into`) is because multiple ones can match:

```
note: multiple `impl`s satisfying `IndexControllerError: ~const std::convert::From<_>` found
   --> meilisearch-lib/src/index_controller/error.rs:18:17
    |
18  | #[derive(Debug, thiserror::Error)]
    |                 ^^^^^^^^^^^^^^^^
...
40  | internal_error!(IndexControllerError: JoinError, UpdateFileStoreError);
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
41  |
42  | impl From<actix_web::error::PayloadError> for IndexControllerError {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: and another `impl` found in the `core` crate: `impl<T> std::convert::From<T> for T;`
    = note: required because of the requirements on the impl of `FromResidual<std::result::Result<Infallible, _>>` for `std::result::Result<tasks::task::Task, IndexControllerError>`
```

Kerollmops (changes requested):
This PR is simply amazing, it is simple to review with different commits for each change, it is short, and does the job pretty well. You even have exposed the issues you had with some parts of the code and the attempts you've made.

I would say GG and thank you 😃 🎉

meilisearch-http/src/task.rs
walterbm: done!

meilisearch-lib/src/index_resolver/mod.rs
walterbm: done! Clippy suggested `uid.is_empty()` instead of `uid.len() < 1` so I used that but let me know if you want me to change it

meilisearch-lib/src/index_controller/mod.rs
walterbm: Ah that makes sense! Thank you for the explanation!

walterbm:
@Kerollmops — just made all the updates you suggested! Let me know what you think!

Kerollmops (approved):
Thank you very much for your work, that's amazing!
bors merge

PR #2464, opened 2022-06-06
Simplify the `star_or` function usage

Kerollmops:
This PR simplifies the usage of the `star_or` function that was originally introduced in #2399. The `serde-cs` https://github.com/naughie/serde-cs/pull/1 PR was merged and was implementing the `IntoIterator` trait on the `CS` types, which makes it possible to directly collect without converting a `CS` into the inner type (vec).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3078,9 +3078,9 @@ dependencies = [
 
 [[package]]
 name = "serde-cs"
-version = "0.2.2"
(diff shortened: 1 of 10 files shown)


ManyTheFish (approved):
nice refacto. I find it more pleasant to read.
Thanks @Kerollmops!

bors merge

PR #2476, opened 2022-06-07
fix(test): Reduce default index size in tests

ManyTheFish:


meilisearch-http/tests/index/get_index.rs
```diff
@@ -77,13 +77,12 @@ async fn get_and_paginate_indexes() {
     const NB_INDEXES: usize = 50;
     for i in 0..NB_INDEXES {
         server.index(&format!("test_{i:02}")).create(None).await;
+        server
+            .index(&format!("test_{i:02}"))
+            .wait_task(i as u64)
+            .await;
```
MarinPostma, line 83: any idea why this works?

MarinPostma (approved):

meilisearch-http/tests/index/get_index.rs
ManyTheFish: I'm not sure, but, I think the Windows instances have lower performances than the Linux ones. This increases the chances for `wait_task` to timeout:
Capture d’écran 2022-06-07 à 16 14 07

irevoire: windows is really not fun, it's going to slow down everything a little bit more :weary:

PR #2487, opened 2022-06-08
Feat(auth): Use hmac instead of sha256 to derivate API keys from master key

ManyTheFish:
Wrap sha256 in HMAC to derivate new API keys from the master key.

### DIFF
--- a/meilisearch-auth/src/store.rs
+++ b/meilisearch-auth/src/store.rs
@@ -11,7 +11,7 @@ use enum_iterator::IntoEnumIterator;
 use hmac::{Hmac, Mac};
 use milli::heed::types::{ByteSlice, DecodeIgnore, SerdeJson};
 use milli::heed::{Database, Env, EnvOpenOptions, RwTxn};
-use sha2::Sha256;
+use sha2::{Digest, Sha256};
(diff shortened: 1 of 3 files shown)


meilisearch-auth/src/store.rs
```diff
@@ -242,9 +243,12 @@ impl<'a> milli::heed::BytesEncode<'a> for KeyIdActionCodec {
 }
 
 pub fn generate_key_as_base64(uid: &[u8], master_key: &[u8]) -> String {
-    let key = [uid, master_key].concat();
-    let sha = Sha256::digest(&key);
-    base64::encode_config(sha, base64::URL_SAFE_NO_PAD)
+    let master_key_sha = Sha256::digest(master_key);
+    let mut mac = Hmac::<Sha256>::new_from_slice(master_key).unwrap();
```
ManyTheFish, line 247: this can be unwrapped safely because `Hmac::new_from_slice` never returns an error:
https://docs.rs/hmac/0.12.1/src/hmac/optim.rs.html#152-174

MarinPostma (approved):
Lgtm 

Bors merge

PR #2489, opened 2022-06-08
Fix(auth): Authorization test were not testing keys unrestricted on index

ManyTheFish:


Kerollmops (approved):
ok, c'est bien!
bors merge

PR #2507, opened 2022-06-11
Create a docker tag without patch version if git tag has 0 patch vers…

### LINKED ISSUE #2497: Publish docker image tag without the patch version
Currently, when pushing a tag to this git repository, [this CI](https://github.com/meilisearch/meilisearch/blob/main/.github/workflows/publish-docker-images.yml) is triggered, publishing a docker image with the name of the tag (and `latest` if it's a GitHub release).
It means if the `v0.27.0` tag is pushed, the `v0.27.0` docker image will be available.

We would like to update [this CI](https://github.com/meilisearch/meilisearch/blob/main/.github/workflows/publish-docker-images.yml) to create/update a tag without the patch version when a tag is pushed.

Ex:
- the `v0.27.0` git tag is pushed to the repository -> the `v0.27` and `v0.27.0` tags are available for the docker image and are equivalent.
- then later, the `v0.27.1` git tag is pushed to the repository -> the `v0.27.1` docker tag is pushed, but also the `v0.27` tag is updated for the docker image to correspond to the `v0.27.1` tag and not to the `v0.27.0` anymore

janithPet:
…ion.

Added a line that conditionally adds a version number without the patch version (i.e, `v<major>.<minor>`)  if the version given by the git tag ends with `.0`. 

# Pull Request

## What does this PR do?
Fixes #2497 

## PR checklist

### DIFF
--- a/.github/workflows/publish-docker-images.yml
+++ b/.github/workflows/publish-docker-images.yml
@@ -37,6 +37,7 @@ jobs:
           flavor: latest=false
           tags: |
             type=ref,event=tag
+            type=semver,pattern=v{{major}}.{{minor}},enable=${{ endsWith(github.ref_name, '.0') }}
             type=raw,value=latest,enable=${{ github.event_name == 'release' }}
 
       - name: Build and push


janithPet:
Hello, 

This is my first time contributing to a large project like this! 

I hope I had understood the requirments described in #2497 correctly. I would grateful for any feedback regarding this. 

Thanks!

curquiza:
Hello @janithpet thanks a lot for your contribution, I'll try to review this asap! 😄

.github/workflows/publish-docker-images.yml
```diff
@@ -37,6 +37,7 @@ jobs:
           flavor: latest=false
           tags: |
             type=ref,event=tag
+            type=semver,pattern=v{{major}}.{{minor}},enable=${{ endsWith(github.ref_name, '.0') }}
```
curquiza, line 40: As far as I understand, it will not work if I publish a tag name `v0.27.2` because it does not end with `.0`
Could it be possible to create (maybe with a regexp?) this shorter Docker tag if the git tag finishes by `.` and a number? In the best case, this number could be multiple digits, like `v0.27.2`, but also `v0.27.12`

curquiza (changes requested):
Thanks again for the PR. If I'm correct it will not work in the case of the tag does not finish by `0`.

.github/workflows/publish-docker-images.yml
janithpet: You are right! I had misunderstood the requirements (sorry) -> I thought that it would only be needed if the patch ended with a ".0". 

I have made a new commit that adds an additional workflow step runs the regex matching. It had to be done this way because, (as far as I understand), we can't run pattern matching in an expression. 

The pattern I'm matching to is 'v<number>.<number>.<number>', so that it doesn't catch tags that could look like 'v0.26.1-cloud-team.1', which still end with '.<number>'.

curquiza: Thanks a lot for your reactivity! And the explanations!

curquiza:
Thanks a lot @janithpet for this PR! 😄 
I'll merge it in my temporary branch so that I can do tests and then I will merge my branch on main. Be sure you will be in the release changelogs anyway! Thank you very much! ❤️

janithpet:
> Thanks a lot @janithpet for this PR! 😄 
> 
> I'll merge it in my temporary branch so that I can do tests and then I will merge my branch on main. Be sure you will be in the release changelogs anyway! Thank you very much! ❤️ 

No problem at all! Please let me know if any issues come up 😃

PR #2512, opened 2022-06-13
Allow task abortion

MarinPostma:
This pr introduces the ability to cancel pending tasks.

Two new routes are added to conveniently cancel pending tasks: tasks that aren't processed nor finished. A call to any of those routes will create a new task aborting the requested tasks. This task has the highest priority, meaning that it will be processed as soon as the current processing task is finished.

Aborting an already processing/processed task is a no-op, and the abort task will be marked as successful, even though the requested task has not been aborted.

The newly created routes are:
- `POST /tasks/abort`: takes a JSON array of the task ids to be aborted
- `POST /tasks/clear`: clears the task queue by marking all pending updates as aborted.

on success, an abortion task will return the number of tasks that have been aborted in the `abortedTasks` field of the `details` field of the task view.

When a task is marked as aborted, it is not removed from the scheduler right away. That is because cleaning might be necessary to do so (e.g removing update files for the document updates). Rather than that, the update is left in the scheduler and is ignored by the batch handler. It is the `BatchHandler` job to clean after all the tasks it is handed.

### DIFF
--- a/meilisearch-auth/src/action.rs
+++ b/meilisearch-auth/src/action.rs
@@ -24,6 +24,8 @@ pub enum Action {
     IndexesDelete = actions::INDEXES_DELETE,
     #[serde(rename = "tasks.get")]
     TasksGet = actions::TASKS_GET,
+    #[serde(rename = "tasks.abort")]
+    TasksAbort = actions::TASKS_ABORT,
     #[serde(rename = "settings.get")]
     SettingsGet = actions::SETTINGS_GET,
     #[serde(rename = "settings.update")]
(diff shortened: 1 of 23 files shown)


ShivamJoker:
@MarinPostma any ideas when this is going to be merged I started a big task and I can stop it.

curquiza:
Hello @ShivamJoker, sorry for the confusion, this PR is outdated and will not be merged. 
However we will add this possibility in Meilisearch v0.30.0, you can follow the corresponding issue: https://github.com/meilisearch/meilisearch/issues/2749

Sorry again for the confusion, I close this PR then in favor of the futures PRs associated to https://github.com/meilisearch/meilisearch/issues/2749

PR #2523, opened 2022-06-15
Improve the tasks error reporting when processed in batches

### LINKED ISSUE #2478: Change the behavior of auto-batching error catching
As a user I want the auto-batching feature to have the same behavior as if it wasn't enabled and the updates would have been processed one by one, the auto-batching feature is an optimization that must not change the usage of Meilisearch.

However, the error reporting of the current implementation is a little bit bad, when an error occurs in a batch (a group of tasks that are processed as a whole) the whole set of related tasks is aborted with the same error message. It is an issue when one single malformed document can make the whole task queue fail. If the auto-batching would have not been enabled the malformed document would have made only one task fail and the others pass.

According to this behavior to change, the goal is to make the auto-batching seamless to use, the error reporting must be the same or near the same as when the auto-batching is not enabled. We discussed the way we will implement this new behavior in the engine in https://github.com/meilisearch/milli/issues/546.

### LINKED ISSUE #2582: Autobatching: improve error handler + make it by default
## What is auto-batching?

The auto-batching feature is a feature we introduced in v0.26.0 to improve the indexation speed when users push their data in multiple small batches (ex: 100 batches of 1 document). Pushing the data in small batches instead of one big considerably impacts (negatively) the indexation speed.
That's why we released the auto-batching as an experimental feature: automatically regroup in one big batch the small batches of the users.

The auto-batching feature is currently present as experimental, meaning users have to activate it using `--enable-auto-batching` when launching the Meilisearch binary.

This discussion has been created to gather feedback: https://github.com/meilisearch/meilisearch/discussions/2070

We estimate this feature is necessary to increase the user experience by reducing the indexation time. We now want to provide it by default, I mean, not as an experimental feature anymore.

## Goal of the issue

This issue gathers the changes we need to make with the auto-batching feature in order to remove its "experimental" label:
- [ ] Improve the error handler regarding the auto-batching feature: https://github.com/meilisearch/meilisearch/issues/2478
- [ ] Update the settings related to the feature
  - [ ] Auto-batching is activated by default: `--enable-auto-batching` is `true` by default. If the users don't want to use autobatching they have to pass this value at `false`.
  - [ ] `--debounce-duration-sec` should be remove because currently buggy when using the snapshots
  - [ ] Other (`--max-batch-size` and `--max-documents-per-batch`) can be reconsider: should we keep it? We need the @gmourier validation for this

## Process

💫 This issue will be completed with links and all the necessary information all along the feature creation process

- [ ] Create a prototype by providing a docker image
- [ ] Apply changes according to feedback
- [ ] Create the corresponding specification and link it to this issue

-----

Pinging @meilisearch/docs-team and @meilisearch/cloud-team

Kerollmops:
This fixes #2478 by changing the behavior of the task handler when there is an error in a batch of document addition or update.

What changes is that when there is a user error in a task in a batch we now report this task as failed with the right error message but we continue to process the other tasks. A user error can be when a geo field is invalid, a document id is invalid, or missing.

fixes #2582, #2478


meilisearch-lib/src/dump/error.rs
```diff
@@ -11,7 +11,7 @@ pub enum DumpError {
     #[error("An internal error has occurred. `{0}`.")]
     Internal(Box<dyn std::error::Error + Send + Sync + 'static>),
     #[error("{0}")]
-    IndexResolver(#[from] IndexResolverError),
+    IndexResolver(Box<IndexResolverError>),
```
irevoire, line 14: Why do we need to do that instead of the `#[from]`?

meilisearch-lib/src/options.rs
```diff
@@ -41,27 +41,10 @@ pub struct IndexerOpts {
...
-    /// Debounce duration in seconds
-    ///
-    /// When a new task is enqueued, the scheduler waits for `debounce_duration_sec` seconds for new updates before
-    /// starting to process a batch of updates.
-    #[clap(long, requires = "enable-auto-batching", hide = true)]
-    pub debounce_duration_sec: Option<u64>,
+    /// The engine will disable task auto-batching,
+    /// and will sequencialy compute each task one by one.
+    #[clap(long, env = "DISABLE_AUTO_BATCHING")]
+    pub disable_auto_batching: bool,
```
irevoire, line 47: I don’t remember if we’re supposed to hide this parameter or not? poke @gmourier

meilisearch-lib/src/index/updates.rs
```diff
@@ -323,19 +323,34 @@ impl Index {
...
 
-        txn.commit()?;
+            results.push(user_result);
+        }
 
-        info!("document addition done: {:?}", addition);
+        if results.iter().any(Result::is_ok) {
+            let _addition = builder.execute()?;
+            txn.commit()?;
+        }
```
irevoire, line 351: This is where everything happens, and this little if at the end is actually really important.
Maybe we should add some tests to ensure we never forget how useful it is?
- Push 3 updates that all fails and ensure nothing happened
- Push 2 failed updates + 1 successful update and ensure it really processed while the two other got an error

irevoire (changes requested):
The PR looks really good overall, I think it would be good to add a test on the « skip » of failed tasks that didn’t exist before

meilisearch-lib/src/options.rs
gmourier: It should be visible since we want to stabilize the auto batching by making it activated by default

irevoire (approved):
Nice!

meilisearch-http/src/lib.rs
```diff
@@ -30,9 +29,9 @@ pub static AUTOBATCHING_ENABLED: AtomicBool = AtomicBool::new(false);
 pub fn setup_meilisearch(opt: &Opt) -> anyhow::Result<MeiliSearch> {
     let mut meilisearch = MeiliSearch::builder();
 
-    // enable autobatching?
+    // disable autobatching?
     AUTOBATCHING_ENABLED.store(
-        opt.scheduler_options.enable_auto_batching,
+        opt.scheduler_options.disable_auto_batching,
         std::sync::atomic::Ordering::Relaxed,
     );
```
Kerollmops, line 36: Are we sure about the names and booleans values, here. We changed the boolean but not the name of the global.

ManyTheFish: Yup something where wrong in the display of tasks

irevoire (approved):
bors merge

PR #2530, opened 2022-06-16
Check the version in Cargo.toml before publishing

### LINKED ISSUE #2079: Add a CI job to check release tag = version in Cargo.toml
To avoid this kind of issue in the future: https://github.com/meilisearch/MeiliSearch/issues/2078, we need to prevent a release to be published if the tag of the release is not the same as the one in the Cargo.toml.

Multiple CI workflows publish the release, so we have to stop all of them if this statement is not verified.
- [publish the docker image with the `latest` tag](https://github.com/meilisearch/MeiliSearch/blob/main/.github/workflows/publish-docker-latest.yml)
- [publish the docker image with the tag of the release](https://github.com/meilisearch/MeiliSearch/blob/main/.github/workflows/publish-docker-tag.yml)
- [publish the binaries](https://github.com/meilisearch/MeiliSearch/blob/main/.github/workflows/publish-binaries.yml): windows + linux + macos
- [publish the aarch64 binary](https://github.com/meilisearch/MeiliSearch/blob/main/.github/workflows/publish-crossbuild.yml)
- [publish to Homebrew and APT](https://github.com/meilisearch/MeiliSearch/blob/main/.github/workflows/publish-deb-brew-pkg.yml)

Since we cannot ask a workflow to wait for another workflow in GitHub Action, we have to find another smart solution.

curquiza:
Fixes #2079 

Also
- improves the current docker CI for v0.28.0: the current implementation will make run 2 CI instead of just one for the official release
- move the `is-latest-releaes.sh` script, and update the documentation comment
- fix version of permissive-json-pointer

How to test the script?

```
export GITHUB_REF='refs/tags/v0.28.0'
sh .github/scripts/check-release.sh
```

### DIFF
--- a/.github/scripts/check-release.sh
+++ b/.github/scripts/check-release.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+# Checking if current tag matches the package version
+current_tag=$(echo $GITHUB_REF | tr -d 'refs/tags/v')
+file1='meilisearch-auth/Cargo.toml'
+file2='meilisearch-http/Cargo.toml'
+file3='meilisearch-lib/Cargo.toml'
+file4='meilisearch-types/Cargo.toml'
+file5='Cargo.lock'
+
+file_tag1=$(grep '^version = ' $file1 | cut -d '=' -f 2 | tr -d '"' | tr -d ' ')
+file_tag2=$(grep '^version = ' $file2 | cut -d '=' -f 2 | tr -d '"' | tr -d ' ')
+file_tag3=$(grep '^version = ' $file3 | cut -d '=' -f 2 | tr -d '"' | tr -d ' ')
+file_tag4=$(grep '^version = ' $file4 | cut -d '=' -f 2 | tr -d '"' | tr -d ' ')
+file_tag5=$(grep -A 1 'name = "meilisearch-auth"' $file5 | grep version | cut -d '=' -f 2 | tr -d '"' | tr -d ' ')
+
+if [ "$current_tag" != "$file_tag1" ] || [ "$current_tag" != "$file_tag2" ] || [ "$current_tag" != "$file_tag3" ] || [ "$current_tag" != "$file_tag4" ] || [ "$current_tag" != "$file_tag5" ]; then
+  echo "Error: the current tag does not match the version in package file(s)."
+  echo "$file1: found $file_tag1 - expected $current_tag"
+  echo "$file2: found $file_tag2 - expected $current_tag"
+  echo "$file3: found $file_tag3 - expected $current_tag"
+  echo "$file4: found $file_tag4 - expected $current_tag"
+  echo "$file5: found $file_tag5 - expected $current_tag"
+  exit 1
+fi
+
+echo 'OK'
+exit 0
--- a/.github/scripts/check-release.sh
+++ b/.github/scripts/check-release.sh
@@ -3,24 +3,26 @@
 # check_tag $current_tag $file_tag $file_name
 function check_tag {
   if [ "$1" != "$2" ]; then
-      echo "Error: the current tag does not match the version in $3:"
-      echo "Found $1 - expected $2"
-      exit 1
+      echo "Error: the current tag does not match the version in $3: found $1 - expected $2"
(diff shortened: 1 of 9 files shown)


.github/scripts/check-release.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+# Checking if current tag matches the package version
+current_tag=$(echo $GITHUB_REF | tr -d 'refs/tags/v')
+file1='meilisearch-auth/Cargo.toml'
+file2='meilisearch-http/Cargo.toml'
+file3='meilisearch-lib/Cargo.toml'
+file4='meilisearch-types/Cargo.toml'
+file5='Cargo.lock'
```
irevoire, line 9: Actually, I was wondering, why are we bumping the version of meilisearch-[`auth`, `http`, `lib`, `types`] when the only interesting thing is the meilisearch-lib?

Maybe we should take a decision and decide if we;
1. Bump the version number of ALL the crate in the repo,  including `permissive-json-pointer`.
That would allow us to do a;
```suggestion
# Checking if current tag matches the package version
current_tag=$(echo $GITHUB_REF | tr -d 'refs/tags/v')
files='*/Cargo.toml'
lock_file='Cargo.lock'
```
2. Only bump `meilisearch-http` since it's the one that's going to set the version number of the binary.

---------

I don't know what's the best but here is a few examples of projects that do one or the other;
1. [helix-editor](https://github.com/helix-editor/helix/)
2. [ripgrep](https://github.com/BurntSushi/ripgrep/tree/master/crates), [tauri](https://github.com/tauri-apps/tauri/tree/dev/core), [serde](https://github.com/serde-rs/serde), [actix-web](https://github.com/actix/actix-web), [rust](https://github.com/rust-lang/rust/tree/master/library) (-> they make everything be in 0.0.0), [deno](https://github.com/denoland/deno)

.github/scripts/check-release.sh
```diff
@@ -0,0 +1,28 @@
...
+
+if [ "$current_tag" != "$file_tag1" ] || [ "$current_tag" != "$file_tag2" ] || [ "$current_tag" != "$file_tag3" ] || [ "$current_tag" != "$file_tag4" ] || [ "$current_tag" != "$file_tag5" ]; then
+  echo "Error: the current tag does not match the version in package file(s)."
+  echo "$file1: found $file_tag1 - expected $current_tag"
+  echo "$file2: found $file_tag2 - expected $current_tag"
+  echo "$file3: found $file_tag3 - expected $current_tag"
+  echo "$file4: found $file_tag4 - expected $current_tag"
+  echo "$file5: found $file_tag5 - expected $current_tag"
+  exit 1
+fi
```
irevoire, line 25:
```suggestion

file5=$(grep -A 1 'name = "meilisearch-auth"' $file5 | grep version)

for file in $file1 $file2 $file3 $file4 $file5;
do
    file_tag=$(grep '^version = ' $file | cut -d '=' -f 2 | tr -d '"' | tr -d ' ')
    if [ "$current_tag" != "$file_tag" ]; then
      echo "Error: the current tag does not match the version in package file(s)."
      echo "$file: found $file_tag - expected $current_tag"
      exit 1
    fi
done
```

Maybe we could factorize this code a little bit like that what do you think?

irevoire (changes requested):
I think we should make a decision on my first comment, do we really want to check all the crates or only the one responsible for the version of the binary? Or only the ones starting with `meilisearch-` like you did here.

.github/scripts/check-release.sh
curquiza: You're right, I forgot `permissive-json-pointer`: so far we are doing the `1` solution, so I need to add it.
I will update the code to do solution 1 first, and I will open the discussion with core-team!

.github/scripts/check-release.sh
curquiza: Yeeeees definitely thank you! Sorry I did this too quickly, your refactor is way better

irevoire (approved):
purfect thank you :+1: 
bors merge

PR #2543, opened 2022-06-22
fix all the array on the search get route and improve the tests

### LINKED ISSUE #2527: V0.28.0rc0 attributesToHighlight/attributesToCrop: * not working on GET /search route
**Describe the bug**
When making a request on the `GET` /search routes adding `attributesToHighlight: *` does not create a `_formatted` object.
At the contrary, adding a specific field `attributesToHighlight: title` for example, works correctly.

**To Reproduce**
1. Make a search request using the GET method
2. add `attributesToHighlight: *` as a query parameter
4. _formatted is not created nor is there anything highlighted

**Expected behavior**
`attributesToHighlight` should be taken into account.

**Screenshots**
Screenshot 2022-06-15 at 18 31 45

**Meilisearch version:** [e.g. v0.20.0]
v0.28.0rc0

irevoire:
fix #2527


meilisearch-http/src/routes/indexes/search.rs
```diff
@@ -30,16 +29,16 @@ pub struct SearchQueryGet {
     q: Option<String>,
     offset: Option<usize>,
     limit: Option<usize>,
-    attributes_to_retrieve: Option<CS<StarOr<String>>>,
-    attributes_to_crop: Option<CS<StarOr<String>>>,
```
Kerollmops, line 34: So, it is no more possible to use the `StarOr` helper? 🥹

Kerollmops (changes requested):

meilisearch-http/src/routes/indexes/search.rs
irevoire: The issue with this particular route is that its fields are moved into a really similar structure that handles the post route directly. And thus, it's the other one that takes the wildcard.

Kerollmops: Ho yeah, you are right, thank you for the explanation!

Kerollmops (approved):
This PR looks great, thank you!

curquiza:
bors merge

PR #2544, opened 2022-06-22
Fix content of dump/assets for testing

loiclec:
This is just a change to the content of two .dump files used in the integration tests.

Those files contained settings with the criterion `desc(fame)`, which is invalid
for a v3 or higher dump.

The change took place in the updates.json file inside the decompressed
.dump files. Instances of `desc(field)` or `asc(field)` were changed to 
`field:desc` and `field:asc`.

The tests were (wrongly) passing because the ranking rules were never parsed.


irevoire (approved):
perfect thank you!
bors merge

curquiza:
bors merge

PR #2545, opened 2022-06-22
meilisearch-lib was missing a feature in its cargo.toml

irevoire:
Meilisearch-lib was missing a feature to compile on its own


Kerollmops (approved):
Thank you very much!
bors merge

Issue #2028, opened 2021-12-20
Docker CI: should not display `unknown` as commit SHA and commit date


curquiza:
Since v0.25.0 and the [introduction of the new CI](https://github.com/meilisearch/MeiliSearch/pull/1897), MeiliSearch running on Docker does not display the commit sha and the commit date:

<img width="851" alt="Capture d’écran 2021-12-20 à 17 28 37" src="https://user-images.githubusercontent.com/20380692/146800363-3ae54dc3-c623-453b-978f-2461d927832d.png">

When building locally, I don't have any commit sha and date issue, they are well displayed at MeiliSearch launch.

MarinPostma:
@curquiza, after giving it a quick glance, it seems that [this PR](https://github.com/meilisearch/MeiliSearch/pull/1658) removes the use of env variables for building the image, in favor of importing the `.git` directory. But it seems that we you are still expecting to find these variables in the dockerfile [here](https://github.com/meilisearch/MeiliSearch/blob/release-v0.25.0/Dockerfile#L28-L30), and I see nowhere where the `.git` is actually copied. Not sure how all of that workds, but it may be worth to take a look at that.

amab8901:
I'm unable to reproduce the error for the [latest version of meilisearch](https://hub.docker.com/layers/getmeili/meilisearch/v0.30.0/images/sha256-56d263069d9308a1d75a750f90c3a935b97a37dd888344220ce89f1823f754de?context=explore) on dockerhub. 

<details>
<summary>$ sudo docker run --rm -e MEILI_NO_ANALYTICS=true -p 770:770 getmeili/meilisearch:v0.30.0 ./meilisearch --master-</summary>
```
key=masterKey
Unable to find image 'getmeili/meilisearch:v0.30.0' locally
v0.30.0: Pulling from getmeili/meilisearch
ca7dd9ec2225: Pull complete 
3628dd4a8988: Pull complete 
e5d3cc55b1bb: Pull complete 
b0545ee3798f: Pull complete 
d9faa5b12225: Pull complete 
Digest: sha256:364782ac9f870eefb18259b6c0f4652dcb7c8d375233b6e358a51047a9fc4272
Status: Downloaded newer image for getmeili/meilisearch:v0.30.0
[FATAL tini (7)] exec ./meilisearch failed: No such file or directory

```

</details>

curquiza:
[...]

PR #2550, opened 2022-06-24
Setup COMMIT_SHA and COMMIT_DATE build args in the Docker image

brunoocasali:
Summary of the changes:

- Rename from `check-tag-format` to `build-metadata` because now it does more than just check tag format (I was not sure about creating one step for just assigning a variable).
- Add the `build-args` definition from https://github.com/docker/build-push-action#inputs which is supposed to work exactly as `docker build --build-arg`.
- I'm not sure about the usage of `git` on line 34 because I don't know if it will be available at that point.

Fixes https://github.com/meilisearch/meilisearch/issues/2028

### DIFF
--- a/dump/src/reader/mod.rs
+++ b/dump/src/reader/mod.rs
@@ -299,12 +299,12 @@ pub(crate) mod test {
         assert!(indexes.is_empty());
 
         // products
-        insta::assert_json_snapshot!(products.metadata(), { ".createdAt" => "[now]", ".updatedAt" => "[now]" }, @r###"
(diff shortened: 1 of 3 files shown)


curquiza (changes requested):
Thanks @brunoocasali for the PR

First I would keep 2 separated jobs
- let the `Check tag format` -> this should only run for `github.event_name != 'schedule'`
- another new one, your `Check build metadata format` -> this should run without any condition

Secondly, when I tried to run the CI by pushing a random tag (`test-git-tag`) I got an error:
https://github.com/meilisearch/meilisearch/runs/7388047405?check_suite_focus=true

Issue #2555, opened 2022-06-27
Task uid limit: can not parse "99999999999" to a u32


brunoocasali:
**Describe the bug**
I'm not sure if this "is" a bug but in any case, I would like to have more explanations about the behavior.

I'm working on the PHP repo, and I encountered a bug when trying to make it work in this new version, we have a test to force an error that accesses a really big task id the number is 99.999.999.999 (I've added the dots to make it easier to read).

Looking in the last stable version I could check the task#uid used to support u64. Knowing this is something we don't expect to reach so soon, some users may have to handle this bug in the future.

So the question is why changing to u64 (if it was conscious)

**To Reproduce**
Access a task route: `GET /tasks/99999999999` 

**Expected behavior**
~Access the desired task or an 404 error not a 500.~. 
Access the desired task or a 404 error, not a 404 with a different content-type

**Meilisearch version:** v0.28.0rc1

**Additional context**
Additional information that may be relevant to the issue.
Docker on macOS

curquiza:
Thank you @MarinPostma for the answer ❤️ 
I close this issue then!

brunoocasali:
@MarinPostma I liked your example, but you're considering just a serial execution, not a parallel one correct? Because if we have a multi-threaded environment for example a background job process configured in a rails app like sidekiq, you can exponentially increase the number of tasks per minute.

In a recent job experience I had, we used to send at least 1mm emails per day, if I indexed that + the other data we used to have in background jobs we would reach that in a couple of years.

@curquiza WDYT to open a new issue to handle the parsing error? The content type of the response is not compatible with the request's content type.

MarinPostma:
@brunoocasali all writes are serialized, so that wouldn't change much :)

MarinPostma:
@brunoocasali I mean that writing to disk is single-threaded, so even if you submit tasks concurrently, they will be registered sequentially.

The reason it is done this way is that the filtering of tasks is backed by roaring bitmaps. They have the advantage of being easily serializable to disk, zero-copy reads, efficient set operations, and great compression. The drawback is that we can only use u32s. The advantage of using bitmap largely outbalances that of having u64 here

I understand your concern, but other limits are probably gonna be hit long before this one

curquiza:
@brunoocasali when the PR is merged, the specificiation regarding task error message needs to be updated with the new error message 😇

curquiza:
Closed in favor of https://github.com/meilisearch/meilisearch/issues/2593

PR #2562, opened 2022-06-28
Handle the error when user request an invalid id

brunoocasali:
Context: https://github.com/meilisearch/meilisearch/issues/2555

Even keeping the u32 for performance reasons, the user still can reach a dead end by accessing a route with a big id, if they do that, it will respond with a 404 + internal error with an incorrect content-type.

This PR provides a way to handle this kind of error.

Code tips are appreciated 😬

### DIFF
--- a/meilisearch-http/src/routes/tasks.rs
+++ b/meilisearch-http/src/routes/tasks.rs
@@ -170,7 +170,7 @@ async fn get_tasks(
 
 async fn get_task(
     meilisearch: GuardedData<ActionPolicy<{ actions::TASKS_GET }>, MeiliSearch>,
-    task_id: Option<web::Path<TaskId>>,
+    task_id: web::Path<String>,
     req: HttpRequest,
     analytics: web::Data<dyn Analytics>,
 ) -> Result<HttpResponse, ResponseError> {
@@ -191,7 +191,9 @@ async fn get_task(
         Some(filters)
     };
 
-    let id = task_id.ok_or(TaskError::OutOfBounds)?;
+    let id = task_id
+        .parse::<TaskId>()
+        .map_err(|_| TaskError::OutOfBounds)?;
 
     let task: TaskView = meilisearch.get_task(id.into_inner(), filters).await?.into();
 
--- a/meilisearch-http/src/routes/tasks.rs
+++ b/meilisearch-http/src/routes/tasks.rs
@@ -1,4 +1,5 @@
 use actix_web::{web, HttpRequest, HttpResponse};
+use meilisearch_lib::tasks::error::TaskError;
 use meilisearch_lib::tasks::task::{TaskContent, TaskEvent, TaskId};
 use meilisearch_lib::tasks::TaskFilter;
 use meilisearch_lib::MeiliSearch;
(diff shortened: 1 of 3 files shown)


meilisearch-lib/src/tasks/error.rs
```diff
@@ -14,6 +14,8 @@ pub enum TaskError {
     UnexistingTask(TaskId),
     #[error("Internal error: {0}")]
     Internal(Box<dyn std::error::Error + Send + Sync + 'static>),
+    #[error("Task id is out of bounds.")]
+    OutOfBounds(),
```
MarinPostma, line 18: no parenthesis is necessary here
```suggestion
    OutOfBounds,
```

meilisearch-lib/src/tasks/error.rs
```diff
@@ -29,6 +31,7 @@ impl ErrorCode for TaskError {
         match self {
             TaskError::UnexistingTask(_) => Code::TaskNotFound,
             TaskError::Internal(_) => Code::Internal,
+            TaskError::OutOfBounds() => Code::TaskNotFound,
```
MarinPostma, line 34:
```suggestion
            TaskError::OutOfBounds => Code::TaskNotFound,
```

meilisearch-http/src/routes/tasks.rs
```diff
@@ -190,10 +191,9 @@ async fn get_task(
         Some(filters)
     };
 
-    let task: TaskView = meilisearch
-        .get_task(task_id.into_inner(), filters)
-        .await?
-        .into();
+    let id = task_id.ok_or(TaskError::OutOfBounds())?;
```
MarinPostma, line 194:
```suggestion
    let id = task_id.ok_or(TaskError::OutOfBounds)?;
```

meilisearch-http/src/routes/tasks.rs
```diff
@@ -169,7 +170,7 @@ async fn get_tasks(
 
 async fn get_task(
     meilisearch: GuardedData<ActionPolicy<{ actions::TASKS_GET }>, MeiliSearch>,
-    task_id: web::Path<TaskId>,
+    task_id: Option<web::Path<TaskId>>,
```
MarinPostma, line 173: this is actually very strange. It works, but it's more like a ugly hack. If you really want to catch any kind of overflow, we will want to parse it to a `String` first, and then try to parse a `TaskId` from that

```suggestion
task_id: web::Path<String>,
```

meilisearch-http/src/routes/tasks.rs
MarinPostma, line 194:
```suggestion
    let id = task_id.parse::<TaskId>().map_err(|_| TaskError::OutOfBounds)?;
```

meilisearch-http/src/routes/tasks.rs
brunoocasali: I learned this [here](https://github.com/actix/actix-web/discussions/1986) 🤔

Kerollmops:
I am closing this in favor of #2593, thank you for your contribution @brunoocasali and your review @MarinPostma 🙏

PR #2579, opened 2022-07-05
API keys: adds action * for actions

### LINKED ISSUE #2560: Create an API key with the action with `*` for `indexes`, `settings`, `tasks`, `dumps`, `stats`
Checked with @gmourier, we can let the possibility to the users to set keys with the following actions
- `indexes.*`: all the index actions are available, so `indexes.create`, `indexes.get`, `indexes.update`, and `indexes.delete`
- `settings.*`: all the settings actions are available, so `settings.get`, `settings.update`
- `tasks.*`: all the tasks actions are available, it means only `tasks.get` so far.
- `dumps.*`: all the tasks actions are available, it means only `dumps.get` so far.
- `stats.*`: all the tasks actions are available, it means only `stats.get` so far.

This possibility has already been done for documents (`documents.*`), check out [this PR](https://github.com/meilisearch/meilisearch/pull/2525#issuecomment-1168791835) as a clue for the implementation

## For Meili people

Since we let this issue (so far) to the community, we don't know yet in which release this new feature will be available (at minimum v0.29.0, not before). The specifications already have been changed according to this, see https://github.com/meilisearch/specifications/blob/develop/text/0085-api-keys.md

phdavis1027:
# Pull Request
This is PR builds on @janithpet's addition to DocumentsAll; it's basically a copy-and-paste job, except I used ```iter.filter()``` to avoid the possibility for duplication that they mentioned. I'm not sure how much that matters.

Also, hi! This is my first open-source contribution and my first attempt to write Rust for anyone other than myself, so any feedback whatsoever is appreciated. 

## What does this PR do?
Fixes #2560

## PR checklist


meilisearch-auth/src/store.rs
```diff
@@ -92,13 +92,68 @@ impl HeedAuthStore {
...
+                        && *action != Action::IndexesDelete
+                        && *action != Action::IndexesUpdate
+                })
+                .chain(vec![
+                    Action::IndexesAdd,
+                    Action::IndexesGet,
+                    Action::IndexesDelete,
+                    Action::IndexesUpdate,
+                ])
+                .collect()
```
irevoire, line 126: It was okay-ish to do a contain when there was only one, but now we’re iterating over and over on all the actions, and it’s taking a lot of lines of code.

I thought it would be easier to iterate on all the actions, transform the `.*` into what we want, and let the other as-is.

And finally, to remove the duplicate entry, I would prefer a simpler collect to a HashSet + a conversion to a vec at the very end of the computation.

Something like that;
```rust
let actions = key.actions.iter().flat_map(|action| match action {
    IndexesAll => [IndexesAdd, IndexesGet, IndexesDelete, IndexesUpdate].iter(),
    SettingsAll => todo!(),
    *All => todo!(),
...
    other => std::iter::once(other),
  })
  .collect::<HashSet<Action>>();

let action = actions.into_iter().collect::<Vec<Action>>();
```

(this code probably has issues with the iterator in the match, but you get the idea)

irevoire (changes requested):
Hey @phdavis1027, thanks for your contribution!

It’s overall pretty good, but I would like to simplify a bit the pipeline you copy-pasted, it was good when there was only one case, but now it’s getting pretty big and may be hard to maintain.

I proposed one half-working solution, but if you have other ideas, feel free to tell me 👍

phdavis1027:
Thanks for the pointer! That suggestions looks much cleaner and more idiomatic -- currently looking into getting the ActionEnumIterator to behave nicely in that flat_map.

phdavis1027:
> Thanks for the pointer! That suggestions looks much cleaner and more idiomatic -- currently looking into getting the ActionEnumIterator to behave nicely in that flat_map.

This solution isn't nearly as pretty, but I think it puts food on the table.

irevoire (dismissed):
> Thanks for the pointer! That suggestion looks much cleaner and more idiomatic -- currently looking into getting the ActionEnumIterator to behave nicely in that flat_map.

Yeah, I was afraid it wouldn’t play nicely with that `flat_map`, but your solution looks good to me; thank you!

curquiza:
bors try

phdavis1027:
> @phdavis1027 @irevoire looks like there are tests to fix before merging blush

Embarassed to admit how long it took me to spot that the issue was just a typo lol.

curquiza:
bors try

irevoire (approved):
> Embarassed to admit how long it took me to spot that the issue was just a typo lol.

Ahah that happens! Thanks for your time ❤️ 
bors merge

Issue #2569, opened 2022-06-29
Add new CI: cron GHA to create an issue "Update dependencies"


curquiza:
In Meilisearch we don't want to use Dependabot since we have a loooooot of dependencies and Dependabot updates them one by one

Using the `cron` event in a GitHub Actions, we want a CI that creates a new issue every 3 months in this repository with
- title: "Upgrade dependencies"

- description:
"We need to update the dependencies of the Meilisearch repository, and, if possible, the dependencies of all the core-team repositories that Meilisearch depends on (milli, charabia, heed...).

⚠️ This issue should only be done at the beginning of the sprint!"

- labels: `dependencies` and `maintenance`

VasiliySoldatkin:
If there is no one assigned I'll do it this week

curquiza:
Hello @VasiliySoldatkin 
Thanks for your interest in this project :)

FYI, we prefer not assigning people to our issues because sometimes people ask to be assigned and never come back, which discourages the volunteer contributors to open a PR to fix this issue.
We will accept and merge the first PR that fixes correctly and well implements the issue following our [contributing guidelines](https://github.com/meilisearch/meilisearch/blob/main/CONTRIBUTING.md).   

We are looking forward to reviewing your PR 😊

PR #2585, opened 2022-07-05
Add CI creates issue updating dependencies

VasiliySoldatkin:
Adds CI, which uses cron GHA to create Issue "Upgrade dependencies" every 3 months 
Context: [#2569](https://github.com/meilisearch/meilisearch/issues/2569)


.github/workflows/create-issue-dependencies.yml
```diff
@@ -0,0 +1,16 @@
+name: Create Issue "Upgrade dependencies"
```
curquiza, line 1:
```suggestion
name: Create issue to upgrade dependencies
```

.github/workflows/create-issue-dependencies.yml
```diff
@@ -0,0 +1,16 @@
...
+    - cron: '0 0 1 */3 *'
+  workflow_dispatch:
+  
+jobs:
+  create-issue:
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v1
```
curquiza, line 13:
```suggestion
      - uses: actions/checkout@v3
```

curquiza (changes requested):
Hello @VasiliySoldatkin

Thanks for this PR! This is really helpful
I would rather we don't need to create a new template, otherwise it will be suggested when creating an issue, which is not something we want
I've done some research and I think we could you this GHA: https://github.com/actions-ecosystem/action-create-issue

Thanks again for your PR!! ❤️

curquiza:
👆I've applied directly my small suggestion of change on the branch, so don't forget to `git pull` on your branch before doing any new changes 😄

VasiliySoldatkin:
Thank you @curquiza for review! I made some changes in .yml, is that what you need?

PR #2592, opened 2022-07-07
Chores: Add a dedicated section for Language Support in the issue template

ManyTheFish:
This new section in put upper than the feature proposal because language support is kind of a sub-category of it,
and so, in the reading order, we choose to create a feature proposal only if it is not related to Language.


curquiza (approved):
Thank you @ManyTheFish, definitely helpful

curquiza:
bors merge

Issue #2593, opened 2022-07-07
Improve error messages when parsing URL arguments


Kerollmops:
There are some issues with the parsing errors regarding URL arguments, they are not returned as JSON objects but as single text. We found this when reviewing #2562.

![cannot parse hello](https://user-images.githubusercontent.com/3610253/177791129-fa7d4abe-d4cc-4a7c-aa28-9aad40840a83.png)
![cannot parse 99999](https://user-images.githubusercontent.com/3610253/177791142-3d674ac4-a5d4-49ea-929a-ce0a13455f6d.png)

curquiza:
This issue will be fixed at the same time we fix this one: https://github.com/meilisearch/meilisearch/issues/2325, I mean the Jayson integration to improve the Meilisearch's error messages.

I updated the issue with the requirements

PR #2604, opened 2022-07-12
Improve handling invalid taskIds and return descriptive response

chammika-become:
## Summary
This PR fixes issue #2593 . 
Followed the suggestion made by  @MarinPostma in #2562 to pass `TaskId` as a `String` into the route handler and try parsing it there.

## Example

```bash
❯ curl -s -X GET 'http://localhost:7700/tasks/hello' | jq
{
  "message": "Invalid task id `hello`.",
  "code": "task_not_found",
  "type": "invalid_request",
  "link": "https://docs.meilisearch.com/errors#task_not_found"
}
❯
❯ curl -s -X GET 'http://localhost:7700/tasks/999999999999999999999' | jq
{
  "message": "Invalid task id `999999999999999999999`.",
  "code": "task_not_found",
  "type": "invalid_request",
  "link": "https://docs.meilisearch.com/errors#task_not_found"
}
```

## PR checklist

### DIFF
--- a/meilisearch-lib/src/tasks/error.rs
+++ b/meilisearch-lib/src/tasks/error.rs
@@ -12,6 +12,8 @@ pub type Result<T> = std::result::Result<T, TaskError>;
 pub enum TaskError {
     #[error("Task `{0}` not found.")]
     UnexistingTask(TaskId),
+    #[error("Invalid task id `{0}`.")]
+    InvalidTask(String),
     #[error("Internal error: {0}")]
     Internal(Box<dyn std::error::Error + Send + Sync + 'static>),
 }
@@ -28,6 +30,7 @@ impl ErrorCode for TaskError {
     fn error_code(&self) -> Code {
         match self {
             TaskError::UnexistingTask(_) => Code::TaskNotFound,
+            TaskError::InvalidTask(_) => Code::TaskNotFound,
             TaskError::Internal(_) => Code::Internal,
         }
     }
--- a/meilisearch-lib/src/tasks/error.rs
+++ b/meilisearch-lib/src/tasks/error.rs
@@ -30,7 +30,7 @@ impl ErrorCode for TaskError {
     fn error_code(&self) -> Code {
         match self {
             TaskError::UnexistingTask(_) => Code::TaskNotFound,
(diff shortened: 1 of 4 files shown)


meilisearch-lib/src/tasks/error.rs
```diff
@@ -28,6 +30,7 @@ impl ErrorCode for TaskError {
     fn error_code(&self) -> Code {
         match self {
             TaskError::UnexistingTask(_) => Code::TaskNotFound,
+            TaskError::InvalidTask(_) => Code::TaskNotFound,
```
Kerollmops, line 33: Do we want to return a 404 or more a 400 for malformed query?

Kerollmops (changes requested):
Thank you very much for your work here! 😍

meilisearch-lib/src/tasks/error.rs
gmourier: Since you are requesting a task that will never exist, it seems consistent and ok to return a 404 IMO.

chammika-become: Thanks for the reviews. I honestly didn't give this any thought to the error code. Looking at `InvalidIndexUid` and `InvalidDocumentId` errors I feel 400 is better, with new error `Code::InvalidTaskId`. If you guys agree I will make that change.

Kerollmops: I honestly prefer that we continue to be consistent with the `InvalidIndexUid` and `InvalidDocumentId` errors, thus returning a 400. What do you think @gmourier?

curquiza: FYI @chammika-become, @gmourier and @Kerollmops are on Holiday for a few weeks, so they will answer when they come back 😄

Kerollmops:
Hey @chammika-become,

I am sorry to announce to you that we prefer to close your PR as we would prefer to tackle the subject of _having better error messages_ in another way. Also, we do not like the fact that we are untyping the routes by directly accepting `String` and parsing that by hand on the route side. It increases the amount of code to maintain.

Thank you very much for the time you took to work on that issue, and sorry for closing this PR 😢
I hope you'll continue to help us with other subjects 🙏

PR #2655, opened 2022-08-02
Use mold linker in CIs

curquiza:
Using https://github.com/rui314/setup-mold


curquiza:
Close because it does not impact the performance at all

PR #2657, opened 2022-08-04
prometheus and grafana dashboards implemented

pavo-tusker:
Implemented Basic Prometheus Metrics and Grafana Dashboard using this [Prometheus Crate](https://crates.io/crates/prometheus) [#496](https://github.com/meilisearch/product/issues/496)
Screenshot from 2022-08-04 19-59-06
Screenshot from 2022-08-04 19-58-56
Screenshot from 2022-08-04 19-58-40

### DIFF
--- a/meilisearch-http/src/lib.rs
+++ b/meilisearch-http/src/lib.rs
@@ -6,6 +6,7 @@ pub mod task;
 #[macro_use]
 pub mod extractors;
 pub mod option;
+pub mod route_metrics;
 pub mod routes;
 
 use std::sync::{atomic::AtomicBool, Arc};
@@ -140,56 +141,33 @@ pub fn dashboard(config: &mut web::ServiceConfig, _enable_frontend: bool) {
     config.service(web::resource("/").route(web::get().to(routes::running)));
 }
 
+pub fn configure_metrics_route(config: &mut web::ServiceConfig, enable_metrics_route: bool) {
+    if enable_metrics_route {
+        config.service(web::resource("/metrics").route(web::get().to(routes::get_metrics)));
+    }
+}
+
 #[macro_export]
 macro_rules! create_app {
     ($data:expr, $auth:expr, $enable_frontend:expr, $opt:expr, $analytics:expr) => {{
         use actix_cors::Cors;
         use actix_web::dev::Service;
+        use actix_web::middleware::Condition;
         use actix_web::middleware::TrailingSlash;
         use actix_web::App;
         use actix_web::{middleware, web};
         use meilisearch_http::error::MeilisearchHttpError;
         use meilisearch_http::metrics;
+        use meilisearch_http::route_metrics;
         use meilisearch_http::routes;
-        use meilisearch_http::{configure_data, dashboard};
+        use meilisearch_http::{configure_data, configure_metrics_route, dashboard};
         use meilisearch_types::error::ResponseError;
-
-        use lazy_static::lazy_static;
-        use prometheus::{
-            opts, register_histogram_vec, register_int_counter_vec, register_int_gauge,
-        };
-        use prometheus::{HistogramTimer, HistogramVec, IntCounterVec, IntGauge};
 
         App::new()
             .configure(|s| configure_data(s, $data.clone(), $auth.clone(), &$opt, $analytics))
             .configure(routes::configure)
             .configure(|s| dashboard(s, $enable_frontend))
-            .wrap_fn(|req, srv| {
-                let mut histogram_timer: Option<HistogramTimer> = None;
-                let request_path = req.path();
-                let is_registered_resource = req.resource_map().has_resource(request_path);
-                if is_registered_resource {
-                    let request_method = req.method().to_string();
-                    histogram_timer = Some(
-                        metrics::HTTP_RESPONSE_TIME_SECONDS
-                            .with_label_values(&[&request_method, request_path])
-                            .start_timer(),
-                    );
-                    metrics::HTTP_REQUESTS_TOTAL
-                        .with_label_values(&[&request_method, request_path])
-                        .inc();
-                }
-
-                let fut = srv.call(req);
-
-                async {
-                    let res = fut.await?;
-                    if let Some(histogram_timer) = histogram_timer {
-                        histogram_timer.observe_duration();
-                    };
-                    Ok(res)
-                }
-            })
+            .configure(|s| configure_metrics_route(s, $opt.enable_metrics_route))
             .wrap(
                 Cors::default()
                     .send_wildcard()
@@ -202,6 +180,10 @@ macro_rules! create_app {
             .wrap(middleware::Compress::default())
             .wrap(middleware::NormalizePath::new(
                 middleware::TrailingSlash::Trim,
+            ))
+            .wrap(Condition::new(
+                $opt.enable_metrics_route,
+                route_metrics::RouteMetrics,
             ))
     }};
 }
--- a/meilisearch-http/src/lib.rs
+++ b/meilisearch-http/src/lib.rs
@@ -7,7 +7,6 @@ pub mod task;
 pub mod extractors;
 pub mod option;
 pub mod routes;
-pub mod metrics;
 
 use std::sync::{atomic::AtomicBool, Arc};
 use std::time::Duration;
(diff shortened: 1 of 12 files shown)


pavo-tusker:
@irevoire Fixed the issues in the review.

irevoire:
Hey, I'm off today I'll look into it tomorrow!
But thanks a lot for your contribution it looks sick 🤘

meilisearch-http/src/routes/mod.rs
```diff
@@ -269,3 +272,33 @@ struct KeysResponse {
...
 }
+
+pub async fn get_metrics(
+    meilisearch: GuardedData<ActionPolicy<{ actions::STATS_GET }>, MeiliSearch>,
+) -> Result<HttpResponse, ResponseError> {
+    let search_rules = &meilisearch.filters().search_rules;
+    let response = meilisearch.get_all_stats(search_rules).await?;
+
+    crate::metrics::MEILISEARCH_DB_SIZE.set(response.database_size as i64);
+
```
irevoire, line 283:
```suggestion
```

meilisearch-http/src/lib.rs
```diff
@@ -145,18 +146,52 @@ pub fn dashboard(config: &mut web::ServiceConfig, _enable_frontend: bool) {
...
+                    let request_method = req.method().to_string();
+                    histogram_timer = Some(
+                        metrics::HTTP_RESPONSE_TIME_SECONDS
+                            .with_label_values(&[&request_method, request_path])
+                            .start_timer(),
+                    );
+                    metrics::HTTP_REQUESTS_TOTAL
+                        .with_label_values(&[&request_method, request_path])
+                        .inc();
+                }
```
irevoire, line 183: From what I see/understand, Prometheus is using [`Atomic`](https://docs.rs/prometheus/latest/src/prometheus/atomic64.rs.html) to update its state.
Most access are relaxed, but the `inc` is an `Acquire` for example (which can be slower than a mutex iirc).

And this code is executed for EVERY call to ANY route.

Maybe it could slow down mielisearch when it’s under high pressure?
What do you think @meilisearch/core-team?

Personally, I still think we should accept this PR, my question is, should it work like that, or should we put the feature behind a cli args so it only slows down those who need prometheus?

meilisearch-http/src/routes/mod.rs
```diff
@@ -269,3 +272,33 @@ struct KeysResponse {
...
+    }
+
+    let encoder = TextEncoder::new();
+    let mut buffer = vec![];
+    encoder
+        .encode(&prometheus::gather(), &mut buffer)
+        .expect("Failed to encode metrics");
+
+    let response = String::from_utf8(buffer.clone()).expect("Failed to convert bytes to string");
+    buffer.clear();
```
irevoire, line 299:
```suggestion
```

We’re not using this buffer later, we can let it be dropped later

meilisearch-http/src/routes/mod.rs
```diff
@@ -269,3 +272,33 @@ struct KeysResponse {
...
+            .set(value.number_of_documents as i64);
+    }
+
+    let encoder = TextEncoder::new();
+    let mut buffer = vec![];
+    encoder
+        .encode(&prometheus::gather(), &mut buffer)
+        .expect("Failed to encode metrics");
+
+    let response = String::from_utf8(buffer.clone()).expect("Failed to convert bytes to string");
```
irevoire, line 298:
```suggestion
    let response = String::from_utf8(buffer).expect("Failed to convert bytes to string");
```
Not sure but since we don’t need the buffer later we can simply give it to the string without cloning it maybe?

irevoire (changes requested):
Hey, @pavo-tusker thank you for your contribution it’s awesome, I want to see what the rest of the @meilisearch/core-team thinks of it before merging it.

But overall great contribution, thanks!

meilisearch-http/src/lib.rs
ManyTheFish: Good question @irevoire, I agree with you that this PR is a great improvement.

I like the idea of setting an opt-in flag for this feature as we made for the auto-batching beta feature.
Moreover, this feature didn't pass the usual feature process and didn't have any approval from the product team.

So, I'd find it safer to deactivate by default this feature. 

But again, it's a great feature that could be useful for some of our users and for us. 😊

davelarkan: This PR looks great and would be incredibly helpful for us on the cloud side.

Having this be an opt-in feature would be fine (we would likely enable it by default for all projects)

gmourier: @ManyTheFish It seems it could be a strong basis for a more [detailed exporter](https://github.com/meilisearch/product/issues/496) that is considered for the future.

irevoire:
Ok @pavo-tusker, so after chatting with the team we would really like to integrate your change but with the option to disable it.
Could you create an `--enable-metrics-route` cli option as @ManyTheFish said that disables both the `metrics` route and the hook in actix (=the `wrap_fn`).

irevoire:
Hey, just so you know, [clippy was updated](https://github.com/meilisearch/meilisearch/pull/2665) last weekend thus you'll need to rebase on main to make the CI pass again.

pavo-tusker:
> Hey, just so you know, [clippy was updated](https://github.com/meilisearch/meilisearch/pull/2665) last weekend thus you'll need to rebase on main to make the CI pass again.

Hey, @irevoire  Just added the " --enable-metrics-route " option as per the above discussions. By default, metrics will be disabled. passing "--enable-metrics-route" arg will enable /metrics route and Req/Res metrics. Removed wrap_fn and changed it as a conditional wrap middleware. 

and what do I need to do on Clippy? can you give some more details about this?

meilisearch-http/src/option.rs
```diff
@@ -146,6 +146,10 @@ pub struct Opt {
     #[clap(long, env = "MEILI_LOG_LEVEL", default_value = "info")]
     pub log_level: String,
 
+    // Enables Prometheus metrics and /metrics route.
+    #[clap(long, requires = "enable-metrics-route")]
```
irevoire, line 150:
```suggestion
    /// Enables Prometheus metrics and /metrics route.
    #[clap(long)]
```

irevoire (changes requested):
Hey @pavo-tusker we're almost there!

After accepting my last change you will need to rebase on main (there should be no conflict from what I see) and I’ll be able to merge your change!
It should be something like that depending on your workflow.
```
git checkout main
git pull upstream/main
git checkout -
git rebase main
git push -f
```

irevoire:
bors merge

meilisearch-http/src/option.rs
pavo-tusker: Hey, I did rebase the main to the feature branch. Still,
Screenshot from 2022-08-17 21-09-57
 the Clippy is failing.

irevoire: It's super strange; I can't reproduce locally.
But if you can go on, just fix it here and make it stop.

If you're not familiar with Clippy;
You can run `cargo clippy --fix` to automatically fix most of the warnings.

Then don't forget to run `cargo fmt` and you should be good!

irevoire (dismissed):
Thanks for your contribution!
bors merge

pavo-tusker:
> Thanks for your contribution! bors merge

Again there is an issue with clippy :( on new bors merged code.
Just fixed that too.

irevoire (approved):
yeah I don't know what's happening, I'll try again and if it doesn't work we're probably going to bypass bors or remove clappy from the CI 🤔 
bors merge

irevoire:
Hey @pavo-tusker, finally, we decided we wanted to take more time to refine the feature, thus for the next release (happening today), we're going to move the metrics behind a feature flag (which requires you to build your own meilisearch binary in order to use it).

For the next weeks, we’re going to heavily use it and see what other information we can provide, and if, in the end, we are satisfied with the outcome, we’ll probably make it official for the v1.0 (in 4 months) :tada: 
We’ll see how it goes, but I can't promise anything.

I would love to hear your use case, though; why did you feel the need to implement this route, and is it ok for you to recompile meilisearch from scratch in the meantime?

pavo-tusker:
> Ease observability meilisearch/product#496

Sure, 🫡. I was exploring meilisearch for our needs, and observability is something I felt helped understand the queries we did, so I added the same and thought I would make a  pull request to contribute. Using as a binary now.
Would be great to see the merge when it happens, let me know if anything is needed from me.

PR #2663, opened 2022-08-11
Compute the real size of indexes

irevoire:
When returning the size used by an index, instead of returning the size on disk we return the real size used by meilisearch without the cached pages contained in heed


curquiza:
@irevoire can we close this?

PR #2689, opened 2022-08-23
Use mimalloc as the global allocator

loiclec:
milli has switched its global allocator to mimalloc already, and we have seen some performance gains as a result. Furthermore, we can use mimalloc as the global allocator on all platforms whereas jemalloc was only activated on Linux. 

This PR brings mimalloc to Meilisearch as well.


Kerollmops (approved):
That is a nice improvement, thank you very much!
bors merge

PR #2690, opened 2022-08-23
Add LTO and codegen-units=1 to release compile options

loiclec:
This PR brings Meilisearch's release compile options in line with milli (see https://github.com/meilisearch/milli/pull/606 ). 

Adding LTO and codegen=units=1 will make compile times longer, but they also speed up the final binary significantly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,10 @@ members = [
     "permissive-json-pointer",
 ]
 
+[profile.release]
+codegen-units = 1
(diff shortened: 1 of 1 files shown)


irevoire (approved):
noice

Kerollmops (approved):
Release builds will take more time to build, but we don't care we will have fast binaries!
bors merge

curquiza:
How significantly? 😇

PR #2697, opened 2022-08-24
Accept an environment variable to enable the metrics route

Kerollmops:
With the PR Meilisearch is able to accept the `MEILI_ENABLE_METRICS_ROUTE` environment variable to enable the newly introduces metrics route.


ManyTheFish (approved):
bors merge

PR #2715, opened 2022-08-29
Implement tasks cancellation

Kerollmops:
This PR implements a basic task cancellation system, it only works for already processed tasks (for now?):
 - [x] Introduce the `Tasks.cancel` action right for the API Keys.
 - [x] You can cancel one task at a time by calling the `DELETE /indexes/tasks/{taskId}`.
 - [x] Remove the tasks content when canceled.


Kerollmops:
I am closing this PR in favor of #2763, in which we are rewriting the task scheduler to support task cancelation.

Issue #2718, opened 2022-08-29
Enable LTO again


curquiza:
We first enable LTO in this PR https://github.com/meilisearch/meilisearch/pull/2690 to improve the performance of the search engine.
Then, we had to disable LTO in this PR https://github.com/meilisearch/meilisearch/pull/2717 because we were not able to compile the `aarch64` binary anymore.

We would introduce LTO again and prevent using it when compiling the `aarch64` binary.
Some beginning of solution [here](https://github.com/meilisearch/meilisearch/pull/2717#issue-1354487056)

If anyone has a better solution, we would happy to read you!

zamazan4ik:
Any updates on this issue? :)

curquiza:
Hello @zamazan4ik 
The issue is still open and relevant for our project. Feel free to help us on it if you have any clue 😊

zamazan4ik:
I can just reintroduce changes from the https://github.com/meilisearch/meilisearch/pull/2690/files PR. Is it okay for the project? At the very least we could check, if LTO still fails for the `aarch64` platform or not.

curquiza:
I'm not sure to understand what you want to do exactly. Because we already tried changes on #2690 and we already know it does not work with `aarch64` platform
Did you think about something else?

zamazan4ik:
> I'm not sure to understand what you want to do exactly. Because we already tried changes on https://github.com/meilisearch/meilisearch/pull/2690 and we already know it does not work with aarch64 platform

There are several things to consider since the https://github.com/meilisearch/meilisearch/pull/2690: 
* It was more than 2 years ago. This bug (I suppose it was a bug on the compiler side) could already be fixed in Rustc
* As far as I can see, no one has tried to use Fat LTO instead of Thin LTO. Switching from Thin to Fat also could help with resolving the issue (if it still appears)

curquiza:
Feel free to try what you think is the best @zamazan4ik and we are ready to review 😊

zamazan4ik:
@curquiza here we go! https://github.com/meilisearch/meilisearch/pull/5106

PR #2726, opened 2022-08-31
Add dry run for publishing binaries: check the compilation works

curquiza:
To avoid realizing the compilation of one type of binary does not work during the release, I create a dry run for binary compilation every day at 2am 😇 

See the problem we had recently because missing this CI: https://github.com/meilisearch/meilisearch/issues/2718


Kerollmops (approved):
Nice, thank you!
bors merge

Issue #2680, opened 2022-08-18
Improve the error message when adding documents


irevoire:
Currently, when adding json documents to meilisearch, you can either add them as a single document;
```bash
curl -X POST 'http://localhost:7700/indexes/mieli/documents'  -H 'Content-type: application/json' -d '{ "id": 1, "pet": "dog" }'
```
Or as an array of documents;
```bash
curl -X POST 'http://localhost:7700/indexes/mieli/documents'  -H 'Content-type: application/json' -d '[{ "id": 1, "pet": "dog" }, { "id": 2, "pet": "doggo" }]'
```

That's cool, but the error messages are not explicit at all;
```bash
curl -X POST 'http://localhost:7700/indexes/mieli/documents'  -H 'Content-type: application/json' -d '"sending a string instead of a json"'
```
For example, with this request we get this error message;
```
The `json` payload provided is malformed. `Couldn't serialize document value: data did not match any variant of untagged enum Either`.
```

I would like to improve this message and get rids entirely of the notion of « untagget enum » and « Either ».

onyxcherry:
Have you encountered any other meaningless messages?

That issue doesn't seem to will be resolved by the serde team (as they have a low priority on it).

Although the message could be improved by implementing own `Deserialize` for `ArrayOrSingleObject`, isn't it better to simply replace the message:

```rust
let mut serde_msg = se.to_string();
if serde_msg == "data did not match any variant of untagged enum Either" {
    serde_msg = "data are neither an object nor a list of objects".to_string();
}
```
?

inkss:
I encountered the same file upload error, how can I fix it?

```sh
curl -X POST 'https://xxxx/indexes/xxxx/documents/?primaryKey=xxxx' \
  -H "Authorization: Bearer ${{secrets.MEILISEARCH_KEY}}" \
  -H "Content-Type: application/json" \
  --data-binary 'content.json'
```

``` json
{
    "message": "The `json` payload provided is malformed. `Couldn't serialize document value: expected value at line 1 column 1`.",
    "code": "malformed_payload",
    "type": "invalid_request",
    "link": "https://docs.meilisearch.com/errors#malformed_payload"
}
```

onyxcherry:
> I encountered the same file upload error, how can I fix it?
> 
> ```sh
> curl -X POST 'https://xxxx/indexes/xxxx/documents/?primaryKey=xxxx' \
[...]

PR #2727, opened 2022-08-31
Don't panic when the error length is slightly over 100

onyxcherry:
# Pull Request

## What does this PR do?
Fixes PR #2207 as [the last commit](https://github.com/meilisearch/meilisearch/pull/2207/commits/7ece7a9d9ee3015a84f3b19f6b7360b7099b5388) has changed number of the characters at the end to leave in place from `50` to `85` **but the lower limit of a string length wasn't changed**.
Therefore, any data (e.g. example string from issue #2680) was causing `meilisearch` to **panic**.

So I simply raised the minimum value from `100` to `135` (`50 + 85`) to ensure that `replace_range()` won't panic due to an inverted range.
At the same time I am in favor of the `85` value which was changed in the @CNLHC's last commit.

## PR checklist
- [x] Does this PR fix an existing ~issue~ pull request?

### DIFF
--- a/meilisearch-lib/src/document_formats.rs
+++ b/meilisearch-lib/src/document_formats.rs
@@ -47,7 +47,8 @@ impl Display for DocumentFormatError {
                     let trim_input_prefix_len = 50;
                     let trim_input_suffix_len = 85;
 
-                    if serde_msg.len() > trim_input_prefix_len + trim_input_suffix_len + ellipsis.len()
+                    if serde_msg.len()
+                        > trim_input_prefix_len + trim_input_suffix_len + ellipsis.len()
                     {
                         serde_msg.replace_range(
                             trim_input_prefix_len..serde_msg.len() - trim_input_suffix_len,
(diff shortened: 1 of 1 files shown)


Kerollmops (changes requested):
Wow, thank you very much for your work! Fixing a bug we weren't even aware of Ahah!
Would it be possible to declare both of the numbers in variables just above, please?

meilisearch-lib/src/document_formats.rs
```diff
@@ -136,16 +143,23 @@ pub fn read_json(input: impl Read, writer: impl Write + Seek) -> Result<usize> {
...
     for object in content.inner.map_right(|o| vec![o]).into_inner() {
+        println!("{:#?}", object);
         builder
             .append_json_object(&object)
             .map_err(Into::into)
             .map_err(DocumentFormatError::Internal)?;
     }
 
     let count = builder.documents_count();
+    println!("{count}");
```
Kerollmops, line 161: Would you please remove the `println`s?

Kerollmops (changes requested):

meilisearch-lib/src/document_formats.rs
onyxcherry: Yes, obviously. 
Very sorry about that 😅

Kerollmops (approved):
Thank you very much, LGTM!
bors merge

PR #2733, opened 2022-09-01
Move if conditions to the steps and not to the whole job

curquiza:
Follows https://github.com/meilisearch/meilisearch/pull/2726
Fixes the CI. Moves the `if:` condition to the steps and not the whole `check-version` jobs, otherwise, the following jobs checking the compilation cannot run.
See: https://github.com/meilisearch/meilisearch/actions/runs/2968543407


Kerollmops (approved):
I can only trust you 🤷‍♂️
bors merge

PR #2735, opened 2022-09-01
Metrics

irevoire:


irevoire:
Hey,

I'm not sure, but there is a lot of code that we might be able to reuse if we decide to use Prometheus in the future.
Maybe we could keep it until we take a decision on the metrics?

curquiza:
Ok for me!

irevoire:
Prometheus is now used in prod

PR #2744, opened 2022-09-05
Minor fixes in the just added update-version CI

curquiza:
These fixes does not prevent us to use the current CI


ManyTheFish (approved):
bors merge

PR #2759, opened 2022-09-07
Bump milli to 0.33.3

### LINKED ISSUE #2743: MDB_BAD_VALSIZE: Unsupported size of key/DB name/data, or wrong DUPFIXED size.
**Describe the bug**
When we decided to test a new version of MeiliSearch on our forum (XenForo CMS with custom addon) we have began importing all documents to it. So, when the indexation has began, we have ran into some MeiliSearch internal error: MDB_BAD_VALSIZE.

**To Reproduce**
I really have no idea why it happens. We don't have such issue on MeiliSearch 0.28.1, but have in any 0.29 versions. I thought that problem could occur due to auto-batching feature, tried to turn it off but still. Tried to set indexation threads count to 0 but whatever, the problem is still is. Another interesting fact is that this problem doesn't happen when sending problem request manually via curl. I can attach (problem?) request body if you want.

**Expected behavior**
Indexation process completes successfully.

**Screenshots**
No screenshots. I can reproduce this problem and record a video if you want.

**Meilisearch version:**
[v0.29.0rc2](https://github.com/meilisearch/meilisearch/releases/tag/v0.29.0rc2)
Problem occurs on any 0.29 version.

**Additional context**
WSL2, Windows 11

Kerollmops:
This PR fixes #2743.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1139,8 +1139,8 @@ dependencies = [
 
 [[package]]
 name = "filter-parser"
-version = "0.33.2"
-source = "git+https://github.com/meilisearch/milli.git?tag=v0.33.2#f7c352a32db002fdd6aa7f5ce544043c0acca202"
+version = "0.33.3"
+source = "git+https://github.com/meilisearch/milli.git?tag=v0.33.3#549fa12d5a52ce6cee821ffcd89dc8d465f24b42"
 dependencies = [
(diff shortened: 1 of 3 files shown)


curquiza (approved):
Thanks

bors merge

Kerollmops:
bors merge

Issue #2725, opened 2022-08-31
Refactoring the task queue and scheduler


Kerollmops:
In this issue, we simply list the requirements for a new task queue. The current one is pretty complex and adding new functionalities is cumbersome. We would like to design a new scheduler system that fixes and improves over the current one.

For that, we must evaluate the time we need to take to develop most of the features below:

- [x] Durability of the tasks once an answer has been sent to the user.
- [x] Disable the auto-batching system.
- [ ] Cancel one or _multiple_ enqueued or processing tasks.
- [ ] Remove already processed tasks from the task store.
   - [ ] don't delete tasks with an id higher or equal to the deleting task
   - [ ] use bitmaps for the matched task ids
- [ ] Swapping indexes.
- [ ] Support different task types:
    - [ ] Import dumps at startup time.
    - [ ] Export dumps i.e. export the tasks queue.
    - [x] Document addition (replace or update)
    - [x] Document deletion.
    - [x] Clear all documents.
    - [x] Update the settings of an index.
    - [x] Index update e.g. Create an Index, change an index name, delete an index.
    - [ ] Cancel a processing dump.
- [ ] Auto-batching of enqueued tasks:
    - [x] Document addition.
    - [x] Document updates.
    - [ ] Settings.
    - [ ] Document deletion.
- [ ] Prioritize and reorder tasks i.e. Index deletion, Delete all the documents.
- [x] Dump the tasks queue itself.
- [x] Always accept new tasks without blocking.
- [x] Fairly share the loads over the different indexes e.g. Always process the index queue with the lowest id.
- [ ] Easily testable.
- [ ] Well tested i.e. tasks reordering, tasks prioritizing, use atomic barriers to block the tasks for tests.
- [ ] Realtime progressing status e.g. Websocket events.
- [x] List and filter tasks:
    - [x] Paginate the tasks.
    - [x] Filter by index name.
    - [x] Filter on the status, the enqueued, processing, and finished tasks.
    - [x] Filter on the type of task.

Kerollmops:
We have done a meeting yesterday (@irevoire, @loiclec, and myself) to evaluate the time it would take to design the new scheduler designed to support all of the features listed above and with the less code complexity possible.

We decided that we will use LMDB as the task and index name mapper store. The task ids are incremental `u32`s. As a value, we have all the information about the task (see below). When the task type is a document addition, a file with the content is created and stored outside of the LMDB store.

[...]

PR #2763, opened 2022-09-08
Index scheduler

irevoire:
Fix https://github.com/meilisearch/meilisearch/issues/2725

- [x] Durability of the tasks once an answer has been sent to the user.
- [x] Fix the analytics
- [x] Disable the auto-batching system.
- [x] Make sure the task scheduler run if there are tasks to process.
- [x] Auto-batching of enqueued tasks:
    - [x] Do not batch operations from two different indexes.
    - [x] Document addition.
    - [x] Document updates.
    - [x] Settings.
    - [x] Document deletion.
    - [x] Make sure that we only merge batches with the same index-creation rights:
        - [x] the batch either starts with a `yes`
        - [x] [we only batch `no`s together and stop batching when we encounter a `yes`](https://www.youtube.com/watch?v=O27mdRvR1GY)
        - [x] Unify the logic about `false` and `true` index creation rights.
- [ ] Execute all batch kind:
    - [x] Import dumps at startup time.
    - [x] Export dumps i.e. export the tasks queue.
    - [x] Document addition
    - [x] Document update
    - [x] Document deletion.
    - [x] Clear all documents.
    - [x] Update the settings of an index.
    - [ ] Merge multiple settings into a single one.
    - [x] Index update e.g. Create an Index, change an index primary key, delete an index.
    - [x] Cancel enqueued or processing tasks (with filters) (don't count tasks from forbidden indexes) (can't cancel a task with a higher or equal task_id than your own).
    - [x] Delete processed tasks from the task store (with filters) (don't count tasks from forbidden indexes) (can't flush a task with a higher or equal task_id than your own)
    - [x] Document addition + settings
    - [x] Document addition + settings + clear all documents
    - [x] anything + index deletion
    - [x] Snapshot
       - [x] Make the `SnapshotCreation` task visible.
       - [x] Snapshot tasks are scheduled by a detached thread.
       - [x] Only include update files that are useful.
    - [x] Check that statuses and details are correctly set. (ie; if you enqueue a `documentAddition`, is the `documentReceived` well set?)
- [x] Prioritize and reorder tasks i.e. Index deletion, Delete all the documents.
- [x] Always accept new tasks without blocking.
- [x] Fairly share the loads over the different indexes e.g. Always process the index queue with the lowest id.
- [x] Easily testable.
- [x] Well tested i.e. tasks reordering, tasks prioritizing, use atomic barriers to block the tasks for tests.
- [x] Dump
    - [x] Serialize the uuid as string in the keys
    - [x] Create a dump crate with getters and setters
    - [x] Serialize the API key in the dump task
    - [x] Get the instance-uuid in the dump task
- [x] List and filter tasks:
    - [x] Paginate the tasks.
    - [x] Filter by index name.
    - [x] Filter on the status, the enqueued, processing, and finished tasks.
    - [x] Filter on the type of task.
    - [x] Check that it works in `meilisearch-http`.
- [x] Think about [the index wrapper](https://github.com/meilisearch/meilisearch/blob/2c4c14caa8997f1034c1535875dcc9021fed1873/index/src/updates.rs#L269) and probably move or remove it.
- [x] Reduce the amount of copy/paste for the batched operations by creating a sub-enum for the `Batch` enum.
- [x] Move the `IndexScheduler` in the lib.rs file.
- [x] Think about the `MilliError` type and probably remove it.
- [x] Remove the `index` crate entirely
- [x] Remove the `Kind` type from the `TaskView` and introduce another type, remove the `<Kind as FromStr>`.
- [x] Once the point above is done; remove the unreachable variant from the autobatchingkind
- [x] Rename the `Settings` task `Kind` to `SettingsUpdate`
- [x] Rename the `DumpExport` task `Kind` to `DumpExport`
- [x] Path the error message when deserializing a `Kind` and `Status`.
- [x] Check the version file when starting.
- [x] Copy the version file when creating snapshots.

---------

Once everything above is done;
- [x] Check what happens with the update files i.e. when are they deleted.
    - [x] When a TaskDeletion occurs
    - [x] When a TaskCancelation
    - [x] When a task is finished
    - [x] When a task fails
- [ ] When importing a dump forward the date to milli
- [ ] Add tests for the snapshots.
- [ ] Look at all the places where we put _TODOs_.
- [ ] Rename a bunch of things, see https://github.com/meilisearch/meilisearch/pull/2917
- [ ] Ensure that when compiling meilisearch-http with `no-default-features` it doesn’t pull lindera etc
- [ ] Run a bunch of operations in a `tokio::spawn_blocking`
    - [ ] The search requests
- [ ] Issue to create once this is merged:
    - [ ] Realtime progressing status e.g. Websocket events (optional).
    - [ ] Implement an `Uuid` codec instead of using a `Bincode<Uuid>`.
    - [ ] Handle the dump-v1
    - [ ] When importing a dump v1 we could iterate over the whole task queue to find the creation and last update date
    - [ ] When importing a dump v2 we could iterate over the whole task queue to find the creation and last update date
    - [ ] When importing a dump v3 we could iterate over the whole task queue to find the creation and last update date
    - [ ] When importing a dump v4 we could iterate over the whole task queue to find the creation and last update date
    - [ ] When importing a dump v5 we could iterate over the whole task queue to find the creation and last update date


Kerollmops (approved):
bors merge

Kerollmops (approved):
bors merge

Kerollmops (dismissed):
bors merge

Kerollmops (approved):
bors merge

kyrias:
FWIW this broke building with the `metrics` feature because `meilisearch/src/route_metrics.rs` still uses `meilisearch_auth::actions`.

curquiza:
Hello @kyrias thanks for the report, I opened an issue here: https://github.com/meilisearch/meilisearch/issues/3469

Since the engine team's priority is to maintain and improve the final binary we provide in the [release section](https://github.com/meilisearch/meilisearch/releases), we will not spend time fixing this for the moment, but we are fully available to review and integrate any contribution 😊

PR #2772, opened 2022-09-14
Improve issue template display to avoid support in Meilisearch issues

curquiza:
Move the support line higher to make it more visible
Like this: https://github.com/meilisearch/meilisearch/discussions/2780


gmourier (approved):

curquiza:
bors merge

PR #2773, opened 2022-09-14
Allow building without specialized tokenizations

### LINKED ISSUE #2774: Reduce size of the meilisearch binary
Meilisearch binary has over 100 MiB, which is quite huge. About 90% (!) of the size is made up of specialized tokenizations for Chinese, Hebrew, Japanese, and Thai – they bundle large dictionaries. Many Meilisearch users don’t need support for these languages, so it would make sense to make them optional.

jirutka:
Fixes #2774

(Some of) these specialized tokenizations include huge dictionaries that currently account for 90% (!) of the meilisearch binary size.

This commit adds `chinese`, `hebrew`, `japanese`, and `thai` feature flags that are propagated via `milli` down to the `charabia` crate. To keep it backwards compatible, they are enabled by default.

Related to meilisearch/milli#632

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1119,8 +1119,8 @@ dependencies = [
 
 [[package]]
 name = "filter-parser"
-version = "0.33.0"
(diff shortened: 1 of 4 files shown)


curquiza:
bors merge

PR #2775, opened 2022-09-15
Bump lmdb-rkv-sys to the latest revision

jirutka:
See meilisearch/lmdb-rs#10

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1903,7 +1903,7 @@ dependencies = [
 [[package]]
 name = "lmdb-rkv-sys"
 version = "0.15.0"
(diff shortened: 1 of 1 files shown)


Kerollmops (commented):
Hey @jirutka,

I am currently working on [a new version of heed](https://github.com/meilisearch/heed/pull/128) that removes most of the issues we had in the previous versions of it. Therefore, I would prefer that we don't work on anything related to LMDB until I bring the new version in milli/meilisearch.

curquiza:
Thanks for the PR @jirutka anyway, I saw you had made some of them in different repos for a couple of days! Thanks for your involvement in the whole project, it's really appreciated 😄

PR #2812, opened 2022-09-28
Add support for logging to syslog

### LINKED ISSUE #2811: Add support for logging to local syslog
Meilisearch HTTP currently logs all messages to the stderr output. When Meilisearch is run as a system service, users usually want to aggregate logs from all system services and reroute them as needed (to different files, remote log servers etc.). Unix systems typically use syslog for this.

jirutka:
Fixes #2811

If `MEILI_SYSLOG` is set or `--syslog` provided, the log messages will be sent to the local syslog instead of the stderr. The log level can be controlled using `MEILI_LOG_LEVEL` (or `--log-level`) as with logging to the stderr, but the value must be a valid syslog level, otherwise the program will panic.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1097,6 +1097,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "error-chain"
+version = "0.12.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2d2f06b9cac1506ece98fe3231e3cc9c4410ec3d5b1f24ae1c8946f0742cdefc"
+dependencies = [
+ "version_check",
+]
+
+[[package]]
 name = "fastrand"
 version = "1.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


jirutka:
The Clippy error is unrelated to this PR:

```
error: unnecessary closure used with `bool::then`
   --> meilisearch-lib/src/tasks/task_store/mod.rs:118:29
    |
118 |               Some(filter) => filter
    |  _____________________________^
119 | |                 .pass(&task)
120 | |                 .then(|| task)
    | |__________________------------^
    |                    |
    |                    help: use `then_some(..)` instead: `then_some(task)`
```

curquiza:
Hello again @jirutka 👋 

For traceability, more information about the state of this PR in this comment: https://github.com/meilisearch/meilisearch/issues/2811#issuecomment-1262166541

jirutka:
BTW, this is already used in [meilisearch](https://pkgs.alpinelinux.org/packages?name=meilisearch) package on Alpine Linux, to provide users with good integration to the system.

jirutka:
You could have just asked me to update it… I’ve already rebased it for the Alpine Linux package, just haven’t pushed it yet.

curquiza:
We don't know yet if we will definitely merge it one day. That's I don't want to keep it open for nothing. We will reuse the branch if we finally want to integrate it like you implemented it.

jirutka:
Rebased patch: https://github.com/jirutka/meilisearch/commit/e1c3a1f7ccd6d2a891b2360630c0999aa4bb7295

PR #2818, opened 2022-09-29
Implement the IndexScheduler basic operations

Kerollmops:
This PR implements the basic batched operations:
  - `DocumentClear`
  - `DocumentImport` (it now has a `method` field which is either `Replace` or `Update`)
  - `SettingsAndDocumentImport`
  - `DocumentDeletion`
  - `Settings`
  - `DocumentClearAndSetting`

It introduces an `IndexOperation` sub-enum that allows having a clean recursive system which reduces the number of lines of code to write when processing multiple index-related operations.

This PR also exported the write transaction to allow us to execute multiple operations on a single index atomically e.g. batched settings and document additions.

### DIFF
--- a/index-scheduler/src/batch.rs
+++ b/index-scheduler/src/batch.rs
@@ -12,15 +12,10 @@ pub(crate) enum Batch {
     Cancel(Task),
     Snapshot(Vec<Task>),
     Dump(Vec<Task>),
-    DocumentAddition {
+    DocumentImport {
         index_uid: String,
         primary_key: Option<String>,
-        content_files: Vec<Uuid>,
-        tasks: Vec<Task>,
-    },
-    DocumentUpdate {
-        index_uid: String,
-        primary_key: Option<String>,
+        method: IndexDocumentsMethod,
         content_files: Vec<Uuid>,
         tasks: Vec<Task>,
     },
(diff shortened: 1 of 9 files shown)


index-scheduler/src/batch.rs
```diff
@@ -431,27 +376,129 @@ impl IndexScheduler {
...
+                    IndexOperation::DocumentImport { ref index_uid, .. }
+                    | IndexOperation::Settings { ref index_uid, .. }
+                    | IndexOperation::DocumentClearAndSetting { ref index_uid, .. }
+                    | IndexOperation::SettingsAndDocumentImport { ref index_uid, .. } => {
+                        // create the index if it doesn't already exist
+                        let mut wtxn = self.env.write_txn()?;
+                        let index = self.index_mapper.index(&mut wtxn, index_uid)?;
+                        wtxn.commit()?;
+                        index
+                    }
```
irevoire, line 396:
```suggestion
                        // TODO: check if the user was allowed to create an index.
                        
                        // create the index if it doesn't already exist
                        let mut wtxn = self.env.write_txn()?;
                        let index = self.index_mapper.index(&mut wtxn, index_uid)?;
                        wtxn.commit()?;
                        index
                    }
```

Sorry, that's something I forgot to forward in the `IndexOperation` / `Batch`, but we must keep it somewhere.
Also, maybe we should not batch tasks from a user that can create an index and a user that can't?
For now, I think this `TODO` is enough, but at some point, we'll need to look into that.

irevoire (changes requested):

irevoire (approved):

PR #2838, opened 2022-10-03
Index scheduler Clean Ups

Kerollmops:
This PR will clean up the index scheduler by removing some useless tasks and moving some struct from here and there.

Be careful reviewer. I have used the `index_scheduler::Error` struct everywhere with the `.map_err(...)` combo, it is not good, and I don't know what to do about it.


irevoire (approved):
noice

PR #2852, opened 2022-10-04
Fix milestone-workflow.yml

curquiza:
Using the right repo


curquiza:
Merging manually, does not impact the code and I need to move quickly

PR #2884, opened 2022-10-10
Small changes on the Index scheduler

Kerollmops:
This PR introduces a test that check that we are processing tasks that were sent very fast (and triggered only one signal) and we cleaned-up some code.

### DIFF
--- a/index-scheduler/src/lib.rs
+++ b/index-scheduler/src/lib.rs
@@ -138,6 +138,9 @@ pub struct IndexScheduler {
     /// Get a signal when a batch needs to be processed.
     pub(crate) wake_up: Arc<SignalEvent>,
 
+    /// Weither autobatching is enabled or not.
+    pub(crate) autobatching_enabled: bool,
+
     // ================= test
     /// The next entry is dedicated to the tests.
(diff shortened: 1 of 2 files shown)


index-scheduler/src/lib.rs
```diff
@@ -604,6 +603,43 @@ mod tests {
...
+        index_scheduler
+            .register(KindWithContent::IndexDeletion {
+                index_uid: S("doggos"),
+            })
+            .unwrap();
+
+        handle.wait_till(Breakpoint::Start);
+        handle.wait_till(Breakpoint::AfterProcessing);
+        handle.wait_till(Breakpoint::AfterProcessing);
+        handle.wait_till(Breakpoint::AfterProcessing);
```
irevoire, line 633:
```suggestion
        handle.dont_block();
```

I'm not sure, but I think that's what you're trying to do?

irevoire (changes requested):

index-scheduler/src/lib.rs
Kerollmops: Not sure about that!

PR #2907, opened 2022-10-13
Implement task deletion, continued

loiclec:
This PR continues the work of https://github.com/meilisearch/meilisearch/pull/2885 .

1. Renamed `DeleteTasks` to `TaskDeletion`
2. `TaskDeletion` stores its matched tasks as a `RoaringBitmap`
   * added a function to get a roaring bitmap of task ids from a `Query`
   * this required serialising a roaring bitmap with serde -> need to use `roaring = "0.10.0"` -> need to depend on the `update-dependencies` branch of milli
3. Use `meili-snap` for snapshot tests
4. Fix an infinite loop caused by processing `TaskDeletion` tasks even if they were not enqueued
5. Make `limit` in `Query` optional (because it is not used in `TaskDeletion`)
6. Fix a bug where we'd consider the list of all tasks to be `0 .. last_task_id`. That is not correct, since some tasks between `0` and `last_task_id` may have been deleted.
7. Return a `TaskView` instead of `Task` in the `GET /tasks` route
8. Implement the `DELETE /tasks` route
   * it is not unit-tested or integration-tested, but the basics seem to work
   * the date filter is not implemented
9. Optimised task deletion by deleting them all at once when possible
10. Use `yaup` to serialise the Query in the details of the `TaskDeletion`
    * the query that is saved there is the "filtered" query (the one that doesn't allow deletion of tasks in inaccessible indexes for the API key), not the original one


index-scheduler/src/batch.rs
```diff
@@ -816,49 +820,52 @@ impl IndexScheduler {
...
+            })?;
+        }
+        for kind in affected_kinds {
+            self.update_kind(wtxn, kind, |bitmap| {
+                *bitmap -= &to_delete_tasks;
+            })?;
+        }
+        for task in to_delete_tasks.iter() {
+            self.all_tasks.delete(wtxn, &BEU32::new(task))?;
+        }
```
irevoire, line 868: This is pretty nice :+1:

index-scheduler/src/lib.rs
```diff
@@ -242,47 +242,68 @@ impl IndexScheduler {
...
+            for status in [
+                Status::Enqueued,
+                Status::Processing,
+                Status::Succeeded,
+                Status::Failed,
+            ] {
+                all_tasks |= self.get_status(&rtxn, status)?;
+            }
+            all_tasks
+        };
```
irevoire, line 261: Do you think it's a good idea to create a new bitmap to store the whole available tasks? (in another PR)

meilisearch-auth/src/action.rs
```diff
@@ -0,0 +1 @@
+
```
irevoire, line 1: This file should be deleted :eyes:

meilisearch-http/src/routes/tasks.rs
```diff
@@ -170,6 +175,60 @@ pub struct TasksFilterQuery {
...
+
+    let filtered_query = filter_out_inaccessible_indexes_from_query(&index_scheduler, &query);
+
+    let tasks = index_scheduler.get_task_ids(&filtered_query)?;
+    let filtered_query_string = yaup::to_string(&filtered_query).unwrap();
+    let task_deletion = KindWithContent::TaskDeletion {
+        query: filtered_query_string,
+        tasks,
+    };
+    // TODO: Lo: analytics
```
irevoire, line 224:
```suggestion
```

Currently, no analytics have been defined by gmourier I think.
We'll see later when he gives us directions on what we want to send.

meilisearch-types/src/keys.rs
```diff
@@ -224,6 +224,8 @@ pub enum Action {
     IndexesDelete,
     #[serde(rename = "tasks.*")]
     TasksAll,
+    #[serde(rename = "tasks.*")]
```
irevoire, line 227:
```suggestion
    #[serde(rename = "tasks.delete")]
```

meilisearch-types/Cargo.toml
```diff
@@ -9,9 +9,10 @@ actix-web = { version = "4.0.1", default-features = false }
 csv = "1.1.6"
 either = { version = "1.6.1", features = ["serde"] }
 enum-iterator = "0.7.0"
-milli = { git = "https://github.com/meilisearch/milli.git", tag = "v0.33.0" }
+milli = { git = "https://github.com/meilisearch/milli.git", branch = "update-dependencies" }
 proptest = { version = "1.0.0", optional = true }
 proptest-derive = { version = "0.3.0", optional = true }
+roaring = { version = "0.10.0", features = ["serde"] }
```
irevoire, line 15: noice

meilisearch-http/src/routes/tasks.rs
```diff
@@ -288,8 +351,42 @@ async fn get_task(
...
     if let Some(task) = index_scheduler.get_tasks(filters)?.first() {
-        Ok(HttpResponse::Ok().json(task))
+        let task_view = TaskView::from_task(&task);
+        Ok(HttpResponse::Ok().json(task_view))
     } else {
         Err(index_scheduler::Error::TaskNotFound(task_id).into())
     }
 }
+
+fn filter_out_inaccessible_indexes_from_query<const ACTION: u8>(
```
irevoire, line 361: In the future I think we could call this function in a lot of places

irevoire (changes requested):

index-scheduler/src/lib.rs
Kerollmops: Here it looks like we are enumerating the whole set of variants of `Status` right? Shouldn't it be better to [use enum-iterator](https://lib.rs/crates/enum-iterator) or not?

meilisearch-http/src/routes/tasks.rs
Kerollmops: Wow, then we should probably think about a shorter name!

index-scheduler/src/lib.rs
irevoire: Yep why not, and we need it for the error message as well thus, it's not even a new dependency.

meilisearch-types/Cargo.toml
irevoire: Because now we store a bitmap in the `Details` or `KindWithContent` I already forgot, and we need to write this in heed with the `SerdeJson` codec.

index-scheduler/src/lib.rs
```diff
@@ -737,25 +773,63 @@ mod tests {
...
+                    query: "test_query".to_owned(),
+                    tasks: RoaringBitmap::from_iter(&[0]),
+                })
+                .unwrap();
+        }
+        for _ in 0..2 {
+            handle.wait_till(Breakpoint::AfterProcessing);
+        }
+        // The three deletion tasks are marked as succeeded, and all their details say that one
+        // task has been deleted. Is this the correct behaviour? Probably not!
```
irevoire, line 831: Yes, only the first task should be able to say that.
The two other deletion tasks should say they effectively matched one task to delete but deleted 0.

index-scheduler/src/batch.rs
```diff
@@ -816,49 +820,56 @@ impl IndexScheduler {
...
-                self.update_index(wtxn, index, |bitmap| {
-                    bitmap.remove(task.uid);
-                })?;
+        for task_id in to_delete_tasks.iter() {
+            if let Some(task) = self.get_task(wtxn, task_id)? {
+                if let Some(task_indexes) = task.indexes() {
+                    affected_indexes.extend(task_indexes.into_iter().map(|x| x.to_owned()));
+                }
+                affected_statuses.insert(task.status);
+                affected_kinds.insert(task.kind.as_kind());
```
irevoire, line 852: You should write an else stmt where you effectively count how may tasks has been deleted

irevoire (changes requested):

index-scheduler/src/batch.rs
loiclec: ouch, yep, I meant to do that but completely forgot :/

irevoire (approved):
Nice, thank you @loiclec

PR #2917, opened 2022-10-17
Document `index_scheduler` a bit + rename some of its types

loiclec:
1. Added some documentation to the modules, functions, and types in `index_scheduler`
2. Renamed `KindWithContent` to `TaskOperation`. This is the only place where we use the term "operation", so I don't think it would cause any confusion.
3. Renamed `AutobatchKind` to `SimplifiedTaskOperation` since its only purpose is to remove extraneous information from `TaskOperation
4. Renamed `BatchKind` to `AutoBatch` since it is essentially a `Batch` created by the autobatcher


loiclec:
Closed in favour of https://github.com/meilisearch/meilisearch/pull/2936 + another future PR

PR #2921, opened 2022-10-18
Fix import resource leak

### LINKED ISSUE #2878: Importing dumps on Windows giving os error 32
**Describe the bug**
After the recent update (0.29) I decided to try using dumps when upgrading.
I'm running into an issue where I get this error `Error: The process cannot access the file because it is being used by another process. (os error 32)`  when using `--import-dump`

I've verified that the file is read and writeable and it's still giving me the error.

**To Reproduce**
1. Create a dump from 0.28
2. Import the created dump to the new 0.29 version
3. This error should pop up after trying to import: `Error: The process cannot access the file because it is being used by another process. (os error 32)`

**Expected behavior**
I expect the import to be completed and not fail

**Screenshots/Logs**
Full log of the import: https://pastebin.com/tnVB5v6f

**Meilisearch version:** [e.g. v0.20.0]
Dump version: 0.28.0
Import dump version: 0.29.0

**Additional context**
OS: Windows 10 and Windows Server 2019

mdonoughe:
# Pull Request

## Related issue
Fixes #2878 

## What does this PR do?
After importing a database dump, the environment must be closed for two reasons:
1. If the dump is renamed while it is open, this allows the environment to be opened multiple times because the code that prevents this is based on the file path.
2. Windows does not normally allow a file to be renamed while it is open.

Unfortunately, there's a problem with the design of the `env` module in heed such that if you open a file and then drop the `Env` object, the environment will not be closed unless you try to open the same environment again with the same options and then call `prepare_for_closing`.

I noticed there are some places in meilisearch that do `if Arc::strong_count(&self.env) == 1 { self.env.as_ref().clone().prepare_for_closing(); }`, but not all places. This pattern is not thread safe and it's easy to mess up because the last code to execute must own the last reference to the `Arc`. With the current state of the code, that means `TaskStore::load_dump` needs to carefully manage the lifetime of it's `env` variable. If you just put the `if` statement in a `Drop` implementation for `TaskStore` and also at the bottom of `load_dump` it will not work. The last code in `load_dump` will run while there are two references, then the `drop` implementation will run and there will still be two references. The reference count will only reach 1 after Rust finishes dropping `TaskStore` as part of leaving `load_dump`, after executing the condition added to end of `load_dump`, and then the last reference will be dropped without executing the handler. There's also a problem because `Env` is `Clone` and this allows there to be more references to the environment than there are counted in the `Arc`. This pattern should probably be removed, but I worry just deleting it will cause additional problems.

I have a version of heed where dropping the last `Env` without calling `prepare_for_closing` on that `Env` or another pointing to the same path causes the environment to be closed, but the automated tests demonstrate a problem where multiple threads opening and closing the same environment without calling `prepare_for_closing` used to work and no longer does because the next thread to open the same path while `EnvInner::drop` is executing gets a `DatabaseClosing` error. Setting the tests to run single threaded would also fix it, but obviously any real code using threads the same way would have the same problem. I guess I could make it block until the close completes and open a new PR to heed if that's good.

## PR checklist

### DIFF
--- a/meilisearch-http/tests/dumps/mod.rs
+++ b/meilisearch-http/tests/dumps/mod.rs
@@ -8,7 +8,6 @@ use self::data::GetDump;
 
 // all the following test are ignored on windows. See #2364
 #[actix_rt::test]
-#[cfg_attr(target_os = "windows", ignore)]
 async fn import_dump_v1() {
     let temp = tempfile::tempdir().unwrap();
 
@@ -31,7 +30,6 @@ async fn import_dump_v1() {
 }
 
 #[actix_rt::test]
-#[cfg_attr(target_os = "windows", ignore)]
 async fn import_dump_v2_movie_raw() {
     let temp = tempfile::tempdir().unwrap();
 
@@ -95,7 +93,6 @@ async fn import_dump_v2_movie_raw() {
 }
 
 #[actix_rt::test]
-#[cfg_attr(target_os = "windows", ignore)]
 async fn import_dump_v2_movie_with_settings() {
     let temp = tempfile::tempdir().unwrap();
 
@@ -159,7 +156,6 @@ async fn import_dump_v2_movie_with_settings() {
 }
 
 #[actix_rt::test]
-#[cfg_attr(target_os = "windows", ignore)]
 async fn import_dump_v2_rubygems_with_settings() {
     let temp = tempfile::tempdir().unwrap();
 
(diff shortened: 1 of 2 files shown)


curquiza:
Hello @mdonoughe 

Thank you so much for this PR! I discussed with the core team, and it helps us a lot 🙏 

Since it's an important and impactful PR, we are not sure we will have the time to review and merge it before the end of the hacktoberfest. That's why I will add the `hactkoberfest-accepted` label on it since it's definitely a valid job on our side!

Also, since you contributed to improve our project, if you would like to receive some gift from Meilisearch too, please complete [this form](https://meilisearch.typeform.com/to/idt5qugY).

irevoire:
Hey @mdonoughe!
We finally merged our huge refactor which fixes the issue in another way but anyway **you**’re the one who understood why it was failing before so thanks a lot!

Now as you can [see](https://github.com/meilisearch/meilisearch/pull/2993) we’re able to uncomment almost all the ignored tests.

Since we removed the leak I’m going to close the issue.

PR #2926, opened 2022-10-18
Index scheduler cancel tasks

Kerollmops:
This PR has been rebased on the [*index-scheduler-delete-persisted-file](https://github.com/meilisearch/meilisearch/tree/index-scheduler-delete-persisted-file) branch and should not be merged for now.

What does this PR do:
 - It introduces the tasks cancelation operation that can be triggered by using the `POST /tasks/cancel?uid=1,2,3&status=enqueued&type=documentAdditionOrUpdate&indexUid=movies` route.
 - It can cancel processing or enqueued tasks and will change a `must_stop` boolean internally to inform a possible indexation to stop as quickly as possible.
- It adds a new `canceledBy` field to the `TaskView` and the `DumpTaskView`.
- It also supports dumping tasks with a canceled status and the `canceledBy` field.

We need to add more tests, but overall it looks good.


index-scheduler/src/lib.rs
```diff
@@ -381,11 +424,21 @@ impl IndexScheduler {
         match wtxn.commit() {
             Ok(()) => (),
             _e @ Err(_) => {
-                todo!("remove the data associated with the task");
+                self.delete_persisted_task_data(&task)?;
                 // _e?;
```
irevoire, line 428: We should throw the error after that

index-scheduler/src/lib.rs
```diff
@@ -606,6 +676,25 @@ impl IndexScheduler {
...
+            | KindWithContent::IndexDeletion { .. }
+            | KindWithContent::IndexCreation { .. }
+            | KindWithContent::IndexUpdate { .. }
+            | KindWithContent::IndexSwap { .. }
+            | KindWithContent::TaskCancelation { .. }
+            | KindWithContent::TaskDeletion { .. }
+            | KindWithContent::DumpExport { .. }
+            | KindWithContent::Snapshot => Ok(()),
+        }
+    }
```
irevoire, line 697:
```suggestion
    if let Some(content_file) = task.content_uuid() {
        self.delete_update_file(*content_file)?;
    }
    Ok(())
```

We have a function for that

index-scheduler/src/utils.rs
```diff
@@ -2,24 +2,18 @@
...
-            Status::Processing,
-            Status::Succeeded,
-            Status::Failed,
-        ] {
-            all_tasks |= self.get_status(&rtxn, status)?;
-        }
-        Ok(all_tasks)
+        enum_iterator::all()
+            .map(|s| self.get_status(&rtxn, s))
+            .r#union()
```
irevoire, line 14:
```suggestion
            .union()
```

We don't need anything specific to handle the handle the `union` word as a method?

irevoire (changes requested):

index-scheduler/src/lib.rs
```diff
@@ -117,6 +119,40 @@ impl Query {
...
+}
+
+impl ProcessingTasks {
+    /// Stores the currently processing tasks, the date time at which it started
+    /// and resets the _must stop_ flag.
+    fn start_processing_at(&mut self, started_at: OffsetDateTime, processing: RoaringBitmap) {
+        self.started_at = started_at;
+        self.processing = processing;
+        self.must_stop.store(false, Relaxed);
+    }
```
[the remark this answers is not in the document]
Kerollmops: We can't do so as the date must be known by the `tick` method for when it finishes (step 2.).

index-scheduler/src/utils.rs
Kerollmops: But it breaks my editor syntax highlighting 😢 Ok... I'll remove that...

PR #2932, opened 2022-10-19
Implement task date filters

loiclec:
This PR implements the following filters in a task query:
1. `beforeEnqueuedAt`
2. `afterEnqueuedAt`
3. `beforeStartedAt`
4. `afterStartedAt`
5. `beforeFinishedAt`
6. `afterFinishedAt`

The date is parsed using rfc3993 or iso8601.

Three new databases were created whose key/value pairs are:
- key: a datetime
- value: the tasks which were enqueuedAt/startedAt/finishedAt that time (99.999999999999% of the time it is only a single task, but I made it a `CboRoaringBitmap` to be sure)


meilisearch-http/src/routes/tasks.rs
```diff
@@ -462,3 +536,126 @@ fn filter_out_inaccessible_indexes_from_query<const ACTION: u8>(
...
+            let query = serde_json::from_str::<TaskDeletionQuery>(json).unwrap();
+            snapshot!(format!("{:?}", query.dates.after_enqueued_at.unwrap()), @"1997-11-12 9:55:06.0 -06:00:00");
+        }
+        {
+            let json = r#" { "afterEnqueuedAt": "1997-11-12T09:55:06.000000000Z" } "#;
+            let query = serde_json::from_str::<TaskDeletionQuery>(json).unwrap();
+            snapshot!(format!("{:?}", query.dates.after_enqueued_at.unwrap()), @"1997-11-12 9:55:06.0 +00:00:00");
+        }
+    }
+}
```
irevoire, line 661: Last issue, the date needs to always be exclusive

irevoire (changes requested):
Pretty nice overall

meilisearch-http/src/routes/tasks.rs
```diff
@@ -180,26 +180,79 @@ impl From<Details> for DetailsView {
...
+    )]
+    after_started_at: Option<OffsetDateTime>,
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        serialize_with = "time::serde::rfc3339::option::serialize",
+        deserialize_with = "rfc3339_date_or_datetime::deserialize"
+    )]
+    before_started_at: Option<OffsetDateTime>,
+
```
Kerollmops, line 214:
```suggestion
```

meilisearch-http/src/routes/tasks.rs
```diff
@@ -180,26 +180,79 @@ impl From<Details> for DetailsView {
...
 #[derive(Deserialize, Debug)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub struct TasksFilterQuery {
     #[serde(rename = "type")]
-    type_: Option<CS<StarOr<Kind>>>,
+    kind: Option<CS<StarOr<Kind>>>,
+    uid: Option<CS<u32>>,
     status: Option<CS<StarOr<Status>>>,
-    index_uid: Option<CS<StarOr<IndexUid>>>,
+    index_uid: Option<CS<StarOr<String>>>,
```
Kerollmops, line 238: Why have we changed that?

index-scheduler/src/utils.rs
```diff
@@ -158,3 +183,73 @@ impl IndexScheduler {
         Ok(())
     }
 }
+
+pub(crate) fn insert_task_datetime(
+    wtxn: &mut RwTxn,
+    database: Database<OwnedType<BEI128>, CboRoaringBitmapCodec>,
+
```
Kerollmops, line 190:
```suggestion
```

index-scheduler/src/utils.rs
```diff
@@ -158,3 +183,73 @@ impl IndexScheduler {
...
+    let mut task_ids = if let Some(existing) = database.get(&wtxn, &timestamp)? {
+        existing
+    } else {
+        RoaringBitmap::new()
+    };
+    task_ids.insert(task_id);
+    database.put(wtxn, &timestamp, &RoaringBitmap::from_iter([task_id]))?;
+    Ok(())
+}
+pub(crate) fn remove_task_datetime(
```
Kerollmops, line 204:
```suggestion

pub(crate) fn remove_task_datetime(
```

Kerollmops (changes requested):
I am not sure that it works well when the field is `afterEnqueuedAt` and `beforeEnqueuedAt` should not give the same `OffsetDateTime` for the input of `2021-12-03` for example. The former should give `2021-12-03 0:00:00.0 +00:00:00` and the later `2021-12-04 0:00:00.0 +00:00:00`.

meilisearch-http/src/routes/tasks.rs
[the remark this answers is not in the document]
loiclec: IndexUID is a String wrapper with additional validation rules that are not applied when deserialised via serde if I understood it correctly. It made it a little bit harder to collect all the index ids and give to the query, which expects a String.

loiclec: See https://github.com/meilisearch/meilisearch/pull/2932#discussion_r999520318 , and also for consistency with the way I had implemented it for TaskDeletion

meilisearch-http/src/routes/tasks.rs
loiclec: done, I hope :)

meilisearch-http/src/routes/tasks.rs
```diff
@@ -536,65 +536,149 @@ fn filter_out_inaccessible_indexes_from_query<const ACTION: u8>(
...
+                DeserializeDateOption::After => {
+                    let datetime = datetime
+                        .checked_add(Duration::days(1))
+                        .ok_or(serde::de::Error::custom("date overflow"))?;
+                    Ok(datetime)
+                }
+            }
+        } else {
+            Err(serde::de::Error::custom(
+                "could not parse a date with the RFC3339 or YYYY-MM-DD format",
```
Kerollmops, line 578: Did you notice [that YYYY-MM-DD is part of both](https://ijmacd.github.io/rfc3339-iso8601/)?

meilisearch-http/src/routes/tasks.rs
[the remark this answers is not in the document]
irevoire: Ah I see...
Actually I think we should use the `IndexUid` everywhere but let’s do this later

irevoire (approved):
Good to me, we can merge if @Kerollmops is ok with it

Kerollmops (approved):
Looks good to me but have you checked [this document that explains what we should support (at the bottom)](https://www.figma.com/file/5QihlofGxxJnRXNOldaCLM/EXECUTE---Solution-Evaluation---Delete-Tasks?node-id=0%3A1)?

meilisearch-http/src/routes/tasks.rs
loiclec: The `time` crate only parses a fully specified `YYYY-MM-DDTHH:MM:SS(Z|[+-]HH:MM)` for rfc3339, which seems to be the closely interpretation of the specification ( https://tools.ietf.org/html/rfc3339#section-5.6 ) to me.

loiclec: But yes, I saw the Figma document and believe that the implementation of the deserialiser follows it. Regarding the error message, maybe we should say "a full date and time with the RFC3993 format or a date with YYYY-MM-DD format".

Kerollmops: No it is ok, I was just wondering if you saw the document or you were doing that by memory :)

Kerollmops (approved):
We can merge this @irevoire

PR #2935, opened 2022-10-20
Remove key from index_tasks database when the value is empty

loiclec:
When there are no more tasks associated with an index name, delete the key from the `index_tasks` DB.


irevoire (approved):
noice as well

Kerollmops (approved):

PR #2937, opened 2022-10-20
Make sure that we don't delete or cancel future tasks

loiclec:
This should already have been the case before, but there is no harm in adding another check.


irevoire (commented):
Hummm not sure I see the point of that.
I think I would prefer if we did this double check in the `register` before writing the bitmap on disk, what do you think? 🤔

loiclec:
Yes, we could do it in the register as well. I'll update the PR. As for the purpose of the check, I think it's really not clear when looking at the code that the task ids in `matched_tasks` are `< task.uid`, so I don't want to rely on it for the correctness of the implementation.

irevoire (approved):
Perfect thank you!

PR #2949, opened 2022-10-24
Implement /swap-indexes — continued

loiclec:
1. Change the route from `/indexes-swap` to `/swap-indexes`
2. Change the payload so the field is called `swap` instead of `indexes`
3. Add two custom error messages for the route
4. Fix auth errors


meilisearch-http/src/routes/swap_indexes.rs
```diff
@@ -0,0 +1,131 @@
...
+            }
+            Ok(())
+        }
+    }
+    impl std::error::Error for IndexesNotFoundError {}
+    impl ErrorCode for IndexesNotFoundError {
+        fn error_code(&self) -> Code {
+            Code::IndexNotFound
+        }
+    }
```
irevoire, line 101: Could you delete this custom struct and directly define everything in the `meilisearch-http/src/error.rs` module?
Also, instead of redifining the `IndexNotFound` error, maybe we could call the one from the `IndexScheduler` so we don’t duplicate the message for now?

We already have this kind of error which looks pretty similar to me;
```rust
    #[error("A Content-Type header is missing. Accepted values for the Content-Type header are: {}",
            .0.iter().map(|s| format!("`{}`", s)).collect::<Vec<_>>().join(", "))]
    MissingContentType(Vec<String>),
```


And for the `indexes` we’re already

meilisearch-http/src/routes/swap_indexes.rs
```diff
@@ -0,0 +1,131 @@
...
+                )?;
+                for index in self.indexes.iter().skip(1) {
+                    write!(f, ", `{}`", index)?;
+                }
+                write!(f, "were specified several times.")?;
+            }
+
+            Ok(())
+        }
+    }
```
irevoire, line 124: Same it would be nice if we could use our already defined error type with maybe two more variants to handle both cases?

irevoire (changes requested):

irevoire (approved):

PR #2950, opened 2022-10-24
Add some tests to the index scheduler and fix a few bugs

loiclec:
There are a couple of changes in this pull request.

## Add manual failure points for testing purposes
This looks like this in the main code:
```rust
   pub(crate) fn create_next_batch(&self, rtxn: &RoTxn) -> Result<Option<Batch>> {
        #[cfg(test)]
        self.maybe_fail(crate::tests::FailureLocation::InsideCreateBatch)?;
        // ...
   }
```
and this in the tests:
```rust
  // this index scheduler will have a simulated failure on the first call to `tick` inside the `create_batch` method
  let (index_scheduler, handle) =
            IndexScheduler::test(true, vec![(1, FailureLocation::InsideCreateBatch)]);
```
The point is to be able to test that the index scheduler handles exceptional failures well.

## Refactor test breakpoints
Breakpoints in the main code are now called via a helper function that sends two messages in the channel. This is to ensure that the index scheduler never proceeds ahead of the tests. 

## Resilience improvement 1
When an irrecoverable error occurs, the index scheduler will pause for one second before starting the next iteration of the run loop. This is to avoid wasting CPU cycles when it's extremely likely that every call to `tick` will quickly return the same error as last time. 

## Resilience improvement 2
If a task's content file fails to be deleted, an error message is logged, but the task is still marked as succeeded and no error is thrown.

## Resilience improvement 3
`process_batch` is run inside its own thread so that panics can be caught and be prevented from crashing the index scheduler itself.

## Test improvement
A new method is available on `IndexScheduler` in `#[cfg(test)]` called `assert_internally_consistent()`. It ensures that the content of the index scheduler is correct according **if** the content of its `all_tasks` database is correct. This is a best effort at the moment, and stricter checks could still be performed.

## Bug Fix: details of the IndexSwap task 
An index swap task's details are now correctly updated by future index swap tasks.

## Bug fix: `get_task_ids` handle processing tasks correctly
Previously, the `get_task_ids` implementation completely ignored the fact that processing tasks are not stored in the database. Now it does.


irevoire (approved):

index-scheduler/src/lib.rs
```diff
@@ -378,37 +421,93 @@ impl IndexScheduler {
...
+                };
+                let start = map_bound(bounds.0, |b| b.unix_timestamp_nanos());
+                let end = map_bound(bounds.1, |b| b.unix_timestamp_nanos());
+                let is_within_dates = RangeBounds::contains(
+                    &(start, end),
+                    &started_at_processing.unix_timestamp_nanos(),
+                );
+                if !is_within_dates {
+                    filtered_processing_tasks.clear();
+                }
```
Kerollmops, line 497: Instead of using a loop trick to be able to early return, why not putting that in a closure and calling it for the three cases? Or why not using an optional instead?

```
let x = match y {
    good => Some(y),
    bad => None,
};

if let Some(y) = x {
    // I do the thing here
}
```

index-scheduler/src/lib.rs
```diff
@@ -243,7 +248,44 @@ pub struct IndexScheduler {
     /// The next entry is dedicated to the tests.
     /// It provide a way to break in multiple part of the scheduler.
     #[cfg(test)]
-    test_breakpoint_sdr: crossbeam::channel::Sender<Breakpoint>,
+    test_breakpoint_sdr: crossbeam::channel::Sender<(Breakpoint, bool)>,
```
Kerollmops, line 251: Could you please explain what this bool is for, here?

index-scheduler/src/utils.rs
```diff
@@ -256,6 +256,22 @@ pub fn swap_index_uid_in_task(task: &mut Task, swap: (&str, &str)) {
         K::TaskCancelation { .. } | K::TaskDeletion { .. } | K::DumpExport { .. } | K::Snapshot => {
         }
     };
+    match &mut task.details {
+        Some(details) => match details {
+            Details::IndexSwap { swaps } => {
```
Kerollmops, line 261:
```suggestion
        Some(Details::IndexSwap { swaps }) => 
```

index-scheduler/src/lib.rs
```diff
@@ -323,34 +373,27 @@ impl IndexScheduler {
...
+                    // Wait one second when an irrecoverable error occurs.
+                    match e {
+                        Error::CorruptedTaskQueue
+                        | Error::TaskDatabaseUpdate(_)
+                        | Error::HeedTransaction(_)
+                        | Error::CreateBatch(_) => {
+                            std::thread::sleep(Duration::from_secs(1));
+                        }
+                        _ => {}
+                    }
```
Kerollmops, line 395: can we use [the matches macro here](https://doc.rust-lang.org/std/macro.matches.html), please?

index-scheduler/src/lib.rs
```diff
@@ -378,37 +421,93 @@ impl IndexScheduler {
     pub fn get_task_ids(&self, query: &Query) -> Result<RoaringBitmap> {
         let rtxn = self.env.read_txn()?;
 
-        // This is the list of all the tasks.
-        let mut tasks = self.all_task_ids(&rtxn)?;
+        let ProcessingTasks { started_at: started_at_processing, processing: tasks_processing } =
```
Kerollmops, line 424:
```suggestion
        let ProcessingTasks { started_at: started_at_processing, processing: processing_tasks } =
```

index-scheduler/src/lib.rs
```diff
@@ -378,37 +421,93 @@ impl IndexScheduler {
...
+                    // special case for Processing tasks
+                    Status::Processing => {
+                        include_processing_tasks = true;
+                        status_tasks |= &tasks_processing;
+                    }
+                    status => status_tasks |= &self.get_status(&rtxn, *status)?,
+                };
+            }
+            if !include_processing_tasks {
+                tasks -= &tasks_processing;
```
Kerollmops, line 443:
```suggestion
                    Status::Processing => status_tasks |= &tasks_processing,
                    status => status_tasks |= &self.get_status(&rtxn, *status)?,
                }
            }
            if status.contains(&Status::Processing) {
                tasks -= &tasks_processing;
```

You can remove this `include_processing_tasks` boolean and simplify a bit, here.

Kerollmops (changes requested):
This PR is very good, it fixes a lot of bugs and makes the index scheduler even safer. However, would it be possible to use `()` instead of `{}` for the match branches and would it be possible to simplify some of the code that I highlighted?

loiclec:
Thanks for the review, I applied all your suggestions :)

irevoire (approved):

PR #2956, opened 2022-10-25
Fix crash when batching an index swap task containing 0 swaps

loiclec:
The crash occurred because we expected autobatched tasks to always be associated with at least one index.

The fix consists of checking whether the first autobatched task has an `IndexSwap` kind, and then creating the index swap batch separately in the `create_batch` method. 

( I also deleted some unreferenced insta-snapshot files in this PR.)


index-scheduler/src/batch.rs
```diff
@@ -416,41 +412,46 @@ impl IndexScheduler {
...
+        // 5.1. If the task is an index swap, create the batch directly.
+        if matches!(task.kind.as_kind(), Kind::IndexSwap) {
+            return Ok(Some(Batch::IndexSwap { task }));
+        }
+
+        // 5.2. Otherwise, get the index name associated with the task and use the autobatcher
+        // to batch the enqueued tasks associated with this index.
+
+        // This is safe because all the remaining task are associated with an index.
+        let index_name = task.index_uid().unwrap();
```
Kerollmops, line 428: Wouldn't it be better to use this unwrap to our advantage and match on the output of `index_uid`, here?
Like checking that it returns an index uid and if not check that it is, indeed, an IndexSwap.

index-scheduler/src/batch.rs
```diff
@@ -416,41 +412,46 @@ impl IndexScheduler {
             )));
         }
 
-        // 5. We take the next task and try to batch all the tasks associated with this index.
-        if let Some(task_id) = enqueued.min() {
-            let task = self.get_task(rtxn, task_id)?.ok_or(Error::CorruptedTaskQueue)?;
+        // 5. We make a batch from the unprioritised tasks. Start by taking the next enqueued task.
+        let task_id = if let Some(task_id) = enqueued.min() { task_id } else { return Ok(None) };
```
Kerollmops, line 416: Does it pass rustfmt, here? 😂 I love that but not sure it will pass!

loiclec:
Oops, ignore it for now, I just realised it is very wrong 😬

Kerollmops (changes requested):
Thank you for finding the bug here! Loves the fix!

index-scheduler/src/batch.rs
loiclec: I wrote it on several lines and rustfmt put it all in one line haha >:-(

index-scheduler/src/batch.rs
loiclec: The initial goal was to always create the IndexSwap batch manually here, even when it is associated with some indexes. But now I realised that we can't remove `IndexSwap` from the `BatchKind` type, since we need to stop batching tasks together when the autobatcher sees an IndexSwap task.

So now I have reverted the autobatcher to what it was, and I will only create the `IndexSwap` batch manually in the `create_batch` method when it is empty.

index-scheduler/src/batch.rs
```diff
@@ -416,41 +416,47 @@ impl IndexScheduler {
...
+        // If the task is not associated with any index, verify that it is an index swap and
+        // create the batch directly. Otherwise, get the index name associated with the task
+        // and use the autobatcher to batch the enqueued tasks associated with it
+
+        let index_name = if let Some(&index_name) = task.indexes().first() {
+            index_name
+        } else {
+            assert!(matches!(&task.kind, KindWithContent::IndexSwap { swaps } if swaps.is_empty()));
+            return Ok(Some(Batch::IndexSwap { task }));
+        };
```
irevoire, line 432: I’m not sure I understand.
Why is the index_swap not returning the two indexes as it was doing before?

Kerollmops (approved):
Thank you very much, merging!
Just for my info, index swap tasks are associated with no index in the task list 

bors merge

index-scheduler/src/batch.rs
irevoire: Ah nvm just got it

loiclec:
@Kerollmops no, they are! it is only the *empty* index swap tasks that are not associated with an index

PR #2960, opened 2022-10-25
Add index scheduler tests for task cancelation

loiclec:
Nothing special to say about this one. I checked that:
- enqueued tasks are canceled
- processing document addition tasks can be canceled as well, and the AbortedIndexation error is intercepted correctly
- succeeded tasks are not canceled


irevoire (approved):
Nice thank you @loiclec

PR #2967, opened 2022-10-26
test the details of all tasks type

irevoire:


meilisearch-http/src/routes/indexes/settings.rs
```diff
@@ -439,6 +439,7 @@ pub async fn update_all(
         Some(&req),
     );
 
+    println!("Registering a setting update");
```
Kerollmops, line 442:
```suggestion
```

Kerollmops, line 442:
```suggestion
```

Kerollmops (changes requested):

Kerollmops (changes requested):
That is very nice, do we also want to add:
 - a test for a `snapshotCreation` ?
 - a test for a swap with multiple swap operations?
 - a test with an invalid swap operation i.e. length != 2?

PR #2968, opened 2022-10-26
Index scheduler snapshots

Kerollmops:
This PR introduces the `snapshotCreation` tasks, the import of a snapshot, the scheduled snapshots system, and reintroduces the `VERSION` file.


meilisearch-types/src/versioning.rs
```diff
@@ -0,0 +1,61 @@
+use std::fs;
```
irevoire, line 1: nice

irevoire (approved):

meilisearch-types/src/versioning.rs
Kerollmops: 👍

PR #2977, opened 2022-10-26
fix the index creation

irevoire:
When an index is being created we insert it in the index_map straight away to avoid someone else from trying to re-open it. The definitive fix should be made on milli's side

### DIFF
--- a/index-scheduler/src/index_mapper.rs
+++ b/index-scheduler/src/index_mapper.rs
@@ -75,8 +75,11 @@ impl IndexMapper {
 
     /// Get or create the index.
     pub fn create_index(&self, mut wtxn: RwTxn, name: &str) -> Result<Index> {
-        let ret = match self.index(&mut wtxn, name) {
-            Ok(index) => Ok(index),
+        match self.index(&mut wtxn, name) {
+            Ok(index) => {
+                wtxn.commit()?;
+                Ok(index)
+            }
             Err(Error::IndexNotFound(_)) => {
                 let uuid = Uuid::new_v4();
                 self.index_mapping.put(&mut wtxn, name, &uuid)?;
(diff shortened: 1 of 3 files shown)


index-scheduler/src/index_mapper.rs
```diff
@@ -83,7 +83,14 @@ impl IndexMapper {
 
                 let index_path = self.base_path.join(uuid.to_string());
                 fs::create_dir_all(&index_path)?;
-                self.create_or_open_index(&index_path)
+                let index = self.create_or_open_index(&index_path)?;
+
+                // TODO: this is far from perfect. If the caller don't commit or fail his commit
+                // then we end up with an available index that should not exist and is actually
+                // not available in the index_mapping database.
+                self.index_map.write().unwrap().insert(uuid, Available(index.clone()));
```
Kerollmops, line 91: Can't we take the `wtxn` by ownership and commit on our side, only inserting the index in the hashmap if it worked?

Kerollmops: Also, shouldn't we check that the value is not already in a `BeingDeleted` state, and don't insert it if so?

Kerollmops (changes requested):

index-scheduler/src/index_mapper.rs
irevoire: But we actually want to insert it even if it’s in a BeingDeleted state no?
It'll still be deleted later.

> Can't we take the wtxn by ownership and commit on our side, only inserting the index in the hashmap if it worked?

Maybe we could, maybe not, I think it would change a lot of code and I don’t really have the time rn but we can check it together in a few minutes

index-scheduler/src/index_mapper.rs
```diff
@@ -74,16 +74,29 @@ impl IndexMapper {
...
-                self.index_mapping.put(wtxn, name, &uuid)?;
+                self.index_mapping.put(&mut wtxn, name, &uuid)?;
 
                 let index_path = self.base_path.join(uuid.to_string());
                 fs::create_dir_all(&index_path)?;
-                self.create_or_open_index(&index_path)
+                let index = self.create_or_open_index(&index_path)?;
+
+                wtxn.commit()?;
+                // TODO: it would be better to lazyly create the index. But we need an Index::open function for milli.
```
Kerollmops, line 92:
```suggestion
                // TODO: it would be better to lazily create the index. But we need an Index::open function for milli.
```

Kerollmops (approved):

Issue #2970, opened 2022-10-26
Filtering tasks with an `indexUid` returns global tasks having `null` as `indexUid`


gmourier:
**Describe the bug**

When filtering tasks with an `indexUid`, I get global tasks having `null` as `indexUid` while I shouldn't.

**To Reproduce**

1. Swap indexes (e.g `A` and `B`)
2. GET `/tasks?indexUid=A`
3. I get the `indexSwap` task in the response while I shouldn't since the `indexUid` is equal to `null`.

**Expected behavior**

I should get tasks having an `indexUid` field being set to the given value, in that case `A` 

**Meilisearch version:** `v0.30.0-new-index-scheduler.beta.1`

loiclec:
Since an indexSwap task is associated with many indexes, should the `indexUid` field contain an array then? For example:
```json
{
      "details": {
          [
              { "swap": ["A", "B"] },
              { "swap": ["C", "D"] }
              
          ]
      },
      "indexUid": ["A", "B", "C", "D"],
      "status": "succeeded",
      "type": "indexSwap",
      "uid": 0,
      ...
    }
```

PR #2982, opened 2022-10-27
Reimplement task queries to account for special index swap rules

loiclec:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/2970 

## What does this PR do?
- Replace the `get_tasks` method with a `get_tasks_from_authorized_indexes` which returns the list of tasks matched by the query **from the point of view of the user**. That is, it takes into consideration the list of authorised indexes as well as the special case of `IndexSwap` which should not be returned if an index_uid is specified or if any of its associated indexes are not authorised.
- Adapt the code in other places following this change
- Add some tests
- Also the method `get_task_ids_from_authorized_indexes` now takes a read transaction as argument. This is because we want to make sure that the implementation of `get_tasks_from_authorized_indexes` only uses one read transaction. Otherwise, we could (1) get a list of task ids matching the query, then (2) one of these task ids is deleted by a taskDeletion task, and finally (3) we try to get the `Task`s associated with each returned task ids, and get a `CorruptedTaskQueue` error.


index-scheduler/src/lib.rs
```diff
@@ -561,11 +563,73 @@ impl IndexScheduler {
...
+    /// Return true iff there is at least one task associated with this index
+    /// that is processing.
+    pub fn is_index_processing(&self, index: &str) -> Result<bool> {
+        let rtxn = self.env.read_txn()?;
+        let processing_tasks = self.processing_tasks.read().unwrap().processing.clone();
+        let index_tasks = self.index_tasks(&rtxn, index)?;
+
+        let nbr_index_processing_tasks = processing_tasks.intersection_len(&index_tasks);
+
+        Ok(nbr_index_processing_tasks > 0)
```
Kerollmops, line 575:
```suggestion
        let index_tasks = self.index_tasks(&rtxn, index)?;
        let processing_tasks = self.processing_tasks.read().unwrap().processing;
        let nbr_index_processing_tasks = processing_tasks.intersection_len(&index_tasks);
        Ok(nbr_index_processing_tasks > 0)
```

index-scheduler/src/lib.rs
```diff
@@ -561,11 +563,73 @@ impl IndexScheduler {
...
+        if query.index_uid.is_some() {
+            tasks -= self.get_kind(rtxn, Kind::IndexSwap)?
+        }
+
+        // Any task that is internally associated with a non-authorized index
+        // must be discarded.
+        if let Some(authorized_indexes) = authorized_indexes {
+            let all_indexes_iter = self.index_tasks.iter(rtxn)?;
+            for iter_el in all_indexes_iter {
+                let (index, index_tasks) = iter_el?;
```
Kerollmops, line 605:
```suggestion
            for result in all_indexes_iter {
                let (index, index_tasks) = result?;
```

index-scheduler/src/lib.rs
```diff
@@ -561,11 +563,73 @@ impl IndexScheduler {
...
+                if !authorized_indexes.contains(&index.to_owned()) {
```
Kerollmops, line 606:
```suggestion
                if !authorized_indexes.contains(index) {
```

Kerollmops (changes requested):

index-scheduler/src/lib.rs
```diff
@@ -561,11 +563,73 @@ impl IndexScheduler {
         Ok(tasks)
     }
 
-    /// Returns the tasks matched by the given query.
-    pub fn get_tasks(&self, query: Query) -> Result<Vec<Task>> {
-        let tasks = self.get_task_ids(&query)?;
+    /// Return true iff there is at least one task associated with this index
```
irevoire, line 566:
```suggestion
    /// Return true if there is at least one task associated with this index
```

index-scheduler/src/lib.rs
```diff
@@ -561,11 +563,73 @@ impl IndexScheduler {
...
+    }
+
+    /// Return the task ids matching the query from the user's point of view.
+    ///
+    /// There are two differences between an internal query and a query executed by
+    /// the user.
+    ///
+    /// 1. IndexSwap tasks are not publicly associated with any index, but they are associated
+    /// with many indexes internally.
+    /// 2. The user may not have the rights to access the tasks (internally) associated wuth all indexes.
```
irevoire, line 585:
```suggestion
    /// 2. The user may not have the rights to access the tasks (internally) associated with all indexes.
```

index-scheduler/src/lib.rs
```diff
@@ -561,11 +563,73 @@ impl IndexScheduler {
...
+    }
+
+    /// Return the tasks matching the query from the user's point of view.
+    ///
+    /// There are two differences between an internal query and a query executed by
+    /// the user.
+    ///
+    /// 1. IndexSwap tasks are not publicly associated with any index, but they are associated
+    /// with many indexes internally.
+    /// 2. The user may not have the rights to access the tasks (internally) associated wuth all indexes.
```
irevoire, line 622:
```suggestion
    /// 2. The user may not have the rights to access the tasks (internally) associated with all indexes.
```

index-scheduler/src/lib.rs
```diff
@@ -561,11 +563,73 @@ impl IndexScheduler {
...
+        query: &Query,
+        authorized_indexes: &Option<Vec<String>>,
+    ) -> Result<RoaringBitmap> {
+        let mut tasks = self.get_task_ids(rtxn, query)?;
+
+        // If the query contains a list of index_uid, then we must exclude IndexSwap tasks
+        // from the result (because it is not publicly associated with any index)
+        if query.index_uid.is_some() {
+            tasks -= self.get_kind(rtxn, Kind::IndexSwap)?
+        }
```
irevoire, line 598: Ok, so that’s pretty good. But actually, after testing the old meilisearch a little bit, I found out we’re supposed to ignore *every* kind of task that is not associated with an index_uid.
Thus:
- TaskCancelation
- TaskDeletion
- DumpCreation (only this one was present in the previous version of meilisearch and I missed it)
- SnapshotCreation

irevoire (changes requested):

index-scheduler/src/lib.rs
irevoire: Ok so ACTUALLY after rerethinking about it with @gmourier it’s a little bit more complicated.

When checking the allowed indexes **by the API-keys**, if there is a limitation on the indexes (`index_uid!=*` or `None` in the code I think).
We must filter out the « global tasks »:
- TaskCancelation
- TaskDeletion
- DumpCreation

In the case of the `IndexSwap` we can see it only if in the user query we typed `index_uid=*` so this is good.

Now the bad thing I think is... what if there is a swap(A, B) and your API key only allows you to access to A?
You should not see the swap but here from what I understand of your code we will receive it :thinking:
And I don't really knows how to do this efficiently...

Maybe @Kerollmops you'll have an idea?

index-scheduler/src/lib.rs
loiclec: but "iff" means "if and only if" :(

irevoire (approved):
Nice thank you @loiclec

irevoire:
bors merge

index-scheduler/src/lib.rs
loiclec: I am not sure I understand the problem.
If someone specifies `indexUid=catto`, then we'll hit line 495:
```rust
        if let Some(index) = &query.index_uid {
            let mut index_tasks = RoaringBitmap::new();
            for index in index {
                index_tasks |= self.index_tasks(rtxn, index)?;
            }
            tasks &= &index_tasks;
        }
```
and the taskDeletion, taskCancelation, etc. won't be returned.

If there is a `indexSwap(A, B)` and the API key only allows me to access A, then we'll hit line 605:
```rust
      // Any task that is internally associated with a non-authorized index
        // must be discarded.
        if let Some(authorized_indexes) = authorized_indexes {
            let all_indexes_iter = self.index_tasks.iter(rtxn)?;
            for result in all_indexes_iter {
                let (index, index_tasks) = result?;
                if !authorized_indexes.contains(&index.to_owned()) {
                    tasks -= index_tasks;
                }
            }
        }
```
which will remove the task from the results as well. (In the loop there, `index == B` and `authorised_indexes == [A]`, so `[A].contains(B)` will return false and we'll run `tasks -= [indexSwap(A,B)]`)

PR #2990, opened 2022-10-27
isolate the search in another task

irevoire:
In case there is a failure on milli's side that should avoid blocking the tokio main thread


Kerollmops (approved):
I like it like that 🎼
bors merge

PR #3002, opened 2022-10-30
Fix dump import without instance uid

irevoire:
When creating a dump without any instance-uid (that can happen if you’ve always run meilisearch with the `--no-analytics` flag), you could get an error when trying to load the dump.


Kerollmops (approved):
Thank you, it looks good to me ❤️
bors merge

Kerollmops:
This is a flaky test on Windows...
bors merge

PR #3010, opened 2022-10-31
temporary uses nextest in the CI

irevoire:
When we running cargo test we have a lot of tests that become flaky and trigger a deadlock. The issue seems to be linked to the fact that cargo test run all the tests in multiple threads instead of multiple environment. Thus, as a temporary fix I would like to move our test suite over nextest that run each test in a separate process.

Once this is merged we should open an issue to remove nextest from our CI.
The reason being that it doesn't run the doctests. We don't have a lot of doctests in meilisearch since we ship a binary and not a lib but I think it would be better to check the full test suite in the CI.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -31,6 +31,11 @@ jobs:
       with:
         command: build
         args: --locked --release --no-default-features
+    - name: Install cargo-nextest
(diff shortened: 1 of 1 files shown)


irevoire:
As we can see in the CI, it looks like there are still some tests hanging, and thus this PR doesn't fix the issue.

PR #3012, opened 2022-10-31
Extract the dates out of the dumpv4.

### LINKED ISSUE #2987: Extract the date ouf of the dumpv4
When importing a dump v4, we should extract the creation and last updated timestamp.
https://github.com/meilisearch/meilisearch/blob/main/dump/src/reader/v4/mod.rs#L150-L169

The idea is to look at the whole task queue until you find the first task concerning your index (=last_updated_at), and you already have the task_uid of the task_creating it. Thus you then need to continue until you reach this task and extract its date as the creation_date.

funilrys:
Hi there, 

please review this PR that tries to fix #2987. I'm still learning Rust and I found that #2987 is an excellent way for me to read and learn what others do with Rust. So please excuse my semantics ...

Stay safe and healthy.

---

# Pull Request

This patch possibly fixes #2987.

This patch introduces a way to fill the IndexMetadata.created_at and IndexMetadata.updated_at keys from the tasks events. This is done by reading the creation date of the first event (created_at) and the creation date of the last event (updated_at).

## Related issue
Fixes #2987

## What does this PR do?
- Extract the dates out of the dumpv4.

## PR checklist

### DIFF
--- a/dump/src/reader/v4/mod.rs
+++ b/dump/src/reader/v4/mod.rs
@@ -23,6 +23,7 @@ pub type Checked = settings::Checked;
 pub type Unchecked = settings::Unchecked;
 
 pub type Task = tasks::Task;
+pub type TaskEvent = tasks::TaskEvent;
 pub type Key = keys::Key;
 
 // everything related to the settings
@@ -100,6 +101,7 @@ impl V4Reader {
             V4IndexReader::new(
                 index.uid.clone(),
                 &self.dump.path().join("indexes").join(index.index_meta.uuid.to_string()),
+                BufReader::new(self.tasks.get_ref().try_clone().unwrap()),
             )
         }))
     }
@@ -147,16 +149,31 @@ pub struct V4IndexReader {
 }
 
 impl V4IndexReader {
-    pub fn new(name: String, path: &Path) -> Result<Self> {
+    pub fn new(name: String, path: &Path, tasks: BufReader<File>) -> Result<Self> {
         let meta = File::open(path.join("meta.json"))?;
         let meta: DumpMeta = serde_json::from_reader(meta)?;
+
+        let mut index_tasks: Vec<Task> = vec![];
+
+        for line in tasks.lines() {
+            let task: Task = serde_json::from_str(&line?)?;
+
+            if task.index_uid.to_string() == name {
+                index_tasks.push(task)
+            }
+        }
 
         let metadata = IndexMetadata {
             uid: name,
             primary_key: meta.primary_key,
-            // FIXME: Iterate over the whole task queue to find the creation and last update date.
-            created_at: OffsetDateTime::now_utc(),
-            updated_at: OffsetDateTime::now_utc(),
+            created_at: match index_tasks.first().unwrap().events.first() {
+                Some(TaskEvent::Created(ts)) => *ts,
+                _ => OffsetDateTime::now_utc(),
+            },
+            updated_at: match index_tasks.last().unwrap().events.last() {
+                Some(TaskEvent::Created(ts)) => *ts,
+                _ => OffsetDateTime::now_utc(),
+            },
         };
 
         let ret = V4IndexReader {


dump/src/reader/v4/mod.rs
```diff
@@ -147,16 +149,31 @@ pub struct V4IndexReader {
...
 
+        let mut index_tasks: Vec<Task> = vec![];
+
+        for line in tasks.lines() {
+            let task: Task = serde_json::from_str(&line?)?;
+
+            if task.index_uid.to_string() == name {
+                index_tasks.push(task)
+            }
+        }
```
irevoire, line 164: Could you compute the `created_at` and `updated_at` here directly to avoid allocating a `Vec` with all tasks and avoid scanning the whole tasks queue if there is only a few tasks concerning the index.

```suggestion
        let created_at = None;
        let updated_at = None;

        for line in tasks.lines() {
            let task: Task = serde_json::from_str(&line?)?;

            if task.index_uid == &name {
                if updated_at.is_none() {
                    updated_at = Some(...); // get the timestamp of the first task matching our index name
                }
                if task.type == Created { // find a way to check if the task type was an IndexCreation
                    created_at = Some(...); // get the timestamp of the task creation.
                    break; // we don't need to read any more task and can exit.
                }
            }
        }
```

dump/src/reader/v4/mod.rs
```diff
@@ -147,16 +149,31 @@ pub struct V4IndexReader {
...
-            created_at: OffsetDateTime::now_utc(),
-            updated_at: OffsetDateTime::now_utc(),
+            created_at: match index_tasks.first().unwrap().events.first() {
+                Some(TaskEvent::Created(ts)) => *ts,
+                _ => OffsetDateTime::now_utc(),
+            },
+            updated_at: match index_tasks.last().unwrap().events.last() {
+                Some(TaskEvent::Created(ts)) => *ts,
+                _ => OffsetDateTime::now_utc(),
+            },
```
irevoire, line 176:
```suggestion
            created_at: created_at.unwrap_or(OffsetDateTime::now_utc()),
            updated_at:  updated_at.unwrap_or(OffsetDateTime::now_utc()),
```

Once the previous change are implemented we should be able to do this

irevoire (changes requested):
Hey @funilrys, I think your PR works but I don't have the time to test it rn.
Anyway, I think we could extract the dates in a less intensive way, see my comment.

Basically, if we were trying to get the dates of the index A what I would like to do is something like that:
```python
DocumentAddition(B) # the index uid doesn't match, we ignore this task
DocumentAddition(A) # The index uid matches and we don't have any updated_at set, we set our date to this task whatever its type is.
DocumentAddition(B) # index uid don't match we ignore it
DocumentAddition(A) # The updated_at is already set, we ignore this task
IndexCreation(A) # the index uid matches and it's an indexCreation, we set the created_at to this timestamp and can exit the loop.
IndexCreation(B) # This task is not read at all
```

irevoire:
Also, since your PR was probably working and you've probably spend a lot of time on it, I've marked it as hacktoberfest-accepted. To receive some swag from Meilisearch, please complete [this form](https://meilisearch.typeform.com/to/idt5qugY).

dump/src/reader/v4/mod.rs
irevoire: Thanks for implementing the changes @funilrys :tada:

dump/src/reader/v4/mod.rs
```diff
@@ -147,16 +150,45 @@ pub struct V4IndexReader {
...
+
+                if created_at.is_none() {
+                    created_at = match task.content {
+                        TaskContent::IndexCreation { primary_key } => match task.events.first() {
+                            Some(TaskEvent::Created(ts)) => Some(*ts),
+                            _ => None,
+                        },
+                        _ => None,
+                    };
+                }
```
irevoire, line 179: hey @funilrys sorry, when I wrote my first review I totally forgot something about meilisearch.
And index can be created by multiple events:
- An indexCreation
- A documentAddition
- A settingsAddition

Thus this simple if actually doesn't work and finding the creation date is harder than I thought.
BUT, when we wrote this dump format back in the past we already thought about it and created a special struct to retrieve the `creation_task_id`.
Here is a print of the `index_uuids/data.jsonl` file:
```json
{
  "uid": "dnd_spells",
  "index_meta": {
    "uuid": "684b4662-e3c3-4440-8d57-07ee42b0b901",
    "creation_task_id": 6
  }
```

And in the code of the dumpV4 I'm already deserializing this structure:
```rust
#[derive(Deserialize, Debug)]
#[cfg_attr(test, derive(serde::Serialize))]
pub struct IndexUuid {
    pub uid: String,
    pub index_meta: IndexMeta,
}

#[derive(Deserialize, Debug)]
#[cfg_attr(test, derive(serde::Serialize))]
pub struct IndexMeta {
    pub uuid: Uuid,
    pub creation_task_id: usize,
}
```


Thus the idea for
1. Stopping the loop
2. Setting the creation date
Is to simply stop when we reach the `creation_task_id` of the index match the `task_id` of the task.

```suggestion
                if task.id == self.creation_task_id { // I have no idea of what fields are accessible sorry
                    created_at = task.processed_at(); // we'll probably need to write a function that extract the date whatever is the variant
                    break; // we don't need to check anything else and can stop right away
                }
```

irevoire (changes requested):
Hey @funilrys, sorry I forgot one little detail about the way an index can be created :pray: 

I think that's the last change you need to implement, and then we're good to merge.
Thanks for your work! :1st_place_medal:

dump/src/reader/v4/mod.rs
funilrys: @irevoire Sorry for the waiting time... Last week was a long one! ... 

I can't find any `TaskContent::SettingsAddition` ... Do you mean `TaskContent::SetingsUpdate` ?

funilrys: By the way, is is a reason behind the incompatibility of `task.id` and `IndexMeta.creation_task_id` ?

One is `u32` and the later `usize` ... That lost me ...

funilrys: @irevoire in your suggestion what would `processed_at` be? Right now I will only shift my match to that method ... I don't really understand what you mean with "whatever is the variant" ... 

Right now - from my understanding - the only way to read the creation or update time is to go through the list of events - as I implemented ...

irevoire: > Sorry for the waiting time...

No worries! I’m already super happy to see someone tackling this hard problem and getting into the codebase!

> Do you mean TaskContent::SettingsUpdate ?

Yes sorry! 🙏 

> By the way, is is a reason behind the incompatibility of task.id and IndexMeta.creation_task_id ?

The reason is because the way I implemented the dumps was by taking the exact code that generated these dumps at the time. And it looks like, at the time of the dumpv4 the `creation_task_id` was a `usize` for no reason, thus we need to convert it as `u32` or the other way around before checking for equality 😔 

> in your suggestion what would processed_at be?

It should be a new method implemented on the `Task` type in the `v4::task` module that returns, at which time the task has been processed.
Being processed means, the task was either a success or a failure. (little note on this: meilisearch isn’t entirely ACID, and when adding documents, an index will be automatically created, but then the document addition can fail; thus, an index can be created by a failed task).

I imagine something like that would be enough:
```rust
impl Task {
    /// Return at which time the task finished processing.
    pub fn processed_at(&self) -> Option<OffsetDateTime> {
        match self.events.last() {
            Some(TaskEvent::Succeded { timestamp, .. } | TaskEvent::Failed { timestamp, .. }) => {
                Some(timestamp.clone())
            }
            _ => None,
        }
    }
}
```

-------

> the only way to read the creation or update time is to go through the list of events - as I implemented ...

Well, yes and no. We still need to go through the list of events, as you said, thus, the code shouldn’t change a lot.
But we need to use the stored `created_at` value to get the time of the task that created the index. And for the last_update part I think your code is working 👍

curquiza:
bors try

funilrys:
@irevoire Any change requested? I'm asking so that I can go ahead with #3013 😄

irevoire:
Hey, yeah, it looks pretty good to me.
But it looks like it introduced a bug, and now the tests fail without being able to get the task queue.

funilrys:
Hey @irevoire, I looked at the CI output and I'm not sure why this is happening...

I think it's probably because of the new arguments of the `V4IndexReader` but I can't explain how could that happen when the upstream implementation only calls a V4Reader ... From my understanding, `V4IndexReader` is only called by `V4Reader` so the new arguments shouldn't cause any issue ... 

Can you give me some theoretical pointers? 🤔

irevoire (approved):
Hey, @funilrys, that's NICE; I'll try to merge as-is.

But _I think_ you'll need to rebase yourself on main and fix one or two tests on the date.
These tests were disabled when you started working on this subject but have been re-enabled in the meanwhile.

You shouldn't have any conflicts thought because no one worked on the dump-v4 since you opened the PR :+1: 

bors merge

PR #3019, opened 2022-11-02
Fix error code of the "duplicate index found" error on the index swap route

loiclec:
According to the spec, the code of the error should be `duplicate_index_found`, but it was `bad_request` instead.


irevoire (approved):
Thanks @loiclec 
bors merge

PR #3022, opened 2022-11-02
Store the `started_at` for a task that is canceled when processing

Kerollmops:
This PR changes the current behavior of the displayed tasks. When a processing task is canceled, the `started_at` date time is kept and displayed to the user. Otherwise, if the task was just enqueued, the `started_at` remains `null`. If a task is processing, the engine is ctrl-c, starts again, and the task becomes enqueued again, so if it is canceled, its `started_at` will be `null`.

You can read more [in this discussion](https://github.com/meilisearch/specifications/pull/195/files#r1009602335).


index-scheduler/src/lib.rs
```diff
@@ -947,6 +947,12 @@ impl IndexScheduler {
                 #[cfg(test)]
                 self.breakpoint(Breakpoint::AbortedIndexation);
                 wtxn.abort().map_err(Error::HeedTransaction)?;
+
+                // We make sure that we don't call `stop_processing` on the `processing_tasks`,
+                // this is because we want to let the tick call `cerate_next_batch` and keep
```
irevoire, line 952:
```suggestion
                // this is because we want to let the next tick call `cerate_next_batch` and keep
```

irevoire (approved):
nice

irevoire (approved):
thanks @Kerollmops 

bors merge

irevoire:
bors merge

PR #3032, opened 2022-11-08
Fix Index name parsing error message to fit the specification

### LINKED ISSUE #2754: Benchmark some usecases and find a way to measure the team's performance
_This implementation is only for our internal usage and does not impact the users._

Following this issue: https://github.com/meilisearch/devrel/issues/334

We need to implement benchmark tools to be able to measure our performance improvements. Currently, we have benchmarks on the [milli library](https://github.com/meilisearch/milli/actions/workflows/manual_benchmarks.yml), but it does not correspond to the reality of the user scenarios.

We want to measure some of these scenarios automatically and compare them between two versions.

### LINKED ISSUE #2924: Wrong variant returned for `invalid_api_key_indexes` error when creating an API key.
> Sending a value that is not a valid index (e.g. regarding the index uid format) in the indexes field returns an [invalid_api_key_indexes](https://github.com/meilisearch/meilisearch/issues/0061-error-format-and-definitions.md#invalid_api_key_indexes) with a variant error message. See "Sending an invalid index uid format in indexes field" variant.

It should have been solved by https://github.com/meilisearch/meilisearch/issues/2158, but I did not get the new variant testing on the latest version.

```json
{
    "message": "`indexes` field value `[\"invalid index # / \\name with spaces\"]` is invalid. It should be an array of string representing index names.",
    "code": "invalid_api_key_indexes",
    "type": "invalid_request",
    "link": "https://docs.meilisearch.com/errors#invalid_api_key_indexes"
}
```

while I should receive

```json
{
    "message": "` [\"invalid index # / \\name with spaces\"]` is not a valid index uid. Index uid can be an integer or a string containing only alphanumeric characters, hyphens (-) and underscores (_).",
    "code": "invalid_api_key_indexes",
    "type": "invalid_request",
    "link": "https://docs.meilisearch.com/errors#invalid_api_key_indexes"
}
```

Relates to that catch-up specification https://github.com/meilisearch/specifications/pull/181 (The original one was never merged 🤔 So it makes me doubt if it was solved)

ManyTheFish:
fixes #2924

### DIFF
--- a/meilisearch-http/tests/auth/api_keys.rs
+++ b/meilisearch-http/tests/auth/api_keys.rs
@@ -352,7 +352,7 @@ async fn error_add_api_key_invalid_parameters_indexes() {
     assert_eq!(400, code, "{:?}", &response);
 
     let expected_response = json!({
-        "message": r#"`{"name":"products"}` is not a valid index uid. It should be an array of string representing index names."#,
+        "message": r#"`indexes` field value `{"name":"products"}` is invalid. It should be an array of string representing index names."#,
         "code": "invalid_api_key_indexes",
         "type": "invalid_request",
         "link": "https://docs.meilisearch.com/errors#invalid_api_key_indexes"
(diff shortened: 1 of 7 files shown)


curquiza:
@ManyTheFish looks like GitHub does not like us, despite you add `fixes #...` in your PR description, can you check the issues you fix are well added to the `Development` section of the PRs you open please? 😇

PR #3070, opened 2022-11-16
Remove core and use engine

curquiza:
Following the new team name
Not mandatory since GitHub is doing redirection, but more consistent


Kerollmops (approved):
LGTM, that's amazing 🎉 
bors merge

PR #3071, opened 2022-11-16
Analytics on the tasks route

irevoire:
Implement the missing analytics on the delete and cancel task routes.
+ Batch the analytics on the `GET tasks` route to avoid flooding ourselves while polling meilisearch.

### DIFF
--- a/meilisearch-http/src/analytics/segment_analytics.rs
+++ b/meilisearch-http/src/analytics/segment_analytics.rs
@@ -27,6 +27,7 @@ use super::{config_user_id_path, DocumentDeletionKind, MEILISEARCH_CONFIG_PATH};
 use crate::analytics::Analytics;
 use crate::option::{default_http_addr, IndexerOpts, MaxMemory, MaxThreads, SchedulerConfig};
 use crate::routes::indexes::documents::UpdateDocumentsQuery;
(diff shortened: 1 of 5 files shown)


meilisearch-http/src/analytics/segment_analytics.rs
```diff
@@ -840,6 +855,91 @@ impl DocumentsDeletionAggregator {
...
+    timestamp: Option<OffsetDateTime>,
+
+    // context
+    #[serde(rename = "user-agent")]
+    user_agents: HashSet<String>,
+
+    filtered_by_uid: bool,
+    filtered_by_index_uid: bool,
+    filtered_by_type: bool,
+    filtered_by_status: bool,
```
Kerollmops, line 870: Are we sure we want these names and not the plural ones?
Are we sure we don't prefer a count here, instead?

Kerollmops (changes requested):

meilisearch-http/src/analytics/segment_analytics.rs
Kerollmops: As seen internally, we don't need that.

Kerollmops (approved):
LGTM 🎉 🚀 
bors merge

PR #3085, opened 2022-11-17
refactorize the whole test suite

irevoire:
1. Make a call to assert_internally_consistent automatically when snapshotting the scheduler. There is no point in snapshotting something broken and expecting the dumb humans to notice.
2. Replace every possible call to assert_internally_consistent with a snapshot of the scheduler. It uses the same amount of lines and ensures we never change something without noticing in any tests ever.
3. Name every snapshot: it's easier to debug when something goes wrong and easier to review in general.
4. Stop skipping breakpoints; it's too easy to miss something. Now you must explicitly show which path the scheduler is supposed to use.
5. Add a timeout on the channel.recv, it eases the process of writing tests; now, when something files, you get a failure instead of a deadlock.


curquiza:
bors try

curquiza:
@irevoire the tests fail

loiclec (approved):
Suuuuuper nice! Love it. 
I trust that you didn't actually change the behaviour of the tests and that the content of the insta-snapshot files didn't change?
If so, LGTM once the failing test is fixed 👍

Kerollmops (approved):
This looks way better than before, we will be able to catch more bugs, and I am happy that the tests are passing right now 😅

bors merge

irevoire:
bors merge

PR #3100, opened 2022-11-21
Add a dispatch to the publish binaries workflow

curquiza:
Add `worklfow_dispatch` event to publish binaries workflow to allow the manually trigger


Kerollmops (approved):
I hope it works!
bors merge

PR #3105, opened 2022-11-22
Fix publish release CI

curquiza:
Fix CI to avoid release creation when triggering manually the CI with `workflow_dispatch` -> only trigger the upload of binary when the event is `release` instead of different of `schedule`


Kerollmops (approved):
Thank you, don't forget to cherry-pick on main ☺️
bors merge

Issue #1831, opened 2021-10-19
Error when deleting index: DirectoryNotEmpty


stefkes:
**Description**
When deleting an index using `$client->deleteIndex('Posts')` I see in the MeiliSearch output the error below. However, the index does seem to be deleted in the sense that `$client->getAllIndexes()` no longer shows the one I just tried to delete.

`ERROR meilisearch_lib::index_controller::index_resolver] Error while deleting index: Internal(Os { code: 39, kind: DirectoryNotEmpty, message: "Directory not empty" })`

**Expected behavior**
There should not be an error

**Current behavior**
There's a `DirectoryNotEmpty` error

**Screenshots or Logs**
See above

**Environment (please complete the following information):**
- OS: Ubuntu running in a Virtualbox (Laravel Homestead) on MacOS
- MeiliSearch version: 0.23.1
- meilisearch-php version: ^0.19.2

veerendrapolepalli:
I am also getting same kind of error , we have persisted the data.ms folder in the server , when there is new update we are trying to delete the existing index and create a new index with same name , logs are attached below

[2021-11-10T09:45:12Z INFO  actix_web::middleware::logger] 10.0.5.213:47334 "DELETE /indexes/price HTTP/1.1" 204 0 "-" "python-requests/2.25.1" 0.056508
[2021-11-10T09:45:12Z ERROR meilisearch_lib::index_controller::index_resolver] Error while deleting index: Internal(Os { code: 39, kind: DirectoryNotEmpty, message: "Directory not empty" })
[2021-11-10T09:45:12Z INFO  actix_web::middleware::logger] 10.0.5.213:47338 "DELETE /indexes/attribute HTTP/1.1" 204 0 "-" "python-requests/2.25.1" 0.059311
[2021-11-10T09:45:12Z ERROR meilisearch_lib::index_controller::index_resolver] Error while deleting index: Internal(Os { code: 39, kind: DirectoryNotEmpty, message: "Directory not empty" })
[2021-11-10T09:45:12Z INFO  actix_web::middleware::logger] 10.0.5.213:47336 "DELETE /indexes/auto HTTP/1.1" 204 0 "-" "python-requests/2.25.1" 0.089133
[2021-11-10T09:45:12Z ERROR meilisearch_lib::index_controller::index_resolver] Error while deleting index: Internal(Os { code: 39, kind: DirectoryNotEmpty, message: "Directory not empty" })
[...]

PR #3107, opened 2022-11-22
Fix index directory deletion

ManyTheFish:
In this PR I suggest retrying the Index directory deletion before returning an Error.
I **think** that, sometimes, mounted file systems may not remove a file making Meilisearch return an Error because the directory it tries to remove is not empty.


Kerollmops:
Closing as [the bug seems to be already fixed](https://github.com/meilisearch/meilisearch/issues/1831#issuecomment-1323752827).

PR #3114, opened 2022-11-23
Update the analytics on the ranking rules

### LINKED ISSUE #3113: Add missing telemetry
This issue is about adding the missing telemetry before v0.30.0

See the related changes

https://github.com/meilisearch/specifications/commit/31b437badfd3db1dcc0fdad132e460ea83c9e127

@meilisearch/docs-team sorry for the last minute addition

irevoire:
fix #3113

### DIFF
--- a/meilisearch-http/src/routes/indexes/settings.rs
+++ b/meilisearch-http/src/routes/indexes/settings.rs
@@ -344,7 +344,7 @@ make_setting_route!(
                     "attribute_position": setting.as_ref().map(|rr| rr.iter().position(|s| s == "attribute")),
                     "sort_position": setting.as_ref().map(|rr| rr.iter().position(|s| s == "sort")),
                     "exactness_position": setting.as_ref().map(|rr| rr.iter().position(|s| s == "exactness")),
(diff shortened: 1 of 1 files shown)


meilisearch-http/src/analytics/segment_analytics.rs
```diff
@@ -373,8 +373,8 @@ impl Segment {
     }
 
     async fn run(mut self, index_scheduler: Arc<IndexScheduler>) {
-        const INTERVAL: Duration = Duration::from_secs(60 * 60); // one hour
-                                                                 // The first batch must be sent after one hour.
+        const INTERVAL: Duration = Duration::from_secs(60); // one hour
```
dureuill, line 376: It looks to me like the comment is no longer current, because the interval has been changed to 1 minute rather than 1h. Is that meant to stay that way in the final PR?

irevoire: Oooops not at all thank you!

dureuill (approved):
Assuming you tested and the analytics looked OK, this looks good to me.

irevoire:
Yep, I tested it with @gmourier thanks! :crab: 

bors merge

PR #3122, opened 2022-11-23
Display the `dumpUid` as `null` until we create it

### LINKED ISSUE #3117: Implement the details on the dump tasks
**Describe the bug**
With the big refactor of the index-scheduler it looks like we forgot to implement the details of the dump tasks.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a dump

------
BUT, we want to update when we display the `dumpId`, since it's confusing to provide a `dumpId` when the task is `enqueued` or `failed`. We're only going to set the details when the task succeeds, and something is effectively written on disk.
This is more consistent with the other tasks.
It's breaking since it wasn't implemented like that in the previous release, but it was already specified for the previous release and had simply been missed.
poke @curquiza 

See this little description by @gmourier:
- `enqueued`  -> Obj `details` with `dumpUid` being null
- `processing`  -> Obj `details` with `dumpUid` being null
- `succeeded`  -> Obj `details` with `dumpUid` being specified
- `failed`  -> Obj `details` with `dumpUid` being null
- `canceled` -> Obj `details` with `dumpUid` being null

Kerollmops:
This PR fixes #3117 by displaying the `DumpCreation` `dumpUid` details field as `null` until we compute the dump and the task is finished.

### DIFF
--- a/dump/src/lib.rs
+++ b/dump/src/lib.rs
@@ -87,7 +87,7 @@ pub struct TaskDump {
     pub finished_at: Option<OffsetDateTime>,
 }
 
-// A `Kind` specific version made for the dump. If modified you may break the dump.
+// A `Kind` specific version made for the dump. If modified you may break the dump.
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub enum KindDump {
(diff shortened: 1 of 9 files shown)


irevoire (approved):
noice thank you @Kerollmops 
bors merge

PR #2662, opened 2022-08-11
Fix(cli): Clamp databases max size to a multiple of system page size

ManyTheFish:
fix #2659

PR #3160, opened 2022-11-29
Clamp databases max size to the page size

### LINKED ISSUE #3150: Config sizes in GiB instead of Gb
Meilisearch 0.29:
```--max-task-db-size=150Gb --max-index-size=250Gb```

Meilisearch 0.30:
```--max-task-db-size=150GiB --max-index-size=250GiB```

https://docs.meilisearch.com/learn/configuration/instance_options.html#max-index-size
**Current documentation:**
Sets the maximum size of the index. Value must be given in bytes or explicitly stating a base unit. For example, the default value can be written as `107374182400`, `'107.7Gb'`, or `'107374 Mb'`.

Kerollmops:
This PR fixes #3150 (again #2662). We fix it again, here, as we entirely rewrote the index scheduler and forgot about this little detail.

@irevoire Can I have your input on where we create the indexes in the tests? I want to use a non-page-size rounded value in the tests. This way, we can see this issue in the tests next time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1625,7 +1625,7 @@ dependencies = [
  "libc",
  "lmdb-rkv-sys",
  "once_cell",
- "page_size",
+ "page_size 0.4.2",
  "synchronoise",
  "url",
  "zerocopy",
@@ -1783,6 +1783,7 @@ dependencies = [
  "meili-snap",
  "meilisearch-types",
  "nelson",
+ "page_size 0.5.0",
  "roaring",
  "serde",
  "serde_json",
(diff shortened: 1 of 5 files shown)


irevoire:
Hey @Kerollmops, the code looks nice.
And here is where we create the index-scheduler in our test:
https://github.com/meilisearch/meilisearch/blob/main/index-scheduler/src/lib.rs#L1100-L1140

Kerollmops:
Thank you @irevoire, I just pushed a commit that makes the test use MB instead to see if the code fixes the issue. I hope we will catch this bug in the future if it comes back 🤞

irevoire (approved):
Nice thank you @Kerollmops 

bors merge

PR #3164, opened 2022-11-29
Improve the way we receive the documents payload

### LINKED ISSUE #3037: Improve the way we receive the documents payload
We currently use the above-linked method to receive the documents payload sent by the users. This function stores everything into memory, this is not optimal, we could probably do much better and stream that to disk at least with [a `BufWriter` or equivalent](https://doc.rust-lang.org/std/io/struct.BufWriter.html).

https://github.com/meilisearch/meilisearch/blob/9e189f5041fad0312fc0aa94038a890ad922e92c/meilisearch-http/src/routes/indexes/documents.rs#L197-L293

jiangbo212:
# Pull Request

## Related issue
Fixes #3037 

## What does this PR do?
- writing the playload to a temporary file via BufWritter
- deserialising the json tempporary file to an array of Objects by means of a memory map
- deserialising thie csv tempporary file by means of a memory map
- Adapted some read_json tests

## PR checklist


curquiza:
bors try

meilisearch-types/Cargo.toml
```diff
@@ -23,6 +23,8 @@ thiserror = "1.0.30"
 time = { version = "0.3.7", features = ["serde-well-known", "formatting", "parsing", "macros"] }
 tokio = "1.0"
 uuid = { version = "1.1.2", features = ["serde", "v4"] }
+memmap = "0.7.0"
+log = "0.4.17"
```
Kerollmops, line 27: Could you please order them alphabetically, please?

meilisearch-types/src/document_formats.rs
```diff
@@ -99,10 +102,10 @@ impl ErrorCode for DocumentFormatError {
 internal_error!(DocumentFormatError: io::Error);
 
 /// Reads CSV from input and write an obkv batch to writer.
-pub fn read_csv(input: impl Read, writer: impl Write + Seek) -> Result<usize> {
+pub fn read_csv(file: &File, writer: impl Write + Seek) -> Result<usize> {
     let mut builder = DocumentsBatchBuilder::new(writer);
-
-    let csv = csv::Reader::from_reader(input);
+    let mmap = unsafe { MmapOptions::new().map(file).unwrap() };
```
Kerollmops, line 107: Why not return an error here instead of unwrapping?

meilisearch-types/src/document_formats.rs
```diff
@@ -111,17 +114,49 @@ pub fn read_csv(input: impl Read, writer: impl Write + Seek) -> Result<usize> {
...
     let mut builder = DocumentsBatchBuilder::new(writer);
-    let reader = BufReader::new(input);
-
-    for result in serde_json::Deserializer::from_reader(reader).into_iter() {
-        let object = result.map_err(Error::Json).map_err(|e| (PayloadType::Ndjson, e))?;
-        builder
-            .append_json_object(&object)
-            .map_err(Into::into)
-            .map_err(DocumentFormatError::Internal)?;
+    let mmap = unsafe { MmapOptions::new().map(file).unwrap() };
```
Kerollmops, line 134: Same here, could you return the error instead of unwrapping?

meilisearch-types/src/document_formats.rs
```diff
@@ -130,30 +165,46 @@ pub fn read_ndjson(input: impl Read, writer: impl Write + Seek) -> Result<usize>
...
-        inner: Either<Vec<Object>, Object>,
-    }
-
-    let content: ArrayOrSingleObject =
-        serde_json::from_reader(reader).map_err(Error::Json).map_err(|e| (PayloadType::Json, e))?;
+/**
+ * https://serde.rs/stream-array.html
+ * https://github.com/serde-rs/json/issues/160
+ */
+fn array_each<'de, D, T, F>(deserializer: D, f: F) -> std::result::Result<io::Result<u64>, D::Error>
```
Kerollmops, line 172: Could you add a simple description of what this function do?
Also, it looks like it will reject empty arrays of documents, I am not sure we want that, it is not an issue to send zero documents to Meilisearch.

meilisearch-types/src/error.rs
```diff
@@ -323,6 +324,10 @@ impl Code {
             DuplicateIndexFound => {
                 ErrCode::invalid("duplicate_index_found", StatusCode::BAD_REQUEST)
             }
+            ReceivePayloadErr => ErrCode::internal(
+                "receive_payload_internal_exceptions",
```
Kerollmops, line 328: This is a new error code that we never defined in the doc. I would say that we should use an already known one [like the _internal_ one](https://docs.meilisearch.com/reference/errors/error_codes.html#internal).

meilisearch-types/src/document_formats.rs
```diff
@@ -111,17 +114,49 @@ pub fn read_csv(input: impl Read, writer: impl Write + Seek) -> Result<usize> {
...
-            .map_err(Into::into)
-            .map_err(DocumentFormatError::Internal)?;
+    let mmap = unsafe { MmapOptions::new().map(file).unwrap() };
+    let mut deserializer = serde_json::Deserializer::from_slice(&mmap);
+
+    match array_each(&mut deserializer, |obj: Object| builder.append_json_object(&obj)) {
+        Ok(Ok(count)) => debug!("serde json array size: {}", count),
+        Ok(Err(e)) => return Err(DocumentFormatError::Internal(Box::new(e))),
+        Err(_e) => {
+            debug!("deserialize single json");
```
Kerollmops, line 141: Would it be possible to replace this debug log by a little more explicit comment, here?

meilisearch-types/src/document_formats.rs
```diff
@@ -111,17 +114,49 @@ pub fn read_csv(input: impl Read, writer: impl Write + Seek) -> Result<usize> {
...
-        let object = result.map_err(Error::Json).map_err(|e| (PayloadType::Ndjson, e))?;
-        builder
-            .append_json_object(&object)
-            .map_err(Into::into)
-            .map_err(DocumentFormatError::Internal)?;
+    let mmap = unsafe { MmapOptions::new().map(file).unwrap() };
+    let mut deserializer = serde_json::Deserializer::from_slice(&mmap);
+
+    match array_each(&mut deserializer, |obj: Object| builder.append_json_object(&obj)) {
+        Ok(Ok(count)) => debug!("serde json array size: {}", count),
```
Kerollmops, line 138: Could you remove this debug log, too and return a `()`, please?

meilisearch-http/src/routes/indexes/documents.rs
```diff
@@ -1,12 +1,18 @@
-use std::io::{Cursor, ErrorKind};
-
+use crate::analytics::{Analytics, DocumentDeletionKind};
+use crate::error::MeilisearchHttpError;
+use crate::error::PayloadError::ReceivePayloadErr;
+use crate::extractors::authentication::policies::*;
+use crate::extractors::authentication::GuardedData;
+use crate::extractors::payload::Payload;
+use crate::extractors::sequential_extractor::SeqHandler;
+use crate::routes::{fold_star_or, PaginationView, SummarizedTaskView};
```
Kerollmops, line 8: This should be kept separated by a newline under the foreign-library import.

meilisearch-http/src/routes/indexes/documents.rs
```diff
@@ -227,26 +227,51 @@ async fn document_addition(
...
+                return err;
+            }
+        }
     }
-    if buffer.is_empty() {
+
+    if let Err(e) = buffer.flush() {
+        error!("bufWriter flush error: {}", e);
+        return err;
+    };
```
Kerollmops, line 263: Would it be possible to always declare the whole `Err(MeilisearchHttpError::Payload(ReceivePayloadErr))` instead of declaring it as an `err` at the top, please?

Kerollmops: What do you think about storing the error information directly as a string in the `ReceivePayloadErr` variant?

Kerollmops (changes requested):

meilisearch-http/src/routes/indexes/documents.rs
irevoire, line 263: We're in an async context here, creating a file is already far from perfect, but doing write and flush is absolutely forbidden!

You should try to use the tokio io-util API for all these operations;
- [BufWriter](https://docs.rs/tokio/latest/tokio/io/struct.BufWriter.html)
- [flush](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWriteExt.html#method.flush)

meilisearch-types/src/document_formats.rs
irevoire, line 172: Ouuuh this is niiiice :ok_hand:

irevoire (changes requested):

meilisearch-http/src/routes/indexes/documents.rs
jiangbo212: ok. I will try it with the tokio io-util API

Cargo.lock
```diff
@@ -2396,6 +2398,16 @@ version = "2.5.0"
...
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6585fd95e7bb50d6cc31e20d4cf9afb4e2ba16c5846fc76793f11218da9c475b"
+dependencies = [
+ "libc",
+ "winapi",
+]
+
 [[package]]
 name = "memmap2"
```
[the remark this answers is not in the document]
jiangbo212: I'll adjust it.

meilisearch-types/Cargo.toml
jiangbo212: I'll adjust it.

meilisearch-types/src/error.rs
jiangbo212: Adjusting it to an internal code is no problem. I originally meant to show this better

meilisearch-types/src/document_formats.rs
jiangbo212: I'll adjust it.

meilisearch-types/src/error.rs
irevoire: Yep sorry, currently, the way we see the error code is something along the lines of:
If there is an error code, there is a web page associated with it that explains why it happened and how you can solve your issue: https://docs.meilisearch.com/reference/errors/error_codes.html#error-codes

And since you can't do anything about an internal error like yours, we just output the internal error code and then tries to be explicit with the error message so when users encounter it, they can help us to understand where it happened.

meilisearch-http/src/routes/indexes/documents.rs
jiangbo212: Storing a different error message to `ReceivePayloadErr` does not seem necessary here. Exceptions that cause storing paylod to a temporary file are almost always an OS or network problem for the user. What is more important to see in this case is the error log.

jiangbo212:
@Kerollmops @irevoire Requested changes have been resolved

meilisearch-http/src/routes/indexes/documents.rs
```diff
@@ -227,26 +230,54 @@ async fn document_addition(
 
     let (uuid, mut update_file) = index_scheduler.create_update_file()?;
 
-    // TODO: this can be slow, maybe we should spawn a thread? But the payload isn't Send+Sync :weary:
-    // push the entire stream into a `Vec`.
-    // If someone sends us a never ending stream we're going to block the thread.
-    // TODO: Maybe we should write it to a file to reduce the RAM consumption
-    // and then reread it to convert it to obkv?
-    let mut buffer = Vec::new();
+    let temp_file = match NamedTempFile::new() {
```
irevoire, line 233: There are a lot of drawbacks to using a `NamedTempFile`, and I'm not sure I see the point. We're not persisting the file or anything; it's only used as a temporary buffer while we wait to write the data in another file :thinking: 
Why don't we use a simpler [TempFile](https://docs.rs/tempfile/latest/tempfile/fn.tempfile.html)?

meilisearch-http/src/routes/indexes/documents.rs
```diff
@@ -227,26 +230,54 @@ async fn document_addition(
...
+        }
+    };
+    debug!("temp file path: {:?}", temp_file.as_ref());
+    let buffer_file = match temp_file.reopen() {
+        Ok(buffer_file) => buffer_file,
+        Err(e) => {
+            error!("reopen payload temporary file error: {}", e);
+            return Err(MeilisearchHttpError::Payload(ReceivePayloadErr));
+        }
+    };
```
irevoire, line 247: I don't understand why we're doing that either?
Shouldn't a simple `as_file_mut` do the trick?

Or, if we can use a normal `TempFile` then we can delete this whole block.

meilisearch-http/src/routes/indexes/documents.rs
irevoire: Hey, sorry, but I'm 100% with @Kerollmops on this one.
> What is more important to see in this case is the error log.

Most users don't read or can't access their logs, they're going to report their bug from the HTTP error we return, and thus, we definitely want to be able to differentiate all the possible errors.

You should update the error type to something like that:
```
#[error("Error while writing the playload to disk: {}", .0)]
```

Also, we don't want to struggle with lifetimes in case of errors; store the message as a `String`.

irevoire (changes requested):
Hey, we're almost there, there is just a few things I would like to change, but in terms of algorithm it looks good to me :1st_place_medal:

meilisearch-http/src/routes/indexes/documents.rs
jiangbo212: I also tried `tempfile()`  and it doesn't seem to work well for this scenario. To persist the byte stream of the payload to the tempfile, and read it again, the `seek` method must be used after the entire byte stream has been transferred, which doesn't seem like a good choice especially if the payload is particularly large. Using `tempfile()`and `seek()` introduces a new problem, when `read_json` fails if it tries to deserialize to seq first, it currently uses `serde_json::from_reader` to deserialize, and in this case the contents of tempfile are not available.

meilisearch-http/src/routes/indexes/documents.rs
jiangbo212: `as_file_mut` returns a reference to` file`, `tokio::fs::file` requires file. tempfile will have both read and write handles, just using temp_file is not enough, so it need `reopen` . I made some adjustments, you can review it again.

meilisearch-http/src/routes/indexes/documents.rs
irevoire: Huum, not sure I see the issue with using `seek` on a very large file; it's made for that and pretty much answers in constant time, right?

> when read_json fails if it tries to deserialize to seq first, it currently uses serde_json::from_reader to deserialize, and in this case the contents of tempfile are not available.

And I'm also not sure I understood this part; sorry :grimacing: 
`tempfile()` returns a real `File`. Thus you can read and write in it or give it to `serde_json::from_reader`?
How are the contents of tempfile not available?

irevoire:
Oops, also second big comment, for the async part we should definitely use an `AsyncBufWriter` or we're going to have terrible performances: https://docs.rs/tokio/latest/tokio/io/struct.BufWriter.html !

meilisearch-http/src/routes/indexes/documents.rs
jiangbo212: I used the same code as you yesterday, but when I tested it, I found many problems during deserialization. When debugging, it was found that the file read during deserialization was empty. I tried it again today and it works perfectly. Thank you for your help.👍

jiangbo212:
Sorry, I seem to have mis-pressed some buttons. I originally wanted you to review the changes. @irevoire @Kerollmops

irevoire (approved):
I think we're good for me!

curquiza:
Hello @jiangbo212 sorry there is git conflict with your PR, can you solve them?

jiangbo212:
conflicts resolved @curquiza

meilisearch/src/routes/indexes/documents.rs
```diff
@@ -227,26 +231,51 @@ async fn document_addition(
...
-    // and then reread it to convert it to obkv?
-    let mut buffer = Vec::new();
+    let temp_file = match tempfile() {
+        Ok(temp_file) => temp_file,
+        Err(e) => {
+            return Err(MeilisearchHttpError::Payload(ReceivePayloadErr(Box::new(e))));
+        }
+    };
+
+    let mut buffer = File::from_std(temp_file);
```
irevoire, line 241: > Oops, also second big comment, for the async part we should definitely use an AsyncBufWriter or we're going to have terrible performances: https://docs.rs/tokio/latest/tokio/io/struct.BufWriter.html !

I think we lost this change? Or I forgot about it and approved the last time :grimacing: 

```suggestion
    let async_file = File::from_std(temp_file);
    let mut buffer = BufWriter::new(async_file);
```

irevoire (changes requested):

curquiza:
Hello @jiangbo212 
Sorry but there is a (new?) git conflict.. sorry for this and thanks for your reactivity!

meilisearch/src/routes/indexes/documents.rs
jiangbo212: Already processed

irevoire (approved):
PERFECT, thanks a lot @jiangbo212 🎉

meilisearch/src/error.rs
```diff
@@ -95,6 +95,8 @@ pub enum PayloadError {
     MalformedPayload(serde_json::error::Error),
     #[error("A json payload is missing.")]
     MissingPayload,
+    #[error("Error while writing the playload to disk: `{0}`.")]
+    ReceivePayloadErr(Box<dyn std::error::Error + Send + Sync + 'static>),
```
Kerollmops, line 99: We need to rename this variant. There already is a `PayloadErr` part in the enum name.

meilisearch-types/src/document_formats.rs
```diff
@@ -111,17 +115,56 @@ pub fn read_csv(input: impl Read, writer: impl Write + Seek) -> Result<usize> {
...
+        }
+        Ok(Err(e)) => return Err(DocumentFormatError::Internal(Box::new(e))),
+        // Prefer deserialization as a json array. Failure to do deserialisation using the traditional method.
+        Err(_e) => {
+            #[derive(Deserialize, Debug)]
+            #[serde(transparent)]
+            struct ArrayOrSingleObject {
+                #[serde(with = "either::serde_untagged")]
+                inner: Either<Vec<Object>, Object>,
+            }
```
Kerollmops, line 155: We need to only deserialize the `Object` and not the `Vec<Object>` type.

meilisearch/src/routes/indexes/documents.rs
```diff
@@ -227,26 +231,52 @@ async fn document_addition(
 
     let (uuid, mut update_file) = index_scheduler.create_update_file()?;
 
-    // TODO: this can be slow, maybe we should spawn a thread? But the payload isn't Send+Sync :weary:
-    // push the entire stream into a `Vec`.
-    // If someone sends us a never ending stream we're going to block the thread.
-    // TODO: Maybe we should write it to a file to reduce the RAM consumption
-    // and then reread it to convert it to obkv?
-    let mut buffer = Vec::new();
+    let temp_file = match tempfile() {
```
Kerollmops, line 234: We need to rewrite this a little bit. There are too many uneeded `;`.

Kerollmops (approved):
Thank you very much for your PR. It will make Meilisearch memory efficient as heck!
bors merge

PR #3170, opened 2022-11-30
Re-enable importing from dumps v1

### LINKED ISSUE #2985: Add the support of the dump v1 (Meilisearch v0.20.0 and before)
Currently, users using Meilisearch v0.20.0 and before need to do TWO migrations to upgrade Meilisearch to v0.30.0. [See the related documentation](https://docs.meilisearch.com/learn/advanced/updating.html#update-to-the-latest-meilisearch-version).

This is due to the fact we dropped the support of dump v1 some releases ago.

It means it will not be easy for all our users to pass from their current version to v1. This is something we would like to avoid because in the best-case scenario, all the Meilisearch users will use Meilisearch v1.0.0 and after.

We would like to re-introduce the dump v1 support of Meilisearch to ensure people using Meilisearch v0.20.0 and before can migrate to Meilisearch v1.0.0 in one migration and not two.

## Impacted teams

## TODO

edited by @dureuill

- [ ] Implement the changes and merge to the `main` branch => https://github.com/meilisearch/meilisearch/pull/3170
- [ ] Update the [related spec](https://github.com/meilisearch/specifications/blob/main/text/0105-dumps-api.md#34-dump-version-support) => https://github.com/meilisearch/specifications/pull/203

dureuill:
# Pull Request

## Related issue
Fixes #2985 

### User standpoint

- Allows importing dumps version 1 (exported with Meilisearch <=v0.20) to modern-day Meilisearch
- Tasks of type "Customs" are skipped, with a warning
- Tasks of status "enqueued" are skipped, with a warning
- The "WordsPosition" ranking rule is skipped when encountered in the ranking rules, with a warning.

After an import from a v1 dump, it is recommended that a user checks each index and its settings.

### Implementation standpoint

- Add a dump v1 reader based on the one by @irevoire 
- Add a v1_to_v2 compatibility layer based on the v2_to_v3 one
  - as v2 requires UUIDs, the v1 indexes are mapped to UUIDs built from their position in the metada file: the first index is given UUID all zeroes, the second one UUID `00000000-0000-0000-0000-000000000001`, and so on... This should have no bearing on the final indexes because v6 is not using UUIDs, but this allows us to correctly identify which tasks belong to which index.
- Modify the v2_to_v3 compatibility layer to account for the fact that the reader can actually be a v1_to_v2 compat layer
- Make some base dump types Clone
- impl Display for v2::settings::Criterion

## PR checklist


dump/src/error.rs
```diff
@@ -3,8 +3,6 @@ use thiserror::Error;
 
 #[derive(Debug, Error)]
 pub enum Error {
-    #[error("The version 1 of the dumps is not supported anymore. You can re-export your dump from a version between 0.21 and 0.24, or start fresh from a version 0.25 onwards.")]
-    DumpV1Unsupported,
```
irevoire, line 7: niice

dump/src/reader/v1/mod.rs
```diff
@@ -1,173 +1,265 @@
...
-                        .file_name()
-                        .to_str()
-                        .ok_or(Error::BadIndexName)?
-                        .to_string(),
-                    &entry.path(),
-                )?);
-            }
+        for index in metadata.indexes.into_iter() {
+            let index_path = dump.path().join(&index.uid);
+            indexes.push(V1Index::new(index_path, index));
```
irevoire, line 94: From what I see, the indexes are actually ordered in the dump, right?

dump/src/reader/v1/mod.rs
```diff
@@ -1,173 +1,265 @@
...
+
+        let dump = V1Reader::open(dir).unwrap();
+
+        // top level infos
+        assert_eq!(dump.date(), None);
+
+        // indexes
+        let mut indexes = dump.indexes().unwrap().collect::<Result<Vec<_>>>().unwrap();
+        // the index are not ordered in any way by default
+        indexes.sort_by_key(|index| index.metadata().uid.to_string());
```
irevoire, line 200:
```suggestion
```

Thus we don't need these lines

dump/src/reader/compat/v1_to_v2.rs
```diff
@@ -0,0 +1,414 @@
...
+        self.from
+            .index_uuid()
+            .into_iter()
+            .enumerate()
+            // we use the index of the index 😬 as UUID for the index, so that we can link the v2::Task to their index
+            .map(|(index, index_uuid)| v2::meta::IndexUuid {
+                uid: index_uuid.uid,
+                uuid: uuid::Uuid::from_u128(index as u128),
+            })
+            .collect()
```
irevoire, line 39: If the indexes end up not being ordered (which I think is not the case, see my other comment).
We should probably sort them here.

irevoire (changes requested):
Hey, it looks super good.

Could you add one test at the bottom of `dump/src/reader/mod.rs` that imports our dump v1 through all our compat layers, please? :grin:

dump/src/reader/v1/mod.rs
dureuill: yes, that is the case: in the `metadata.json` of the dumps v1, we have a list of indexes, so it is definitely ordered.

I guess I copy pasted this from v2 where indexes weren't ordered

dump/src/reader/compat/v1_to_v2.rs
dureuill: the indexes are definitely ordered, see https://github.com/meilisearch/meilisearch/pull/3170#discussion_r1037066016

irevoire: Nice thanks!

dump/src/reader/v1/mod.rs
dureuill: Lines are removed, thanks for spotting that!

dureuill:
> Could you add one test at the bottom of dump/src/reader/mod.rs that imports our dump v1 through all our compat layers, please? 😁

Oh sorry I didn't write a message to specify, but I did that.

Also I detailed some of the quirks of the current implementation, format-wise, in https://github.com/meilisearch/meilisearch/issues/2985#issuecomment-1333840498

irevoire (approved):
Nice, we're good to me!

bors merge

irevoire:
bors merge

Issue #3147, opened 2022-11-28
Internal process: remove the usage of the `stable` branch


curquiza:
Currently, when doing releases of Meilisearch, we use and provide a `stable` branch. Basically, the `stable` branch contains the latest stable commit, so it corresponds to the latest Meilisearch release.
See [our full process](https://github.com/meilisearch/engine-team/blob/main/resources/meilisearch-release.md).

However, we found some issues using and providing this `stable` branch 
- this adds complexity by adding an intermediate step for each release -> merging the `release-vX.X.X` branch into stable
- this leads to a struggle with some git conflicts

## Pros and cons of removing `stable`

Pro
- we remove complexity: we only do GitHub release on the `release-vX.X.X` and the `release-vX.X.X` branch is brought back to `main` once the release is done. When a patched release is needed (ex: v0.30.1), we start the new `release-v0.30.1` branch from `release-v0.30.0` instead of `stable`
- we remove some git conflict. However, not all will be avoided, for example, when bringing back `release-vX.X.X` into `main` after a release.
- Integration team, cloud team, and docs team do not use `stable` in their internal processes.

Cons
- we don't provide a git reference (`stable`) anymore to access the latest stable version of Meilisearch without caring of the version number
- the official documentation of Meilisearch refers to it [here](https://docs.meilisearch.com/learn/getting_started/quick_start.html#setup-and-installation) (in the `Source` tab)

## Final possibility

- [x] Stop using `stable` and rename it `stable-archived` to avoid removing it, and make anyone understand not to use it.
- [x] Only create GitHub releases on the `release-vX.X.X` branch and bring back the `release-vX.X.X` commits into `main`.
- [x] When releasing a stable GitHub release of Meilisearch, create a git tag `latest` to replace the `stable` reference we provided with the `stable` branch -> PR submitted https://github.com/meilisearch/meilisearch/pull/3172
- [x] Update the documentation part accordingly -> https://github.com/meilisearch/documentation/pull/2026
- [x] Update our [internal process guide](https://github.com/meilisearch/engine-team/blob/main/resources/meilisearch-release.md) -> https://github.com/meilisearch/engine-team/pull/35

curquiza:
Closing this issue, `stable` branch is officially out of our process

PR #3172, opened 2022-12-01
Add CI to push a latest git tag for every stable Meilisearch release

curquiza:
Fixes partially #3147.

- Add a CI to add/update a `latest` git tag ONLY when releasing a stable version of Meilisearch (not for pre-release or for custom tag for prototypes)
- Update the Docker CI to avoid being triggered when `latest` git tag is pushed: the CI is still triggered when a git tag is pushed, except for the `latest` tag.
Reminder: the `latest` Docker image is already created and pushed when releasing a stable version of Meilisearch. This step is already present in our current Docker CI.

### DIFF
--- a/.github/workflows/latest-git-tag.yml
+++ b/.github/workflows/latest-git-tag.yml
@@ -1,0 +1,19 @@
+# Create or update a latest git tag when releasing a stable vesrsin of Meilisearch
+name: Update latest git tag
+on:
+  workflow_dispatch:
+  release:
+    types: [released]
+
+jobs:
+  update-latest-tag:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: rickstaa/action-create-tag@v1
+        with:
+          tag: "latest"
+          message: "Latest stable release of Meilisearch"
+          # Move the tag if `latest` already exists
+          force_push_tag: true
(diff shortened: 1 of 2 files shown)


.github/workflows/latest-git-tag.yml
```diff
@@ -0,0 +1,19 @@
...
+  update-latest-tag:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: rickstaa/action-create-tag@v1
+        with:
+          tag: "latest"
+          message: "Latest stable release of Meilisearch"
+          # Move the tag if `latest` already exists
+          force_push_tag: true
```
irevoire, line 18: Does this remove the previous latest tag?
That's why we don't need to delete it?

irevoire (approved):
From what I understood it should works :grin:

.github/workflows/latest-git-tag.yml
curquiza: It moves the tag, you cannot push two similar tags in the same git repo. So it uses the `--force` command
https://stackoverflow.com/questions/8044583/how-can-i-move-a-tag-on-a-git-branch-to-a-different-commit

curquiza:
bors merge

.github/workflows/latest-git-tag.yml
irevoire: Nice thank you!

PR #3174, opened 2022-12-01
Allow wildcards at the end of index names for API Keys and Tenant tokens

### LINKED ISSUE #2788: Use wildcard (`*`) at the end of the index names when creating API keys
Following
- this product discussion: https://github.com/meilisearch/product/discussions/491
- this spec: https://github.com/meilisearch/specifications/pull/189

## Goal

Allow using the wildcard `*` at the end of the index name (in `indexes` field) when creating API keys

## How does it work?

When [creating an API key](https://docs.meilisearch.com/reference/api/keys.html#create-a-key), the `*` should be available this ways:

```bash
curl \
  -X POST 'http://localhost:7700/keys' \
  -H 'Authorization: Bearer MASTER_KEY' \
  --data-binary '{
    "description": "Add documents: Products API key",
    "actions": ["documents.add"],
    "indexes": ["products_*"],
    "expiresAt": "2042-04-02T00:42:42Z"
  }'
```

This means the newly created keys will be available for the `products_main`, `products_dev`, but not for `products`

#### Additional explanation of the matching

The `books*` expression will match the following indexes: `books`, `books_staging`, but NOT `book` (without `s`)

#### Errors

- if no indexes are matched, no error are thrown during the API key creation. However this key will be useless until a matching index is created

## Impacted teams

Pinging @meilisearch/docs-team, but also @meilisearch/integration-team (I'm not sure for you)

--- 

## Todo
- [ ] Implementation the changes -> Open to external contributions
- [X] Update the spec -> https://github.com/meilisearch/specifications/pull/189

---
Feel free to ask any questions to implement this issue. The core team is available to guide you through this contribution, even if it's your first one.
We recommend referring to the [CONTRIBUTING.md guidelines](https://github.com/meilisearch/meilisearch/blob/main/CONTRIBUTING.md), especially if it's your first contribution to this repository.
Thanks in advance for your involvement ❤️

### LINKED ISSUE #2908: Allow wildcard (`*`) at the end of the index names when creating tenant tokens
Following this product discussion: https://github.com/meilisearch/product/discussions/491#discussioncomment-3048342
Related to this spec: https://github.com/meilisearch/specifications/pull/189

## Goal

Allow using the wildcard `*` at the end of the index name (in `searchRules` field) when creating tenant tokens.

## How does it work?

To create a tenant token, you need to provide a `searchRules` field containing one or several index names.

```json
{
    "searchRules": {
        "medical_records": {
            "filter": "user_id = 1"
        },
        "medical_records_staging": {
            "filter": "user_id = 1"
        },     
    }
}
```

Also, the `*` is already allowing as following, meaning "all the indexes":

```json
{
    "searchRules": {
        "*": {
            "filter": "user_id = 1"
        }
    }
}
```

But the users need an intermediate usage of the `*` character: allowing to use the `*` character at the end of the index name.

Instead of passing:

```json
{
    "searchRules": {
        "medical_records": {
            "filter": "user_id = 1"
        },
        "medical_records_staging": {
            "filter": "user_id = 1"
        },     
    }
}
```

The users could pass:

```json
{
    "searchRules": {
        "medical_records*": {
            "filter": "user_id = 1"
        }
}
``` 

And this tenant will be available for all indexes matching `medical_records*`, so `medical_records` and `medical_records_staging` will match, but not `medical_record` (without `s`)

#### Errors

- if no indexes are matched, no error are thrown during the tenant token creation. However this token will be useless until a matching index is created

## Todo

- [ ] Implementation the changes by opening a PR in this repository -> open to external contributions. You can be inspired by [this PR](https://github.com/meilisearch/meilisearch/pull/2869)
- [X] Update the spec regarding this -> https://github.com/meilisearch/specifications/pull/189

## Impacted teams

Pinging @meilisearch/docs-team, but also @meilisearch/integration-team (I'm not sure for you)

Kerollmops:
This PR introduces the wildcards at the end of the index names when identifying indexes in the API Keys and tenant tokens. It fixes #2788 and fixes #2908. This PR is based on @akhildevelops' work.

Note that when a tenant token filter is chosen to restrict a search, it is always the most restrictive pattern that is chosen. If we have an index pattern _prod*_ that defines _filter1_ and _p*_ that defines _filter2_, the engine will choose _filter1_ over _filter2_ as it is defined for a most restrictive pattern, _prod*_. This restrictiveness is defined by 1. is it exact, without _*_ 2. the length of the pattern.

It is a continuation of work that has already started and should close #2869.


meilisearch-types/src/index_uid_pattern.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+impl fmt::Display for IndexUidPatternFormatError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "`{}` is not a valid index uid pattern. Index uid patterns \
+            can be an integer or a string containing only alphanumeric \
+            characters, hyphens (-), underscores (_), and \
+            optionally end with a star (*).",
```
Kerollmops, line 112: Hey @gmourier, What do you think of this new error message? I adapted the one form the index uid errors. It if raised when an invalid pattern is used, e.g., `"sal*s"`, `""` in the API Keys or Tenant Token indexes fields.

Kerollmops:
Hey @ManyTheFish and @irevoire,

Could you review this PR, please? It adds the possibility to specify an API Key and a Tenant Token with a pattern that can optionally end with a `*`. I advise you to review the first commits first and only the final merge commit at the end.

dump/src/reader/compat/v5_to_v6.rs
```diff
@@ -181,10 +181,8 @@ impl CompatV5ToV6 {
                     .indexes
                     .into_iter()
                     .map(|index| match index {
-                        v5::StarOr::Star => v6::StarOr::Star,
-                        v5::StarOr::Other(uid) => {
-                            v6::StarOr::Other(v6::IndexUid::new_unchecked(uid.as_str()))
-                        }
+                        v5::StarOr::Star => v6::IndexUidPattern::all(),
+                        v5::StarOr::Other(uid) => v6::IndexUidPattern::new_unchecked(uid.as_str()),
```
irevoire, line 185: Not sure it should be a `new_unchecked`.
If someone had an API key with something invalid, we should ignore his key and send a warning, I think?

Currently, let me know if I'm wrong, but `indexes_*_doggo` is a valid API key name that should be forbidden in the new version, right?

meilisearch-auth/src/store.rs
```diff
@@ -214,11 +214,28 @@ impl HeedAuthStore {
...
+        match self.action_keyid_index_expiration.get(&rtxn, &tuple)? {
+            Some(expiration) => Ok(Some(expiration)),
+            None => {
+                let tuple = (&uid, &action, None);
+                for result in self.action_keyid_index_expiration.prefix_iter(&rtxn, &tuple)? {
+                    let ((_, _, index_uid_pattern), expiration) = result?;
+                    if let Some((pattern, index)) = index_uid_pattern.zip(index) {
+                        let index_uid_pattern = str::from_utf8(pattern)?.to_string();
+                        // TODO I shouldn't unwrap here but rather return an internal error
+                        let pattern = IndexUidPattern::try_from(index_uid_pattern).unwrap();
```
irevoire, line 230: Can we do it in this PR?

meilisearch-auth/src/lib.rs
```diff
@@ -192,42 +174,54 @@ impl Default for AuthFilter {
...
                     None
                 }
             }
             Self::Map(map) => {
-                map.get(index).or_else(|| map.get("*")).map(|isr| isr.clone().unwrap_or_default())
+                // We must take the most retrictive rule of this index uid patterns set of rules.
+                map.iter()
+                    .filter(|(pattern, _)| pattern.matches_str(index))
+                    .max_by_key(|(pattern, _)| (pattern.is_exact(), pattern.len()))
+                    .and_then(|(_, rule)| rule.clone())
```
irevoire, line 217: That's nice! Didn't realize that could work with a simple loop :rock:

irevoire (approved):
Nice PR and nice tests! :hedgehog: 

I left a few questions that are not critical. I don't know if we want to do something about it

meilisearch-types/src/index_uid_pattern.rs
```diff
@@ -0,0 +1,125 @@
...
+        )
+    }
+}
+
+impl Error for IndexUidPatternFormatError {}
+
+impl ErrorCode for IndexUidPatternFormatError {
+    fn error_code(&self) -> Code {
+        // TODO should I return a new error code?
+        Code::InvalidIndexUid
```
ManyTheFish, line 123: here it's no more an IndexUid but an IndexUidPattern, so I think it would be better to have a new error code. 🤔 (poke @gmourier)

meilisearch/src/extractors/authentication/mod.rs
```diff
@@ -199,14 +199,17 @@ pub mod policies {
             token: &str,
             index: Option<&str>,
         ) -> Option<AuthFilter> {
+            // Tenant token will always define an index.
```
ManyTheFish, line 202:
```suggestion
            // Only the search route is accessible to Tenant tokens, this route always defines an index.
```

ManyTheFish (changes requested):

dump/src/reader/compat/v5_to_v6.rs
irevoire: Oh yes, you're right, perfect then :ok_hand:

meilisearch-types/src/index_uid_pattern.rs
irevoire: Personally, I don't think this error code matter that much for the end user, but I may be wrong :thinking:

irevoire (dismissed):
Good for me :poultry_leg: 
Let's wait for the error_code thingy before merging though :+1:

meilisearch-types/src/index_uid_pattern.rs
gmourier: Here is what we have for the swap indexes resource when you give a wrong pattern for an index uid in the index array.

We could follow the same rule to stay consistent.

https://github.com/meilisearch/specifications/blob/main/text/0191-swap-indexes-api.md#34131-synchronous-errors

> Sending an indexes array containing an invalid index uid format returns an [invalid_swap_indexes](https://github.com/meilisearch/specifications/blob/main/text/0061-error-format-and-definitions.md#invalid_swap_indexes) error.

It means keeping `invalid_api_key_indexes` for the error code and injecting the `invalid_index_uid` variant for the `message` for the sake of consistency.

Kerollmops: @ManyTheFish @irevoire, the `IndexUid` struct was returning an `InvalidIndexUid`. I think I should keep what I wrote here and also return this `Code` for when an `IndexUidPattern` is wrong, right? Doing that will keep the consistency of the API and the error message is custom and really easy to understand.

ManyTheFish: @Kerollmops, I'm not sure to understand.
Now you always take an `IndexUidPattern` in parameter. No matter if there is a `*` at the end or not, it's an `IndexUidPattern` no? 🤔

Kerollmops: After a quick call with @ManyTheFish we decided that we will keep the error code I currently use: `Code::InvalidIndexUid`. The error message is sent by Meilisearch and is clear enough for a user to understand. Thank you everybody 😍

irevoire (approved):
perfect, thanks :octopus: 

bors merge

meilisearch-types/src/index_uid_pattern.rs
gmourier: I think it's pretty good if that variant is only present when the `invalid_api_key_indexes` is only shown for that specific case. Sorry to answer late!

gmourier: Another question that comes to mind but when a tenant token is invalid (no way to check it at generation yet since they are not generated server-side), we have decided in the past that it will return an `invalid_api_key` error. Does it change with that implem?

Kerollmops: The only thing that should change is the error message that you can read above. Nothing will be different than the previous version of the engine that was not supporting index uid patterns.

PR #3187, opened 2022-12-05
Use ubuntu-latest for Docker CI

curquiza:
Ensuring it's worth to still use a personal GitHub runner

Just push `test-docker-ci` tag on this branch, to check using `ubuntu-latest`, so the native GitHub runner, is way more slower than our personal GitHub runner
https://github.com/meilisearch/meilisearch/actions/runs/3621631699


curquiza:
The CI crashes, with a `137` error, so an issue with the memory consumption. We will keep using our own self-hosted GitHub runner.

curquiza:
Closing this then!

PR #3188, opened 2022-12-05
re-enable the dump test on the dates

irevoire:
I just noticed that we have the real date in the dump-v1 contrarily to the dump-v2/3/4/5, thus we can ensure it doesn't change unexpectedly :+1:


dureuill (approved):
Nice catch! Assuming CI is green, this looks good to me!

irevoire:
bors merge

PR #3207, opened 2022-12-07
Add release check when starting latest CI

curquiza:
Adding this to have the same kind of check before starting to move the latest tag

Capture d’écran 2022-12-07 à 12 18 33

Also, removing an un-unused script


.github/workflows/latest-git-tag.yml
```diff
@@ -1,11 +1,20 @@
-# Create or update a latest git tag when releasing a stable vesrsin of Meilisearch
```
Kerollmops, line 1: Lol NOPE

Kerollmops (approved):
Otherwise I like that!

Kerollmops (approved):
LGTM!
bors merge

PR #3208, opened 2022-12-07
Stop snapshotting the version of meilisearch in the dump

irevoire:
It might change, and we don't want to update this test every time we make a new release.


Kerollmops (approved):
Thank you, merging!
bors merge

PR #3213, opened 2022-12-07
Fix the instance-uid in the data.ms

### LINKED ISSUE #3214: The `instance-uid` file is badly formatted
**Describe the bug**
Since the `v0.30.0`, the file containing the `instance-uid` is badly formatted.
It's supposed to be written as a simple string so our users can share it with us easily if they need to, but after the huge refactor of the index-scheduler, it was written as bytes directly, which are a bit cumbersome to copy-paste/share.

**To Reproduce**
1. Run meilisearch
2. Read the `data.ms/instance-uid`

**Expected behavior**
It's a human-readable string

**Screenshots**

**Meilisearch version:**
Meilisearch `v0.30.0` and `v0.30.1`

**Additional context**
Introduced in https://github.com/meilisearch/meilisearch/pull/2763

irevoire:
We were writing the instance-uid as bytes instead of strings in the data.ms, and thus we were unable to parse it later. Also, it was less practical for our users to retrieve it and send it to us.

Fix #3214


curquiza:
I validate `git` part, the branch is pointing to the right one

Kerollmops (approved):
Noice, thank you!
bors merge

PR #3229, opened 2022-12-11
Add a nightly CI: create every day a `nightly` Docker tag based on the latest commit on `main`

### LINKED ISSUE #3195: `latest` Docker image update: do it only if the GitHub release is considered as the latest one
Currently our [current CI](https://github.com/meilisearch/meilisearch/blob/main/.github/workflows/publish-docker-images.yml) (to create Docker images when a tag is pushed) will push the `latest` Docker image to DockerHub in term of times and not in term of version.

For example:
- we release v1.10.0 -> `latest` Docker image will correspond to the `v1.10.0` release
- then, later, we release v1.1.12 -> `latest` Docker image will now correspond to the `v1.1.12` release despite this is not the latest release, `v1.10.0` is!

We need to create the `latest` Docker image and push it to DockerHub only if the release is considered the latest one by GitHub.

The latest release is defined this way on GitHub when creating a new GitHub release.

Capture d’écran 2022-12-01 à 17 01 59

Currently the CI is triggered when a tag is created. We might need to change this (trigger for releases) to achieve the goal of the issue.

curquiza:
Also, fixes #3195

Easy to follow with the commits
- In the Docker CI:
  - create every day a `nightly` Docker tag based on the latest commit on `main`
  - check if the release is the latest one, before creating the `latest` Docker tag. A script has been added.
  - add the `worflow_dispatch` event to trigger the CI to build the `nightly` tag when we want (always on the latest commit on `main`)
- In multiple CIs: replace the `released` type by `published`, see [here](https://stackoverflow.com/questions/59319281/github-action-different-between-release-created-and-published) why. Will not impact anything, but will prevent to fail our future automation
- Remove a useless CI (code coverage, not used for 1 year)
- Remove useless lines (comments and CI logic) that don't have any impact


.github/workflows/milestone-workflow.yml
```diff
@@ -31,8 +31,6 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       is-patch: ${{ steps.check-patch.outputs.is-patch }}
-    env:
-      MILESTONE_VERSION: ${{ github.event.milestone.title }}
```
curquiza, line 35: No impact, the value is already available in the same file, line 18

.github/workflows/publish-docker-images.yml
```diff
@@ -65,19 +74,19 @@ jobs:
...
+            type=raw,value=nightly,enable=${{ github.event_name != 'push' }}
             type=semver,pattern=v{{major}}.{{minor}},enable=${{ steps.check-tag-format.outputs.stable == 'true' }}
-            type=raw,value=latest,enable=${{ steps.check-tag-format.outputs.stable == 'true' }}
+            type=raw,value=latest,enable=${{ steps.check-tag-format.outputs.stable == 'true' && steps.check-tag-format.outputs.latest == 'true' }}
 
       - name: Build and push
         uses: docker/build-push-action@v3
         with:
           # We do not push tags for the cron jobs, this is only for test purposes
-          push: ${{ github.event_name != 'schedule' }}
```
curquiza, line 80: We now want to push when we run a cron job, since `nighlty` tag is run at the moment.

Kerollmops: > since `nightly` tag is run at the moment.
I don't understand this part. Can you explain what you mean?

Kerollmops (commented):
Looks good to me. Not an easy job to write or even review bash scripts combined with YAML.
I am waiting for your answer to my comment.

.github/workflows/publish-docker-images.yml
curquiza: Sorry, not clear indeed,

Before: we ran a cron job every day, that built the docker image on the `main` branch, but did not push any image related to this build. it was only for a check.

Now: we ran a cron job every day, that built the docker image on the `main` branch, but also push this built. The name of the image we push is `nightly`.

PR #3245, opened 2022-12-14
Enable create_raw_index(...) to specify time

### LINKED ISSUE #2983: When importing a dump. Forward the date to milli.
When importing a dump we recreate all the milli indexes.
Since milli v0.35 it’s possible to specify a creation and last update date when opening a milli index.

---
# TODO

- [x] Implement changes in [Milli](https://github.com/meilisearch/milli/)
- [x] Release a Milli version containing these changes
- [x] Bump this new Milli version in Meilisearch and merge it into `main`

amab8901:
# Pull Request

## Related issue
Partially fixes #2983 

## What does this PR do?
- Enables [`create_raw_index`](https://github.com/meilisearch/meilisearch/blob/660be071b5fd5823fbd5d8095fd0600169889354/index-scheduler/src/lib.rs#L868) to specify time

## PR checklist
- [ X ] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [ X ] Have you read the contributing guidelines?
- [ X ] Have you made sure that the title is accurate and descriptive of the changes?


irevoire (changes requested):
Hey @amab8901,

Your solution _could_ work, but we don't like the impact it has on the codebase; there is a lot of `None, None` that could be miss used.


So, we thought of another implementation:
* We keep a public equivalent of the `create_index` method that lets milli handles the dates
* We create a new method that handles the date and that is only called by the `create_raw_index` method
* To avoid duplicating too much code, we create a third **private** method, this time that contains all the logic.

Here is an example of what the third and private method could look like and what we could call it:
```rust
    /// Get or create the index from a closure and then insert it in the index mapper.
    fn get_or_insert_index_from_fn(
        &self,
        mut wtxn: RwTxn,
        name: &str,
        get_or_create: impl Fn(&Self, &Path) -> Result<Index>,
    ) -> Result<Index> {
        match self.index(&wtxn, name) {
            Ok(index) => {
                wtxn.commit()?;
                Ok(index)
            }
            Err(Error::IndexNotFound(_)) => {
                let uuid = Uuid::new_v4();
                self.index_mapping.put(&mut wtxn, name, &uuid)?;

                let index_path = self.base_path.join(uuid.to_string());
                fs::create_dir_all(&index_path)?;
                let index = get_or_create(self, &index_path)?;

                wtxn.commit()?;
                // TODO: it would be better to lazily create the index. But we need an Index::open function for milli.
                if let Some(BeingDeleted) =
                    self.index_map.write().unwrap().insert(uuid, Available(index.clone()))
                {
                    panic!("Uuid v4 conflict.");
                }

                Ok(index)
            }
            error => error,
        }
    }

    /// Get or create the index.
    pub fn create_index(&self, wtxn: RwTxn, name: &str) -> Result<Index> {
        self.get_or_insert_index_from_fn(wtxn, name, |s, name| s.create_or_open_index(name))
    }
```

---------

To avoid duplicating too much code, another thing we could do is delete the `create_or_open_index` method and instead create a `generate_env_open_options` that only does the important part of this function:
```rust
    fn generate_env_open_option(&self, path: &Path) -> Result<Index> {
        let mut options = EnvOpenOptions::new();
        options.map_size(clamp_to_page_size(self.index_size));
        options.max_readers(1024);
        options
    }
```

-----------

One last little detail is; it would be easier to store the dates in an `Option<(OffsetDateTime, OffsetDateTime)>` instead of two options, we should never lose one of the two dates, but if that happens, it's not a huge deal to reset both dates to the current time.


Let me know if you have any questions.

amab8901:
is it good now?

curquiza:
bors try

curquiza:
Hello @amab8901 thanks for your PR
Looks there is an issue with Rustfmt https://github.com/meilisearch/meilisearch/actions/runs/3711186372/jobs/6326679843

The issue with clippy is not related to your PR, and is also on the `main` branch.

curquiza:
bors try

index-scheduler/src/lib.rs
```diff
@@ -867,7 +867,8 @@ impl IndexScheduler {
     /// Create a new index without any associated task.
     pub fn create_raw_index(&self, name: &str) -> Result<Index> {
         let wtxn = self.env.write_txn()?;
-        let index = self.index_mapper.create_index(wtxn, name)?;
+        let date = Some((time::OffsetDateTime::now_utc(), time::OffsetDateTime::now_utc()));
+        let index = self.index_mapper.create_index(wtxn, name, date)?;
```
irevoire, line 871:
```suggestion
        let index = self.index_mapper.create_index(wtxn, name, date)?;
```

The caller is supposed to forward you the date from the dump here

index-scheduler/src/index_mapper.rs
```diff
@@ -66,15 +66,30 @@ impl IndexMapper {
...
         options.map_size(clamp_to_page_size(self.index_size));
         options.max_readers(1024);
-        Ok(Index::new(options, path)?)
+
+        if date == None {
+            Ok(Index::new(options, path)?)
+        } else {
+            let (created, updated) = date.unwrap();
+            Ok(Index::new_with_creation_dates(options, path, created, updated)?)
+        }
```
irevoire, line 83:
```suggestion
        if let Some(date) = date {
            Ok(Index::new_with_creation_dates(options, path, created, updated)?)
        } else {
            Ok(Index::new(options, path)?)
        }
```

When we can we try to avoid using `.unwrap`

irevoire (changes requested):

curquiza:
@amab8901 looks like you finally have an issue with clippy, sorry for the confusion. We did fix the clippy issues on `main`, so you probably still have some on your branch. Can you fix them?

amab8901:
is it good now?

curquiza:
bors try

amab8901:
let's try again

amab8901:
bors try

curquiza:
bors try

meilisearch/src/lib.rs
```diff
@@ -289,7 +289,14 @@ fn import_dump(
...
         log::info!("Importing index `{}`.", metadata.uid);
-        let index = index_scheduler.create_raw_index(&metadata.uid)?;
+        let date;
+        if let Some(dump_reader_date) = dump_reader.date() {
+            date = Some((dump_reader_date, dump_reader_date));
+        } else {
+            date = Some((time::OffsetDateTime::now_utc(), time::OffsetDateTime::now_utc()));
+        }
+
+        let index = index_scheduler.create_raw_index(&metadata.uid, date)?;
```
irevoire, line 299: Once again, we're not interested in knowing the current time or at which date the dump was created but when the **index** was created.

You must use the date contained in the index, not anywhere else, and if you don't have any date, don't try to autogenerate one just send `None` to the `create_raw_index` function.

irevoire (changes requested):

meilisearch/src/lib.rs
amab8901: where is the date field located in the index? There's tons of items that contain the word "index" and also plenty of search results for "date"

irevoire: In this part of the code, the only variable related to the index is the `index_reader`; if you check its implementation, you'll see it contains only three functions, one to get the documents, one to get the settings and one to get the metadata. And here is what the metadata looks like:
```rust
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
pub struct IndexMetadata {
    pub uid: String,
    pub primary_key: Option<String>,
    #[serde(with = "time::serde::rfc3339")]
    pub created_at: OffsetDateTime,
    #[serde(with = "time::serde::rfc3339")]
    pub updated_at: OffsetDateTime,
}
```

amab8901: I'm trying to learn the process by which I could figure this out if I didn't have access to someone more experienced.

So [`index` variable is created as the output of `create_raw_index`](https://github.com/meilisearch/meilisearch/blob/99253094929745037ca63dcce73e3fafdf731e8f/meilisearch/src/lib.rs#L292). In the implementation, the `index_reader` seems to be [an element of iterator created from dump_reader.indexes()](https://github.com/meilisearch/meilisearch/blob/99253094929745037ca63dcce73e3fafdf731e8f/meilisearch/src/lib.rs#L288).

But the `indexes` method is part of `V6Reader`, which has plenty of functions and not just three. What should I do differently in my problem-solving process to arrive at the places you're referring to?

amab8901:
I think it looks good now. Let's try again.

sudo bors try

irevoire (approved):
Nice, thank you @amab8901 
bors merge

Issue #2785, opened 2022-09-19
Create `too_many_open files` errors


curquiza:
When Meilisearch opens too many FDs, you get the following error

```json
{
    "message": "An internal error has occurred. `Too many open files`.",
    "code": "internal",
    "type": "internal",
    "link": "https://docs.meilisearch.com/errors#internal"
}
```

We want to replace
- the code (not type) `internal` by `too_many_open_files`
- the link `#internal` by #too_many_open_files

## TODO

- [ ] Implement changes in [Milli](https://github.com/meilisearch/milli/)
- [ ] Release a Milli version containing these changes
- [ ] Bump this new Milli version in Meilisearch and merge it into `main`
- [ ] Update the [associated specification](https://github.com/meilisearch/specifications/blob/main/text/0061-error-format-and-definitions.md)

## Impacted teams

Pinging every teams that might use this, @meilisearch/docs-team (a [related issue](https://github.com/meilisearch/documentation/issues/1874)) and @meilisearch/integration-team 
Also @meilisearch/cloud-team I don't know if you would be impacted by this change

semi-infiknight:
please assign me this issue

curquiza:
Hello @semi-infiknight 

Thanks for your interest in this project 🔥 

FYI, we prefer not assigning people to our issues because sometimes people ask to be assigned and never come back, which discourages the volunteer contributors from opening a PR to fix this issue.
We will accept and merge the first PR that fixes correctly and well implements the issue following our [contributing guidelines](https://github.com/meilisearch/meilisearch-dotnet/blob/main/CONTRIBUTING.md).   

We are looking forward to reviewing your PR 😊

Rutam21:
@curquiza Can you help me with finding the place where this change needs to be done? Thanks.

irevoire:
Hey @Rutam21,

I think you'll need to add a variant in the [`Code`](https://github.com/meilisearch/meilisearch/blob/main/meilisearch-types/src/error.rs#L116-L173) to represent the error.
And then in the [`MilliError`](https://github.com/meilisearch/meilisearch/blob/main/meilisearch-lib/src/error.rs#L18-L55) enum you should look for every variant that can output a `TooManyOpenFiles` error and map it to the code you created before :+1:

curquiza:
Moving this to v1 milestones, we will not have the time to fix it for v0.30.0

brunoocasali:
[...]

Issue #2255, opened 2022-03-21
Unclear error when there is not enough space left


nicolasvienot:
**Describe the bug**
When trying to index documents in a **kubernetes** volume that does not have enough space left, Meilisearch returns the following error:

```
{
      “details”: {
        “indexedDocuments”: 0,
        “receivedDocuments”: 46206
      },
      “duration”: “PT512.665928885S”,
      “enqueuedAt”: “2022-03-17T14:58:24.465745303Z”,
      “error”: {
        “code”: “internal”,
        “link”: “https://docs.meilisearch.com/errors#internal”,
        “message”: “I/O error (os error 5)“,
        “type”: “internal”
      },
      “finishedAt”: “2022-03-17T15:06:57.168374834Z”,
      “indexUid”: “bgg”,
      “startedAt”: “2022-03-17T14:58:24.502445949Z”,
      “status”: “failed”,
      “type”: “documentAddition”,
      “uid”: 3
}
```

This error might also happen outside of a kubernetes environment, not tested.

**To Reproduce**
1. Create a Meilisearch instance in a Kubernetes cluster and use a persistent volume to store the `data.ms`
2. Index documents until the volume is full

**Expected behavior**
A clear error that is documented should be returned

**Meilisearch version:**
v0.26.0

---

EDIT from @curquiza 

## How?

When Meilisearch does not have enough place on the machine, you get the following error":

```json
{
    "message": "I/O error (os error 5).",
    "code": "internal",
    "type": "internal",
    "link": "https://docs.meilisearch.com/errors#internal"
}
```

We want to replace
- the code (not type) `internal` by `no_space_left_on_device`
- the link `#internal` by `#no_space_left_on_device`

⚠️ This is what the specification already mentioned and it looks like Meilisearch does not follow it yet: https://github.com/meilisearch/specifications/blob/main/text/0061-error-format-and-definitions.md#no_space_left_on_device

## Impacted teams

Since this error is already in the spec (and so in the docs) and supposed to exist, no ping to do.
https://docs.meilisearch.com/reference/errors/error_codes.html#no-space-left-on-device

MarinPostma:
Uhm, this error is probably triggered when a file is being written to disk, alas, the only thing we can know is that a write as failed, so I'm not so sure how we could improve that... @meilisearch/core-team any idea?

Kerollmops:
[...]

PR #3263, opened 2022-12-19
Handle most io error instead of tagging everything as an internal

irevoire:
Fix https://github.com/meilisearch/meilisearch/issues/2255
Fix https://github.com/meilisearch/meilisearch/issues/2785
Close https://github.com/meilisearch/milli/pull/580

- [ ] Find a way to catch the io::Error contained in serde_json::Error
- [ ] Check the grenad::Error as well
- [ ] Ensure the error code mean the same thing under windows


dump/src/error.rs
```diff
@@ -19,9 +19,9 @@ pub enum Error {
 impl ErrorCode for Error {
     fn error_code(&self) -> Code {
         match self {
-            // Are these three really Internal errors?
-            // TODO look at that later.
-            Error::Io(_) => Code::Internal,
+            Error::Io(e) => e.error_code(),
+
+            // These error come from an internal mis
```
dureuill, line 24: what's a "mis"?

meilisearch-types/src/document_formats.rs
```diff
@@ -36,13 +35,15 @@ impl fmt::Display for PayloadType {
...
 pub enum DocumentFormatError {
+    Io(io::Error),
     Internal(Box<dyn std::error::Error + Send + Sync + 'static>),
     MalformedPayload(Error, PayloadType),
 }
 
 impl Display for DocumentFormatError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
+            Self::Io(e) => write!(f, "{e}"),
```
dureuill, line 46: Do we want to say here that the error is usually (but not always) caused by "no space left on disk" or "permission errors"?

I'm also wondering if we could feasibly add more context here: on which file did the access operation fail, while carrying out which logical operation, etc. This imposes to store more than just the IoError though.

meilisearch-types/src/error.rs
```diff
@@ -415,13 +425,28 @@ impl ErrorCode for milli::Error {
...
+impl ErrorCode for file_store::Error {
+    fn error_code(&self) -> Code {
+        match self {
+            Self::IoError(e) => e.error_code(),
+            Self::PersistError(e) => e.error_code(),
+        }
+    }
+}
+
+impl ErrorCode for tempfile::PersistError {
```
dureuill, line 437: I'm not very clear on what this implementation brings. Is this related to IoErrors?

dureuill (changes requested):
Thanks for the PR :-)! I have a few questions, mostly.

meilisearch-types/src/document_formats.rs
irevoire: Nice, good catch. I totally missed the `documentFormatError` thingy.

We don't need to because that's handled when we create the `ResponseError` from the `Code`

meilisearch-types/src/error.rs
irevoire: Yep, the `PersistError` can contain an `IoError` that must be handled properly

dump/src/error.rs
irevoire: Oops, sorry, look like yesterday I left in the middle of writing a comment :grimacing: 
It's fixed now

dureuill (approved):
some typos, otherwise looks good to me

dureuill (approved):
Thanks for taking my feedback! LGTM

bors merge

PR #3267, opened 2022-12-21
Bump milli to v0.37.5

### LINKED ISSUE #3021: Duplicate documents in the search route
The engine can return multiple versions of the same document (with the same document id). It should not be possible as a document with a specific document id must erase the previous version of itself. This bug appears in version v0.29.0 but not v0.28.1.

A strange coincidence is that [we released the soft-deleted feature in this exact version](https://github.com/meilisearch/milli/pull/557). Here is [a thread explaining the issue users have](https://meilicommunity.slack.com/archives/CP9DVS1RQ/p1665160818982859). We have private access to the settings one user was using when triggering this bug.

This bug could also show documents that had nothing in common with a user query. It was due to the fact that documents were wrongly deleted and, as ids are reused, associated to a completely different document.

---

EDIT by @curquiza 

- [x] Implement changes in [Milli](https://github.com/meilisearch/milli/): https://github.com/meilisearch/milli/pull/723
- [x] Release a Milli version containing these changes
- [ ] Bump this new Milli version in Meilisearch and merge it into `main`

curquiza:
Fixes #3021

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1332,8 +1332,8 @@ dependencies = [
 
 [[package]]
 name = "filter-parser"
-version = "0.37.4"
-source = "git+https://github.com/meilisearch/milli.git?tag=v0.37.4#e1d7d7231338803f940734b4d54a2de99f6b3c6b"
+version = "0.37.5"
+source = "git+https://github.com/meilisearch/milli.git?tag=v0.37.5#0bc7f704c8b7421875338ed3ea996d155446eeda"
 dependencies = [
  "nom",
  "nom_locate",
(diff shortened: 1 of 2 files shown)


loiclec:
The dependencies are correct, but do you know why there is a clippy error just now?

curquiza:
Clippy error is due to a rust upgrade, we fixed it on `main` (https://github.com/meilisearch/meilisearch/pull/3262), and now I fixed it here: https://github.com/meilisearch/meilisearch/pull/3268

loiclec (approved):

curquiza:
bors merge

PR #3278, opened 2022-12-26
Remove `--max-index-size` and `--max-task-db-size` flags

### LINKED ISSUE #3231: Remove `--max-index-size` and `--max-task-db` configuration options
Following [Product discussion](https://github.com/meilisearch/product/discussions/567#discussioncomment-4322126)

Specification: https://github.com/meilisearch/specifications/pull/207
As part of the [CLI stabilization](https://github.com/meilisearch/meilisearch/issues/2752), we want to remove these options as we don't want to commit to them in their current shape for v1.

### Impacted teams

- @meilisearch/docs-team for sure
- I believe @meilisearch/cloud-team is using these options.
- Not sure if the integrations are impacted, however.

### Open questions

Before working on this, we need to resolve the following open questions positively:

1. Is Windows still pre-allocating map files for modern-day lmdb? It appears as if MacOS isn't.
2. Is lmdb taking more space than necessary when working with a large mapsize? We would need to benchmark this
3. We need another way to express the size available to soft-deleted documents, as it currently uses the --max-index-size option for this purpose

dureuill:
# Pull Request

## Related issue
Fixes #3231 

## What does this PR do?
- Remove `--max-index-size` and `--max-task-db-size` flags from the CLI, config file and environment variable
- Set the size of all indexes to **500GiB** and the size of the task DB to **10GiB**.  Reviewers might want to review these values carefully.

## PR checklist


dureuill:
Update the index size to 400GiB to try and address the tests failing in Windows. 

Alternatives:

1. reduce size only if cfg windows (will do if 400GiB still triggers the issue)
2. limit test parallelism (at least under Windows)

dureuill:
New attempt with 200GiB for Windows, 500GiB for other OSes.

The issues under windows underline the necessity of providing another mechanism to handle mapsizes. My preference goes towards automatically resizing environments, but we need to think about how we want to do this because this cannot be fully transparent at the heed level: as MDB_MAP_FULL is returned inside of a transaction (after a call to put), one has to abort that transaction, resize the environment, and then attempt again that transaction. This is best handled at the milli or even meilisearch level. I'm unsure how to make sure we always handle that condition and restart the full transaction in that case, I'm afraid that manually handling each call to db/cursor::put in the code is going to cause us to miss some (or in the future after a refactor). Ideally when aborting a txn due to an error we'd remember if it was caused to the max mapsize being reached and we'd resize at this point, before attempting the txn again.

dureuill:
Still not OK even with 200GiB for Windows. Adjustments made in the tests seem to suggest a situation quite dire in the CI: https://github.com/meilisearch/meilisearch/pull/2476/files

Kerollmops:
> I'm not sure how to proceed.

The only thing I suggest is reducing the size of the indexes for Windows only.

The solution you were proposing about closing and reopening the index with a bigger max index size is great. We could try thinking about that or ask Howard for help on how we should proceed with LMDB running on Windows.

dureuill:
Since the Windows CI tests keep failing with strange JavaScript heap errors, I changed strategy:

- The options are removed from config, env variable and CLI, but not from the `Opt` struct
- The tests uses the options like before, allowing to keep the same values
- The values are set to 500GiB per index, 10GiB for the db task queue, for all OSes.

meilisearch/src/option.rs
```diff
@@ -70,6 +66,13 @@ const MEILI_MAX_INDEXING_THREADS: &str = "MEILI_MAX_INDEXING_THREADS";
 const DISABLE_AUTO_BATCHING: &str = "DISABLE_AUTO_BATCHING";
 const DEFAULT_LOG_EVERY_N: usize = 100000;
 
+// Each environment (index and task-db) is taking space in the virtual address space
+//
+// The size of the virtual address sapce is limited by OS. About 100TB for Linux, about 10TB for Windows.
+// This means that the number of indexes is limited to about 200 for Linux, and about 20 for Windows.
```
Kerollmops, line 72:
```suggestion
// Each environment (index and task-db) is taking space in the virtual address space.
//
// The size of the virtual address space is limited by the OS. About 100TB for Linux and about 10TB for Windows.
// This means that the number of indexes is limited to about 200 for Linux and about 20 for Windows.
```

Kerollmops (changes requested):
I like what you've done! Will merge just after this small doc fix :)

Kerollmops (approved):
Nice, thank you 🏝️
bors merge

dureuill:
fixed the merge conflict

Kerollmops (approved):
Haha, ok, nice 🤖
bors merge

PR #3281, opened 2022-12-27
Merge `--schedule-snapshot` and `--snapshot-interval-sec` options

### LINKED ISSUE #3131: CLI stabilization: change the behavior of `--schedule-snapshot`
Following this discussion: https://github.com/meilisearch/product/discussions/566
Specification: https://github.com/meilisearch/meilisearch/pull/3281

As part of the [CLI stabilization](https://github.com/meilisearch/meilisearch/issues/2752), we want to improve the snapshot command.

1. Remove `--snapshot-interval-sec`
2. `--schedule-snapshot` now accepts an interval in seconds
3. if `--schedule-snapshot` is given without any value, snapshots are activated with the default value (24h)

```bash
cargo run -- # gives no snapshot
cargo run -- --schedule-snapshot # gives snapshot every 24h
cargo run -- --schedule-snapshot 3600 # gives snapshot every hours
```

⚠️ Don't forget to change the [default config](https://github.com/meilisearch/meilisearch/blob/main/config.toml) file as well.

## Impacted teams

@meilisearch/docs-team 

## TODO

- [ ] Implement the changes and merge to the `main` branch
- [ ] Update the related spec

dureuill:
# Pull Request

## Related issue
Fixes #3131

## What does this PR do?
- Removes `--snapshot-interval-sec`
- `--schedule-snapshot` now accepts an optional integer value specifying the interval in seconds
- The config file no longer has a snapshot_interval_sec key.  Instead, the schedule_snapshot key now additionally accepts an integer value specifying the interval in seconds
- The env variable MEILI_SNAPSHOT_INTERVAL no longer exists
- The env variable MEILI_SCHEDULE_SNAPSHOT is always specified to the interval of the snapshot in seconds when defined. If snapshots are disabled the variable is undefined.

---

Relevant part of the `--help`

Capture d’écran 2022-12-27 à 18 22 32

---

### Tests

| `schedule_snapshot` in config.toml | `--schedule-snapshot` flag on CLI | `MEILI_SCHEDULE_SNAPSHOT` | `opt.schedule_snapshot` |
|--|--|--|--|
| missing | missing | missing | `Disabled`
| `false` | missing | missing | `Disabled`
| `true` | missing | missing | `Enabled(86400)`
| `1234` | missing | missing | `Enabled(1234)`
| missing | `--schedule-snapshot` | missing | `Enabled(86400)`
| `false` | `--schedule-snapshot` | missing | `Enabled(86400)` 
| missing | `--schedule-snapshot 2345` | missing | `Enabled(2345)`
| `false` | `--schedule-snapshot 2345` | missing | `Enabled(2345)`
| `true` | `--schedule-snapshot 2345` | missing | `Enabled(2345)`
| `1234` | `--schedule-snapshot 2345` | missing | `Enabled(2345)`
| `false` | `--schedule-snapshot 2345` | 3456 | `Enabled(2345)`
| `false` | `--schedule-snapshot` | 3456 | **`Enabled(86400)`**
| `1234` | missing | 3456 | `Enabled(3456)`
| `false` | missing | 3456 | `Enabled(3456)`

## PR checklist

### DIFF
--- a/meilisearch/src/analytics/segment_analytics.rs
+++ b/meilisearch/src/analytics/segment_analytics.rs
@@ -25,7 +25,7 @@ use uuid::Uuid;
 
 use super::{config_user_id_path, DocumentDeletionKind, MEILISEARCH_CONFIG_PATH};
 use crate::analytics::Analytics;
-use crate::option::{default_http_addr, IndexerOpts, MaxMemory, MaxThreads};
+use crate::option::{default_http_addr, IndexerOpts, MaxMemory, MaxThreads, ScheduleSnapshot};
 use crate::routes::indexes::documents::UpdateDocumentsQuery;
 use crate::routes::tasks::TasksFilterQueryRaw;
 use crate::routes::{create_all_stats, Stats};
@@ -220,9 +220,8 @@ struct Infos {
     ignore_missing_dump: bool,
     ignore_dump_if_db_exists: bool,
     import_snapshot: bool,
-    schedule_snapshot: bool,
+    schedule_snapshot: Option<u64>,
     snapshot_dir: bool,
-    snapshot_interval_sec: u64,
     ignore_missing_snapshot: bool,
     ignore_snapshot_if_db_exists: bool,
     http_addr: bool,
@@ -267,7 +266,6 @@ impl From<Opt> for Infos {
             ignore_snapshot_if_db_exists,
             snapshot_dir,
             schedule_snapshot,
-            snapshot_interval_sec,
             import_dump,
             ignore_missing_dump,
             ignore_dump_if_db_exists,
@@ -279,6 +277,11 @@ impl From<Opt> for Infos {
             #[cfg(all(not(debug_assertions), feature = "analytics"))]
                 no_analytics: _,
         } = options;
+
+        let schedule_snapshot = match schedule_snapshot {
+            ScheduleSnapshot::Disabled => None,
+            ScheduleSnapshot::Enabled(interval) => Some(interval),
+        };
 
         let IndexerOpts {
             log_every_n: _,
@@ -299,7 +302,6 @@ impl From<Opt> for Infos {
             import_snapshot: import_snapshot.is_some(),
             schedule_snapshot,
             snapshot_dir: snapshot_dir != PathBuf::from("snapshots/"),
-            snapshot_interval_sec,
             ignore_missing_snapshot,
             ignore_snapshot_if_db_exists,
             http_addr: http_addr != default_http_addr(),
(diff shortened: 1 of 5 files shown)


meilisearch/src/analytics/segment_analytics.rs
```diff
@@ -220,9 +220,8 @@ struct Infos {
     ignore_missing_dump: bool,
     ignore_dump_if_db_exists: bool,
     import_snapshot: bool,
-    schedule_snapshot: bool,
+    schedule_snapshot: Option<Option<u64>>,
```
loiclec, line 223: I might have missed it, but it looks like the variant `Some(None)` is never used, could the type of this field be simplified to `Option<u64>` or maybe a new custom enum?

loiclec (approved):
Thanks Louis 😄

dureuill:
Updated the implementation to use a custom struct rather than an `Option<Option<u64>>`. This is clearer and does not create an irrepresentable state.

meilisearch/src/analytics/segment_analytics.rs
```diff
@@ -220,9 +220,8 @@ struct Infos {
     ignore_missing_dump: bool,
     ignore_dump_if_db_exists: bool,
     import_snapshot: bool,
-    schedule_snapshot: bool,
+    schedule_snapshot: Option<u64>,
     snapshot_dir: bool,
-    snapshot_interval_sec: u64,
```
dureuill, line 225: This PR modifies the `Infos` struct, @gmourier  is that OK?

gmourier: OK; We just have to update this [file](https://github.com/meilisearch/specifications/blob/main/text/0034-telemetry-policies.md) in the associated PR specification.

As it's an `Option<u64>` type, we will receive `none` when the feature is not used, and in the opposite case, the value of the interval (the default one or the value specified by the user), is it right?

Signal to @irevoire

dureuill: > As it's an Option<u64> type, we will receive none when the feature is not used, and in the opposite case, the value of the interval (the default one or the value specified by the user), am I right?

Yes that's correct.

loiclec (approved):
👍

meilisearch/src/analytics/segment_analytics.rs
dureuill: This makes me realize that I don't think we made a spec PR for this one. We might have to do so.

dureuill:
bors merge

meilisearch/src/analytics/segment_analytics.rs
gmourier: Sure!

You can track the spec advancements [here](https://github.com/meilisearch/specifications/pull/201) (note: it may be a bit outdated).

I will ask for help regarding the missing specs

dureuill:
Spurious failure?

bors merge

PR #3251, opened 2022-12-14
Add a specific test on finite pagination placeolder search with disti…

ManyTheFish:
Add a specific test on finite pagination placeholder search with distinct attributes

related to https://github.com/meilisearch/milli/pull/743
related to https://github.com/meilisearch/meilisearch/issues/3200

poke @curquiza 

> note that the destination branch should be changed

PR #3288, opened 2023-01-02
Replace underscores with hyphens in documentation link to error code

### LINKED ISSUE #3097: Error links hash is formatted incorrectly
**Describe the bug**
When a user gets an error, as part of the error object Meilisearch returns a link to the docs (e.g. `"link": "https://docs.meilisearch.com/errors#index_not_found"`). The problem is that the hash does not take the user to the correct section of the errors page. This is because VuePress's slugification changes `_` to `-`.

(Actually, the hash is formatted correctly, and it's VuePress that works strangely.)

**To Reproduce**
1. Open Meilisearch and cause an error
2. Try to follow the link to the docs
3. Observe that it takes you to the error page, but not to the correct section corresponding to the specific error code
4. Change the `_` characters in the link hash to `-`
5. ...?
6. Profit

**Expected behavior**
Errors page should scroll automatically to the correct hash. Should be achievable by transforming the `_` character to `-` before concatenating the `error_code` to the link. 

**Screenshots**

**Meilisearch version:** any

**Additional context**
Thank you engine team 🙏🏻 🚂

loiclec:
# Pull Request

## Related issue
Fixes #3097 

## Implementation
Add a new dependency to `convert_case` (already used transitively by `deserr`) so that the link can be generated using:
```rust
    /// return the doc url associated with the error
    fn url(&self) -> String {
        format!(
            "https://docs.meilisearch.com/errors#{}",
            self.name().to_case(convert_case::Case::Kebab)
        )
    }
```

## Review
I'd like the reviewer to check whether it is expected that the content of some `dump` snapshot tests changed :-)


dureuill:
Taking review over @irevoire

dureuill (approved):
> I'd like the reviewer to check whether it is expected that the content of some dump snapshot tests changed :-)

Yes, the dumps are read in their original version, and then the data is converted to the current "live" data, which is what the snapshots are about, and that is affected by the change to the URL.

I approve of these changes ⭐

loiclec:
bors merge

loiclec:
I rebased the changes onto `main`. @dureuill could you approve and merge again?

loiclec:
strange, I just run them on my computer and it worked 🤔 looking into it...

dureuill:
> Add a specific test on finite pagination placeolder search with disti… 

I'm not sure why this commit is in the list of commits from this PR... Maybe it is an artifact from GitHub's interface? The commit was merged to main a few minutes ago: https://github.com/meilisearch/meilisearch/pull/3251#event-8158447566

dureuill (approved):
Let's retry with an approval 🤦 

bors merge

dichotommy:
Hello all, I am wondering if it is possible/easy to revert this PR 🫣

@bidoubiwa created a PR on the docs that allows the error codes page to accept links containing either `-` or `_` in the anchor. This solves the problem without requiring any change on the engine side. Since `_` is more conventional for an anchor, and we plan to migrate to a new site very soon, I think it makes more sense to keep using `_` in the engine-side error links after all.

I really hope that this is not too much of an inconvenience and apologize for the time lost fixing this issue 🙇🏻

dureuill:
Hello Tommy,

I don't have a definitive answer to provide, just wondering:

>  Since _ is more conventional for an anchor

Is it? It seems to me that most slugification libraries ([python](https://pypi.org/project/python-slugify/), [rust](https://docs.rs/slug/latest/slug/fn.slugify.html), [Ruby](https://github.com/cocur/slugify#usage)) and CMSes (e.g. jekyll, [mdbook](https://rust-lang.github.io/mdBook/guide/reading.html#reading-books), [GitHub](https://github.com/meilisearch/meilisearch/#-getting-started)) use `-` rather than `_` for slugification.

In that sense I personally prefer we migrate to `-`, as it seems more likely to me that the next system we'll use will force `-` or at least prefer it by default.

irevoire:
Personally, I'm 100% towards coming back to the `_`.
1. It removes code on the engine, and every line of code we don't write can't cause a bug
2. It seems easier for the user to understand how we generate our link and manipulate them if he can copy/paste the error_code instead of moving it into a case-changing function

dureuill:
So we've been discussing this with the engine team,

we decided to revert the behavior to `_` in documentation links.

PRs to revert this one will come up shortly.

dichotommy:
Thank you @meilisearch/engine-team , you are heroes 🙏🏻 🧑‍🚒

Issue #3292, opened 2023-01-03
Explicitly restrict log level options to those that are documented


loiclec:
See: https://docs.meilisearch.com/learn/configuration/instance_options.html#log-level

Currently we forward the given log level as-is to `env_logger`'s builder. This inadvertently makes `env_logger`'s API part of meilisearch's interface. We should be more explicit in the code about the accepted log levels.

PR #3293, opened 2023-01-03
Explicitly restrict log level options to those that are documented

loiclec:
Fixes https://github.com/meilisearch/meilisearch/issues/3292


dureuill (dismissed):
Thanks, nice to use an enum rather than only the strings; it simplifies a bit the default handling.

dureuill (approved):

PR #3296, opened 2023-01-03
Remove `--disable-auto-batching` CLI option

### LINKED ISSUE #3294: Remove `--disable-auto-batching` CLI option
We should not keep the `--disable-auto-batching` CLI option, because:
1. I don't think there are clear advantages to disabling autobatching other than for debug purposes.
2. We can't introduce breaking changes after v1.0 is released. Keeping the flag will commit us to keeping this feature until v2.0 is released.
3. We can reintroduce it as an unstable feature later on.

## Impacted teams
@meilisearch/docs-team

loiclec:
Fixes #3294 

The `index-scheduler` code is not modified, only the CLI options have changed.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -47,11 +47,6 @@ max_task_db_size = "100 GiB"
 # Sets the maximum number of threads Meilisearch can use during indexing.
 # https://docs.meilisearch.com/learn/configuration/instance_options.html#max-indexing-threads
 
-disable_auto_batching = false
-# Deactivates auto-batching when provided.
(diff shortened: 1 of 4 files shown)


irevoire (approved):
@gmourier, you can bors merge if you want to be part of the contributors list 🎉

Kerollmops (approved):
It looks good to me, waiting for the approval of @gmourier 🐝

gmourier:
bors merge

PR #3305, opened 2023-01-05
Remove hidden but usable CLI arguments

Kerollmops:
@curquiza found out that we were exposing some internal CLI arguments: `nb-max-chunks` and `log-every-n`. In this PR I removed those two, the only two ones that I found. Those options shouldn't be accessible as non-documented in the documentation or the `--help` message.

Fixes https://github.com/meilisearch/meilisearch/issues/3307

### DIFF
--- a/meilisearch/src/option.rs
+++ b/meilisearch/src/option.rs
@@ -489,7 +489,7 @@ impl Opt {
     }
 }
 
-#[derive(Debug, Clone, Parser, Deserialize)]
+#[derive(Debug, Default, Clone, Parser, Deserialize)]
 pub struct IndexerOpts {
     /// Sets the maximum amount of RAM Meilisearch can use when indexing. By default, Meilisearch
     /// uses no more than two thirds of available memory.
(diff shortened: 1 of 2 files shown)


meilisearch/src/option.rs
```diff
@@ -63,7 +63,7 @@ const DEFAULT_DUMP_DIR: &str = "dumps/";
 
 const MEILI_MAX_INDEXING_MEMORY: &str = "MEILI_MAX_INDEXING_MEMORY";
 const MEILI_MAX_INDEXING_THREADS: &str = "MEILI_MAX_INDEXING_THREADS";
-const DEFAULT_LOG_EVERY_N: usize = 100000;
+const DEFAULT_LOG_EVERY_N: usize = 100_000;
```
dureuill, line 66: I feel like we could remove this, as it doesn't appear to be used anywhere now. What do you think?

dureuill (changes requested):
The PR looks good, I have one question (see inline comment).

meilisearch/src/option.rs
Kerollmops: It is used by milli in the indexer part to log for indexed documents.

dureuill (approved):
Found the `DEFAULT_LOG_EVERY_N` 😅

Kerollmops:
bors merge

PR #3308, opened 2023-01-05
Remove `--generate-master-key` option

dureuill:
# Pull Request

## Related issue

Related to https://github.com/meilisearch/specifications/pull/210#issuecomment-1372035525

## What does this PR do?
- Remove the short-lived `--generate-master-key` flag that was too beautiful for this world :D.

Removal of this option proceeds of the following reasoning:

1. It is the only option that starts meilisearch and then immediately exits
2. We are unsure if we want to keep it under this form in the future or switch to a subcommand.
3. Releasing this option in v1 would make it insta-stable.
5. The option is only marginally useful, as users will be presented with freshly generated key directly in the error messages if their master key is absent/too short.
6. If we remove this option now, we can still add it back in a future v1 release. If we add it now, we won't be able to remove it in any future v1 version.

## PR checklist

### DIFF
--- a/meilisearch/src/analytics/segment_analytics.rs
+++ b/meilisearch/src/analytics/segment_analytics.rs
@@ -271,7 +271,6 @@ impl From<Opt> for Infos {
             log_level,
             indexer_options,
             config_file_path,
-            generate_master_key: _,
             #[cfg(all(not(debug_assertions), feature = "analytics"))]
                 no_analytics: _,
         } = options;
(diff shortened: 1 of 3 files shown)


Kerollmops (approved):
Nice, I looove what you've done ❤️ Thank you!
Will the new messages be introduced in another PR?

dureuill:
I proposed discussing the new messages in this PR, but we can also discuss them later in the specs.

Advantage of later is that we can merge as-is for RC1.

Kerollmops:
Ok, so I will merge this but we can continue discussing here or in another PR!
bors merge

PR #3246, opened 2022-12-14
Implement most of the error handling enhancement planned for v1.0

irevoire:
Fix #3095 and #2325
Close https://github.com/meilisearch/meilisearch/pull/2540

Implements most of https://github.com/meilisearch/specifications/pull/212

## Generic error message we re-implements (in deserr):

- [x] Json
  - [x] Incorrect value kind
  - [x] Missing field
  - [x] Unknown key
  - [x] Unexpected
  - [x] Reimplement the way we show the location

- [x] Query parameter
  - [x] Incorrect value kind
  - [x] Missing field
  - [x] Unknown key
  - [x] Unexpected

## Routes to implements:
- [x] Get search
- [x] Post search
- [x] Settings
- [x] Swap indexes
- [x] Task API
- [x] Documents ressource

Error codes to implements;
## Swap API

- [x] `duplicate_index_found` → `invalid_swap_duplicate_index_found`

## Search API

- [x] `invalid_search_q`
- [x] `invalid_search_offset`
- [x] `invalid_search_limit`
- [x] `invalid_search_page`
- [x] `invalid_search_hits_per_page`
- [x] `invalid_search_attributes_to_retrieve`
- [x] `invalid_search_attributes_to_crop`
- [x] `invalid_search_crop_length`
- [x] `invalid_search_attributes_to_highlight`
- [x] `invalid_search_show_matches_position`
- [x] `invalid_search_filter`
- [x] `invalid_search_sort`
- [x] `invalid_search_facets`
- [x] `invalid_search_highlight_pre_tag`
- [x] `invalid_search_highlight_post_tag`
- [x] `invalid_search_crop_marker`
- [x] `invalid_search_matching_strategy`

## Settings API

- [x] invalid_settings_displayed_attributes
- [x] invalid_settings_searchable_attributes
- [x] invalid_settings_filterable_attributes
- [x] invalid_settings_sortable_attributes
- [x] invalid_settings_ranking_rules
- [x] invalid_settings_stop_words
- [x] invalid_settings_synonyms
- [x] invalid_settings_distinct_attribute
- [x] Add invalid_settings_typo_tolerance
    - [x] ~~invalid_settings_typo_tolerance_min_word_size_for_typos~~ (Merge in **invalid_settings_typo_tolerance**)
- [x] invalid_settings_faceting
- [x] invalid_settings_pagination

## Task API

- [x] invalid_task_date_filer → invalid_task_before_enqueued_at_filter (for all date filter) ?

## Document Resource

[...]

PR #3317, opened 2023-01-09
Remove the unused error codes

irevoire:
Remove some unused error code + fix the usage of the search+settings sort and filter error_code


curquiza:
Hello @irevoire 

Could you be a little more precise? What does this PR changes exactly?

> Remove some unused error code

Does it impact the users? (to know how I "highlight" it in the changelog) -> I would say no: these are error `code`s that are not used anymore. Can you confirm?

> fix the usage of the search+settings sort and filter error_code

Does it impact the users? I checked the code to try to better understand it, and I understand the `link` was wrong for the sort and filter errors. Is it correct?

irevoire:
Hey @curquiza,

Actually, it's a followup / bugfix of https://github.com/meilisearch/meilisearch/pull/3246, which wasn't completely finished.
What it does is:
1. Convert the old `search` and `filter` error codes from the dump to the new one
2. Returns the expected `invalid_search_filter/sort` and `invalid_settings_filterable/sortable_attributes` (which we thought was implemented but wasn't working in every cases (it was still returned _sometimes_)

bors merge

curquiza:
Thanks @irevoire for the precision 👍

PR #3324, opened 2023-01-10
Add a test on the search route for each possible error codes

irevoire:


meilisearch/tests/search/errors.rs
```diff
@@ -37,104 +38,360 @@ async fn search_unexisting_parameter() {
...
+    snapshot!(code, @"400 Bad Request");
+    snapshot!(json_string!(response), @r###"
+    {
+      "message": "invalid type: Sequence `[\"doggo\"]`, expected a String at `.q`.",
+      "code": "invalid_search_q",
+      "type": "invalid_request",
+      "link": "https://docs.meilisearch.com/errors#invalid-search-q"
+    }
+    "###);
+    // Can't make the `q` fail with a get search since it'll accept anything as a string.
```
loiclec, line 55: hm, interesting! I hadn't thought about it

meilisearch/tests/search/errors.rs
```diff
@@ -37,104 +38,360 @@ async fn search_unexisting_parameter() {
...
-            "type": String("invalid_request"),
-            "link": String("https://docs.meilisearch.com/errors#invalid-search-crop-marker"),
-        },
-        400,
-    )
+    let (response, code) = index.search_post(json!({"q": ["doggo"]})).await;
+    snapshot!(code, @"400 Bad Request");
+    snapshot!(json_string!(response), @r###"
+    {
+      "message": "invalid type: Sequence `[\"doggo\"]`, expected a String at `.q`.",
```
loiclec, line 49: We should have a better way to print `Sequence [`\"doggo\"`]` in the future

meilisearch/tests/search/errors.rs
```diff
@@ -37,104 +38,360 @@ async fn search_unexisting_parameter() {
...
+#[actix_rt::test]
+async fn search_bad_attributes_to_crop() {
+    let server = Server::new().await;
+    let index = server.index("test");
+
+    let (response, code) = index.search_post(json!({"attributesToCrop": "doggo"})).await;
+    snapshot!(code, @"400 Bad Request");
+    snapshot!(json_string!(response), @r###"
+    {
+      "message": "invalid type: String `\"doggo\"`, expected a Sequence at `.attributesToCrop`.",
```
loiclec, line 179: Other future improvement: expected a Sequence of what?
Maybe we could customise these messages with something like:
```rust
struct Search {
    #[deserr(expected = "an array containing the names (String) of the attributes to crop")]
    attributes_to_crop: Vec<String>
}
```

meilisearch/tests/search/errors.rs
```diff
@@ -37,104 +38,360 @@ async fn search_unexisting_parameter() {
...
+
+    // Can't make the `filter` fail with a get search since it'll accept anything as a strings.
+    let (response, code) = index.search_post(json!({ "filter": true })).await;
+    snapshot!(code, @"404 Not Found");
+    snapshot!(json_string!(response), @r###"
+    {
+      "message": "Index `test` not found.",
+      "code": "index_not_found",
+      "type": "invalid_request",
+      "link": "https://docs.meilisearch.com/errors#index-not-found"
```
loiclec, line 247: That's strange. Why is the error `index_not_found`?

loiclec (approved):
I approved it because the tests are well-written, but could you look at why the error returned by the bad filter request is `index_not_found`?

meilisearch/tests/search/errors.rs
irevoire: ah yeah that would be nice

meilisearch/tests/search/errors.rs
irevoire: Since we can deserialise the payload, actix jump to the next error that is; the index specified in the route doesn’t exists

meilisearch/tests/search/errors.rs
irevoire: Like what?
I would like to get rids of the term `Sequence` for json, but I think it’s a pretty good thing that we’re able to print `["doggo"]` is a json value

loiclec (dismissed):
Yayyy 🙏

irevoire:
> I approved it because the tests are well-written, but could you look at why the error returned by the bad filter request is index_not_found?

This was expected actually BUT after a good night of sleep I realised I could also test the error we generate by hand if I were to create an index + provide a bad value.
Which I did and it highlighted a bug so, good catch 🎉

loiclec (approved):

irevoire:
bors merge

irevoire:
bors merge

Issue #1841, opened 2021-10-22
Cannot allocate memory when creating indexes


neylsongularte:
I'm trying to create 10000+ indexes with 500+ documents and I get this error on a machine with 16GB of ram:

```
MeiliSearch ApiException: Http Status: 500 - Message: Cannot allocate memory (os error 12) - Error code: internal - Error type: internal_error - Error link: https://docs.meilisearch.com/errors#internal
```

![image](https://user-images.githubusercontent.com/4109823/138458881-08b247de-6587-4428-9a48-fccc64ee5106.png)

MeiliSearch version: 0.23.1

**Additional context**
Amazon Linux 2 - x86_64 - 16GB ram - With 10GB of ram free

1,1G	data.ms/

tanmaymaheshwari97:
@neylsongularte Was this memory allocation issue fixed?

curquiza:
Hello @tanmaymaheshwari97!
Can you check if you still have enough space on your machine for Meilisearch?

tanmaymaheshwari97:
Hi @curquiza. My prod instance is on hosted on google cloud platform. And only around 4.5 GB of memory is utilized out of 16 GB.
![508A62CC-AA96-4B9A-BD31-C9097680A3AB_1_201_a](https://user-images.githubusercontent.com/66941891/175527243-e9c2b1aa-614d-4457-80f3-5a3835fc0a41.jpeg)

curquiza:
Thanks for your quick answer. I was not asking about the RAM (16Gb), but the total about of space you have on your disk, and if there is enough space for Meilisearch to index your data 😊

tanmaymaheshwari97:
I had a look at the disk space utilisation also. I don't think space is the issue here. Please suggest how to fix this issue as our search entirely depends on it.
![02C4D705-0E71-446C-AE14-FA5BC129B46E_1_201_a](https://user-images.githubusercontent.com/66941891/175536339-8d324adb-3d1a-4b90-a537-cd8bfb34df3c.jpeg)

neylsongularte:
> I had a look at the disk space utilisation also. I don't think space is the issue here. Please suggest how to fix this issue as our search entirely depends on it. Thanks.

I changed the strategy and put it in a single index as suggested.

curquiza:
For information, related issues: https://github.com/meilisearch/meilisearch/issues/1841
And it does not seem to be a huge amount of indexes

dmitriid:
[...]

PR #3331, opened 2023-01-11
Limit the number of concurrently opened indexes

### LINKED ISSUE #3382: Don't limit the number or size of indexes
## Context

- There is currently an inherent tradeoff between the number of opened indexes and their max size. In v1.0.0, this tradeoff is "fixed" due to the removal of the `--max-index-size` option. We would like to not expose this tradeoff to end users anymore.
- For more context: https://github.com/meilisearch/meilisearch/discussions/3280
- Also related to issue https://github.com/meilisearch/meilisearch/issues/1841 that's already planned for the v1.1 milestone

## Implementation

- Automatically resize indexes: this allows to start with a lower "`--max-index-size`" and plan accordingly to the actual space needs of the index.
- Limit the concurrent number of opened indexes: this ensures that we can reach bigger `--max-index-size` by closing least recently used indexes.

I have some draft PRs implementing both of these: https://github.com/meilisearch/meilisearch/pull/3319 and https://github.com/meilisearch/meilisearch/pull/3331 

## Specification

A specification PR was opened: see https://github.com/meilisearch/specifications/pull/226

dureuill:
# Pull Request

## Related issue
Relevant to #1841, fixes #3382

### User standpoint

- Limit the number of concurrently opened indexes (currently, the number of indexes that can be concurrently opened is computed at startup)
- When too many an index is opened, the least recently used one is closed and its virtual memory released.
- This allows a user to have an arbitrary number of indexes of an arbitrary size

### Implementation standpoint

- Added a LRU cache map in `index-scheduler::lru`. A more complete implementation  (eg with helper functions not used here) is available but would better fit a dedicated crate.
- Use the LRU cache map in the `IndexScheduler`. To simplify the lifecycle of indexes, they are never removed from the cache when they are in the middle of a resize or delete operation. To achieve this, an intermediate `Vec` stores the UUIDs of the indexes that are in the middle of such an operation.
- Upon creating the index scheduler object, compute the total virtual memory that is adressable by using a dichotomic search on the max size of an index. Use this as a base to compute the number of indexes that can be open with 2TiB per index. If the virtual memory address space is lower than 2TiB, then only allow for 1 index of a fraction of that size.


dureuill:
I feel like with the tests I did, this PR is now ready for review.

index-scheduler/src/index_mapper.rs
```diff
@@ -26,28 +25,400 @@ const INDEX_MAPPING: &str = "index-mapping";
...
+        /// To do so, a `ReopenableIndex` is returned, that can be used to either definitely close or definitely open
+        /// the index without waiting anymore.
+        pub fn wait_timeout(self, timeout: Duration) -> ReopenableIndex {
+            self.closing_event.wait_timeout(timeout);
+            ReopenableIndex {
+                uuid: self.uuid,
+                map_size: self.map_size,
+                generation: self.generation,
+            }
+        }
```
Kerollmops, line 126: Not sure to understand but if you use a `wait_timeout` shouldn't you take the result of it into account? I mean what if the timeout was reached and not the fact that it was effectively closed?

index-scheduler/src/index_mapper.rs
```diff
@@ -26,28 +25,400 @@ const INDEX_MAPPING: &str = "index-mapping";
...
+                    panic!("Attempt to open an index that was already opened")
+                }
+            }
+            Ok(index)
+        }
+
+        fn next_generation(&mut self) -> usize {
+            self.generation = self.generation.checked_add(1).unwrap();
+            self.generation
+        }
```
Kerollmops, line 246: Could you document the `generation` field and this very method, please?

index-scheduler/src/index_mapper.rs
```diff
@@ -26,28 +25,400 @@ const INDEX_MAPPING: &str = "index-mapping";
...
+                return Ok(Some(index.prepare_for_closing()));
+            }
+            match self.unavailable.remove(uuid) {
+                Some(Some(reopen)) => Err(Some(reopen)),
+                Some(None) => Err(None),
+                None => Ok(None),
+            }
+        }
+
+        /// Marks that an index finished deletion.
```
Kerollmops, line 296:
```suggestion
        /// Marks that an index deletion finished.
        ///
        /// Must be used after calling `start_deletion`.
```

index-scheduler/src/index_mapper.rs
```diff
@@ -26,28 +25,400 @@ const INDEX_MAPPING: &str = "index-mapping";
...
+        fn close(&mut self, uuid: Uuid, index: Index, map_size_growth: usize) {
+            let map_size = index.map_size().unwrap_or(DEFAULT_MAP_SIZE) + map_size_growth;
+            let closing_event = index.prepare_for_closing();
+            let generation = self.next_generation();
+            self.unavailable
+                .insert(uuid, Some(ClosingIndex { uuid, closing_event, map_size, generation }));
+        }
+
+        /// Attempts to delete and index.
+        ///
```
Kerollmops, line 273:
```suggestion
        /// Attempts to delete and index.
        ///
        ///  `end_deletion` must be called just after.
        ///
```

index-scheduler/src/lru.rs
```diff
@@ -0,0 +1,202 @@
...
+        let mut min_generation_index = None;
+        for (index, (generation, _)) in self.data.iter_mut().enumerate() {
+            let generation = *generation.get_mut();
+            if let Some((_, min_generation)) = min_generation_index {
+                if min_generation > generation {
+                    min_generation_index = Some((index, generation));
+                }
+            } else {
+                min_generation_index = Some((index, generation))
+            }
```
Kerollmops, line 88: Can't we replace that with a match, that's shorter? But I don't understand why you can't use the `min_by_key` here as it is not as issue right? As it returns a `Self::Item` in the end even if it was `&Self::Item` in the closure.

```suggestion
            match min_generation_index {
                Some((_, min_generation)) => if min_generation > *generation.get_mut() {
                    min_generation_index = Some((index, generation));
                },
                None => min_generation_index = Some((index, generation)),
            }
```

Kerollmops (changes requested):
Thank you very much, maybe some small changes to do.

dureuill:
- Change default index size to 10GiB
- Default to 100 concurrent indexes in Unixes, 10 in Windows

index-scheduler/src/index_mapper.rs
```diff
@@ -26,28 +25,400 @@ const INDEX_MAPPING: &str = "index-mapping";
...
+    /// The map size an index is opened with on the first time.
+    index_base_map_size: usize,
+    /// The quantity by which the map size of an index is incremented upon reopening, in bytes.
+    index_growth_amount: usize,
     pub indexer_config: Arc<IndexerConfig>,
 }
 
+mod index_map {
+    /// the map size to use when we don't succeed in reading it in indexes.
+    const DEFAULT_MAP_SIZE: usize = 10_737_418_240; // 10 GiB
```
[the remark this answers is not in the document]
dureuill: > And now I understand that we start at 10GiB and not the 1 MiB I proposed in the previous PR.

Yes! The thing is that resizing an index happens after a failed indexing task. The more resize operations we need, the more indexing tasks we have to start again. So it makes sense to start with a bigger value than 1MiB.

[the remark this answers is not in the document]
dureuill: Adapted your preferred format. BTW the actual default map size that is used most of the time is defined in `meilisearch/src/options.rs:167`.

index-scheduler/src/index_mapper.rs
dureuill: You're correct, I updated the code to only return a Reopenable index when the env was effectively closed

index-scheduler/src/index_mapper.rs
dureuill: You're right, this field is needed to address subtle possible races (that I'm unsure can actually happen, but I guess I prefer this to be "theoretically sound").

I added a comment explanation, tell me what you think.

index-scheduler/src/lru.rs
dureuill: Apologies, I see my doc comment was insufficient, and I had to think again what the full reason was.

The reason is performance.

min_by_key gives us a shared reference to the generation, which is an atomic integer. Atomic integers require an atomic load with an implicit memory fence to be correct. It is unnecessary because we have an exclusive reference, meaning either we have a single thread or we have some kind of lock that handled the memory fencing for us.

This is why atomic integers allow direct access "as if they were simple integers" when we can provide an exclusive reference to them. Which is what I use here.

Alternatively, I could have:

- performed loads with an acquire semantics, but that would be subefficient performance wise
- performed loads with a relaxed semantics, and put a an explicit memory fence above the loop, but even this fence is unnecessary
- performed loads with a relaxed semantics, and put a comment explaining why this is not a problem.

I preferred shunting completely the atomic load by using the `get_mut()` method that returns a exclusive reference to the atomic.

To be clear, here the important semantic distinction is exclusivity, not mutability of the `&mut`.

Feel free to tell me if:

1. the explanation is clearer now
2. if you prefer that I change for one of alternatives outlined above
3. how you'd like me to improve the doc comment so the reason becomes clearer. It could also be a module wide comment because I'm using the exclusiveness of `&mut AtomicU64` in quite a few locations.

dureuill:
Added an update with the following:

- We compute an "index budget" when the Index scheduler is created, to know how much cumulated virtual memory we can allocate through memmap before exhausting the virtual address space. A fraction of the budget is then split to determine how many indexes can be opened simultaneously in the cache with a size of 2TiB. If that number is 0, then there can be only one index in the cache with a size of whatever budget is available.
- As computing the budget adds 2 seconds at startup, it is disabled in tests.

This should result in the following:

- No more performance hit due to index resize, except for users with indexes bigger than 2 TiB (if this impedes your use case, please contact us)
- Much lower risk of filling the entirety of the cache
- Adapts to machines with smaller virtual memory address space

I will do a second prototype with these changes soon.

---

### remaining TODO

- Some comments on the PR left unaddressed
- "Invasive" logs that are good for the prototype but should be removed before merging
- Handle the methods that attempt to open all indexes at once
- Check that the added dynamic budget computation doesn't impede usage for Cloud

dureuill:
Pushed an update that handles the methods that attempt to open all indexes at once.

There is now a `try_for_each_index` method that allows to iterate on all indexes. For callers that only needed the names, there is also a new `index_names` method that doesn't open the indexes at all.

The `stats`, `indexes`, `dumps` routes as well as snapshot creation/import and dump import seem to be fixed by this change.

index-scheduler/src/index_mapper.rs
```diff
@@ -26,28 +25,400 @@ const INDEX_MAPPING: &str = "index-mapping";
...
     base_path: PathBuf,
-    index_size: usize,
+    /// The map size an index is opened with on the first time.
+    index_base_map_size: usize,
+    /// The quantity by which the map size of an index is incremented upon reopening, in bytes.
+    index_growth_amount: usize,
     pub indexer_config: Arc<IndexerConfig>,
 }
 
+mod index_map {
```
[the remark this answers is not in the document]
dureuill: done

index-scheduler/src/lru.rs
dureuill: Changed the comment to:

```rust
        // Don't use `Iterator::min_by_key` that provides shared references to its elements,
        // so that we can get an exclusive one.
        // This allows to handles the `AtomicU64`s as normal integers without using atomic instructions.
```

Tell me what you think 👍

index-scheduler/src/index_mapper.rs
```diff
@@ -135,23 +514,30 @@ impl IndexMapper {
...
-            resize_operation.wait();
+            match lock.start_deletion(&uuid) {
+                Ok(env_closing) => break env_closing,
+                Err(Some(reopen)) => {
+                    // drop the lock here so that we don't synchronously wait for the index to close.
+                    drop(lock);
+                    tries += 1;
+                    if tries >= 100 {
+                        panic!("Too many attempts to close index {name} prior to deletion.")
+                    }
```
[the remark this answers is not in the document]
dureuill: I added a comment stating the situations where the panic could possibly occur.

I also tested manually triggering a panic at this location. It fails the deletion task, and kills the batch thread, which is renewed for each task, so no other side effect.

index-scheduler/src/index_mapper.rs
```diff
@@ -275,46 +600,53 @@ impl IndexMapper {
             .ok_or_else(|| Error::IndexNotFound(name.to_string()))?;
 
         // we clone here to drop the lock before entering the match
+        let mut tries = 0;
         let index = loop {
-            let index = self.index_map.read().unwrap().get(&uuid).cloned();
+            tries += 1;
+            if tries > 100 {
+                panic!("Too many spurious wake ups while the index is being resized");
+            }
```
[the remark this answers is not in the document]
dureuill: I added a comment stating the situations where the panic could possibly occur.

I also tested manually triggering the panic and trying to search/add documents in an index. In both cases it kills an "actix arbiter" thread and fails the request (not in a very graceful manner, I basically get `curl: (56) Recv failure: Connection reset by peer`).

It looks like actix respawns the arbiter threads, so we're not at risk of running out of threads in this situation.

index-scheduler/src/lru.rs
Kerollmops: Thank you very much, much clearer!

index-scheduler/src/lib.rs
```diff
@@ -407,6 +430,60 @@ impl IndexScheduler {
...
     }
 
+    fn index_budget(
+        tasks_path: &Path,
+        base_map_size: usize,
+        mut task_db_size: usize,
+        max_index_count: usize,
+    ) -> IndexBudget {
+        let budget = utils::dichotomic_search(base_map_size, |map_size| {
+            Self::is_good_heed(tasks_path, map_size)
```
Kerollmops, line 440: You should probably return an error here if it is not due to an index memory map size issue.

index-scheduler/src/lru.rs
```diff
@@ -0,0 +1,203 @@
+//! Thread-safe `Vec`-backend LRU cache using [`std::sync::atomic::AtomicU64`] for synchronization.
```
Kerollmops, line 1: I hope we will release the crate in the near future, just to avoid having to maintain this file in the repository :)

index-scheduler/src/lib.rs
```diff
@@ -1111,6 +1215,16 @@ pub enum TickOutcome {
     WaitForSignal,
 }
 
+/// How many indexes we can afford to have open simultaneously.
+struct IndexBudget {
+    /// Map size of an index.
+    map_size: usize,
+    /// Index count of an index.
+    index_count: usize,
```
Kerollmops, line 1223: WHat do you mean by _Index count of an index_?

meilisearch/src/lib.rs
```diff
@@ -45,6 +45,33 @@ use option::ScheduleSnapshot;
...
+/// Default number of simultaneously opened indexes,
+/// lower for Windows that dedicates a smaller virtual address space to processes.
+///
+/// The value was chosen this way:
+///
+/// - Windows provides a small virtual address space of about 10TiB to processes.
+/// - The chosen value allows for indexes to reach a safe size of 1TiB.
+/// - This can accomodate an unlimited number of indexes as long as they stay below 1TiB size.
+#[cfg(windows)]
+const DEFAULT_INDEX_COUNT: usize = 10;
```
Kerollmops, line 57:
```suggestion
const DEFAULT_INDEX_COUNT: usize = 10;

```

meilisearch/src/lib.rs
```diff
@@ -45,6 +45,33 @@ use option::ScheduleSnapshot;
 
 use crate::error::MeilisearchHttpError;
 
+/// Default number of simultaneously opened indexes,
```
Kerollmops, line 48: Are these two descriptions up-to-date?

meilisearch/src/lib.rs
```diff
@@ -45,6 +45,33 @@ use option::ScheduleSnapshot;
...
+const DEFAULT_INDEX_COUNT: usize = 10;
+/// Default number of simultaneously opened indexes.
+///
+/// The higher, the better for avoiding reopening indexes.
+///
+/// The value was chosen this way:
+///
+/// - Opening an index consumes a file descriptor.
+/// - The default on many unices is about 256 file descriptors for a process.
+/// - 100 is a little bit less than half this value.
```
Kerollmops, line 66: What I find funny is that we always ask users to `limit -Sn 8000` in order to index a large number of documents.

meilisearch/src/routes/indexes/mod.rs
```diff
@@ -90,13 +90,14 @@ pub async fn list_indexes(
...
-        .collect::<Result<Vec<_>, _>>()?;
-
+    let indexes: Vec<Option<IndexView>> =
+        index_scheduler.try_for_each_index(|uid, index| -> Result<Option<IndexView>, _> {
+            if !search_rules.is_index_authorized(uid) {
+                return Ok(None);
+            }
+            Ok(Some(IndexView::new(uid.to_string(), index)?))
+        })?;
+    let indexes: Vec<_> = indexes.into_iter().flatten().collect();
```
Kerollmops, line 100: Aren't we collecting the opened indexes in a single `Vec` here?

index-scheduler/src/utils.rs
```diff
@@ -538,3 +538,37 @@ impl IndexScheduler {
         }
     }
 }
+
+pub fn dichotomic_search(start_point: usize, mut is_good: impl FnMut(usize) -> bool) -> usize {
```
Kerollmops, line 542: I wanted to propose you use a generic "popular" crate to do this dichotomic search (and avoid logic mistakes), but I am astonished that [no single crate contains the word `dichotomic` at all](https://lib.rs/search?q=dichotomic) 😱

index-scheduler/src/index_mapper/index_map.rs
```diff
@@ -0,0 +1,369 @@
...
+    ///
+    /// # Status table
+    ///
+    /// | Previous Status | New Status |
+    /// |-----------------|------------|
+    /// | Missing | Missing |
+    /// | BeingDeleted | BeingDeleted |
+    /// | Closing | Available or Closing depending on generation |
+    /// | Available | Available |
+    ///
```
Kerollmops, line 99: Could you align those array elements please, just for code reviewers :)

index-scheduler/src/index_mapper/index_map.rs
```diff
@@ -0,0 +1,369 @@
...
+    /// Increases the current generation. See documentation for this field.
+    ///
+    /// In the unlikely event that the 2^64 generations would have been exhausted, we simply wrap-around.
+    ///
+    /// For this to cause an issue, one should be able to stop a reader in time after it got a `ReopenableIndex` and before it takes the lock
+    /// to remove it from the unavailable map, and keep the reader in this frozen state for 2^64 closing of other indexes.
+    ///
+    /// This seems overwhelmingly impossible to achieve in practice.
+    fn next_generation(&mut self) -> usize {
+        self.generation = self.generation.wrapping_add(1);
```
Kerollmops, line 200: Do you want to wrap here or crash instead? Wouldn't it be better to unwrap?

Kerollmops (changes requested):
Thank you very much!

index-scheduler/src/lib.rs
```diff
@@ -407,6 +430,60 @@ impl IndexScheduler {
...
+            index_count = max_index_count;
+        }
+        log::debug!("Up to {index_count} indexes of {base_map_size}B opened simultaneously.");
+        IndexBudget { map_size: base_map_size, index_count, task_db_size }
+    }
+
+    fn is_good_heed(tasks_path: &Path, map_size: usize) -> bool {
+        if let Ok(env) =
+            heed::EnvOpenOptions::new().map_size(clamp_to_page_size(map_size)).open(tasks_path)
+        {
```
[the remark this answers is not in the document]
dureuill: Unfortunately this is not easy to do.

- The dichotomic routine is not written to take into account the possibility of error, although we could change it at the cost of increased complexity
- More annoyingly, while unices are rather well-behaved and will return an OS error 12 when unable to allocate, Windows returns up to 4 different errors, making for a total of 5 cases to handle. I think it is best handled in a follow-up PR.

Looking at various scenarios that can happen here:

1. There is a persistent error wrt to opening the task db that is never resolved. In that case, we will fail opening the task db afterwards and get the error at that point.
2. There is a transient error that is resolved after computing the budget, but before opening the task db. In that case the returned budget will be 0 (1 index of 0 bytes, and the task_db_size will be computed to 0), which will attempt to open the indexes to their "natural size" (or the default size in lmdb if creating the index). This will probably render the instance unusable though as the task db will be MDB_MAP_FULL very early. A restart will be necessary.
3. There is a transient error that is resolved while computing the budget: the budget will be lower than it should be. How much of an issue this poses is dependent on when the issue is resolved. The closer we get to 0 budget, the closer to solution (2) we get.
4. There is a transient error that appear after computing the budget: in that case the budget might not be actual anymore and we'll see allocation errors pop when opening indexes. I try to mitigate that case by taking a fraction of the budget (because the budget is shared with all allocations in the application and can suffer from memory fragmentation), but ultimately this PR cannot address the possible issues caused by e.g. a different part of the programming memory mapping with a big size or making big allocations.

I'll update the PR to comment at least on the "persistent error" scenario, which looks the most likely to me.

index-scheduler/src/lib.rs
dureuill: See above

meilisearch/src/routes/indexes/mod.rs
dureuill: no the type is `IndexView` which doesn't keep a reference to the index (only uses it in its constructor). I will add a comment stating so.

index-scheduler/src/index_mapper/index_map.rs
dureuill: well like documented the situation is probably impossible, so I don't have a strong opinion either way. The advantage of the wrapping_add is that we don't pollute the branch predictor with one branch we'll never take, but I recon that's a bit of a micro-optimization

I can change if you feel we could use the additional caution

index-scheduler/src/lib.rs
```diff
@@ -460,15 +537,42 @@ impl IndexScheduler {
...
         let rtxn = self.env.read_txn()?;
         self.index_mapper.index(&rtxn, name)
     }
 
-    /// Return and open all the indexes.
-    pub fn indexes(&self) -> Result<Vec<(String, Index)>> {
+    /// Return the name of all indexes without opening them.
+    pub fn index_names(self) -> Result<Vec<String>> {
+        let rtxn = self.env.read_txn()?;
+        self.index_mapper.index_names(&rtxn)
```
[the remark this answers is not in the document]
dureuill: Yes, because they're removed from the db before the deletion starts.

index-scheduler/src/lru.rs
dureuill: Yep I'll talk about that during pre-release I think

meilisearch/src/lib.rs
dureuill: No, good catch! Even the values were not up to date with the idea that we would use 2TiB of size!

Will update

index-scheduler/src/index_mapper/index_map.rs
Kerollmops: No, keep it that way :)

index-scheduler/src/lib.rs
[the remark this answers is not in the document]
Kerollmops: Thank you very much @dureuill, for this explanation. You can also link to this GitHub comment if the reader needs more info 😃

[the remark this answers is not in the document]
dureuill: Done, see https://github.com/meilisearch/meilisearch/pull/3331/commits/b5489fdb158e293951c6240ea452096a311dcf73

index-scheduler/src/lib.rs
dureuill: Done https://github.com/meilisearch/meilisearch/pull/3331/commits/92a707b63d3df62d07756ab4d0f7d821c8049d96

meilisearch/src/lib.rs
dureuill: done

https://github.com/meilisearch/meilisearch/pull/3331/commits/234d7541348a29156418cec3a3e75562244109c2

meilisearch/src/routes/indexes/mod.rs
dureuill: done, see https://github.com/meilisearch/meilisearch/pull/3331/commits/535fdc1bc0a943a209e6409b6030d1852fc747f8

index-scheduler/src/index_mapper/index_map.rs
dureuill: :eyes: https://github.com/meilisearch/meilisearch/pull/3331/commits/a294767a8cdeae2aff7e905ca90f0bad7f1ab4a5

dureuill:
TODO:

- It appears that indexes are not being registered correctly as `BeingDeleted` currently, this needs checking
- Rebase on `main`
- Squash redundant commits

Follow-ups for subsequent PRs (to be added in the tracking issue and/or opened as different issues):

- Increased Windows/macOS startup time. This could be mitigated by first trying a budget value that should be "good enough" and skipping the dichotomy if it passes.
- Degraded stats performance. This could be mitigated by caching the stats of indexes (eg in the tasks db).
- Finer-grained error handling in the dichotomy: make a difference between allocation errors (continue the dichotomy after marking this value as "bad") and other kinds of errors (stop the dichotomy and report the error to the caller)

Kerollmops (approved):
Beautiful work @dureuill 🎉 I think we can merge this PR and address your comment (☝️) in other PRs. I let you bors merge.

PR #3336, opened 2023-01-12
Add missingneeds: to the git latest tag workflow

curquiza:
Fixes this problem: the workflow to update the latest git tag was triggered despite the first check failed

Capture d’écran 2023-01-12 à 15 07 00

These leads to update our latest git tag: our latest git tag corresponds to the `v1.0.0-rc.0` tag instead of `v0.30.5`. (I'm fixing this right now)

Capture d’écran 2023-01-12 à 15 08 15

### DIFF
--- a/.github/workflows/latest-git-tag.yml
+++ b/.github/workflows/latest-git-tag.yml
@@ -17,6 +17,7 @@ jobs:
 
   update-latest-tag:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Kerollmops (approved):
Right! Good catch and thank you for the fix!
bors merge

curquiza:
bors merge

Issue #2901, opened 2022-10-12
Bringing milli into the Meilisearch repository


Kerollmops:
We talked about bringing milli into the Meilisearch repository, this has different implications, good and bad effects.

First thing first, should we keep the name of _milli_? Shouldn't it be more explicit in what it does? As is: managing one single index at a time. If we publish it on crates.io one day, shouldn't it be prefixed by _meilisearch_, like _meilisearch-index_? We haven't done that for _charabia_ after all.

About the pros, bringing milli into the Meilisearch repository would reduce the synchronization cost. Right now, we need to publish a new GitHub release associated with a tag and update Meilisearch with the latest published tag. It is cumbersome and could bring bugs as git tags are not immutable. A simple `push --force` could break the releases.

## Publishing milli and Meilisearch on crates.io

This is why I want to talk about publishing milli to crates.io. This is the official crate registry that we should use. Publishing to it would force us to be more mature as we expose our work to the world, and more people would use this library. Once a release has been published to crates.io, it can't be deleted, only yanked. This system is here to be sure that builds are reproducible.

Publishing to crates.io will not bring the synchronization complexity we have right now. We can continue to work [with `path = "../milli"` and `version = "0.32.0"` simultaneously](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations). Crates.io doesn't allow to publish crates that use a dependency specified only with a `path`. So to be sure that Meilisearch can be built in a reproducible way, we should also make sure that we publish it on crates.io too.

Another advantage of having Meilisearch on crates.io is that installing a specific version of the engine, or even the last one would be as easy as `cargo install meilisearch`. It is obviously easy for people who already have a Rust toolchain installed.

## Downsides of having a monorepos

Having everything in one single repository could make us less disciplined and makes the developers more prone to creating more inter-dependencies between the repositories in the same workspace. This is because synchronization no more exists it would be easier just to do whatever we want, when we want, without thinking about good encapsulation of the libraries. But I'll let @ManyTheFish talk about that!

irevoire:
> Having everything in one single repository could make us let disciplined and makes the developers more prone to creating more inter-dependencies between the repositories in the same workspace.

[...]

PR #3334, opened 2023-01-12
Add specific error codes `immutable_...`

loiclec:
Add the following error codes:

When an immutable field of API key is sent to the `PATCH /keys` route: 
- `ImmutableApiKeyUid` 
- `ImmutableApiKeyKey`
- `ImmutableApiKeyActions`
- `ImmutableApiKeyIndexes`
- `ImmutableApiKeyExpiresAt`
- `ImmutableApiKeyCreatedAt`
- `ImmutableApiKeyUpdatedAt`

When an immutable field of Index is sent to the `PATCH /indexes/{uid}` route:
- `ImmutableIndexUid`
- `ImmutableIndexCreatedAt`
- `ImmutableIndexUpdatedAt`

PR #3346, opened 2023-01-16
Import milli 🎉

Kerollmops:
Fixes https://github.com/meilisearch/meilisearch/issues/2901

Main work
- integrate the milli repository as an internal crate into this repo  
- Update the Cargo.toml accordingly
- Ensure meilisearch-type now uses the internal milli crate and not the remote repository
- Update the milli's version to follow the meilisearch one

Also
- Removed the beta tests in test CI (will be re-integrated later if needed)
- Move and modify milli's README into the `milli` folder
- remove the script folder from `milli`
- Removed useless CI (release-drafter and enforce-label)

⚠️ Also, import all the `release-v1.0.0` until [a5c4fb](https://github.com/meilisearch/meilisearch/commit/a5c4fbbceaa39748977880b005df0dbf3ed70b63) included (merged of the PR https://github.com/meilisearch/meilisearch/pull/3334)


irevoire:
image

Nice 🙈 
bros merge

loiclec (commented):
I think it's a mistake to bump the version of every crate to 1.0.0.

While we have stabilised the HTTP+CLI interface of the meilisearch executable, we haven't stabilised the API of any of our crates. So in the (near) future, we will make many breaking changes to them. At that point, we will have two choices:
1. Keep their version in sync with the meilisearch releases, and therefore breaking their use of semantic versioning; or:
2. Updating their version to `2.0`, `3.0`, etc. which would once again make them out-of-sync with the meilisearch version

I still think it's a good idea to sync the version of every crate, for convenience. We could do it by updating their version to `0.100.0` instead. When we release Meilisearch v1.1, we can update their version to `0.101.0`. This respects semantic versioning and still stays in sync (somewhat) with the meilisearch releases.

Unfortunately, this problem also applies to the `meilisearch` crate itself. It is composed of two products: a command line tool (the official product, v1.0) and a Rust library (implementation detail like all the other crates). Therefore, I think we should split `meilisearch` into two crates: 

1. `meilisearch` the CLI, `1.0.0`
2. `meilisearch-lib`, `0.100.0`

---

Edit because I don't want my only comment to be a grumpy one: wooooo 🎉  I am excited!

irevoire:
> At that point, we will have two choices:

Actually, we have a third choice. Since milli is a lib made by us and for us with the sole intent of developing meilisearch faster we could also _not_ version milli.

If we’re not publishing milli to crates.io and it doesn’t have a tag then I don’t see any issue to not give it a version (either follow the version of meilisearch or [give it the version `0.0.0`](https://matklad.github.io/2021/08/22/large-rust-workspaces.html))

Now that goes a little bit against the idea of publishing milli to `crates.io` so let me know what you think of it

curquiza:
I'm personally really afraid of having the internal crate out of sync. Will you have the rigor to update them each time you make a change? It makes no point to having un-synchronized versions in the crates if they make no sense because you forget to update them.

> Actually, we have a third choice. Since milli is a lib made by us and for us with the sole intent of developing meilisearch faster we could also not version milli.

I agree 👍 Our work is not providing a perfect milli crate with a changelog and a version, it's a bonus if we succeed in it one day. Versioning our "internal" crates is not a priority to me.

However I understand the future issue if one day we want to publish it to crates.io 🫠

curquiza:
bors try

dureuill:
> I'm personally really afraid of having the internal crate out of sync. Will you have the rigor to update them each time you make a change? It makes no point to having un-synchronized versions in the crates if they make no sense because you forget to update them.

@curquiza I feel like this could be automated in CI, maybe with https://github.com/obi1kenobi/cargo-semver-checks, or with some custom logic based on cargo metadata.

Meanwhile, I see no real drawback to just using 0.0.0 in milli

curquiza:
Good tools @dureuill 👍 But once we choose to publish milli right? I don't see any point in having this additional tool until we decide to publish milli

> Meanwhile, I see no real drawback to just using 0.0.0 in milli

I just have to change this [automation](https://github.com/meilisearch/meilisearch/blob/main/.github/workflows/update-cargo-toml-version.yml), because, so far, it updates every `*/src/Cargo.toml` file.
Should be ok, I hope. I often struggle to anticipate what is easy or not to do in a CI.

dureuill:
💪 good luck curquiza. In my experience, *everything* tends to be more complicated than it should be in CI.

> But once we choose to publish milli right? I don't see any point in having this additional tool until we decide to publish milli

Agreed!

loiclec:
I am afraid something in my original message was misunderstood. I was not talking about milli specifically, but about EVERY crate, that is:
- meilisearch
- meilisearch-types
- meilisearch-auth
- meili-snap
- index-scheduler
- dump
- file-store
- permissive-json-pointer
- milli
- filter-parser
- flatten-serde-json
- json-depth-checker
- benchmarks

None of them (including `meilisearch`!) have a stable API, and therefore none of them should have the version `1.0.0`. I like the idea of using `0.0.0` for all of them (again, including meilisearch!).

curquiza:
Which cargo.toml version is used in the /version route?

loiclec:
It is `meilisearch`'s Cargo.toml if I am not mistaken, thus the idea of either:
- splitting `meilisearch` into `meilisearch-lib` (0.0.0) and `meilisearch-cli` (1.0.0) (but I am not 100% sure it would work correctly, it's easy to verify though); or:
- defining the version manually in the code

dureuill:
> Which cargo.toml version is used in the /version route?

Excellent question! The version in the /version route uses the CARGO_PKG_VERSION of the `meilisearch` package, however there's a `versioning.rs` file in `meilisearch-types` that also uses this environment variable (in the context of `meilisearch-types`), notably to check the DB version

Kerollmops:
One solution to the fact that the meilisearch crate version could be to simply delete the _lib.rs_ file.
Wouldn't it fix the issue? This way nobody but the _main.rs_ file can access the internal API.

About using the `0.0.0` version everywhere, it is a good idea for crates that are not published on crates.io. Maintaining a valid version of the milli crate is not our main concern, we can continue to ship it by using GitHub tags until it is published officially.

dureuill:
Discussed internally, consensus seems to be:

- All unpublished crates, which currently is all crates, switch to v1.0.0
- The README is updated to specify: the crates in this repository are not currently available on crates.io and do not follow [semver](https://semver.org/) conventions

README.md
```diff
@@ -101,3 +101,7 @@ Meilisearch is a search engine created by [Meili](https://www.welcometothejungle
 - For everything else, please check [this page listing some of the other places where you can find us](https://docs.meilisearch.com/learn/what_is_meilisearch/contact.html)
 
 Thank you for your support!
+
+## 📦 Internal crates and their versioning
```
curquiza, line 105: Hum, I don't really like this title, it's too technical compared to what we have on it. We are a famous project and now the README is almost like a landing page. However I don't want to block the PR, I will make a PR suggestion later
(I notice we don't talk about contribution in the README, it will be also an opportunity to add this 😄)

curquiza (approved):

curquiza:
bors try

Kerollmops:
And here we are! The day has come to import milli once and for all into the Meilisearch repository 🎉
bors merge

PR #3370, opened 2023-01-18
make the swap indexes not found errors return an IndexNotFound error-code

irevoire:


Kerollmops (approved):
Ok, so now when we do a swap and indexes don't exist, we return an `index_not_found`, right? I like that!

irevoire:
Yep exactly! That was the old behaviour, don't know why I changed it, and since we had no integration test on the index-swap, I missed it :grimacing: 

bors merge

PR #3372, opened 2023-01-18
Enhance facet string normalization

### LINKED ISSUE #3260: sort by is incorrect for accented characters
## Describe the bug
According to the Hungarian ABC, certain vowels have accented pairs. For example: a-á, e-é, i-í, ü-ű, o-ó ...
When I sort the index based on text, the accented characters are always the last.

## Version
- Meilisearch engine: `0.30.4`

## Screenshots
The value of the `sortBy.` field can be 1 or 2, number type.
image
image

The letter é should be between the letters e and f.
image

## Additional context
https://en.wikipedia.org/wiki/Hungarian_alphabet

ManyTheFish:
# Pull Request

Use compatibility decomposition normalizer in facet string extraction in order to have a more human friendly sort order.

Now, [é (U+00E9)](https://www.compart.com/fr/unicode/U+00E9) is converted to [e (U+0065)](https://www.compart.com/fr/unicode/U+0065) + [◌́ (U+0301)](https://www.compart.com/fr/unicode/U+0301). This way any word starting with an accented/diacritized version of a character is put just after the words starting with the unaccented version of the character.

## Related issue

Fixes #3260


Kerollmops (approved):
That is great. Thank you!

irevoire:
> Now, [é (U+00E9)](https://www.compart.com/fr/unicode/U+00E9) is converted to [e (U+0065)](https://www.compart.com/fr/unicode/U+0065) + [◌́ (U+0301)](https://www.compart.com/fr/unicode/U+0301). This way any word starting with an accented/diacritized version of a character is put just after the words starting with the unaccented version of the character.

Is that true, though?
```
et
été
ez
```

The end of your message tells me these words would be sorted like that, but if I understood correctly the first part they're actually going to be sorted like that;
```
et
ez
été
```

ManyTheFish:
thanks @Kerollmops and @irevoire,

bors merge

Issue #3384, opened 2023-01-18
While using the facets parameter of the search it returns a bad request when everything is good except your field is not filterable


curquiza:

dureuill:
/shrug Somehow didn't close automatically after #3389

PR #3389, opened 2023-01-19
Return `invalid_search_facets` rather than `bad_request` when using facet on a non filterable attribute

dureuill:
# Pull Request

## Related issue

Fixes https://github.com/meilisearch/meilisearch/issues/3384

## What does the PR does

- title
- also adds a test

## PR checklist

### DIFF
--- a/meilisearch/tests/search/errors.rs
+++ b/meilisearch/tests/search/errors.rs
@@ -322,6 +322,37 @@ async fn search_bad_facets() {
 }
 
 #[actix_rt::test]
+async fn search_non_filterable_facets() {
+    let server = Server::new().await;
+    let index = server.index("test");
+    index.update_settings(json!({"filterableAttributes": ["title"]})).await;
+    // Wait for the settings update to complete
+    index.wait_task(0).await;
+
+    let (response, code) = index.search_post(json!({"facets": ["doggo"]})).await;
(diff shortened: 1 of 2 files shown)


meilisearch/tests/search/errors.rs
```diff
@@ -321,6 +321,25 @@ async fn search_bad_facets() {
...
+    snapshot!(code, @"400 Bad Request");
+    snapshot!(json_string!(response), @r###"
+    {
+      "message": "Invalid facet distribution, the fields `doggo` are not set as filterable.",
+      "code": "invalid_search_facets",
+      "type": "invalid_request",
+      "link": "https://docs.meilisearch.com/errors#invalid-search-facets"
+    }
+    "###);
+    // Can't make the `attributes_to_highlight` fail with a get search since it'll accept anything as an array of strings.
```
irevoire, line 340: Could you write a test with a `search_get` since you actually can in this case

irevoire (changes requested):
Nice, i would just like to slightly improve the test

meilisearch/tests/search/errors.rs
dureuill: done

irevoire (approved):
Nice, thank you :ok: 
bors merge

PR #3405, opened 2023-01-19
Implement geo bounding box

### LINKED ISSUE #2761: Improve geo-search: introduction of `_geoBoundingBox`
Following this product discussion: https://github.com/meilisearch/product/discussions/431
Specification: https://github.com/meilisearch/specifications/pull/197

## API

`_geoBoundingBox((lat, lng), (lat, lng))` using the [`filter` search parameter](https://docs.meilisearch.com/reference/api/search.html#filter)

`{lat, lng}` <- Top left corner,
`{lat, lng}` <- Bottom right corner

## Implementation

- [ ] Implement changes
- [X] Create the spec about name and behavior changes -> https://github.com/meilisearch/specifications/pull/197

## Impacted teams

Pinging @meilisearch/docs-team, also @meilisearch/integration-team even if I'm not sure. Feel free to unsubscribe from this issue if I was wrong to ping you.

---

<s>Pinging @mlemesle since I think you are interested in implementing this 😄 </s>
This issue is opened to anyone!

Feel free to ask any questions to implement this issue. The core team is available to guide you through this contribution, even if it's your first one.
We recommend referring to the [CONTRIBUTING.md guidelines](https://github.com/meilisearch/meilisearch/blob/main/CONTRIBUTING.md), especially if it's your first contribution to this repository.
Thanks in advance for your involvement ❤️

curquiza:
Following https://github.com/meilisearch/milli/pull/672 (work from @gmourier)

Fixes #2761

### DIFF
--- a/filter-parser/src/lib.rs
+++ b/filter-parser/src/lib.rs
@@ -18,6 +18,7 @@
 //! doubleQuoted   = "\"" .* all but double quotes "\""
 //! word           = (alphanumeric | _ | - | .)+
 //! geoRadius      = "_geoRadius(" WS* float WS* "," WS* float WS* "," float WS* ")"
+//! geoBoundingBox = "_geoBoundingBox((" WS * float WS* "," WS* float WS* "), (" WS* float WS* "," WS* float WS* ")")
 //! ```
 //!
 //! Other BNF grammar used to handle some specific errors:
@@ -130,6 +131,7 @@ pub enum FilterCondition<'a> {
     Or(Vec<Self>),
     And(Vec<Self>),
     GeoLowerThan { point: [Token<'a>; 2], radius: Token<'a> },
+    GeoBoundingBox { top_left_point: [Token<'a>; 2], bottom_right_point: [Token<'a>; 2]},
 }
 
 impl<'a> FilterCondition<'a> {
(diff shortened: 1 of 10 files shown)


filter-parser/src/lib.rs
```diff
@@ -70,7 +71,7 @@ const MAX_FILTER_DEPTH: usize = 200;
 #[derive(Debug, Clone, Eq)]
 pub struct Token<'a> {
     /// The token in the original input, it should be used when possible.
-    span: Span<'a>,
+    pub span: Span<'a>,
```
dureuill, line 74: It feels a bit "strange" to me to expose this field publicly, as it feels to me like this struct has an invariant of "uses the original span", unless its value was afterwards modified.

It doesn't feel right for me to be able to mutate the span if e.g. we already have a `value` to `Some`.

Should we consider an `original_span` getter instead?

filter-parser/src/lib.rs
```diff
@@ -100,7 +101,7 @@ impl<'a> Token<'a> {
     }
 
     pub fn parse_finite_float(&self) -> Result<f64, Error> {
-        let value: f64 = self.span.parse().map_err(|e| self.as_external_error(e))?;
+        let value: f64 = self.value().parse().map_err(|e| self.as_external_error(e))?;
```
dureuill, line 104: I propagated the change to the original `parse<T>` function that replace span by value() here. This feels like a bug fix indeed, but I am uncertain of its impact. Can you give me some insight as to why this was changed?

dureuill (changes requested):
Hello :wave:,

just pushed a rebase of this branch, I do have some concerns though! See inline comments.

filter-parser/src/lib.rs
dureuill: OK I saw your :+1: tamo so I implemented this change in a new commit.

irevoire: Noice

filter-parser/src/lib.rs
dureuill: Thanks for the explanation ☺️ Is this documented somewhere in the code? If not, I feel like it possibly ought to?

irevoire: I added a clear explanation (I think) about the implementation of the geoboundingbox + simplified a little bit the usage of the `Token`.

dureuill (approved):
Thank you for the additional comments, I fix a few typos but otherwise OK for me 👌

irevoire (dismissed):

curquiza:
bors merge

irevoire:
bors cancel

irevoire:
I implemented the two « products » change.

1. Now the syntax for the geo bounding box is; `_geoBoundingBox([lat, lng], [lat, lng])`.
2. If the **top** left corner is found below the **bottom** right corner we throw an error

milli/src/index.rs
```diff
@@ -1583,20 +1583,27 @@ pub(crate) mod tests {
             .unwrap();
         insta::assert_debug_snapshot!(search_result.candidates, @"RoaringBitmap<[4]>");
 
+        // the requests that doesn't make sense
```
dureuill, line 1586:
```suggestion
        // the requests that don't make sense
```

or

```suggestion
        // the request that doesn't make sense
```

dureuill (changes requested):
Thank you for the changes. If I understand correctly, `_geoBoundingBox([-80, 0], [80, 0])` is illegal, and should be e.g. `_geoBoundingBox([80, 0], [-80, 0])`?

If so isn't this a bit backward? Shouldn't we specify the smaller numbers first?

Apologies if this has already been discussed, in that case do you have a link to this discussion?

irevoire:
Hey, I understand the struggle you had while reading the code.

It should be easier to understand with this screenshot hehe;
image

As you can see, we're asking for the « top left corner » and _then_ the « bottom right corner ».

dureuill (approved):
Wow thank you for the explanation. Basically I'm used to "inverted axes" and here they're not inverted.

If this is what is usual in the field, then go go go 👍

irevoire:
bors merge

PR #3407, opened 2023-01-19
Add Cargo feature for LMDB's POSIX semaphores

GregoryConrad:
See https://github.com/meilisearch/milli/pull/757

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1747,8 +1747,8 @@ checksum = "2540771e65fc8cb83cd6e8a237f70c319bd5c29f78ed1084ba5d50eeac86f7f9"
 
 [[package]]
 name = "heed"
-version = "0.12.4"
-source = "git+https://github.com/meilisearch/heed?tag=v0.12.4#7a4542bc72dd60ef0f508c89900ea292218223fb"
+version = "0.12.5"
+source = "git+https://github.com/meilisearch/heed?tag=v0.12.5#4158a6c484752afaaf9e2530a6ee0e7ab0f24ee8"
 dependencies = [
  "byteorder",
  "heed-traits",
@@ -1765,12 +1765,12 @@ dependencies = [
 [[package]]
 name = "heed-traits"
 version = "0.7.0"
-source = "git+https://github.com/meilisearch/heed?tag=v0.12.4#7a4542bc72dd60ef0f508c89900ea292218223fb"
+source = "git+https://github.com/meilisearch/heed?tag=v0.12.5#4158a6c484752afaaf9e2530a6ee0e7ab0f24ee8"
 
 [[package]]
 name = "heed-types"
 version = "0.7.2"
-source = "git+https://github.com/meilisearch/heed?tag=v0.12.4#7a4542bc72dd60ef0f508c89900ea292218223fb"
+source = "git+https://github.com/meilisearch/heed?tag=v0.12.5#4158a6c484752afaaf9e2530a6ee0e7ab0f24ee8"
 dependencies = [
  "bincode",
  "heed-traits",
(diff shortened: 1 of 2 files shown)


GregoryConrad:
Also, not sure if this PR should be made to `main` or not; did I choose the correct branch?

irevoire:
Hey @GregoryConrad, we're currently moving milli to the meilisearch repository to reduce the time we spent in useless synchronisation between both repo.
Thus, it would be better to open your pr on meilisearch -- main 👍

GregoryConrad:
Great; I noticed there was a `release-v1.0.0` branch and wasn't sure if this PR should be made there instead.

irevoire:
Yeah, our workflow work as-is;
-> During one month, we're in the sprint phase and implement new features for the next release of meilisearch. During this time we push & merge on main
-> Once this is over we enter the pre-release phase, it last one month as well, during this time we merge ~only bug fixes required for the next release in `release-v1.x.0`

So, since your change is not required for the next release it would go in meilisearch -- main 👍

dureuill:
Hey @GregoryConrad :wave: thanks for your work on this :-)

Questions to whom it may concern on the team:

1. should this be a feature flag at the milli level, or should we just always enable/disable this flag at the milli level? 
2. If we want a flag for milli, same question for meilisearch itself.

More generally, what are the advantages/drawbacks of using POSIX semaphores on macOS? I seem to remember off the top of my head:

- The number of POSIX semaphores is not limited to 10 like the number of sysv semaphores, fixing some of the os errors we sometimes get in tests under macOS (something we could consider for milli/meilisearch tests under macOS)
- As OP noted, only POSIX semaphores are supported when using App Sandbox
- POSIX semaphores only provide "filesystem persistence" rather than "process persistence" like sysv semaphores. What this means is that if milli crashes in an *impromptu* manner, then the semaphore won't be released, which in the case of LMDB if my understanding is correct could result in meilisearch believing that the DB is locked by another process when it actually is not. In Meilisearch we could probably work around this by ignoring the lock (since the same DB is not supposed to be used by multiple concurrent instances just yet) or by asking the user to release the lock by deleting the lock file, but both workarounds add a new failure mode
- I have no idea of whether the performance characteristics of both kinds of semaphores are similar or not

Is the above correct, and am I missing something?

GregoryConrad:
Hi!

> 1. should this be a feature flag at the milli level, or should we just always enable/disable this flag at the milli level?

It really is just a `heed` feature, but due to how Cargo dependencies work, you can't reliably set a feature on a transitive dependency so it also needs to be exposed in milli.

I suppose the `build.rs` implementation over in `lmdb-master-sys` could instead check the binary target directly and enable/disable, but I think we should still leave it up to the end-user who is compiling the binary. There are use cases where you may want to use POSIX semaphores on linux and not SysV (which would probably be the default).

> 2. If we want a flag for milli, same question for meilisearch itself.

Maybe. My use case for this Cargo feature was compiling to macOS and iOS applications. It could work well for Linux targets as well, at least when building containers. That is something you could benchmark in meilisearch and see if it's worth it. I.e., POSIX semaphores provide a slight speed improvement over SysV semaphores at the cost of possibly leaking the semaphore in the event of a crash. However, if dealing with a container, the container would simply restart and any semaphore should be released. Now, this speed improvement (if there is even a noticeable one in benchmarks) might not even be worth changing the build scripts you have; it is likely bottlenecks are elsewhere and not in something as mundane as a semaphore difference.

> More generally, what are the advantages/drawbacks of using POSIX semaphores on macOS? I seem to remember off the top of my head:

You got all of the ones I am aware of. I will note though that SysV semaphores require more overhead and are more complex than POSIX semaphores, and consequently tend to be slower (not sure by what margin).

dureuill (approved):
Thanks for your answers! 

We can consider using POSIX semaphores in Meilisearch independently from this PR. It looks good to me.

bors merge

dureuill:
ubuntu mirrors are not feeling well today :/

let's try again

bors merge

dureuill:
How is Ubuntu faring today?

bors merge

PR #3417, opened 2023-01-24
Allow multiple searches in a single request

### LINKED ISSUE #3427: Perform multiple searches in a single query
Related product discussion: https://github.com/meilisearch/product/discussions/489
Related specification: WIP

### TODO

- [ ] For implementation, check with the product squad to get how it should be implemented
    - A prototype PR is available: #3417
- [ ] Merge the changes into main
- [ ] Update the specification

### Impacted teams

- @meilisearch/integration-team
- @meilisearch/docs-team: new endpoint to document.

dureuill:
# Pull Request

## Related issue
Fixes #3427

### User standpoint

- Adds a new `/multi-search` entry point (not to be confused with the existing `/{index_uid}/search` entry points) that accepts a POST whose body is an object containing an array of queries.
    - Each query must specify on which index it acts by providing its `indexUid`. Other parameters are identical to the one in the existing search routes (`q`, `limit`, etc.).
    - The response is a JSON object containing an array of the results for each search query as if it had been performed using the `/{index_uid}/search` routes.

### Implementation standpoint

- Refactor authentication module:
  -  Allow tenant token to be checked even without an index in URL
  - Add `meilisearch-auth` as a dependency to `index-scheduler` so as to have a working method of checking if the indexes are authorized there that takes into account both the API key and the tenant token (existing method relied on a behavior that was returning the allowed indexes from the API key as long as there weren't any tenant token)
  - Make `AuthFilter` an object with invariants and so its fields are now private
  - Use the methods of `AuthFilter` to know if an index is authorized rather than relying on its internal search rules.
  - Make tenant token search rules optional and `None` when the `AuthFilter` was not built with a tenant token.
- Add a new `routes::index::search::multiple_search` module containing a post handler that performs the same work as the existing `routes::index::search` post handler, but in a loop.
  - Add various tests
  - Add authentication test suite 

### Sample request

<details>
<summary>
Click to see request/response
</summary>

```json
~/datasets
❯ curl \
  -X POST 'http://localhost:7700/multi-search' \
  --data-binary '{"queries": [{ "indexUid": "index-0", "q": "toto", "limit": 1 }, {"indexUid": "index-1", "q": "titi", "limit": 1}]}' | jsonxf
{ "results": [
  {
    "indexUid": "index-0",
    "hits": [
      {
        "id": 20480,
        "title": "Toto - 25th Anniversary - Live in Amsterdam",
        "overview": "Filmed in High Definition in Amsterdam on Toto's 25th Anniversary Tour in 2003, this stunning concert captures the band at their very best, reunited with original vocalist Bobby Kimball. The set combines all their hits with tracks from their latest album \"Through the Looking Glass\" and other live favorites, performed in front of a wildly enthusastic sell-out crowd. Extras include 35 minute behind-the-scenes film following the band through various stages of their world tour including footage from Japan, Thailand, South Korea, and France.  Toto celebrate their 25th anniversary with this blistering live concert, filmed in Amsterdam on February 25th, 2003. Proving they've still got exactly what it takes to move a crowd, the band perform a mixture of medley's, solo spots, and huge hits. Tracks include \"Rosanna,\" \"Africa,\" \"Hold The Line,\" a cover of the Beatles' \"While My Guitar Gently Weeps,\" and many more.",
        "genres": [
          "Music"
        ],
        "poster": "https://image.tmdb.org/t/p/w500/7SCbUPwoB8Z7VUIA1Rn1WWwjNiT.jpg",
        "release_date": 1064275200
      }
    ],
    "query": "toto",
    "processingTimeMs": 1,
    "limit": 1,
    "offset": 0,
    "estimatedTotalHits": 17
  },
  {
    "indexUid": "index-1",
    "hits": [
      {
        "id": 41212,
        "title": "Titicut Follies",
        "overview": "The film is a stark and graphic portrayal of the conditions that existed at the State Prison for the Criminally Insane at Bridgewater, Massachusetts. TITICUT FOLLIES documents the various ways the inmates are treated by the guards, social workers and psychiatrists.",
        "genres": [
          "Documentary"
        ],
        "poster": "https://image.tmdb.org/t/p/w500/2Ju5hn1ofOPeP1eRJtQWakiHuhW.jpg",
        "release_date": -70934400
      }
    ],
    "query": "titi",
    "processingTimeMs": 0,
    "limit": 1,
    "offset": 0,
    "estimatedTotalHits": 7
  }
]}
```

</details>

## PR checklist

### DIFF
--- a/index-scheduler/Cargo.toml
+++ b/index-scheduler/Cargo.toml
@@ -19,6 +19,7 @@ dump = { path = "../dump" }
 enum-iterator = "1.1.3"
 file-store = { path = "../file-store" }
 log = "0.4.14"
+meilisearch-auth = { path = "../meilisearch-auth" }
 meilisearch-types = { path = "../meilisearch-types" }
 page_size = "0.5.0"
 roaring = { version = "0.10.0", features = ["serde"] }
--- a/meilisearch/src/routes/indexes/mod.rs
+++ b/meilisearch/src/routes/indexes/mod.rs
@@ -89,11 +89,11 @@ pub async fn list_indexes(
     index_scheduler: GuardedData<ActionPolicy<{ actions::INDEXES_GET }>, Data<IndexScheduler>>,
     paginate: AwebQueryParameter<ListIndexes, DeserrQueryParamError>,
 ) -> Result<HttpResponse, ResponseError> {
-    let search_rules = &index_scheduler.filters().search_rules;
+    let filters = index_scheduler.filters();
     let indexes: Vec<_> = index_scheduler.indexes()?;
     let indexes = indexes
         .into_iter()
-        .filter(|(name, _)| search_rules.is_index_authorized(name))
+        .filter(|(name, _)| filters.is_index_authorized(name))
         .map(|(name, index)| IndexView::new(name, &index))
         .collect::<Result<Vec<_>, _>>()?;
 
@@ -120,7 +120,8 @@ pub async fn create_index(
 ) -> Result<HttpResponse, ResponseError> {
     let IndexCreateRequest { primary_key, uid } = body.into_inner();
 
-    let allow_index_creation = index_scheduler.filters().search_rules.is_index_authorized(&uid);
+    // FIXME: allow_index_creation?
+    let allow_index_creation = index_scheduler.filters().is_index_authorized(&uid);
     if allow_index_creation {
         analytics.publish(
             "Index Created".to_string(),
--- a/meilisearch/src/routes/indexes/mod.rs
+++ b/meilisearch/src/routes/indexes/mod.rs
@@ -120,8 +120,7 @@ pub async fn create_index(
 ) -> Result<HttpResponse, ResponseError> {
     let IndexCreateRequest { primary_key, uid } = body.into_inner();
 
-    // FIXME: allow_index_creation?
-    let allow_index_creation = index_scheduler.filters().is_index_authorized(&uid);
+    let allow_index_creation = index_scheduler.filters().allow_index_creation(&uid);
     if allow_index_creation {
         analytics.publish(
             "Index Created".to_string(),
--- a/meilisearch/src/routes/indexes/search.rs
+++ b/meilisearch/src/routes/indexes/search.rs
@@ -3,7 +3,6 @@ use actix_web::{web, HttpRequest, HttpResponse};
 use deserr::actix_web::{AwebJson, AwebQueryParameter};
 use index_scheduler::IndexScheduler;
 use log::debug;
-use meilisearch_auth::IndexSearchRules;
 use meilisearch_types::deserr::query_params::Param;
 use meilisearch_types::deserr::{DeserrJsonError, DeserrQueryParamError};
 use meilisearch_types::error::deserr_codes::*;
@@ -17,9 +16,9 @@ use crate::extractors::authentication::policies::*;
 use crate::extractors::authentication::GuardedData;
 use crate::extractors::sequential_extractor::SeqHandler;
 use crate::search::{
-    perform_search, MatchingStrategy, SearchQuery, DEFAULT_CROP_LENGTH, DEFAULT_CROP_MARKER,
-    DEFAULT_HIGHLIGHT_POST_TAG, DEFAULT_HIGHLIGHT_PRE_TAG, DEFAULT_SEARCH_LIMIT,
-    DEFAULT_SEARCH_OFFSET,
+    add_search_rules, perform_search, MatchingStrategy, SearchQuery, DEFAULT_CROP_LENGTH,
+    DEFAULT_CROP_MARKER, DEFAULT_HIGHLIGHT_POST_TAG, DEFAULT_HIGHLIGHT_PRE_TAG,
+    DEFAULT_SEARCH_LIMIT, DEFAULT_SEARCH_OFFSET,
 };
 
 pub fn configure(cfg: &mut web::ServiceConfig) {
@@ -97,26 +96,6 @@ impl From<SearchQueryGet> for SearchQuery {
             highlight_post_tag: other.highlight_post_tag,
             crop_marker: other.crop_marker,
             matching_strategy: other.matching_strategy,
-        }
-    }
-}
-
-/// Incorporate search rules in search query
-fn add_search_rules(query: &mut SearchQuery, rules: IndexSearchRules) {
-    query.filter = match (query.filter.take(), rules.filter) {
-        (None, rules_filter) => rules_filter,
-        (filter, None) => filter,
-        (Some(filter), Some(rules_filter)) => {
-            let filter = match filter {
-                Value::Array(filter) => filter,
-                filter => vec![filter],
-            };
-            let rules_filter = match rules_filter {
-                Value::Array(rules_filter) => rules_filter,
-                rules_filter => vec![rules_filter],
-            };
-
-            Some(Value::Array([filter, rules_filter].concat()))
         }
     }
 }
--- a/meilisearch/src/routes/indexes/search.rs
+++ b/meilisearch/src/routes/indexes/search.rs
@@ -159,9 +159,7 @@ pub async fn search_with_url_query(
     let mut query: SearchQuery = params.into_inner().into();
 
     // Tenant token search_rules.
-    if let Some(search_rules) =
-        index_scheduler.filters().search_rules.get_index_search_rules(&index_uid)
-    {
+    if let Some(search_rules) = index_scheduler.filters().get_index_search_rules(&index_uid) {
         add_search_rules(&mut query, search_rules);
     }
 
@@ -193,9 +191,7 @@ pub async fn search_with_post(
     debug!("search called with params: {:?}", query);
 
     // Tenant token search_rules.
-    if let Some(search_rules) =
-        index_scheduler.filters().search_rules.get_index_search_rules(&index_uid)
-    {
+    if let Some(search_rules) = index_scheduler.filters().get_index_search_rules(&index_uid) {
         add_search_rules(&mut query, search_rules);
     }
 
--- a/meilisearch/src/routes/mod.rs
+++ b/meilisearch/src/routes/mod.rs
@@ -22,6 +22,7 @@ const PAGINATION_DEFAULT_LIMIT: usize = 20;
 mod api_key;
 mod dump;
 pub mod indexes;
+mod multi_search;
 mod swap_indexes;
 pub mod tasks;
 
@@ -33,6 +34,7 @@ pub fn configure(cfg: &mut web::ServiceConfig) {
         .service(web::resource("/stats").route(web::get().to(get_stats)))
         .service(web::resource("/version").route(web::get().to(get_version)))
         .service(web::scope("/indexes").configure(indexes::configure))
+        .service(web::scope("/multi-search").configure(multi_search::configure))
         .service(web::scope("/swap-indexes").configure(swap_indexes::configure));
 }
 
--- a/meilisearch/src/routes/mod.rs
+++ b/meilisearch/src/routes/mod.rs
@@ -237,10 +237,9 @@ async fn get_stats(
     analytics: web::Data<dyn Analytics>,
 ) -> Result<HttpResponse, ResponseError> {
     analytics.publish("Stats Seen".to_string(), json!({ "per_index_uid": false }), Some(&req));
-    let search_rules = &index_scheduler.filters().search_rules;
+    let filters = index_scheduler.filters();
 
-    let stats =
-        create_all_stats((*index_scheduler).clone(), (*auth_controller).clone(), search_rules)?;
+    let stats = create_all_stats((*index_scheduler).clone(), (*auth_controller).clone(), filters)?;
 
     debug!("returns: {:?}", stats);
     Ok(HttpResponse::Ok().json(stats))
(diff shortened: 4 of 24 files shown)


qdequele:
Hello @dureuill, awesome work! Just a question, from what I see on your PR, your result return an array of object containing the response of a normal search request,  but there is no way to know which hits if for which index. You could imagine something like:

dureuill:
Hello @qdequele 👋 thanks for the feedback on my PR!

> there is no way to know which hits if for which index

Not from the response alone, indeed. The array of results is in the same order as the array of queries, so it is guaranteed that the index in the i<sup>th</sup> result is the index of the i<sup>th</sup> request.

That being said, I see that we already return the other parameters of the request in the results, so we could add the `indexUid` too for the sake of consistency.

I'm curious though, what's the typical use case for duplicating this data?

Kerollmops:
I would like to add a little comment about adding the `indexUid` in the responses of each index: it is useless for when it is used in disjunctive facet operations as all the queries are sent to the same index.

However, I don't see any blocking in adding the name of the index. It can help the users for when they just query different indexes.

dureuill:
Done.

<details>
<summary>
Click to see request/response
</summary>

```json
~/datasets
❯ curl \
  -X POST 'http://localhost:7700/search' \
  --data-binary '[{ "indexUid": "index-0", "q": "toto", "limit": 1 }, {"indexUid": "index-1", "q": "titi", "limit": 1}]' | jsonxf
[
  {
    "indexUid": "index-0",
    "hits": [
      {
        "id": 20480,
        "title": "Toto - 25th Anniversary - Live in Amsterdam",
        "overview": "Filmed in High Definition in Amsterdam on Toto's 25th Anniversary Tour in 2003, this stunning concert captures the band at their very best, reunited with original vocalist Bobby Kimball. The set combines all their hits with tracks from their latest album \"Through the Looking Glass\" and other live favorites, performed in front of a wildly enthusastic sell-out crowd. Extras include 35 minute behind-the-scenes film following the band through various stages of their world tour including footage from Japan, Thailand, South Korea, and France.  Toto celebrate their 25th anniversary with this blistering live concert, filmed in Amsterdam on February 25th, 2003. Proving they've still got exactly what it takes to move a crowd, the band perform a mixture of medley's, solo spots, and huge hits. Tracks include \"Rosanna,\" \"Africa,\" \"Hold The Line,\" a cover of the Beatles' \"While My Guitar Gently Weeps,\" and many more.",
        "genres": [
          "Music"
        ],
        "poster": "https://image.tmdb.org/t/p/w500/7SCbUPwoB8Z7VUIA1Rn1WWwjNiT.jpg",
        "release_date": 1064275200
      }
    ],
    "query": "toto",
    "processingTimeMs": 1,
    "limit": 1,
    "offset": 0,
    "estimatedTotalHits": 17
  },
  {
    "indexUid": "index-1",
    "hits": [
      {
        "id": 41212,
        "title": "Titicut Follies",
        "overview": "The film is a stark and graphic portrayal of the conditions that existed at the State Prison for the Criminally Insane at Bridgewater, Massachusetts. TITICUT FOLLIES documents the various ways the inmates are treated by the guards, social workers and psychiatrists.",
        "genres": [
          "Documentary"
        ],
        "poster": "https://image.tmdb.org/t/p/w500/2Ju5hn1ofOPeP1eRJtQWakiHuhW.jpg",
        "release_date": -70934400
      }
    ],
    "query": "titi",
    "processingTimeMs": 0,
    "limit": 1,
    "offset": 0,
    "estimatedTotalHits": 7
  }
]
```

</details>

meilisearch/src/routes/indexes/search.rs
```diff
@@ -82,6 +82,64 @@ pub struct SearchQueryGet {
     matching_strategy: MatchingStrategy,
 }
 
+pub mod multiple_search {
```
irevoire, line 85: I don’t really see the point of doing this module instead of just dropping the function in the current module.

Also we usually try to follow the routes we’re in, thus, if the route associated to this module is `/search`, it shouldn’t be in `/indexes/search`.

I would prefer to;
1. Get rids of this module
2. Move the content of the module in `routes/mod.rs` or create a new file containing this code in `routes/multi_search.rs`

dureuill: The module is used to separate the multi query search from index query.

It reuses [`add_search_rules`](https://github.com/meilisearch/meilisearch/blob/ad49a8121ddb6ba97b57663fdb0bbf65252a3fb5/meilisearch/src/routes/indexes/search.rs#L176) from the parent module. Should I duplicate this function as well or move it to `meilisearch/src/search.rs`?

irevoire: I would say we move it to `/search.rs` since that already contains all the « helper » function we need to make both the get and post search 🤔

dureuill: done

irevoire (commented):
Codewise it looks nice; I guess it's just missing a bunch of tests, but maybe that can come after making the prototype?

I don't think we want to merge before the feature is tested though :eyes:

dureuill:
Added some tests.

If any of the search requests is faulty, the response does not specify which one. This is an ergonomic issue, do you think we should address this?

dureuill:
Fresh weekly update :-)

- Queries and results are now wrapped in a JSON object. This is, notably, for future-proofing and consistency with the rest of the API.
- `index_uid` is now wrapped in an `IndexUid` object for more consistent error handling (illegal index uid name, missing index uid field).
- Updated and added some tests.

LukasKalbertodt:
Do I understand this PR correctly in that it does *not* add support for "aggregated result set" as discussed in [this product discussion](https://github.com/meilisearch/product/discussions/489)? It seems like everyone in that discussion specifically wants the aggregated result set. If this new `/search` endpoint is basically just a loop, I don't really see the point to be honest. I can already write a loop myself and search multiple indexes; the main problem instead is to merge the results of the individual indexes into one list with useful ranking. 

In the software I'm developing as part of my day-job, we are already using Meili but need to have a hacky "merge results of different indexes" algorithm. See [here](https://github.com/elan-ev/tobira/blob/629c6ad36076925539a639497ce9c1da11f3b382/backend/src/api/model/search/mod.rs#L142-L238). The parallel search of both indexes is easy. But there is no good way for us to merge the results. (It would help a lot of there was a `score` attached to each hit by which we could simply merge & sort. But as I understand, Meili is not really using scores for ranking.)

Sorry if this sounds so negative. I appreciate all the work going into Meili! But yes: what about this use case? It seems to be the main reason people requested this feature.

gmourier:
Hello @LukasKalbertodt,

We have identified two use cases.

1. Search multiple indexes and retrieve separate results.
2. Search multiple indexes and retrieve aggregated results in a single list. This will be explored in a second iteration. 

We will start by proposing an endpoint facilitating (1) (a multi-search request endpoint) because:

- It also makes it easier to compute disjunctive facets.
- It removes the workaround of parallel query management, which can be a friction: 
    - for some beginners
    - it generates a lot of queries when a user types in the search bar, so it should better integrate into any cloud usage.

Concerning (2), we will work on retrieving ranking details for a search result. This is the first track that could users to re-rank the results within a unique list on the client side. Given that, we could think about doing this natively on the engine side.

I hope it is clearer. Don't hesitate to tell me if you need more details. Thanks a lot!

LukasKalbertodt:
> This will be explored in a second iteration.

Thank you for the swift reply! That's basically all I wanted to hear; I was simply confused that it was not mentioned in this PR yet. 
So yes, unfortunate for me that I have to wait some more, but glad to hear you are still planning to address the use case (2.). Thanks!

irevoire (approved):
Looks pretty good now 😁 

I guess we should wait for more feedback on the prototype before merging? But you have my approve

dureuill:
Thank you @irevoire ❤️ 

We need to rebase to accommodate for the wildcards in tenant tokens though. We'll take a look with @Kerollmops this very afternoon 😃

dureuill:
Renamed the new route from `/search` to `/multi-search`.

The PR description is up-to-date

dureuill:
Changed event name in analytics

index-scheduler/Cargo.toml
```diff
@@ -19,6 +19,7 @@ dump = { path = "../dump" }
 enum-iterator = "1.1.3"
 file-store = { path = "../file-store" }
 log = "0.4.14"
+meilisearch-auth = { path = "../meilisearch-auth" }
```
irevoire, line 22: What do you think about that @Kerollmops?

I think we were against it some time ago, but I can't remember why.

meilisearch/src/routes/indexes/mod.rs
```diff
@@ -120,7 +120,8 @@ pub async fn create_index(
 ) -> Result<HttpResponse, ResponseError> {
     let IndexCreateRequest { primary_key, uid } = body.into_inner();
 
-    let allow_index_creation = index_scheduler.filters().search_rules.is_index_authorized(&uid);
+    // FIXME: allow_index_creation?
+    let allow_index_creation = index_scheduler.filters().is_index_authorized(&uid);
```
irevoire, line 124: What is this FIXME?

meilisearch/src/routes/mod.rs
```diff
@@ -327,3 +327,5 @@ pub async fn get_health(
 
     Ok(HttpResponse::Ok().json(serde_json::json!({ "status": "available" })))
 }
+
+mod multi_search;
```
irevoire, line 331: Can you move this `mod` to the top of the file with the other ones

irevoire (changes requested):

meilisearch/src/routes/indexes/mod.rs
dureuill: I was surprised by the code here: we determine `allow_index_creation` from `is_index_authorized`. 

I would expect it be determined from `filters().allow_index_creation()`

dureuill: Note that although the PR changes the variable to use the `filters.is_index_authorized()` method, the original code would use `filters.search_rules.is_index_authorized()` which explains why it is surprising

index-scheduler/Cargo.toml
dureuill: I'm not @Kerollmops, but if I had to review this code, I guess my objection would be around *coupling*.

Adding the dependency is a symptom that the `index-scheduler` depends on the implementation of the `meilisearch-auth`, which might not be what we want.

The various approaches I can suggest:

1. Move the `AuthFilter` to `meilisearch-types`. I don't feel like this approach actually reduces the coupling between `meilisearch-auth` and the `index-scheduler`, IMO it just misplaces some of the `AuthFilter` implementation in the `meilisearch-types` crate.
2. Create an `AuthFilter` trait in `meilisearch-types`, implement it on `meilisearch-auth::AuthFilter`, then accept any type implementing the trait in the `index-scheduler`. This actually reduces the coupling (for instance it becomes possible to implement a `MockAuthFilter` in the `index-scheduler` for tests and we no longer need `AuthFilter::with_allowed_indexes`). It increases the complexity, though, and I don't really like adding a trait that will have a single production implementation.
3. Pass a single optional list of `allowed_indexes` like we did before. In the previous implementation, the list is however incorrect when using tenant token, which happened not to be the case, but it looks to me like a footgun to expose it as part of the `AuthFilter` interface. It is unclear to me that we could somehow merge the two lists of allowed indexes between the API key and the tenant token is a way that always works.
4. Pass a single `Option<(Vec<UuidPattern>, Vec<UuidPattern>)>` and have the the caller make sure to implemented the logic to extract the equivalent of `is_authorized_index` from this object. IMO this reduces coupling at the price of encapsulation, because the caller is expected to know how to correctly use a complicated type so as to keep its invariants, rather than the type being named and keeping its invariant on its own.
5. Wrap the type in (4) in a struct that exposes the `is_authorized_index` method. Now we have the same issue of where to put this struct, but for this struct instead of `AuthFilter`. Also, it is unclear how this struct is significantly different from `AuthFilter` itself.
6. Accept the coupling and move on by doing what the current PR does. It was my choice but I can revisit it.

The solutions above rely on the conceptual argument of unnecessary coupling, but maybe there are other arguments against this added dependency. Feel free to discuss I'm ready to implement most of the solutions above if needed.

meilisearch/src/routes/mod.rs
dureuill: done

index-scheduler/Cargo.toml
Kerollmops: That's maybe not a bad idea anymore. I think the reason was that we didn't want cross-dependencies. I am not sure about the drawbacks.

dureuill:
Changed error handling:

- Indicate the index of the query causing the failure inside of the `queries` array
- Return a 400 error for `index_not_found` rather than a 404, for this route only.

meilisearch/src/routes/indexes/mod.rs
dureuill: Update: following internal discussion, I updated the `AuthFilter::allow_index_creation` to take an indexUid and check both for access with this uid and creation permission.

I updated all the call sites and so removed the FIXME

irevoire: noice 👌

irevoire (dismissed):
I'm not super confident on the new auth stuff, but the rest looks fine to me.
You can either merge or try to get some feedback from someone else 👀

dureuill:
Thanks @irevoire !

I'm adding @ManyTheFish to the list of reviewers like we discussed in the team

ManyTheFish (dismissed):
Looks good to me!

Kerollmops (dismissed):
Looks good to me too! Just a "simple" rebase and we are good to merge ❤️

irevoire (approved):
Perfect, thank you @dureuill 🦝 

bors merge

Issue #3201, opened 2022-12-06
The `databaseSize` field of the `/stats` route is wrong


irevoire:
**Describe the bug**
Currently, the `databaseSize` actually represents the sum of the size of all indexes.
It ignores;
1. The size of the api-key database
2. The size of the task queue
3. The size of all the pending tasks (which can get pretty big, so this one is definitely an issue)

curquiza:
@irevoire, thanks for the report

How should we fix it? I mean, all of these 3 points should be taken into account for the fix?

Also, do you think we should fix this for v1?

irevoire:
Yep, we should take the three points into account, but it shouldn't take a lot of locs; thus, it should be done in one PR.

> Also, do you think we should fix this for v1?

Personally, I think it would be nice to fix it for v1, but since it's non-breaking and no one ever complained about it, I would say it's a P3 task.
Also, I think it's a good first issue, so if someone from the community wants to tackle it, I could definitely help him!!

dcm03:
It sounds like you are experiencing an issue with a database where the databaseSize value is not accurately reflecting the actual size of the database. Specifically, you mentioned that the databaseSize value is only considering the size of indexes and is not taking into account the size of the api-key database, the task queue, and pending tasks.

To fix this issue, you will need to modify the code that calculates the databaseSize value to include the sizes of the api-key database, task queue, and pending tasks. Depending on how the database is structured and how the databaseSize value is currently being calculated, this may involve adding additional code to iterate over the api-key database, task queue, and pending tasks and summing their sizes.

It's also worth considering why the databaseSize value is not accurately reflecting the actual size of the database. It's possible that there is a bug in the code that is causing certain elements of the database to be excluded from the calculation, or that the databaseSize value is not being updated properly when the database is modified. Identifying and fixing these issues may also be necessary to ensure that the databaseSize value is accurate.

Kerollmops:
Hey @dcm03,

Please don't use chatGPT or alternative bots/AIs to answer our issues. Answers are empty, don't bring any information or help, and pollute the issue. One day, we can use those AIs to help us fix code problems but not for now. As you have probably noticed, your answer is just writing tree times the content of the original issue content.

Have a happy Christmas 🎄 😃

dcm03:
Understood. Thanks for the response. Merry Christmas.On Dec 25, 2022 4:49 PM, Clément Renault ***@***.***> wrote:

Hey @dcm03,

[...]

PR #3418, opened 2023-01-24
Compute the size of the auth-controller, index-scheduler and all update files in the global stats

irevoire:
Fix https://github.com/meilisearch/meilisearch/issues/3201

### DIFF
--- a/file-store/src/lib.rs
+++ b/file-store/src/lib.rs
@@ -1,4 +1,3 @@
-use std::collections::BTreeSet;
 use std::fs::File as StdFile;
 use std::ops::{Deref, DerefMut};
 use std::path::{Path, PathBuf};
@@ -11,10 +10,14 @@ const UPDATE_FILES_PATH: &str = "updates/updates_files";
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
+    #[error("Could not parse file name as utf-8")]
+    CouldNotParseFileNameAsUtf8,
     #[error(transparent)]
     IoError(#[from] std::io::Error),
     #[error(transparent)]
     PersistError(#[from] tempfile::PersistError),
+    #[error(transparent)]
+    UuidError(#[from] uuid::Error),
 }
 
 pub type Result<T> = std::result::Result<T, Error>;
@@ -33,13 +36,11 @@ impl DerefMut for File {
     }
 }
 
-#[cfg_attr(test, faux::create)]
 #[derive(Clone, Debug)]
 pub struct FileStore {
     path: PathBuf,
 }
 
-#[cfg(not(test))]
 impl FileStore {
     pub fn new(path: impl AsRef<Path>) -> Result<FileStore> {
         let path = path.as_ref().to_path_buf();
@@ -48,7 +49,6 @@ impl FileStore {
     }
 }
 
-#[cfg_attr(test, faux::methods)]
 impl FileStore {
     /// Creates a new temporary update file.
     /// A call to `persist` is needed to persist the file in the database.
@@ -94,6 +94,14 @@ impl FileStore {
         Ok(())
     }
 
+    pub fn update_total_size(&self) -> Result<u64> {
+        let mut total = 0;
+        for uuid in self.all_uuids()? {
+            total += self.get_size(uuid?)?;
+        }
+        Ok(total)
+    }
+
     pub fn get_size(&self, uuid: Uuid) -> Result<u64> {
         Ok(self.get_update(uuid)?.metadata()?.len())
     }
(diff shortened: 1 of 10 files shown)


file-store/src/lib.rs
```diff
@@ -94,6 +94,14 @@ impl FileStore {
         Ok(())
     }
 
+    pub fn update_total_size(&self) -> Result<u64> {
```
dureuill, line 97: the naming of that function is a bit confusing to me.

I parsed it as "please update the total size please", initially, while it means "get the total size of the update files".

I know prefixing with `get` is unidiomatic in Rust, but in the interest of avoiding ambiguity, I wonder if we wouldn't want `get_update_total_size` here. Or even, given this is not a free operation and always recomputed from scratch, maybe use a verb to convey this information? `compute_update_total_size` or `process_update_total_size`?

file-store/src/lib.rs
```diff
@@ -94,6 +94,14 @@ impl FileStore {
...
+        let mut total = 0;
+        for uuid in self.all_uuids()? {
+            total += self.get_size(uuid?)?;
```
dureuill, line 100: I don't know enough context to answer this question by myself:

is there any risk that an update file gets deleted while we are counting the size, in normal operation? If so, is failing to get the size a sufficient reason to fail the entire sum, or should we rather unwrap_or_zero it?

index-scheduler/src/lib.rs
```diff
@@ -898,6 +902,11 @@ impl IndexScheduler {
         Ok(self.file_store.new_update_with_uuid(uuid)?)
     }
 
+    /// List the update files contained in the IndexScheduler.
+    pub fn update_file_size(&self) -> Result<u64> {
```
dureuill, line 906: Same issue as above wrt naming, although the doc comment helps

index-scheduler/src/lib.rs
```diff
@@ -898,6 +902,11 @@ impl IndexScheduler {
         Ok(self.file_store.new_update_with_uuid(uuid)?)
     }
 
+    /// List the update files contained in the IndexScheduler.
```
dureuill, line 905:
```suggestion
    /// The size on disk taken by all the updates files contained in the `IndexScheduler`, in bytes.
```

meilisearch-types/src/error.rs
```diff
@@ -343,6 +343,7 @@ impl ErrorCode for file_store::Error {
         match self {
             Self::IoError(e) => e.error_code(),
             Self::PersistError(e) => e.error_code(),
+            Self::CouldNotParseFileNameAsUtf8 | Self::UuidError(_) => Code::Internal,
```
dureuill, line 346: Does this mean that if the user adds a file with a different name in the update directory they will trigger an internal? Is that an issue?

dureuill (changes requested):
I have a few comments; great PR, I love we get to compute more accurate stats!!!

meilisearch-types/src/error.rs
irevoire: > Does this mean that if the user adds a file with a different name in the update directory they will trigger an internal? 

Yes, but tbh we’ve never seen anyone adding things to the `data.ms`, I’m quite confident on the fact that, if this error happens it’ll be because we effectively wrote something illegal in the first place 😬

file-store/src/lib.rs
irevoire: Yeah I like the compute, I’m going to update both function

file-store/src/lib.rs
irevoire: Yep good catch, it doesn’t look like a good reason to crash the whole stat computation.

file-store/src/lib.rs
```diff
@@ -11,10 +10,14 @@ const UPDATE_FILES_PATH: &str = "updates/updates_files";
...
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
+    #[error("Could not parse file name as utf-8")]
+    CouldNotParseFileNameAsUtf8,
     #[error(transparent)]
     IoError(#[from] std::io::Error),
     #[error(transparent)]
     PersistError(#[from] tempfile::PersistError),
+    #[error(transparent)]
+    UuidError(#[from] uuid::Error),
```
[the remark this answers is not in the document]
irevoire: We could but I feel like this would just make us lose a bit of precision for no real gain of simplicity 🤔

meilisearch-types/src/error.rs
dureuill: OK, let's just keep this fact in mind for now and resolve this

file-store/src/lib.rs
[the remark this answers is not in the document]
dureuill: OK, let's leave it at that for the sake of simplicity

dureuill (approved):
In the end you elected to keep `IndexScheduler::update_file_size`?

OK anyway

dureuill (approved):
Alright, it's now in! Thank you!

bors merge

dureuill:
bors merge

PR #3403, opened 2023-01-19
Add `--all` to test CI

curquiza:
Discussed with @irevoire [here](https://meilisearch.slack.com/archives/G01A1F4KVGU/p1674144546920649?thread_ts=1674144456.561199&cid=G01A1F4KVGU) (internal link)

PR #3422, opened 2023-01-25
Remove cache from test CI

curquiza:
Comment the lines in CIs where we use the test CIs
We indeed have cache issues (lack of space on the machine) when running our test CIs
https://github.com/meilisearch/meilisearch/pull/3403

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -53,8 +53,8 @@ jobs:
         os: [macos-12, windows-2022]
     steps:
     - uses: actions/checkout@v3
-    - name: Cache dependencies
-      uses: Swatinem/rust-cache@v2.2.0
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -31,8 +31,8 @@ jobs:
       with:
         toolchain: stable
         override: true
-    - name: Cache dependencies
-      uses: Swatinem/rust-cache@v2.2.0
+    # - name: Cache dependencies
```
dureuill, line 34: Maybe we could comment on why we're disabling the cache, to spare future readers from guessing

```suggestion
    # Disable cache due to disk space issues with Windows workers in CI
    # - name: Cache dependencies
```

dureuill:
bors merge

PR #3434, opened 2023-01-30
Make clippy happy for Rust 1.67, allow `uninlined_format_args`

dureuill:
# Pull Request

This PR allows `uninlined_format_args` in CI for clippy.

This is due to https://github.com/rust-lang/rust-clippy/issues/10087, which in particular has correctness issues wrt edition 2018 crates, and is a big change altogether. https://github.com/rust-lang/rust-clippy/pull/10265 is already open in order to change the category of this lint to "pedantic", meaning that if this latter PR merges, a future Rust release will accept our code unmodified wrt uninlined format arguments.

As a result, this PR introduces the following changes:

1. Allow `uninlined_format_args` in the clippy command in CI.
2. Use rewind rather than seek(0)
3. Remove lifetimes that clippy deems needless.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -109,7 +109,8 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets -- --deny warnings
+          # allow unlined_format_args https://github.com/rust-lang/rust-clippy/issues/10087
+          args: --all-targets -- --deny warnings --allow clippy::uninlined_format_args
 
   fmt:
     name: Run Rustfmt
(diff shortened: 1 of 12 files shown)


dureuill:
This PR is incorrect, because the `cargo clippy --fix` causes format arguments to be added inline, even for 2018 edition crates, which changes the behavior in an unwanted way. See https://github.com/rust-lang/rust-clippy/issues/10087#issuecomment-1408986074 for details.

curquiza:
I removed the `skip-changelog` because I don't want this kind of PR to be missed in the changelog if still present when doing the v1.1 changelog

Also can you open an issue in the `v1.1` Milestone so that we don't forget to re-introduce it?

dureuill:
Thanks @curquiza, but we changed our mind and only allow (in other words, disable) the problematic lint. As it is unclear if this lint will remain a warning by default, we don't necessarily have to do something in the future now. Other clippy warnings will remain emitted, and we're not too keen in enabling this lint, anyway.

Should I put back the skip changelog label?

Kerollmops (approved):
Again, thank you 💯
bors merge

PR #3439, opened 2023-01-31
Add git config about ownership in Docker CI

curquiza:
The docker CI si failing because of git usage: https://github.com/meilisearch/meilisearch/actions/runs/4053334082/jobs/6973827940

Capture d’écran 2023-01-31 à 12 12 44

> fatal: detected dubious ownership in repository at '/home/meili/actions-runner/_work/meilisearch/meilisearch'

I made some research and I found out this https://github.com/actions/runner-images/issues/6775

### DIFF
--- a/.github/workflows/publish-docker-images.yml
+++ b/.github/workflows/publish-docker-images.yml
@@ -52,6 +52,9 @@ jobs:
       - name: Set build-args for Docker buildx
         id: build-metadata
         run: |
+          # Define ownership
+          git config --global --add safe.directory /home/meili/actions-runner/_work/meilisearch/meilisearch
+
           # Extract commit date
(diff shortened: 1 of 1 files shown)


curquiza:
The fix seems to work: https://github.com/meilisearch/meilisearch/actions/runs/4053454799/jobs/6974088652

curquiza:
But I don't feel confident with this fix @Kerollmops 
Could it be a security issue from your point of view

Kerollmops (approved):
I am not exactly sure why this changed recently, but I would say that the only way to break meilisearch's security would be to change our CI? 🤔

curquiza:
> I am not exactly sure why this changed recently, but I would say that the only way to break meilisearch's security would be to change our CI? 🤔

From what I understand, I would say yes, only changing our CI could impact it...

curquiza:
So
bors merge

PR #3452, opened 2023-02-02
Add tests with rust stable and nightly in CI

### LINKED ISSUE #3402: Add tests with rust `nightly` to our CI
In the [`test-linux`](https://github.com/meilisearch/meilisearch/blob/b3166df7ea597beef798278e56d2a4b4c545c0da/.github/workflows/rust.yml#L18) job, we would like to add two tests
- one run with rust `stable`
- one with rust `nightly`

As we did in this CI:

https://github.com/meilisearch/milli/blob/3521a3a0b287ff788de875b53f3f35d04c7fa155/.github/workflows/rust.yml#L15-L49

Like in the previous CI, the test on `nightly` should only run once a day, like we did before:

https://github.com/meilisearch/milli/blob/3521a3a0b287ff788de875b53f3f35d04c7fa155/.github/workflows/rust.yml#L4-L5

The test with `stable` should keep running for every commit on pull request

hugoamalric:
# Pull Request

## Related issue
Fixes #3402

## What does this PR do?
In the [test-linux](https://github.com/meilisearch/meilisearch/blob/b3166df7ea597beef798278e56d2a4b4c545c0da/.github/workflows/rust.yml#L18) job two tests are added : 

- one run with rust stable
- one with rust nightly that runs once a day

## PR checklist

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -27,8 +27,19 @@ jobs:
       run: |
         apt-get update && apt-get install -y curl
         apt-get install build-essential -y
-    - uses: actions-rs/toolchain@v1
+    - name: Run test with Rust nightly
+      uses: actions-rs/toolchain@v1
       with:
+        profile: minimal
+        toolchain: nightly
+        override: true
+      on:
+          schedule:
+            - cron: '0 0 * * *' # Run daily at midnight (UTC)
+    - name: Run test with Rust stable
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
         toolchain: stable
         override: true
     # Disable cache due to disk space issues with Windows workers in CI


.github/workflows/rust.yml
```diff
@@ -27,8 +27,19 @@ jobs:
...
-    - uses: actions-rs/toolchain@v1
+    - name: Run test with Rust nightly
+      uses: actions-rs/toolchain@v1
       with:
+        profile: minimal
+        toolchain: nightly
+        override: true
+      on:
+          schedule:
+            - cron: '0 0 * * *' # Run daily at midnight (UTC)
```
curquiza, line 38: I did not know we could do this 😱
I don't find this in the GitHub docs, can you show me the link or proof this is working in one of your repo? 😇

curquiza:
bors try

.github/workflows/rust.yml
curquiza: I think your solution @hugoamalric does not work

https://github.com/meilisearch/meilisearch/actions/runs/4133694844

Capture d’écran 2023-02-09 à 12 28 38

curquiza (changes requested):
Thanks for the PR @hugoamalric but the current solution is not working I guess...

.github/workflows/rust.yml
hugoamalric: Yes my bad I thought it was possible to do so. 

The cron job is in fact scoped to the whole workflow. Is it annoying if the whole workflow is triggered every day at midnight? Otherwise, it is also possible to declare a condition for each job so that they are not affected by this cron job.

curquiza:
Closing in favor of https://github.com/meilisearch/meilisearch/pull/3495, since this current PR was not working
Thanks anyway for the PR @hugoamalric!

Issue #3426, opened 2023-01-26
Return stats for numerical facets like `min`, `max`


dureuill:
Related product discussion: https://github.com/meilisearch/product/discussions/117
Related specification: WIP
Also related to: https://github.com/meilisearch/instant-meilisearch/issues/926

### TODO

- [ ] For implementation, check with the product squad to get how it should be implemented
    - A prototype PR with only `min` and `max` is available: #3423. Do we need `avg` and `sum`?
- [ ] Merge the changes into main
- [ ] Update the specification

### Impacted teams

@meilisearch/integration-team: usable for https://github.com/meilisearch/instant-meilisearch/issues/926
@meilisearch/docs-team

dureuill:
Removing avg from the title as it won't make the first iteration planned for v1.

curquiza:
Hello everyone following this issue 👋 

We have just released the first RC (release candidate) of Meilisearch containing this new implementation!
You can test it by using
-  the [release assets](https://github.com/meilisearch/meilisearch/releases)
- the [Meilisearch Docker image](https://hub.docker.com/repository/docker/getmeili/meilisearch/general)
```bash
docker run -it --rm -p 7700:7700 -v $(pwd)/meili_data:/meili_data getmeili/meilisearch:v1.1.0-rc.0
```

You are more than welcome to communicate any feedback about this new implementation in [this discussion](https://github.com/meilisearch/product/discussions/117).
If you encountered any bugs, please report them [here](https://github.com/meilisearch/meilisearch/issues/new?assignees=&labels=&template=bug_report.md&title=).
Thanks in advance for your help and your involvement in Meilisearch ❤️ 

🎉 Official and stable release containing this change will be available on 3rd April 2023

⚠️ RC (release candidates) are not recommended for production

PR #3423, opened 2023-01-25
Add min and max facet stats

dureuill:
# Pull Request

## Related issue
Fixes #3426

### User standpoint

- When using a `facets` parameter in search, the facets that have numeric values are displayed in a new section of the response called `facetStats` that contains, per facet, the numeric min and max value of the hits returned by the search.

<details>
<summary>
Sample request/response
</summary>

```json
❯ curl \
  -X POST 'http://localhost:7700/indexes/meteorites/search?facets=mass' \
  -H 'Content-Type: application/json' \
  --data-binary '{ "q": "LL6", "facets":["mass", "recclass"], "limit": 5 }' | jsonxf
{
  "hits": [
    {
      "name": "Niger (LL6)",
      "id": "16975",
      "nametype": "Valid",
      "recclass": "LL6",
      "mass": 3.3,
      "fall": "Fell"
    },
    {
      "name": "Appley Bridge",
      "id": "2318",
      "nametype": "Valid",
      "recclass": "LL6",
      "mass": 15000,
      "fall": "Fell",
      "_geo": {
        "lat": 53.58333,
        "lng": -2.71667
      }
    },
    {
      "name": "Athens",
      "id": "4885",
      "nametype": "Valid",
      "recclass": "LL6",
      "mass": 265,
      "fall": "Fell",
      "_geo": {
        "lat": 34.75,
        "lng": -87.0
      }
    },
    {
      "name": "Bandong",
      "id": "4935",
      "nametype": "Valid",
      "recclass": "LL6",
      "mass": 11500,
      "fall": "Fell",
      "_geo": {
        "lat": -6.91667,
        "lng": 107.6
      }
    },
    {
      "name": "Benguerir",
      "id": "30443",
      "nametype": "Valid",
      "recclass": "LL6",
      "mass": 25000,
      "fall": "Fell",
      "_geo": {
        "lat": 32.25,
        "lng": -8.15
      }
    }
  ],
  "query": "LL6",
  "processingTimeMs": 15,
  "limit": 5,
  "offset": 0,
  "estimatedTotalHits": 42,
  "facetDistribution": {
    "mass": {
      "110000": 1,
      "11500": 1,
      "1161": 1,
      "12000": 1,
      "1215.5": 1,
      "127000": 1,
      "15000": 1,
      "1676": 1,
      "1700": 1,
      "1710.5": 1,
      "18000": 1,
      "19000": 1,
      "220000": 1,
      "2220": 1,
      "22300": 1,
      "25000": 2,
[...]

PR #3453, opened 2023-02-02
Force exhaustive search when faceting is requested

dureuill:
# Pull Request

## Related issue

Related to #3426.

## What does this PR do?

When not using pagination mode, the search is not exhaustive. This can result in incomplete results for facet distribution and wrong results for facet stats (as implemented in #3423).

This PR forces the search to be as exhaustive as possible[^1], even when not using pagination mode, as soon as a facet distribution (or facet stats) is required.

## Tests

Unfortunately, I was not able to come up with a test case that demonstrates how the search is not exhaustive by default. I'm still opening the PR should the issue pop up.

[^1]: My understanding is that we can still have differences in distribution/stats when distinct attributes are used in the query.

### DIFF
--- a/meilisearch/src/search.rs
+++ b/meilisearch/src/search.rs
@@ -144,14 +144,17 @@ pub fn perform_search(
     }
 
     let is_finite_pagination = query.is_finite_pagination();
+    let has_facet_distribution =
+        if let Some(facets) = query.facets.as_deref() { !facets.is_empty() } else { false };
+
+    search.exhaustive_number_hits(is_finite_pagination || has_facet_distribution);
+
     search.terms_matching_strategy(query.matching_strategy.into());
 
     let max_total_hits = index
         .pagination_max_total_hits(&rtxn)
         .map_err(milli::Error::from)?
         .unwrap_or(DEFAULT_PAGINATION_MAX_TOTAL_HITS);
-
-    search.exhaustive_number_hits(is_finite_pagination);
 
     // compute the offset on the limit depending on the pagination mode.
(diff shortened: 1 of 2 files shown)


ManyTheFish:
Hello @dureuill,

thank you for the PR, I'd personally wait for proof or an issue showing that the facet distribution is not sufficiently exhaustive to merge this PR, I'm a bit concerned by the performance impact. 😕

Did you see any significant difference in the search performances?

dureuill:
> Did you see any significant difference in the search performances?

Testing on my dataset was insufficient to show anything one way or the other. We'd need tests or some more realistic use cases.

> thank you for the PR, I'd personally wait for proof or an issue showing that the facet distribution is not sufficiently exhaustive to merge this PR, I'm a bit concerned by the performance impact. 😕


Sure! Let's keep this PR as draft for now

gmourier:
@dureuill We could test it on the dataset @Strift will use to make an e-commerce demo

Strift:
Hey there. Here is the link to the aforementioned [ecommerce dataset](https://github.com/meilisearch/demos/blob/main/src/ecommerce/seed/data.json).

To be honest, I haven't taken much time to challenge the relevance of the data. So if you see anything weird, you can let me know!

Cheers,

Kerollmops:
I am also very concerned about the performance impact it can have. I want to avoid falling into [this kind of performance issue](https://github.com/typesense/typesense/issues/914). Having estimated counts is enough for some of our users that search performance over the accuracy, but it can be an issue for other users.

dureuill:
closing old PR that's likely unmergeable as is

PR #3463, opened 2023-02-06
CI: attempt to use apt-mirror-updater

dureuill:
Ubuntu mirrors are not responding.

This PR attempts to fix that for the time being, just in case situation doesn't return to normal shortly.


dureuill:
No `pip3` in CI, unless we do an `apt install pip3` first, which defeats the purpose 🫠

PR #3478, opened 2023-02-09
Add `custom-static-files` option

joepio:
Just a draft proposal to toy around with. Probably not a good idea, but it's useful in my usecase: wanting one less component in my setup for a customer.

- Adds the `--custom-static-files` CLI option that allows users to use their own static files, instead of the dashboard. Makes it easier to host a custom front-end.
- Adds the `actix_files` crate as dependency. The impact of this is pretty low, as by far [most](https://github.com/meilisearch/meilisearch/pull/3478/files#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87e) of the sub-dependencies are already used.

See https://github.com/meilisearch/product/discussions/610

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,6 +35,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "actix-files"
+version = "0.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d832782fac6ca7369a70c9ee9a20554623c5e51c76e190ad151780ebea1cf689"
+dependencies = [
+ "actix-http",
+ "actix-service",
+ "actix-utils",
+ "actix-web",
+ "askama_escape",
+ "bitflags",
+ "bytes",
+ "derive_more",
+ "futures-core",
+ "http-range",
+ "log",
+ "mime",
+ "mime_guess",
+ "percent-encoding",
+ "pin-project-lite",
+]
+
+[[package]]
 name = "actix-http"
 version = "3.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


irevoire:
Hey, I like the idea.
That means, if you have a static website, it's super easy to host it with meilisearch with https and without a reverse proxy.

Not sure it's something we want to merge product-wise (pretty sure it's not going to land for the v1.1.0).

What do you think of it, @gmourier?
Personally, I don't think it's too risky to merge it as an experimental feature to see if people use it for now before making a real decision.

joepio:
> Hey, I like the idea. That means, if you have a static website, it's super easy to host it with meilisearch with https and without a reverse proxy.

Mostly yes, but if we want to get rid of the reverse proxy and still have HTTPS with plain meilisearch, we'll also need some form of acme / TLS support.  (May be off-topic for this PR, but still good to share some thoughts as this too relates to the usecase of using Meili as a one-stop-shop for hosting a static site with powerful search.) I added automated acme / TLS setup in atomic-server using [`acme_lib`](https://github.com/algesten/acme-lib/) and later using [`instant-acme`](https://github.com/InstantDomain/instant-acme) (which uses Rustls instead of OpenSSL, removing an important runtime dependency). Feel free to [use my implementation](](https://github.com/atomicdata-dev/atomic-data-rust/blob/master/server/src/https.rs)), which also is built on actix-web.

curquiza:
Thanks again for the work @joepio 
Since the PR is a little bite old and it's not a priority for the team now, I close the PR
However, thanks again for your work, it's definitely valuable and will help us if decide to go with it. Let's see if the [discussion](https://github.com/meilisearch/product/discussions/609) on the product side has more attractiveness in the future.

Issue #3476, opened 2023-02-09
Uffizzi should not rebuild the Meilisearch image from scratch


curquiza:
As reported here https://github.com/meilisearch/meilisearch/pull/3454#issuecomment-1414141115, Uffizzi should not re-build from scratch the Meilisearch image
In the PR, you can see the small change done between each commit, but the image build is still btw 28 and 23 minutes.

Discussed with @waveywaves who is working on it 😊

curquiza:
Of course! Thank you very much @waveywaves

PR #3482, opened 2023-02-10
Optimize meilisearch uffizzi build

waveywaves:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/3476

## What does this PR do?
even though docker cache was being used earlier for uffizzi builds, seems like the cache layers weren't persisting. This commit adds changes to move meilisearch building outside the dockerfile so that we can use the rust cache action. We are also building to the musl target so that the binary for meilisearch which is created can be used for the uffizzi ttyd image which uses alpine.

Meilisearch build time brought to 5 mins example https://github.com/waveywaves/meilisearch/actions/runs/4142776058

we also update the version of uffizzi action used here which fixes another uffizzi bug where the environments are not deployed. https://app.uffizzi.com/github.com/waveywaves/meilisearch/pull/2 was built as a part of a test for this PR and we can be sure that the deployment works well now.

## PR checklist

### DIFF
--- a/.github/workflows/uffizzi-preview-deploy.yml
+++ b/.github/workflows/uffizzi-preview-deploy.yml
@@ -82,7 +82,7 @@ jobs:
     name: Use Remote Workflow to Preview on Uffizzi
     needs:
       - cache-compose-file
-    uses: UffizziCloud/preview-action/.github/workflows/reusable.yaml@desc
+    uses: UffizziCloud/preview-action/.github/workflows/reusable.yaml@v2
     with:
       # If this workflow was triggered by a PR close event, cache-key will be an empty string
       # and this reusable workflow will delete the preview deployment.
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -43,4 +43,4 @@ WORKDIR /meili_data
 EXPOSE  7700/tcp
 
 ENTRYPOINT ["tini", "--"]
-CMD     /bin/meilisearch
+CMD     /bin/meilisearch
```
curquiza, line 46: can you put back the new line at the end of the file? We want to unify our yml files 😇

.github/workflows/uffizzi-preview-deploy.yml
```diff
@@ -82,21 +82,21 @@ jobs:
...
-        the preview from the link Meilisearch Endpoint link given below.
+      # description: |
+      #   The meilisearch preview environment contains a web terminal from where you can run the
+      #   `meilisearch` command. You should be able to access this instance of meilisearch running in
+      #   the preview from the link Meilisearch Endpoint link given below.
          
-        Web Terminal Endpoint : ${{ needs.cache-compose-file.outputs.expected-url }}
-        Meilisearch Endpoint : ${{ needs.cache-compose-file.outputs.expected-url }}/meilisearch
+      #   Web Terminal Endpoint : ${{ needs.cache-compose-file.outputs.expected-url }}
+      #   Meilisearch Endpoint : ${{ needs.cache-compose-file.outputs.expected-url }}/meilisearch
```
curquiza, line 99: Why do you comment this?
Also looks like it does not work currently

Capture d’écran 2023-02-15 à 15 40 00

curquiza (changes requested):
Thanks @waveywaves for the PR 😇

curquiza (approved):
Thanks a lot @waveywaves! 😊 

bors merge

PR #3490, opened 2023-02-13
Fix attributes set candidates

### LINKED ISSUE #3483: Pagination did not return result despite metadata saying so
**Describe the bug**
I imported 150K records and when I search for certain keyword, the totalPages indicate 733, but when I am on page=201, no hits returned:

```ruby
o = MeiliSearch::Rails.client.index("Talent_CandidateProfile").search(query, page: 201, hits_per_page: 15)
{"hits"=>[], "query"=>"cto system", "processingTimeMs"=>1279, "hitsPerPage"=>15, "page"=>201, "totalPages"=>733, "totalHits"=>10989}
```

7ais5_21_16_39

ugxgs_21_17_43

**Expected behavior**
Hits should have array of results.

**Meilisearch version:** v0.30.5

**Additional context**
Using Rails 7, macOS

ManyTheFish:
# Pull Request

Fix attributes set candidates for v1.1.0

## details

The attribute criterion was not returning the remaining candidates when its internal algorithm was been exhausted.
We had a loss of candidates by the attribute criterion leading to the bug reported in the issue linked below.
After some investigation, it seems that it was the only criterion that had this behavior.

We are now returning the remaining candidates instead of an empty bitmap.

## Related issue

Fixes #3483
PR on milli for v1.0.1: https://github.com/meilisearch/milli/pull/777

### DIFF
--- a/milli/src/search/criteria/attribute.rs
+++ b/milli/src/search/criteria/attribute.rs
@@ -123,7 +123,7 @@ impl<'t> Criterion for Attribute<'t> {
                             None => {
                                 return Ok(Some(CriterionResult {
                                     query_tree: Some(query_tree),
-                                    candidates: Some(RoaringBitmap::new()),
+                                    candidates: Some(allowed_candidates),
                                     filtered_candidates: None,
                                     initial_candidates: Some(self.initial_candidates.take()),
                                 }));


curquiza:
I changed the milestones to v1.0.1 to avoid duplicate the information in the v1.0.1 changelog AND in the v1.1.0 changelogs.
This PR (if I'm correct) is about importing the milli changes into Meilisearch

ManyTheFish:
@curquiza yes you're right, I completely understand your thoughts.

curquiza:
bors merge

PR #3499, opened 2023-02-15
Use the workspace inheritance

irevoire:
Use the workspace inheritance [introduced in rust 1.64](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds).

It allows us to define the version of meilisearch once in the main `Cargo.toml` and let all the other `Cargo.toml` uses this version.

@curquiza I added you as a reviewer because I had to patch some CI scripts

And @Kerollmops, I had to bump the `cargo_toml` crates because our version was getting old and didn't support the feature yet.

Also, in another PR, I would like to unify some of our dependencies to ensure we always stay in sync between all our crates.

### DIFF
--- a/.github/scripts/check-release.sh
+++ b/.github/scripts/check-release.sh
@@ -3,7 +3,7 @@
 # check_tag $current_tag $file_tag $file_name
 function check_tag {
   if [[ "$1" != "$2" ]]; then
-      echo "Error: the current tag does not match the version in $3: found $2 - expected $1"
+      echo "Error: the current tag does not match the version in Cargo.toml: found $2 - expected $1"
       ret=1
   fi
 }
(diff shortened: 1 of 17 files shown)


curquiza (dismissed):
The future looks so pretty 🤩 

Thanks Tamo!

Can you also update this part of the CI?

https://github.com/meilisearch/meilisearch/blob/91ce8a5e67553803ea74c26013659acad1a6e274/.github/workflows/update-cargo-toml-version.yml#L26-L32

We don't need `sd` anymore, the already present `sed` should work since only 1 file needs to the be changed

Kerollmops (changes requested):
Thank you very much @irevoire!

Could you also change the [`default-run = "meilisearch"` rule in the workspace TOML](https://doc.rust-lang.org/cargo/reference/manifest.html#the-default-run-field), please? 😊

irevoire:
Hey @Kerollmops,

> Could you also change the [default-run = "meilisearch" rule in the workspace TOML](https://doc.rust-lang.org/cargo/reference/manifest.html#the-default-run-field), please? 😊

I wanted to do that very hard, but default-run is not supported at the workspace level 😔 
The other option was to use `default-members`, which is supported.
But that applies to _every command_, which means when we run the test, by default, it'll only run the tests of the meilisearch crate...

Kerollmops (approved):
Thank you very much Tamo 🤜🍎🤛 
bors merge

Issue #3442, opened 2023-01-31
Configure the CSV delimiter


curquiza:
Suggested by https://github.com/meilisearch/meilisearch/pull/2803

Meilisearch will let the possibility of customizing the CSV delimiter using the `csvDelimiter` parameter when adding or updating documents (PUT and POST routes)

Example when asking for `;` as CSV delimiter:

```bash
POST localhost:7700/indexes/<uid>/documents?csvDelimiter=%3B
```

## TODO

- [ ] Implement changes based on https://github.com/meilisearch/meilisearch/pull/2803. A new PR must be created by the @meilisearch/engine-team, since we want to avoid asking for changes to the users due to git conflicts and the obsolete PR. However, if possible keep the commits of the user.
- [ ] Update the spec

## Impacted teams

@meilisearch/docs-team @meilisearch/integration-team

curquiza:
Hello everyone following this issue 👋 

We have just released the first RC (release candidate) of Meilisearch containing this new implementation!
You can test it by using
-  the [release assets](https://github.com/meilisearch/meilisearch/releases)
- the [Meilisearch Docker image](https://hub.docker.com/repository/docker/getmeili/meilisearch/general)
```bash
docker run -it --rm -p 7700:7700 -v $(pwd)/meili_data:/meili_data getmeili/meilisearch:v1.1.0-rc.0
```

You are more than welcome to communicate any feedback about this new implementation here in this current issue.
If you encountered any bugs, please report them [here](https://github.com/meilisearch/meilisearch/issues/new?assignees=&labels=&template=bug_report.md&title=).
Thanks in advance for your help and your involvement in Meilisearch ❤️ 

🎉 Official and stable release containing this change will be available on 3rd April 2023

⚠️ RC (release candidates) are not recommended for productionhttps://github.com/meilisearch/product/discussions/491

PR #3505, opened 2023-02-16
Csv delimiter

irevoire:
Fixes https://github.com/meilisearch/meilisearch/issues/3442
Closes https://github.com/meilisearch/meilisearch/pull/2803
Specified in https://github.com/meilisearch/specifications/pull/221

This PR is a reimplementation of https://github.com/meilisearch/meilisearch/pull/2803, on the new engine. Thanks for your idea and initial PR @MixusMinimax; sorry I couldn’t update/merge your PR. Way too many changes happened on the engine in the meantime.

**Attention to reviewer**; I had to update deserr to implement the support of deserializing `char`s

-------

It introduces four new error messages;
- Invalid value in parameter csvDelimiter: expected a string of one character, but found an empty string
- Invalid value in parameter csvDelimiter: expected a string of one character, but found the following string of 5 characters: doggo
- csv delimiter must be an ascii character. Found: 🍰 
- The Content-Type application/json does not support the use of a csv delimiter. The csv delimiter can only be used with the Content-Type text/csv.

And one error code;
- `invalid_index_csv_delimiter`

The `invalid_content_type` error code is now also used when we encounter the `csvDelimiter` query parameter with a non-csv content type.

### DIFF
--- a/meilisearch-types/src/document_formats.rs
+++ b/meilisearch-types/src/document_formats.rs
@@ -19,7 +19,7 @@ type Result<T> = std::result::Result<T, DocumentFormatError>;
 pub enum PayloadType {
     Ndjson,
     Json,
-    Csv(u8),
+    Csv { delimiter: u8 },
 }
 
 impl fmt::Display for PayloadType {
@@ -27,7 +27,7 @@ impl fmt::Display for PayloadType {
         match self {
             PayloadType::Ndjson => f.write_str("ndjson"),
             PayloadType::Json => f.write_str("json"),
-            PayloadType::Csv(_) => f.write_str("csv"),
+            PayloadType::Csv { .. } => f.write_str("csv"),
         }
     }
 }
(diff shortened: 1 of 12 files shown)


MixusMinimax:
Thank you for implementing this, looks good! I think this will be very useful, as the csv standard is all over the place...
I'm not sure if even more configuration makes sense, there is a lot in csv:
https://docs.rs/csv/latest/csv/tutorial/index.html#delimiters-quotes-and-variable-length-records
But I think the delimiter is pretty much the only thing that is very frequently different.

irevoire:
> Thank you very much @irevoire. You could probably [co-sign the main commit](https://docs.github.com/fr/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) with @MixusMinimax's email 😊

Good idea! Will do 😁 

>  I'm not sure if even more configuration makes sense, there is a lot in csv:

Bad idea, please everyone, stop using custom-everything for a standard that is well defined and doesn't use any of these shenanigans 😂

MixusMinimax:
> Bad idea, please everyone, stop using custom-everything for a standard that is well defined and doesn't use any of these shenanigans

Hi, if I created my own data, I would totally agree, but in the real world you often have to deal with data sent to you by your customers or public databases. We're talking about csv datasets that are thousands of lines, and no, csv is not a well defined standard 😅

meilisearch-types/src/document_formats.rs
```diff
@@ -19,15 +19,15 @@ type Result<T> = std::result::Result<T, DocumentFormatError>;
 pub enum PayloadType {
     Ndjson,
     Json,
-    Csv,
+    Csv(u8),
```
Kerollmops, line 22: Can we change this to be a struct with a named field i.e. `delimiter`, please? That would be better.

Kerollmops (changes requested):
Thank you very much for the tests too!
Can you change the small change request I asked for, please?

Kerollmops (approved):
Thank you very much! I like what you've done here!
I approve that!

irevoire:
Thanks,
bors merge

PR #3515, opened 2023-02-20
Consider null as a valid geo object

### LINKED ISSUE #3497: Handle `"_geo": null` instead of throwing an error
Related discussion: https://github.com/meilisearch/product/discussions/586#discussioncomment-4969629

When indexing documents with the geosearch enabled, it’s common to send a field set to `null` if it doesn’t exist. (especially in highly typed languages)
Currently, in v1.0, we throw an error if this happens, but it would be nice to accept the payload and register the document as _, not_ geo faceted.

irevoire:
Fix #3497


irevoire:
bors merge

PR #3516, opened 2023-02-20
Use Debian instead of Alpine in Dockerfile

### LINKED ISSUE #3504: Use ubuntu instead of alpine in Docker image
Discussed with @dureuill 

Currently, in our [Dockerfile](https://github.com/meilisearch/meilisearch/blob/main/Dockerfile), we use the alpine image as a base.

We spotted multi potential issues we could have in the future, and here is why we would rather use the ubuntu/debian image:
- Meilisearch is a program doing a lot of system calls and then highly depends on the platform where it runs: we don't run any integration tests with Alpine, and even if it's a Linux distribution, we definitely better know the Ubuntu and Debian usages than Alpine
- a lot of users use our docker image, especially @meilisearch/cloud-team. Perfectly knowing the edge cases of the docker use case is more and more mandatory. Engine team would be more confident with a debian or ubuntu image, since it's what we test every day.
- Alpine uses a musl glic binary, and of course, Meilisearch uses glic a lot -> this musl version of the glibc can bring latency (not verified yet) but also edge cases we cannot identify now.

For all these reasons, we would rather use an ubuntu or debian image to build our own Meilisearch image at top of it

## TODO

- [ ] Test with Debian or Ubuntu instead of Alpine
  - [ ] Is the size of the image still correct?
  - [ ] Give a prototype to the cloud team, and wait for their approval -> is it working as previously?
- [ ] Implement the changes

## Impacted team

@meilisearch/cloud-team

curquiza:
Fixes #3504

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,10 @@ FROM    debian:11.6
 ENV     MEILI_HTTP_ADDR 0.0.0.0:7700
 ENV     MEILI_SERVER_PROVIDER docker
 
-RUN     apt update -q \
-        && apt install -q -y tini
+RUN     set -ex; \
+        apt-get update -q; \
+        apt-get install -q -y --no-install-recommends tini; \
+        rm -rf /var/lib/apt/lists/*
 
 # add meilisearch to the `/bin` so you can run it from anywhere and it's easy
 # to find.
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
 # Compile
-FROM    rust:alpine3.16 AS compiler
-
-RUN     apk add -q --update-cache --no-cache build-base openssl-dev
+FROM    rust:bullseye AS compiler
 
 WORKDIR /meilisearch
 
@@ -13,20 +11,20 @@ ENV     RUSTFLAGS="-C target-feature=-crt-static"
 
 COPY    . .
 RUN     set -eux; \
-        apkArch="$(apk --print-arch)"; \
-        if [ "$apkArch" = "aarch64" ]; then \
+        arch="$(dpkg --print-architecture)"; \
+        if [ "$arch" = "aarch64" ]; then \
             export JEMALLOC_SYS_WITH_LG_PAGE=16; \
         fi && \
         cargo build --release
 
 # Run
-FROM    alpine:3.16
+FROM    debian:11.6
 
 ENV     MEILI_HTTP_ADDR 0.0.0.0:7700
 ENV     MEILI_SERVER_PROVIDER docker
 
-RUN     apk update --quiet \
-        && apk add -q --no-cache libgcc tini curl
+RUN     apt update -q \
+        && apt install -q -y tini
 
 # add meilisearch to the `/bin` so you can run it from anywhere and it's easy
 # to find.
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ ENV     RUSTFLAGS="-C target-feature=-crt-static"
 COPY    . .
 RUN     set -eux; \
         arch="$(dpkg --print-architecture)"; \
-        if [ "$arch" = "aarch64" ]; then \
+        if [ "$arch" = "arm64" ]; then \
             export JEMALLOC_SYS_WITH_LG_PAGE=16; \
         fi && \
         cargo build --release


Dockerfile
```diff
@@ -1,7 +1,5 @@
 # Compile
-FROM    rust:alpine3.16 AS compiler
-
-RUN     apk add -q --update-cache --no-cache build-base openssl-dev
+FROM    rust:bullseye AS compiler
```
curquiza, line 2: Unfortunately I cannot select a docker image from Rust with a pinned version of Debian

Let me know if I'm wrong: https://hub.docker.com/_/rust

Dockerfile
```diff
@@ -13,20 +11,20 @@ ENV     RUSTFLAGS="-C target-feature=-crt-static"
...
-        if [ "$apkArch" = "aarch64" ]; then \
+        arch="$(dpkg --print-architecture)"; \
+        if [ "$arch" = "aarch64" ]; then \
             export JEMALLOC_SYS_WITH_LG_PAGE=16; \
         fi && \
         cargo build --release
 
 # Run
-FROM    alpine:3.16
+FROM    debian:11.6
```
curquiza, line 21: Should we use the same version of debian as the compiler, even if it's not a pinned version?

Dockerfile
```diff
@@ -13,20 +11,20 @@ ENV     RUSTFLAGS="-C target-feature=-crt-static"
...
 
 ENV     MEILI_HTTP_ADDR 0.0.0.0:7700
 ENV     MEILI_SERVER_PROVIDER docker
 
-RUN     apk update --quiet \
-        && apk add -q --no-cache libgcc tini curl
+RUN     apt update -q \
+        && apt install -q -y tini
```
mmachatschek, line 27:
```suggestion
RUN     set -ex; \
        apt-get update -q; \
        apt-get install -q -y --no-install-recommends tini; \
        rm -rf /var/lib/apt/lists/*
```

Dockerfile
```diff
@@ -13,20 +11,20 @@ ENV     RUSTFLAGS="-C target-feature=-crt-static"
 
 COPY    . .
 RUN     set -eux; \
-        apkArch="$(apk --print-arch)"; \
-        if [ "$apkArch" = "aarch64" ]; then \
+        arch="$(dpkg --print-architecture)"; \
+        if [ "$arch" = "aarch64" ]; then \
```
mmachatschek, line 15:
```suggestion
        if [ "$arch" = "arm64" ]; then \
```

Dockerfile
mmachatschek, line 21:
```suggestion
FROM    debian:11.6-slim
```

mmachatschek (changes requested):

Dockerfile
mmachatschek: the `&&` vs `;` is actually preference but I thought why not also use the `;` syntax here as its used in the build step as well

Dockerfile
mmachatschek: this will trim the image by at least 20MB

Dockerfile
curquiza: Thanks @mmachatschek 

What is the purpose of the last line? `rm -rf /var/lib/apt/lists/*`

mmachatschek: The directory is filled with information about packages and versions. In docker builds this is removed to reduce the size of the final image.

(This can be looked up in other docker files that build Debian/Ubuntu images.)

Dockerfile
```diff
@@ -13,20 +11,22 @@ ENV     RUSTFLAGS="-C target-feature=-crt-static"
...
 
 ENV     MEILI_HTTP_ADDR 0.0.0.0:7700
 ENV     MEILI_SERVER_PROVIDER docker
 
-RUN     apk update --quiet \
-        && apk add -q --no-cache libgcc tini curl
+RUN     set -ex; \
+        apt-get update -q; \
+        apt-get install -q -y --no-install-recommends tini; \
+        rm -rf /var/lib/apt/lists/*
```
sanders41, line 29: I haven't run a build here to check actual numbers, but adding an `apt-get clean` in addition to the `rm -rf /var/lib/apt/lists/*` should reduce the image size some more.

```suggestion
RUN     set -ex; \
        apt-get update -q; \
        apt-get install -q -y --no-install-recommends tini; \
        apt-get clean; \
        rm -rf /var/lib/apt/lists/*
```

curquiza:
Old, closing

PR #3524, opened 2023-02-21
Update the metrics route

### LINKED ISSUE #3523: Experimental feature: Metrics
All details are in: https://github.com/meilisearch/meilisearch/discussions/3518
Spec: https://github.com/meilisearch/specifications/pull/174

Get rid of the feature flags we had and provide the metrics route as a cli flag shipped by default in every executable.

## For the @meilisearch/docs-team 
I named the cli flag `--experimental-enable-metrics`.
That can change, obviously. Let me know if you would like to update it. The only constraint is that it **must** starts with `experimental`.

I also wrote a little description for the `--help`:
```
      --experimental-enable-metrics
          Experimental metrics feature. For more information, see: https://github.com/meilisearch/meilisearch/discussions/3518

          Enables the Prometheus metrics on the `GET /metrics` endpoint.

          [env: MEILI_EXPERIMENTAL_ENABLE_METRICS=]
```

Also, take care of documenting this sentence well:
> Provides access to `GET` `/metrics` route. **A restriction on `indexes` does not affect returned metrics.**

It means that someone who should only have access to one index will be able to see the metrics of **all** indexes, including the one he shouldn't know the existence of.

## Impacted teams

@meilisearch/integration-team @meilisearch/docs-team

irevoire:
Fixes #3523

Make the metrics available by default without a feature flag.
+ Rename the cli-flag to `experimental-enable-metrics`.

### DIFF
--- a/meilisearch/src/option.rs
+++ b/meilisearch/src/option.rs
@@ -47,7 +47,6 @@ const MEILI_IGNORE_MISSING_DUMP: &str = "MEILI_IGNORE_MISSING_DUMP";
 const MEILI_IGNORE_DUMP_IF_DB_EXISTS: &str = "MEILI_IGNORE_DUMP_IF_DB_EXISTS";
 const MEILI_DUMP_DIR: &str = "MEILI_DUMP_DIR";
 const MEILI_LOG_LEVEL: &str = "MEILI_LOG_LEVEL";
-#[cfg(feature = "metrics")]
 const MEILI_ENABLE_METRICS_ROUTE: &str = "MEILI_ENABLE_METRICS_ROUTE";
 
 const DEFAULT_CONFIG_FILE_PATH: &str = "./config.toml";
@@ -288,7 +287,6 @@ pub struct Opt {
     pub log_level: LogLevel,
 
     /// Enables Prometheus metrics and /metrics route.
-    #[cfg(feature = "metrics")]
     #[clap(long, env = MEILI_ENABLE_METRICS_ROUTE)]
     #[serde(default)]
     pub enable_metrics_route: bool,
@@ -384,7 +382,6 @@ impl Opt {
             config_file_path: _,
             #[cfg(all(not(debug_assertions), feature = "analytics"))]
             no_analytics,
-            #[cfg(feature = "metrics")]
             enable_metrics_route,
         } = self;
         export_to_env_if_not_present(MEILI_DB_PATH, db_path);
@@ -423,13 +420,7 @@ impl Opt {
 
         export_to_env_if_not_present(MEILI_DUMP_DIR, dump_dir);
         export_to_env_if_not_present(MEILI_LOG_LEVEL, log_level.to_string());
-        #[cfg(feature = "metrics")]
-        {
-            export_to_env_if_not_present(
-                MEILI_ENABLE_METRICS_ROUTE,
-                enable_metrics_route.to_string(),
-            );
-        }
+        export_to_env_if_not_present(MEILI_ENABLE_METRICS_ROUTE, enable_metrics_route.to_string());
         indexer_options.export_to_env();
     }
 
--- a/meilisearch/src/option.rs
+++ b/meilisearch/src/option.rs
@@ -286,7 +286,7 @@ pub struct Opt {
     #[serde(default)]
     pub log_level: LogLevel,
 
-    /// Experimental metrics feature. For more information, see: https://github.com/meilisearch/meilisearch/discussions/3518
+    /// Experimental metrics feature. For more information, see: <https://github.com/meilisearch/meilisearch/discussions/3518>
     ///
     /// Enables the Prometheus metrics on the `GET /metrics` endpoint.
     #[clap(long, env = MEILI_EXPERIMENTAL_ENABLE_METRICS)]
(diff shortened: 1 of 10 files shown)


meilisearch/src/option.rs
```diff
@@ -287,11 +286,12 @@ pub struct Opt {
     #[serde(default)]
     pub log_level: LogLevel,
 
-    /// Enables Prometheus metrics and /metrics route.
-    #[cfg(feature = "metrics")]
-    #[clap(long, env = MEILI_ENABLE_METRICS_ROUTE)]
+    /// Experimental metrics feature. For more information, see: https://github.com/meilisearch/meilisearch/discussions/3518
```
dureuill, line 289:
```suggestion
    /// Experimental metrics feature. For more information, see: <https://github.com/meilisearch/meilisearch/discussions/3518>
```

(rustdocs comments are markdown. Stray URLs in markdown need to be enclosed in `<` and `>`)

meilisearch/src/option.rs
```diff
@@ -287,11 +286,12 @@ pub struct Opt {
...
+    ///
+    /// Enables the Prometheus metrics on the `GET /metrics` endpoint.
+    #[clap(long, env = MEILI_EXPERIMENTAL_ENABLE_METRICS)]
     #[serde(default)]
-    pub enable_metrics_route: bool,
+    pub experimental_enable_metrics: bool,
```
dureuill, line 294: Should we also change the default config file?

dureuill (approved):
A few questions (see inline comments) otherwise looks nice, thank you :-)

meilisearch/src/option.rs
irevoire: Good idea. I added it at the very end in an `experimental features` category

irevoire:
Sorry, @dureuill I had to rebase on main after https://github.com/meilisearch/meilisearch/pull/3417 was merged.
But nothing changed from your last review, only a bunch of imports.
And I added the feature at the end of the config file 👍

dureuill (dismissed):
Looking good, thanks!

irevoire:
bors merge

meilisearch/src/routes/metrics.rs
```diff
@@ -0,0 +1,45 @@
...
+
+pub async fn get_metrics(
+    index_scheduler: GuardedData<ActionPolicy<{ actions::METRICS_GET }>, Data<IndexScheduler>>,
+    auth_controller: GuardedData<ActionPolicy<{ actions::METRICS_GET }>, AuthController>,
+) -> Result<HttpResponse, ResponseError> {
+    let response = create_all_stats(
+        (*index_scheduler).clone(),
+        (*auth_controller).clone(),
+        // we don't use the filters contained in the `ActionPolicy` because the metrics must have the right to access all the indexes.
+        &AuthFilter::default(),
```
dureuill, line 26: 1. Thanks for the docs comment 🙂 
2. I think this has the potential to be surprising to users. I think that in no circumstances should the existence of an index not authorized by an API key be revealed. 
3. If we're going to go this route, maybe we could use `AuthFilter::with_allowed_indexes` instead to at least not have the `allow_index_creation` boolean set to true?

```suggestion
        &AuthFilter::with_allowed_indexes(std::iter::once("*").collect()),
```

irevoire: > I think that in no circumstances should the existence of an index not authorized by an API key be revealed.

Oh, but it'll be revealed whatever happens since there is a middleware telling you which routes were called and how many times 😬 
The only information you'll lose from what I see is the size of the non-authorized indexes.

-----

I think this discussion could be interesting for you @gmourier.
FYI, currently we wrote [Provides access to `GET` `/metrics` route. **A restriction on `indexes` does not affect returned metrics.**](https://github.com/meilisearch/specifications/pull/174/files#diff-efbeaa05fe855150edfa47dc218874a3d03dda4000c79375238288abd551694aR331) but that can change.

dureuill: > Oh, but it'll be revealed whatever happens since there is a middleware telling you which routes were called and how many times 😬

I see.

A maybe more specific issue I can see is that if an API key is restricted on indexes but not actions, one can use the metrics route to get information on which indexes are in existence.

dureuill: Another possibility I'm thinking of is to reject access to the metrics route with an error if the corresponding API key doesn't have `filters().all_indexes_authorized()`, so that a user must explicitly have all indexes authorized + the metrics action to access the metrics route.

irevoire: Ah, yes, that could be a nice solution indeed!

irevoire:
bors merge

meilisearch/src/routes/metrics.rs
gmourier: Since it's experimental, could we keep it as it is and note it for a future iteration? (I've planned to create a dedicated production discussion for the whole Experimental /metrics)

I sense `/metrics` will mostly be consumed with an API Key having access to all indexes initially because the user first use case is to monitor everything happening on the instance.

dureuill: discussed with @gmourier and @irevoire: we'll be switching to the following:

>  Reject access to the metrics route with an error if the corresponding API key doesn't have filters().all_indexes_authorized(), so that a user must explicitly have all indexes authorized + the metrics action to access the metrics route.

gmourier: Thanks, @dureuill!

Issue #3521, opened 2023-02-21
Panic when searching with stop-words in quotes


Northn:
**Describe the bug**
When performing a simple POST request with any stop-word in quote, MeiliSearch returns HTTP Error 500 and it panics

**To Reproduce**
1. Create an index
2. Add any stop-word to it by corresponding method, for example, hello
3. Perform a POST search request with that stop-word in quotes, and add some word before/after these quotes, e.g.: "`some \"hello\"`"
4. See the error

**Expected behavior**
Search request has to be completed without any error

**Screenshots**
None

**Meilisearch version:**
1.0.1, but applies to 1.0.0 too

**Additional context**
Ubuntu 20.04, Debian, OS does not matter
amd64
Also there's no error if nothing else added before/after the quotes, for example, it won't panic if searching "`\"hello\"`" instead of "`some \"hello\"`"

curquiza:
Hello @Northn 
Thanks for the report 🙏

I succeeded in reproducing

- add documents
- add stop-words `["the", "of"]`
- search with `"q": "prince \"of\""`
- get a panic
```bash
{
    "message": "task 38 panicked",
    "code": "internal",
    "type": "internal",
    "link": "https://docs.meilisearch.com/errors#internal"
}
```

curquiza:
Added to v1.0.2 milestones
We will release a patch release asap 😇

curquiza:
Hello again @Northn 

A small question. What was your original use case, I mean, having stop words in the `"`, so for a phrase search? How did you get this? 😊

Northn:
> Hello again @Northn
> 
> A small question. What was your original use case, I mean, having stop words in the `"`, so for a phrase search? How did you get this? 😊

Well, that's really funny situation. We made an addon to integrate Xenforo with Meilisearch, and obviously, all content needs to be indexed. After full indexation it has to find and cache similar threads list. So, uhm, someone accidentally created a thread with stopword in brackets, we got tons of errors in logs and reported 😂😂

curquiza:
Closed by #3531, will be present in v1.0.2 release

PR #3525, opened 2023-02-21
Fix phrase search containing stop words

ManyTheFish:
# Summary
A search with a phrase containing only stop words was returning an HTTP error 500,
this PR filters the phrase containing only stop words dropping them before the search starts, a query with a phrase containing only stop words now behaves like a placeholder search.

fixes https://github.com/meilisearch/meilisearch/issues/3521

related v1.0.2 PR on milli: https://github.com/meilisearch/milli/pull/779

### DIFF
--- a/benchmarks/benches/formatting.rs
+++ b/benchmarks/benches/formatting.rs
@@ -29,7 +29,7 @@ fn bench_formatting(c: &mut criterion::Criterion) {
 	            (vec![Rc::new(MatchingWord::new("thedoord".to_string(), 1, true).unwrap())], vec![0, 1, 2]),
 	            (vec![Rc::new(MatchingWord::new("doord".to_string(), 1, true).unwrap())], vec![1, 2]),
         	]
-            ), TokenizerBuilder::default().build()),
+            ).unwrap(), TokenizerBuilder::default().build()),
 		},
     ];
 
--- a/meilisearch/tests/search/mod.rs
+++ b/meilisearch/tests/search/mod.rs
@@ -148,6 +148,27 @@ async fn simple_search() {
         .await;
 }
 
+#[actix_rt::test]
+async fn phrase_search_with_stop_word() {
+    // related to https://github.com/meilisearch/meilisearch/issues/3521
+    let server = Server::new().await;
+    let index = server.index("test");
+
+    let (_, code) = index.update_settings(json!({"stopWords": ["the", "of"]})).await;
+    meili_snap::snapshot!(code, @"202 Accepted");
+
+    let documents = DOCUMENTS.clone();
+    index.add_documents(documents, None).await;
+    index.wait_task(1).await;
+
+    index
+        .search(json!({"q": "how \"to\" train \"the" }), |response, code| {
+            assert_eq!(code, 200, "{}", response);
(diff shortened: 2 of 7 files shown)


curquiza:
@ManyTheFish can we merge this? 😊

ManyTheFish:
@curquiza yes we will merge this for v1.1.0

dureuill (approved):
Thank you! I approve this PR since similar code was approved by @irevoire for v1.0.2.

Should we directly address the panicking dereference in `matching_words.rs`? We could either add explicit handling of the case where the vec is empty at that point, or, if that doesn't make sense, add some documentation on the `MatchingWords` structure and `MatchingWords::new` function stating the precondition on the input `Vec<(Vec<Rc<MatchingWord>>` (each inner vec contains at least one element).

What do you think?

ManyTheFish:
Hello @dureuill, I think this will not occur anymore, however, to be 100% safe, I can change the code!

Let me know, I'll follow your suggestion.

dureuill:
> I think this will not occur anymore

That's great! I think we can do either of:

 1. keep the panic behavior with some documentation on the precondition of the `MatchingWords` structure:
   ```rust
  impl MatchingWords {
      /// Constructs a `MatchingWords` container from its inner structure.
      ///
      /// # Panics
      ///
      /// - If any of the inner vectors in `matching_words` is empty.
      pub fn new(mut matching_words: Vec<(Vec<Rc<MatchingWord>>, Vec<PrimitiveWordId>)>) -> Self {
  ```

   Then ideally we'd also document above the calls to `MatchingWords::new` why none of the vecs are empty. 
   
 2. let the callers handle the precondition by changing the return value of `MatchingWords::new` to an `Option<Self>`:
 ```rust
     /// Constructs a `MatchingWords` container from its inner structure.
    ///
    /// Returns `None` if any of the inner vectors in `matching_words` is empty.
    pub fn new(mut matching_words: Vec<(Vec<Rc<MatchingWord>>, Vec<PrimitiveWordId>)>) -> Option<Self> {
```

In this second option we'd need to update the callers to handle the option, either to pass up the `None` further, or to `unwrap` it with a comment explaining why the precondition holds (unless it is already established from context).

dureuill (commented):
@ManyTheFish 

OK, I'm unsure which is better between failing-fast like before or filtering non-sensical inputs like now. If going this third route, I suggested using `retain` rather than a new `Vec` to spare the allocation.

ManyTheFish:
Hello @dureuill, I took into account your last comment, we are now returning an internal error in the case of the MatchingWord creation fails.

dureuill (approved):
Hello @ManyTheFish!

Thanks for the changes! 

In general, I'd say that for errors that "should never happen" eg programming errors, we could prefer panicking to `InternalError`s, so as to not create "dead code".

However, I'm not really aware of what our policy is wrt `InternalError` vs `panic`s so I trust that you chose the solution that is in line with the current policy. Thank you again 😃

ManyTheFish:
@dureuill, In an ideal world, I'd say that an internal error should never happen, therefore, if an internal error appears, we expect the user to create an issue referenced as a bug. So from my POV, it seems legitimate that an internal error is thrown in this case.
thank you for your review!

bors merge

PR #3537, opened 2023-02-23
fix a bug where the filestore could try to parse its own tmp file and fail (main)

curquiza:


irevoire (approved):
Wow, this code looks SOOO nice; thanks 🐶 
bors merge

Issue #3540, opened 2023-02-23
Cache the result of the indexes stats 


irevoire:
Since https://github.com/meilisearch/meilisearch/pull/3331 has been merged, we now have the possibility to store thousands of indexes in meilisearch.
But at the same time, opening indexes became slower.

That can be a real issue when calling the global `stats`.
And it will also degrade the usage of the new experimental `/metrics` route.

After chatting with @dureuill we decided it would be better to cache the result of the stats on disk to avoid opening the indexing just to extract a few information out of the index.

gmourier:
Does the cache is invalidated as soon as an index operation impacts its stats number? Just wondering if the information is always up-to-date 🤓

irevoire:
The cache is updated after any update is applied to an index. 
Thus it should always be up to date.

curquiza:
Fixed by https://github.com/meilisearch/meilisearch/pull/3541
Will be released in rc1 today

PR #3541, opened 2023-02-23
Add cache on the indexes stats

irevoire:
Fix https://github.com/meilisearch/meilisearch/issues/3540


dureuill (dismissed):
Thank you for this PR, I think it is a very good first step towards caching the stats of the indexes! 🎉 

I have a few concerns to resolve before we can accept this PR:

1. upgrade path: my understanding from reading the code is that running this code on an instance that has indexes from before this PR will lead to `IndexNotFound` errors until an update is executed on the index. Similarly, failing a stats write on index creation is not a cause for error (only a log), yet it will cause an `IndexNotFound` error in the stats route.
2. support for index swapping: before this PR, since the stats are computed eagerly, swapping two indexes will also swap this stats. Because the new DB uses the name of the index as key, my understanding here is that the swapping the indexes won't update their stats, and so won't swap the stats at the same time than the indexes.
3. persistent database vs in-memory cache: this PR implements the cache as a persistent DB. @Kerollmops expressed a preference for an in-memory cache, to avoid storing this redundant information on disk.

Regarding (1), I think the DB should be made optional such as if an index has no entry in the cache, then it is fetched eagerly and added to the cache. This will make upgrading seamless.

Regarding (2), I think this can be fixed by having the new DB use UUIDs as keys rather than the index names and use the existing name -> UUID table as an intermediary to retrieve the current correspondence between an index name's and its stats.

I don't have a strong opinion one way or the other regarding (3). The main advantage of the disk representation is that there won't be a "cold start" due to the need of populating the cache on startup (or first call to the stats route). The main drawback is that if we get the implementation wrong, any mistake appearing in edge cases will be more persistent and harder to correct. On the other hand, the lifetime of any entry being one index update, I don't think this is too much of a drawback.

For now, I intent to modify this PR to implement solutions for points (1) and (2). I don't intent to switch the DB to an in-memory representation. I feel like this can be easily changed anyway.

dureuill:
>  I think the DB should be made optional such as if an index has no entry in the cache, then it is fetched eagerly and added to the cache.

Actually, the *and added to the cache* part is going to be harder than expected as the stats route is a reader of the DB and so can't typically write to it. This is a motivation for switing to an in-memory database, that can be set behind a RWLock so it can be shared.

dureuill:
Pushed an update with the following changes:

- Rebase on main
- Switch the DB to UUID -> Stats instead of str -> Stats
- Fallback to eager computation of the stats if the stats DB is missing/incomplete
- Refactor all around to avoid reopening indexes
- Fix an issue where we could reopen an index while already holding it, resulting in a deadlock if the index is evicted from the cache between its updates and the stat computation.
- Finish fixing snapshots
- Documented the introduced functions and structures

dureuill:
@irevoire doesn't look like I can request your review (which is a bit silly since I changed your PR, but...).

Also requesting @Kerollmops ' review

index-scheduler/src/snapshots/lib.rs/cancel_enqueued_task/cancel_processed.snap
```diff
@@ -1,6 +1,5 @@
 ---
 source: index-scheduler/src/lib.rs
-assertion_line: 1755
```
dureuill, line 3: I don't really have any idea why there was an `assertion_line` here and what it means and why it is not here anymore.

curquiza:
@irevoire can you rebase the branch to point to `release-v1.1.0`?

index-scheduler/src/index_mapper/mod.rs
```diff
@@ -360,6 +404,45 @@ impl IndexMapper {
         Ok(())
     }
 
+    /// The stats of an index.
+    ///
+    /// If available in the cache, they are directly returned.
+    /// Otherwise, the `Index` is opened to compute the stats on the fly (the result is not cached).
```
irevoire, line 410: Is this possible?
From what I see, we're always creating and updating the stats of the indexes; thus, we should probably return an internal error instead of computing the stats silently?

index-scheduler/src/snapshots/lib.rs/cancel_enqueued_task/cancel_processed.snap
irevoire: Don't know either but if that makes insta happy I'm 100% for it 😂

irevoire (commented):
Nice

index-scheduler/src/index_mapper/mod.rs
dureuill: It happens when e.g. using a 1.0 DB with a 1.1 Meilisearch.

Kerollmops: Yeah, I prefer that we continue to be resilient to missing stats. And I like that we only try to cache it if it is present (we could wait for too long anyway).

index-scheduler/src/index_mapper/mod.rs
```diff
@@ -360,6 +404,45 @@ impl IndexMapper {
...
+        stats: IndexStats,
+    ) -> Result<()> {
+        let uuid = self
+            .index_mapping
+            .get(wtxn, index_uid)?
+            .ok_or_else(|| Error::IndexNotFound(index_uid.to_string()))?;
+
+        self.index_stats.put(wtxn, &uuid, &stats)?;
+        Ok(())
+    }
```
Kerollmops, line 444:
```suggestion
    pub fn store_stats_of(
        &self,
        wtxn: &mut RwTxn,
        index_uid: &str,
        stats: &IndexStats,
    ) -> Result<()> {
        let uuid = self
            .index_mapping
            .get(wtxn, index_uid)?
            .ok_or_else(|| Error::IndexNotFound(index_uid.to_string()))?;

        self.index_stats.put(wtxn, &uuid, stats)?;
        Ok(())
    }
```

meilisearch/src/routes/mod.rs
```diff
@@ -260,32 +259,19 @@ pub fn create_all_stats(
...
-    let processing_index = processing_task.first().and_then(|task| task.index_uid());
-    index_scheduler.try_for_each_index(|name, index| {
-        if !filters.is_index_authorized(name) {
-            return Ok(());
+    for index_uid in index_scheduler.index_names()? {
+        if !filters.is_index_authorized(&index_uid) {
+            continue;
         }
 
-        database_size += index.on_disk_size()?;
```
Kerollmops, line 274: Ho! you removed the increment of the database size.

But, don't worry, that's actually great, you'll be able to **always** increment it even if the user doesn't have the right to access this index 😊 We talked about that with @gmourier and thought it was bad to not show the real database size on disk, this is not a security concern and it conflicts with the [Principle of least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). Could you change that in this PR, please?

Kerollmops (changes requested):

curquiza:
(In case you misstd it, still not the right branch to merge into it, should be `release-v1.1.0` 😇)

index-scheduler/src/index_mapper/mod.rs
irevoire: > It happens when e.g. using a 1.0 DB with a 1.1 Meilisearch.

But that won't happen since you need to use a dump to move from 1.0 to 1.1.

However, if you both prefer keeping it this way, we can. I just feel like it could hide a bug.

meilisearch/src/routes/mod.rs
irevoire: Isn't it the kind of thing that allows a [side-channel attack](https://en.wikipedia.org/wiki/Side-channel_attack)?

Personally, I don't have any strong opinion on this subject, but the fact that this line was removed is definitely a bug 😂

index-scheduler/src/index_mapper/mod.rs
dureuill: > But that won't happen since you need to use a dump to move from 1.0 to 1.1.

Will we actually need to use a dump?

If adding this DB is the only change we did to the data structures, I'd be OK to allow using a 1.0 DB from a 1.1 Meilisearch.

meilisearch/src/routes/mod.rs
Kerollmops: This is not a security issue but more a correctness issue here. The `databaseSize` stats value should display the actual database size, not less. I don't think there is any possible attack with this `/stats` route.

Kerollmops: Thank you @dureuill 😂

dureuill:
Updated:

- Restore contribution of the index sizes to the db size
  - the index size now contributes to the db size even if the index is not authorized
- Pass IndexStat by ref in store_stats_of

meilisearch/src/routes/mod.rs
dureuill: done

index-scheduler/src/index_mapper/mod.rs
dureuill: changed

Kerollmops (approved):
Thank you for the PR!

However, could please you create an issue describing that the stats are now computing the index disk size too? And make this PR close it?

Thank you 🦪

dureuill:
This PR also fixes #3578

dureuill:
bors merge

PR #3544, opened 2023-02-27
Attempt to use default vram budget for faster startup

dureuill:
# Pull Request

## Related issue
Follow-up to #3382: addresses the added startup time on Windows/macOS.

## What does this PR do?
- Attempt to skip budget calculation by using "known good values" instead
- Perform dichotomic budget calculation as fallback only when the known value is not actually good.

### DIFF
--- a/index-scheduler/src/lib.rs
+++ b/index-scheduler/src/lib.rs
@@ -435,9 +435,19 @@ impl IndexScheduler {
         mut task_db_size: usize,
         max_index_count: usize,
     ) -> IndexBudget {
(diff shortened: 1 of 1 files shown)


dureuill:
Tested on macOS, Windows and Linux:

- had to decrease the value in Windows because 12TiB was at the limit and so would sometimes trigger the slow behavior. Since it only resulted in 1 index available anyway, I choose 6TiB instead which should reliably skip dichotomy.
- Startup in now fast in all three OSes :-)

Kerollmops (approved):
I love what you've done ❤️
bors merge

Issue #3477, opened 2023-02-09
Delete and get documents by filter


dureuill:
Product card (internal only): https://www.notion.so/meilisearch/Enhance-document-management-Get-Delete-by-filter-999adc1dc73c4a5381f0b84e523ad3be
Related product discussion: https://github.com/meilisearch/product/discussions/284
Related specifications:
- Get documents by filter: https://github.com/meilisearch/specifications/pull/234
- Delete documents by filter: https://github.com/meilisearch/specifications/pull/236

Implemented in:
- Get documents by filter: https://github.com/meilisearch/meilisearch/pull/3570
- Delete documents by filter: https://github.com/meilisearch/meilisearch/pull/3550

### Impacted teams

@meilisearch/docs-team 
@meilisearch/integration-team

curquiza:
For people following this issue: we probably will not be able to make it for v1.1. In this case, this improvement will be implemented for v1.2 😇

dureuill:
Opened #3550 as an initial exploration of the feature.

This surfaces a few product questions.

1. Should we reuse the existing `/indexes/{:indexUid}/documents/delete-batch` route, or use a new one (e.g. `../delete-filter`)?
2. Similarly, should we reuse the existing `documentDeletion` task type, or introduce a new `documentDeletionByFilter` task type?
3. What should be the "details" of the task when performing a delete by filter?
4. If reusing the `documentDeletion` task, what should we do of the `providedIds` field? Should it be set to `null`, absent altogether, ...?
5. Should the filter be applied on the documents as they are in the index when the query is received ("query time"), or as they are once all the previously sent updates have been asynchronously applied by the scheduler ("processing time")? Whichever solution we choose, I think the documentation should talk about this timing issue and discourage the use of "flaky" filters that are too dependent on an update being applied or not to the document (for instance, filters on fields that are often updated).

ManyTheFish:
@dureuill I'll speak only about 5.
I think that it should be computed at "processing time" to respect the task processing order, and, as explained in a private message:
 - a document modification is in fact a deletion of the old version then, the addition of the new version. That means that if the list of documents to delete is computed at query time, then, all the documents that will be modified before the processing of the deletion will not be deleted. However, this case appears only if you list the internal document ids without fetching the external document ids at "query time".
- what do we do with the document additions between the "query time" list computation and the processing of the task deletion? Should we keep them despite the fact that if the indexing were faster they would be deleted?

[...]

PR #3550, opened 2023-03-01
Delete documents by filter

dureuill:
# Prototype `prototype-delete-by-filter-0`

Usage:
A new route is available under `POST /indexes/{index_uid}/documents/delete` that allows you to delete your documents by filter.
The expected payload looks like that:
```json
{
  "filter": "doggo = bernese",
}
```

It'll then enqueue a task in your task queue that'll delete all the documents matching this filter once it's processed.
Here is an example of the associated details;
```json
  "details": {
    "deletedDocuments": 53,
    "originalFilter": "\"doggo = bernese\""
  }
```

----------

# Pull Request

## Related issue
Related to https://github.com/meilisearch/meilisearch/issues/3477

### User standpoint

- Modifies the `/indexes/{:indexUid}/documents/delete-batch` route to accept either the existing array of documents ids, or a JSON object with a `filter` field representing a filter to apply. If that latter variant is used, any document matching the filter will be deleted.

### Implementation standpoint

- (processing time version) Adds a new BatchKind that is not autobatchable and that performs the delete by filter
- Reuse the `documentDeletion` task with a new `originalFilter` detail that replaces the `providedIds` detail.

## Example

<details>
<summary>Sample request, response and task result</summary>

Request:

```
curl \
  -X POST 'http://localhost:7700/indexes/index-10/documents/delete-batch' \
  --data-binary '{ "filter" : "mass = 600"}'
```

Response:

```
{
  "taskUid": 3902,
  "indexUid": "index-10",
  "status": "enqueued",
  "type": "documentDeletion",
  "enqueuedAt": "2023-02-28T20:50:31.667502Z"
}
```

Task log:

```json
    {
      "uid": 3906,
      "indexUid": "index-12",
      "status": "succeeded",
      "type": "documentDeletion",
      "canceledBy": null,
      "details": {
        "deletedDocuments": 3,
        "originalFilter": "\"mass = 600\""
      },
      "error": null,
      "duration": "PT0.001819S",
      "enqueuedAt": "2023-03-07T08:57:20.11387Z",
      "startedAt": "2023-03-07T08:57:20.115895Z",
      "finishedAt": "2023-03-07T08:57:20.117714Z"
    }
```

</details>

## Draft status

- [ ] Error handling
- [ ] Analytics
- [ ] Do we want to reuse the `delete-batch` route in this way, or create a new route instead?
- [ ] Should the filter be applied at request time or when the deletion task is processed? 
  - The first commit in this PR applies the filter at request time, meaning that even if a document is modified in a way that no longer matches the filter in a later update, it will be deleted as long as the deletion task is processed after that update. 
  - The other commits in this PR apply the filter only when the asynchronous deletion task is processed, meaning that documents that match the filter at processing time are deleted even if they didn't match the filter at request time.
- [ ] If keeping the filter at request time, find a more elegant way to recover the user document ids from the internal document ids. The current way implemented in the first commit of this PR involves getting all the documents matching the filter, looking for the value of their primary key, and turning it into a string by copy-pasting routines found in milli...
- [ ] Security consideration, if any
- [ ] Fix the tests (but waiting until product questions are resolved)
- [ ] Add delete by filter specific tests

### DIFF
--- a/meilisearch/src/analytics/segment_analytics.rs
+++ b/meilisearch/src/analytics/segment_analytics.rs
@@ -949,6 +949,7 @@ pub struct DocumentsDeletionAggregator {
     per_document_id: bool,
     clear_all: bool,
     per_batch: bool,
+    per_filter: bool,
 }
 
 impl DocumentsDeletionAggregator {
@@ -962,6 +963,7 @@ impl DocumentsDeletionAggregator {
             DocumentDeletionKind::PerDocumentId => ret.per_document_id = true,
             DocumentDeletionKind::ClearAll => ret.clear_all = true,
             DocumentDeletionKind::PerBatch => ret.per_batch = true,
+            DocumentDeletionKind::PerFilter => ret.per_batch = true,
         }
 
         ret
@@ -981,6 +983,7 @@ impl DocumentsDeletionAggregator {
         self.per_document_id |= other.per_document_id;
         self.clear_all |= other.clear_all;
         self.per_batch |= other.per_batch;
+        self.per_filter |= other.per_filter;
     }
 
     pub fn into_event(self, user: &User, event_name: &str) -> Option<Track> {
--- a/meilisearch/src/analytics/segment_analytics.rs
+++ b/meilisearch/src/analytics/segment_analytics.rs
@@ -963,7 +963,7 @@ impl DocumentsDeletionAggregator {
             DocumentDeletionKind::PerDocumentId => ret.per_document_id = true,
             DocumentDeletionKind::ClearAll => ret.clear_all = true,
             DocumentDeletionKind::PerBatch => ret.per_batch = true,
-            DocumentDeletionKind::PerFilter => ret.per_batch = true,
+            DocumentDeletionKind::PerFilter => ret.per_filter = true,
         }
 
         ret
--- a/meilisearch/src/routes/indexes/documents.rs
+++ b/meilisearch/src/routes/indexes/documents.rs
@@ -17,7 +17,6 @@ use meilisearch_types::error::{Code, ResponseError};
 use meilisearch_types::heed::RoTxn;
 use meilisearch_types::index_uid::IndexUid;
 use meilisearch_types::milli::update::IndexDocumentsMethod;
-use meilisearch_types::milli::InternalError;
 use meilisearch_types::star_or::OptionStarOrList;
 use meilisearch_types::tasks::KindWithContent;
 use meilisearch_types::{milli, Document, Index};
@@ -381,31 +380,6 @@ pub enum DocumentDeletionQuery {
     Object { filter: Option<Value> },
 }
 
-/// Parses a Json encoded document id and validate it, returning a user error when it is one.
-/// FIXME: stolen from milli
-fn validate_document_id_value(document_id: Value) -> String {
-    match document_id {
-        Value::String(string) => match validate_document_id(&string) {
-            Some(s) if s.len() == string.len() => string,
-            Some(s) => s.to_string(),
-            None => panic!(),
-        },
-        Value::Number(number) if number.is_i64() => number.to_string(),
-        _content => panic!(),
-    }
-}
-
-/// FIXME: stolen from milli
-fn validate_document_id(document_id: &str) -> Option<&str> {
-    if !document_id.is_empty()
-        && document_id.chars().all(|c| matches!(c, 'a'..='z' | 'A'..='Z' | '0'..='9' | '-' | '_'))
-    {
-        Some(document_id)
-    } else {
-        None
-    }
-}
-
 pub async fn delete_documents(
     index_scheduler: GuardedData<ActionPolicy<{ actions::DOCUMENTS_DELETE }>, Data<IndexScheduler>>,
     index_uid: web::Path<String>,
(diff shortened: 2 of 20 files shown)


Kerollmops:
> Should the filter be applied at request time or when the deletion task is processed?

I think that we must follow the behavior of the current update system. The way the deletion work is that it evaluates the ids to delete when it is executed, no error is raised until the update is processed.

Doing that is much easier to work with when using the task queue and one day replicating the queue too.

dureuill:
thanks for the answer @Kerollmops, but unfortunately I'm don't understand your suggestion.

Are you suggesting that the filter be applied at request time, or only when the deletion task is processed?

If applied at request time, that means that we evaluate which documents match the filter synchronously before answering the HTTP request. It means we have to go back to the user (external) document ids because we have no guarantee that a previous document update will not modify the internal document ids. If applying at request time we can reuse the existing tasks without a change because the filter is converted to the ids of the documents synchronously.

If applied at processing time, that means that we evaluate which documents match the filter asynchronously, after answering the HTTP request. It means we have to somehow pass the filter to the task so the filter evaluation can take place in the index scheduler. It also means we have to handle any filter-related error that can happen in the index scheduler (I don't think we're currently filtering inside of the index scheduler). We can directly use the internal ids here because no other task can change them during the processing of our task.

Kerollmops:
> Are you suggesting that the filter be applied at request time, or only when the deletion task is processed?

Only when the deletion task is processed. I want to evaluate the request in the queue as we already do for the other way to batch delete. When we delete the documents by using the array of ids, we don't evaluate the ids at request time but rather at processing time.

> It means we have to somehow pass the filter to the task so the filter evaluation can take place in the index scheduler.

Indeed, it's unfortunately breaking and for the dumps too!

> It also means we have to handle any filter-related error that can happen in the index scheduler (I don't think we're currently filtering inside of the index scheduler).

Indeed, the deletion-by-filter feature will only work if the fields are filterable. However, we could check, at least, the syntax at request time when we receive the filter. We could also accept to delete-by-filter even if some fields are non-filterable, but that's a future improvement.

> We can directly use the internal ids here because no other task can change them during the processing of our task.

If you mean that the index-scheduler will execute the filter to fetch the internal document ids and that the scheduler always process one task at a time, then yes, that's what I was thinking about 😃

dureuill:
> Only when the deletion task is processed. I want to evaluate the request in the queue as we already do for the other way to batch delete. When we delete the documents by using the array of ids, we don't evaluate the ids at request time but rather at processing time.

Sorry this is still unclear to me 😓 

When we delete the documents by using an array of external document ids, it is true that the external document ids are not evaluated at request time. Only at processing time are they used by the batcher to determine the transform output, and after the transform is applied they are turned to internal ids for the actual deletion.

For the delete by filter, we could do either:

1. keep the filter, evaluate it to internal id at processing time
2. evaluate the filter to internal ids at processing time, then go back the corresponding external ids still at processing time, so that we can express the delete by filter like a delete by array, basically.

dureuill:
So, discussing this further with the team, @irevoire suggested we don't auto-batch delete-by-filter tasks, in which case we would be able to delete the documents directly from their internal ids.

Summarizing the 3 solutions I see at the moment:

1. Apply filter at query time (this PR in its current shape).
  - Needs to round trip from internal ids to external ids
  - No need to change the `documentDeletion` task, we can reuse it directly
  - Benefits from autobatching
2. Apply filter at processing time, keeping autobatching.
  - Needs a new kind of task, or to modify `documentDeletion` task.
  - Needs to round trip from internal ids to external ids anyway.
  - Benefits from autobatching
3. Apply filter at processing time, disabling autobatching for this `deleteByFilter` tasks.
  - Needs a new kind of task (or to modify `documentDeletion` task and have it autobatch conditionally)
  - No round trip from internal ids to external ids.
  - No autobatching

At the moment it is unclear to me what the best way to fetch the external ids from the internal ids might be.

We could:

1. Load the full document from the internal id, then read the primary key value (this PR in its current form)
2. Iterate on the external ids -> internal ids FST until we find the correct internal id

dureuill:
Thanks for the input @irevoire !

> Delete a document that the user never wanted to delete in an obscure manner
> Fail to delete a document that the user clearly specified

I don't see how this could happen if we're resolving to external ids at query time. Can you describe the sequence of events that could lead to this happening? 

I agree that it could definitely happen if we resolve the filter to internal ids at query time and then use the internal ids to delete at processing time, but this is not what I'm proposing.

> We also talked about being able to delete non-filterable documents. Has that been canceled/delayed?

Cannot find a reference to it, but yes, we agreed to have a first implementation that doesn't do that. Discussion with users struggled to find a use case for this.

> There is definitely an issue we need to consider with the tenant tokens that can contain a filter!

I was under the impression that tenant tokens would only apply to the search; is that incorrect? Considering that users can delete documents by their external ids today with just the `documents.delete` action + index authorization, does it make sense to limit the filter by tenant token? If yes, then you're correct that we need to take it into account.

dureuill:
Updated with added commits that implement solution (3) from https://github.com/meilisearch/meilisearch/pull/3550#issuecomment-1456073242

dureuill:
Rebased on `main`

meilisearch/src/analytics/segment_analytics.rs
```diff
@@ -962,6 +963,7 @@ impl DocumentsDeletionAggregator {
             DocumentDeletionKind::PerDocumentId => ret.per_document_id = true,
             DocumentDeletionKind::ClearAll => ret.clear_all = true,
             DocumentDeletionKind::PerBatch => ret.per_batch = true,
+            DocumentDeletionKind::PerFilter => ret.per_batch = true,
```
dureuill, line 966:
```suggestion
            DocumentDeletionKind::PerFilter => ret.per_filter = true,
```

meilisearch/src/routes/indexes/documents.rs
```diff
@@ -71,8 +71,11 @@ pub fn configure(cfg: &mut web::ServiceConfig) {
             .route(web::put().to(SeqHandler(update_documents)))
             .route(web::delete().to(SeqHandler(clear_all_documents))),
     )
-    // this route needs to be before the /documents/{document_id} to match properly
-    .service(web::resource("/delete-batch").route(web::post().to(SeqHandler(delete_documents))))
+    // these routes needs to be before the /documents/{document_id} to match properly
```
dureuill, line 74:
```suggestion
    // these routes need to be before the /documents/{document_id} to match properly
```

dureuill (commented):
Looking OK, needing some minor changes (typo analytics)

did not check the dumps

dureuill (commented):
Would approve if it was possible.

Tested exporting and importing a dump, worked OK; did not test importing a dump from a previous version.

Tested deleting documents in the movie dataset by genre. Worked like a charm.

irevoire (approved):
I love my code

bors merge

dureuill:
trying again after spurious failure

bors merge

Issue #3507, opened 2023-02-17
Snapshot file permission issues


Magorx15:
**Describe the bug**
After creating a snapshot, data.ms.snapshot file has incorrect permissions, unable to interact with it without sudo or as root

**To Reproduce**
`docker run -e MEILI_SNAPSHOT_DIR="/snapshots" -e MEILI_SCHEDULE_SNAPSHOT=true -e MEILI_SNAPSHOT_INTERVAL_SEC=10 getmeili/meilisearch:v0.29.0`
The snapshot in /snapshots folder has correct permissions (see screenshots)

`docker run -e MEILI_SNAPSHOT_DIR="/snapshots" -e MEILI_SCHEDULE_SNAPSHOT=true -e MEILI_SNAPSHOT_INTERVAL_SEC=10 getmeili/meilisearch:v0.30.0`
The snapshot in /snapshots folder has incorrect permissions (see screenshots)

**Screenshots**
Using v0.29.0
![image](https://user-images.githubusercontent.com/118532529/219595612-64262ae8-df76-4e10-ab46-175e66ca5ba0.png)
Using v0.30.0-v1.0.0
![image](https://user-images.githubusercontent.com/118532529/219598028-3f7225bf-df43-4a86-8cd7-d3d4168ffd95.png)

**Meilisearch version:** v0.29.0, v0.30.0-v1.0.0

oluademola:
Hello @Magorx15. Thank you for informing us about this problem. Indeed the snapshot permission file changed from v0.30 upward. I'm not sure if this change was intended or not; however, I'd discuss it with the team and get back to you. I understand this might trigger permission issues, but I would like to know specifically how this change impacts you. Thanks again.

Magorx15:
We are trying to upload the snapshot to somewhere else (and therefore we need read permission) which obviously fails as it needs root user for that

Magorx15:
Any updates @oluademola?

irevoire:
Hey @Magorx15, yeah, this was definitely an issue on our side; it should be fixed for the next release!
Thanks for the report 💯

curquiza:
Fixed by #3561, will be released in v1.1.0-rc.1 by the end of the week or at the beginning of next week

PR #3561, opened 2023-03-06
Fix the snapshots permissions on unix system

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/3507

The snapshot permissions were wrong after the v0.30 and the huge refacto of the index scheduler.
Fix this issue + add a test on the permissions on unix


curquiza:
⚠️ Do NOT merge until we defined which release will contain this fix (v1.0.3 or v1.1.0-rc.0 or v1.1.0-rc.1) -> in any case, it will not be on `main`
Internal discussion [here](https://meilisearch.slack.com/archives/G01A1F4KVGU/p1678106995961809?thread_ts=1678102150.740059&cid=G01A1F4KVGU)

meilisearch/tests/snapshot/mod.rs
```diff
@@ -71,7 +80,10 @@ async fn perform_snapshot() {
         // for some reason the db sizes differ. this may be due to the compaction options we have
         // set when performing the snapshot
         //server.stats(),
-        server.tasks(),
+
+        // We can't test all the tasks contained in the snapshot because the on the original instance the snapshotCreation task was added
```
Kerollmops, line 84: Not sure to understand this comment.

index-scheduler/src/batch.rs
```diff
@@ -714,6 +711,14 @@ impl IndexScheduler {
                 // 5.3 Change the permission to make the snapshot readonly
                 let mut permissions = file.metadata()?.permissions();
                 permissions.set_readonly(true);
+                #[cfg(unix)]
+                {
+                    use std::os::unix::fs::PermissionsExt;
+                    #[allow(clippy::non_octal_unix_permissions)]
+                    //                     rwxrwxrwx
+                    permissions.set_mode(0b100100100);
```
Kerollmops, line 719: What is the mode set by`set_readonly`? not sure to understand what they meant by readonly.

Kerollmops (changes requested):

curquiza:
@irevoire can you rebase the branch to point to `release-v1.1.0`?

index-scheduler/src/batch.rs
irevoire: From what I see, it removes the right to write + sets the right to read if you're root.

_I think_ it's sufficient for windows, but it's not for UNIX. I decided to keep it just in case, but tbh, I don't really know if we should.

meilisearch/tests/snapshot/mod.rs
irevoire: It says that we can't test **all** the tasks; we have to start at task `4` because the original index contains something like this;
```
5. snapshotCreation
4. documentAddition
3. indexCreation
2. documentAddition
1. indexCreation
```

While the snapshotted index contains something like that;
```
4. documentAddition
3. indexCreation
2. documentAddition
1. indexCreation
```

Kerollmops: > because the on the original instance the snapshotCreation task was added

I don't understand **this** part specifically 😂

index-scheduler/src/batch.rs
Kerollmops: Thank you! It looks like you are right. We must set the read bits to true for the root part.

Kerollmops (approved):
Thank you very much @irevoire!

curquiza:
(don't know if you notice, but still not the right branch to merge in it 😇)

irevoire:
@Kerollmops, sorry for the push; I realized that I was matching snapshot tasks in my queue with my filter. There are only two tasks in reality.
+ The sleep of one second was not enough for windows; thus, I went back to the two seconds of sleep we had initially.

Kerollmops (approved):
Thank you :)

irevoire:
bors merge

PR #3570, opened 2023-03-08
Get documents by query

dureuill:
# Pull Request

## Related issue

None really, this is more of an extension of #3477: since after this issue we'll be able to delete documents by filter, it makes sense to also be able to get documents by filter. 

### User standpoint

- Add a new `filter` URL parameter to `GET /indexes/{:indexUid}/documents` and a new `POST /indexes/{:indexUid}/documents/fetch` route with the same `offset, limit, fields, filter` parameters. 

### Implementation standpoint

-  Add a new `Index::iter_documents` method to iterate on a set of documents rather than return a vector of these documents.
- Rewrite the other `Index::*documents` methods to use the new `Index::iter_documents` method.

## PR checklist

### DIFF
--- a/meilisearch/src/routes/indexes/documents.rs
+++ b/meilisearch/src/routes/indexes/documents.rs
@@ -4,19 +4,20 @@ use actix_web::http::header::CONTENT_TYPE;
 use actix_web::web::Data;
 use actix_web::{web, HttpMessage, HttpRequest, HttpResponse};
 use bstr::ByteSlice;
-use deserr::actix_web::AwebQueryParameter;
+use deserr::actix_web::{AwebJson, AwebQueryParameter};
 use deserr::Deserr;
 use futures::StreamExt;
 use index_scheduler::IndexScheduler;
 use log::debug;
 use meilisearch_types::deserr::query_params::Param;
-use meilisearch_types::deserr::DeserrQueryParamError;
+use meilisearch_types::deserr::{DeserrJsonError, DeserrQueryParamError};
 use meilisearch_types::document_formats::{read_csv, read_json, read_ndjson, PayloadType};
 use meilisearch_types::error::deserr_codes::*;
 use meilisearch_types::error::{Code, ResponseError};
 use meilisearch_types::heed::RoTxn;
 use meilisearch_types::index_uid::IndexUid;
 use meilisearch_types::milli::update::IndexDocumentsMethod;
+use meilisearch_types::milli::DocumentId;
 use meilisearch_types::star_or::OptionStarOrList;
 use meilisearch_types::tasks::KindWithContent;
 use meilisearch_types::{milli, Document, Index};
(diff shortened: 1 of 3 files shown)


Kerollmops (commented):
Thank you very much, that pretty cool 🧃

You could probably work on implementing [the `adavance_by` `Iterator` trait method on the `RoaringBitmap` type](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.advance_by) for even better `skip` performance, but not sure that it is necessary 😂

Do you correctly handle the filters [as JSON arrays, arrays or arrays, or simply strings](https://docs.meilisearch.com/learn/advanced/filtering_and_faceted_search.html#filter-basics)?

dureuill:
> Do you correctly handle the filters [as JSON arrays, arrays or arrays, or simply strings](https://docs.meilisearch.com/learn/advanced/filtering_and_faceted_search.html#filter-basics)?

Yes it works the same as the other routes (the code is reused).

meilisearch/src/routes/indexes/documents.rs
```diff
@@ -127,29 +130,79 @@ pub async fn delete_document(
...
     #[deserr(default, error = DeserrQueryParamError<InvalidDocumentFields>)]
     fields: OptionStarOrList<String>,
+    #[deserr(default, error = DeserrQueryParamError<InvalidSearchFilter>)]
+    filter: Option<String>,
+}
+
+#[derive(Debug, Deserr)]
+#[deserr(error = DeserrJsonError, rename_all = camelCase, deny_unknown_fields)]
+pub struct BrowseQuery {
+    #[deserr(default, error = DeserrJsonError<InvalidDocumentGetOffset>)]
```
dureuill, line 147: I find the error type naming here strange. Why can't we reuse the `InvalidDocumentOffset` error from `BrowseQueryGet`?

meilisearch/tests/documents/errors.rs
```diff
@@ -82,6 +82,110 @@ async fn get_all_documents_bad_limit() {
     "###);
 }
 
+#[actix_rt::test]
+async fn get_all_documents_bad_filter() {
+    let server = Server::new().await;
+    let index = server.index("test");
+
+    // Since the filter can't be parsed automatically by deserr, we have the wrong error message
+    // if the index does not exists
```
dureuill, line 91: What is the expected error message? Why is the one below wrong?

```suggestion
    // if the index does not exist: we could expect to get an error message about the invalid filter before
    // the existence of the index is checked, but it is not the case.
```

going out on a limb as to what the reason might be ☝️

dureuill (commented):
Looks good, just wondering why we are spawning various `InvalidDocumentFooGetError`s instead of reusing the existing `InvalidDocumentFooError`?

meilisearch/tests/documents/errors.rs
irevoire: The expected error messages are the ones we get once the index is created.
The ones right below are « wrong » because deserr can't parse the filter; thus, we have to do it **after** the index existence has been checked instead of doing it while deserializing the payload as we do for every other field.

meilisearch/src/routes/indexes/documents.rs
irevoire: Solved in the team squad

dureuill (commented):
Thank you @irevoire, looks good to me!

irevoire (approved):
very nice code

bors merge

PR #3571, opened 2023-03-08
Introduce two filters to select documents with `null` and empty fields

Kerollmops:
# Pull Request

## Related issue
This PR implements the `X IS NULL`, `X IS NOT NULL`, `X IS EMPTY`, `X IS NOT EMPTY` filters that [this comment](https://github.com/meilisearch/product/discussions/539#discussioncomment-5115884) is describing in a very detailed manner.

### `IS NULL` and `IS NOT NULL`

This PR will be exposed as a prototype for now. Below is the copy/pasted version of a spec that defines this filter.

- `IS NULL` matches fields that `EXISTS` AND `= IS NULL`
- `IS NOT NULL` matches fields that `NOT EXISTS` OR `!= IS NULL`

1. `{"name": "A", "price": null}`
2. `{"name": "A", "price": 10}`
3. `{"name": "A"}`

`price IS NULL` would match 1
`price IS NOT NULL` or `NOT price IS NULL` would match 2,3
`price EXISTS` would match 1, 2
`price NOT EXISTS` or `NOT price EXISTS` would match 3

common query : `(price EXISTS) AND (price IS NOT NULL)` would match 2

### `IS EMPTY` and `IS NOT EMPTY`

- `IS EMPTY` matches Array `[]`, Object `{}`, or String `""` fields that `EXISTS` and are empty
- `IS NOT EMPTY` matches fields that `NOT EXISTS` OR are not empty.

1. `{"name": "A", "tags": null}`
2. `{"name": "A", "tags": [null]}`
3. `{"name": "A", "tags": []}`
4. `{"name": "A", "tags": ["hello","world"]}`
5. `{"name": "A", "tags": [""]}`
6. `{"name": "A"}`
7. `{"name": "A", "tags": {}}`
8. `{"name": "A", "tags": {"t1":"v1"}}`
9. `{"name": "A", "tags": {"t1":""}}`
10. `{"name": "A", "tags": ""}`

`tags IS EMPTY` would match 3,7,10
`tags IS NOT EMPTY` or `NOT tags IS EMPTY` would match 1,2,4,5,6,8,9
`tags IS NULL` would match 1
`tags IS NOT NULL` or `NOT tags IS NULL` would match 2,3,4,5,6,7,8,9,10
`tags EXISTS` would match 1,2,3,4,5,7,8,9,10
`tags NOT EXISTS` or `NOT tags EXISTS` would match 6

common query : `(tags EXISTS) AND (tags IS NOT NULL) AND (tags IS NOT EMPTY)` would match 2,4,5,8,9

## What should the reviewer do?

- Check that I tested the filters
- Check that I deleted the ids of the documents when deleting documents

### DIFF
--- a/milli/src/update/delete_documents.rs
+++ b/milli/src/update/delete_documents.rs
@@ -245,6 +245,7 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
             field_id_docid_facet_strings: _,
             script_language_docids,
             facet_id_exists_docids,
+            facet_id_is_null_docids,
             documents,
         } = self.index;
 
@@ -520,6 +521,13 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
         remove_docids_from_facet_id_exists_docids(
             self.wtxn,
             facet_id_exists_docids,
+            &self.to_delete_docids,
+        )?;
+
+        // We delete the documents ids that are under the facet field id values.
+        remove_docids_from_facet_id_exists_docids(
+            self.wtxn,
+            facet_id_is_null_docids,
             &self.to_delete_docids,
         )?;
 
--- a/milli/src/update/delete_documents.rs
+++ b/milli/src/update/delete_documents.rs
@@ -246,6 +246,7 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
             script_language_docids,
             facet_id_exists_docids,
             facet_id_is_null_docids,
+            facet_id_is_empty_docids,
             documents,
         } = self.index;
 
@@ -528,6 +529,13 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
         remove_docids_from_facet_id_docids(
             self.wtxn,
             facet_id_is_null_docids,
+            &self.to_delete_docids,
+        )?;
+
+        // We delete the documents ids that are under the facet field id values.
+        remove_docids_from_facet_id_docids(
+            self.wtxn,
+            facet_id_is_empty_docids,
             &self.to_delete_docids,
         )?;
 
--- a/milli/src/update/delete_documents.rs
+++ b/milli/src/update/delete_documents.rs
@@ -518,14 +518,14 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
 
         drop(iter);
         // We delete the documents ids that are under the facet field id values.
-        remove_docids_from_facet_id_exists_docids(
+        remove_docids_from_facet_id_docids(
             self.wtxn,
             facet_id_exists_docids,
             &self.to_delete_docids,
         )?;
 
         // We delete the documents ids that are under the facet field id values.
-        remove_docids_from_facet_id_exists_docids(
+        remove_docids_from_facet_id_docids(
             self.wtxn,
             facet_id_is_null_docids,
             &self.to_delete_docids,
@@ -633,7 +633,7 @@ fn remove_docids_from_field_id_docid_facet_value(
     Ok(all_affected_facet_values)
 }
 
-fn remove_docids_from_facet_id_exists_docids<'a, C>(
+fn remove_docids_from_facet_id_docids<'a, C>(
     wtxn: &'a mut heed::RwTxn,
     db: &heed::Database<C, CboRoaringBitmapCodec>,
     to_remove: &RoaringBitmap,
--- a/milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
+++ b/milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
@@ -21,6 +21,7 @@ use crate::{CboRoaringBitmapCodec, DocumentId, FieldId, Result, BEU32, MAX_FACET
 /// Returns the generated grenad reader containing the docid the fid and the orginal value as key
 /// and the normalized value as value extracted from the given chunk of documents.
 #[logging_timer::time]
+#[allow(clippy::type_complexity)]
 pub fn extract_fid_docid_facet_values<R: io::Read + io::Seek>(
     obkv_documents: grenad::Reader<R>,
     indexer: GrenadParameters,
--- a/milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
+++ b/milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
@@ -95,14 +95,14 @@ pub fn extract_fid_docid_facet_values<R: io::Read + io::Seek>(
                         for (normalized, original) in
                             strings.into_iter().filter(|(n, _)| !n.is_empty())
                         {
-                            let normalised_truncated_value: String = normalized
+                            let normalized_truncated_value: String = normalized
                                 .char_indices()
                                 .take_while(|(idx, _)| idx + 4 < MAX_FACET_VALUE_LENGTH)
                                 .map(|(_, c)| c)
                                 .collect();
 
                             key_buffer.truncate(size_of::<FieldId>() + size_of::<DocumentId>());
-                            key_buffer.extend_from_slice(normalised_truncated_value.as_bytes());
+                            key_buffer.extend_from_slice(normalized_truncated_value.as_bytes());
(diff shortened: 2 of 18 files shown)


milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
```diff
@@ -70,33 +73,40 @@ pub fn extract_fid_docid_facet_values<R: io::Read + io::Seek>(
...
+                                fid_docid_facet_numbers_sorter
+                                    .insert(&key_buffer, ().as_bytes())?;
+                            }
+                        }
+
+                        // insert normalized and original facet string in sorter
+                        for (normalized, original) in
+                            strings.into_iter().filter(|(n, _)| !n.is_empty())
+                        {
+                            let normalised_truncated_value: String = normalized
```
dureuill, line 98:
```suggestion
                            let normalized_truncated_value: String = normalized
```

milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
```diff
@@ -21,11 +21,13 @@ use crate::{CboRoaringBitmapCodec, DocumentId, FieldId, Result, BEU32, MAX_FACET
 /// Returns the generated grenad reader containing the docid the fid and the orginal value as key
 /// and the normalized value as value extracted from the given chunk of documents.
 #[logging_timer::time]
+#[allow(clippy::type_complexity)]
```
dureuill, line 24: I wonder if we could maybe return an intermediate struct from this function, given that it returns a tuple of many items of the same type, but the order is semantically important.

It would mean changing the return value to `Result<ExtractedFacetValues>`, where:

```rust
pub struct ExtractedFacetValues {
    pub docid_fid_facet_numbers_chunk: grenad::Reader<File>,
    pub docid_fid_facet_strings_chunk: grenad::Reader<File>,
    pub fid_facet_is_null_docids_chunk: grenad::Reader<File>,
    pub fid_facet_exists_docids_chunk: grenad::Reader<File>,
}
```

milli/src/update/index_documents/extract/mod.rs
```diff
@@ -55,7 +55,7 @@ pub(crate) fn data_from_obkv_documents(
         .collect::<Result<()>>()?;
 
     #[allow(clippy::type_complexity)]
-    let result: Result<(Vec<_>, (Vec<_>, (Vec<_>, Vec<_>)))> = flattened_obkv_chunks
+    let result: Result<(Vec<_>, (Vec<_>, (Vec<_>, (Vec<_>, Vec<_>))))> = flattened_obkv_chunks
```
dureuill, line 58: tbh I'm a bit with clippy on this one 😅 

I'm not sure what we could do though to express the type differently (using a type alias wouldn't make me feel better and would prevent inference from doing its magic)

milli/src/update/delete_documents.rs
```diff
@@ -523,6 +524,13 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
             &self.to_delete_docids,
         )?;
 
+        // We delete the documents ids that are under the facet field id values.
+        remove_docids_from_facet_id_exists_docids(
```
dureuill, line 528: Should we rename this function since it no longer only remove the docids from `exists_docids`?

dureuill (commented):
Thank you for the PR! Looking good (I trust you on the `grenad` stuff as I couldn't get all it is doing, but it is evidently similar to the existing code)

Regarding tests:

- doesn't appear in `milli/src/snapshot_tests`. Similarly we could have a db_snap with `facet_ids_null_docids` in the tests of delete_documents.rs
- in `milli/src/update/index_documents/mod.rs` we have an `index_documents_check_exists_database`, do we want same thing for the new db? `index_documents_check_null_database`?
- I see the tests in search reimplement a way to extract the expected results from the filter, I'm a bit less comfortable than with tests against known and hardcoded expected results since there could be a bug in *that* implementation. It looks correct upon reading it though.

Regarding the correct deletion of the docids it looks similar to the exists db, so if there are no bugs there there are no bugs here.

I have a few suggestions/comments (see below)

milli/src/update/index_documents/extract/mod.rs
irevoire: What? Don’t you like lisp?

milli/src/update/delete_documents.rs
irevoire: I think the same

irevoire (changes requested):
Everything looks good to me except for the name on one method 🙆

milli/src/update/delete_documents.rs
Kerollmops: Fixed, thank you! I just removed the `exists` part from the name 😄

milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
Kerollmops: Thank you, fixed, Louis 🏵️

milli/src/update/index_documents/extract/mod.rs
Kerollmops: I will keep this type for now as I am not sure how to fix that, it is related to the collect and `FromIterator` standard implementation trait.

dureuill (approved):
lovely ❤️

Kerollmops:
I just created a `prototype-null-filter-0` tag, and [the prototype CI is currently running](https://github.com/meilisearch/meilisearch/actions/runs/4403580820).

ahmednfwela:
Thanks for this great PR! I just tested it on dart client and it works perfectly

dureuill (approved):
Code still looks OK 🫶. I wonder if we should maybe add automated tests on the failed parsing of "value NULL", "value NOT NULL", "value IS", "value IS NOT", "value IS EXISTS", etc.

ahmednfwela:
should `IS EMPTY` match empty strings too? if they can be matched already using `""`, can't users just do 
`(tags EXISTS) AND (tags IS NOT NULL) AND (tags IS NOT EMPTY) AND (tags != "")`

Kerollmops:
Hey @ahmednfwela,

> should `IS EMPTY` match empty strings too? if they can be matched already using `""`, can't users just do `(tags EXISTS) AND (tags IS NOT NULL) AND (tags IS NOT EMPTY) AND (tags != "")`

Indeed `IS EMPTY` matches empty strings, empty arrays, and empty objects. Meilisearch doesn't take empty strings into account, empty facets are not possible. The main reason is mostly because we want to search by prefix of facets and the empty one (`""`) will add more problems than anything.

---

For your information, I pushed a new prototype version named `prototype-null-empty-filters-0` in which I added the new `IS EMPTY` filter and [the prototype CI is currently running](https://github.com/meilisearch/meilisearch/actions/runs/4435312673).

I updated the main message accordingly to reflect the new syntax.

dureuill (approved):
> I wonder if we should maybe add automated tests on the failed parsing of "value NULL", "value NOT NULL", "value IS", "value IS NOT", "value IS EXISTS", etc.

Still would be interested in this kind of tests, but otherwise the code still looks good to me after the `IS EMPTY` addition. Thank you.

ahmednfwela:
what do you think about adding filters for field types as well ?
e.g.:
`IS OBJECT`
`IS ARRAY`
`IS STRING`
`IS NUMBER`
`IS BOOL`

curquiza:
Hello @ahmednfwela 
Thanks for the suggestion! To avoid losing track of them, we recommend not putting them into PRs, but more on the [product repository discussion](https://github.com/meilisearch/product/discussions).
You can create a new discussion if has not already been created. You can let us know about your use case 😊

ahmednfwela:
Thanks for informing me, I created the discussion https://github.com/meilisearch/product/discussions/634

irevoire:
Hey @tonivega, we're just checking a few last things on our side. But this feature should be land in the next release :+1:

curquiza:
Hello @tonivega 

To be more transparent, we were expecting more tests and feedback from the users before merging it, to ensure this is what the users want.
Have you tested it?

As @irevoire said, we indeed plan to merge it, so any feedback would be appreciated

irevoire (approved):
bors merge

Issue #3572, opened 2023-03-08
text/csv bulk insert endpoint does not support booleans


samcfinan:
**Describe the bug**
The endpoint to bulk insert documents fails to parse boolean properties when uploading a CSV. The type annotation is appended to the key name and the value is a string.

**To Reproduce**
1. POST `/indexes/:indexUID/documents?primaryKey=id` with the following body and a Content-Type of `text/csv`:
```
"id:number","active:boolean"
899767,true
```
3. Fetch documents via GET `/indexes/:indexUID/documents`
4. Note that the `active` key will be parsed as `active:boolean` with the value as a string.

**Expected behavior**
Expecting Meilisearch to use the boolean type annotation when creating the document.

**Meilisearch version:** [e.g. v1.0.2]

Thanks everyone! Meilisearch is an excellent product.

-------

## Edit by the meilisearch team

Fixed by https://github.com/meilisearch/meilisearch/pull/3576
The associated spec is here; https://github.com/meilisearch/specifications/pull/228

@meilisearch/docs-team Currently, we’re not super clear on the types we can specify in a csv document; https://docs.meilisearch.com/learn/core_concepts/documents.html#csv
With this addition, I guess it could be the occasion to list all the available types (string, number, boolean, null can’t be specified, but will happens if a field is empty)

PR #3576, opened 2023-03-09
Add boolean support for csv documents

irevoire:
Fixes https://github.com/meilisearch/meilisearch/issues/3572

## What does this PR do?
Add support for the boolean types in csv documents.
The type definition is `boolean` and the possible values are
- `true` for true
- `false` for false
- ` ` for null

Here is an example:
```csv
#id,cute:boolean
0,true
1,false
2,
```

### DIFF
--- a/milli/src/documents/builder.rs
+++ b/milli/src/documents/builder.rs
@@ -114,9 +114,9 @@ impl<W: Write> DocumentsBatchBuilder<W> {
                 self.value_buffer.clear();
 
                 let value = &record[*i];
+                let trimmed_value = value.trim();
                 match type_ {
                     AllowedType::Number => {
-                        let trimmed_value = value.trim();
                         if trimmed_value.is_empty() {
                             to_writer(&mut self.value_buffer, &Value::Null)?;
                         } else if let Ok(integer) = trimmed_value.parse::<i64>() {
@@ -137,7 +137,6 @@ impl<W: Write> DocumentsBatchBuilder<W> {
                         }
                     }
                     AllowedType::Boolean => {
-                        let trimmed_value = value.trim();
                         if trimmed_value.is_empty() {
                             to_writer(&mut self.value_buffer, &Value::Null)?;
                         } else {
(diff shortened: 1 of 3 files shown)


milli/src/documents/builder.rs
```diff
@@ -135,6 +136,25 @@ impl<W: Write> DocumentsBatchBuilder<W> {
                             }
                         }
                     }
+                    AllowedType::Boolean => {
+                        let trimmed_value = value.trim();
+                        if trimmed_value.is_empty() {
```
Kerollmops, line 141: Why not trim above the match once for all? 😄

milli/src/documents/builder.rs
```diff
@@ -181,6 +202,7 @@ fn parse_csv_header(header: &str) -> (&str, AllowedType) {
     match header.rsplit_once(':') {
         Some((field_name, field_type)) => match field_type {
             "string" => (field_name, AllowedType::String),
+            "boolean" => (field_name, AllowedType::Boolean),
```
Kerollmops, line 205: It impacts the documentation team, please warn them 😊

Kerollmops (dismissed):
Looks good to me, thank you 🐈‍⬛

milli/src/documents/builder.rs
irevoire: Because we don’t need to trim strings, and since strings are probably the most used branch by far (since everything defaults to string if you forget to type your csv), I feel like it’s best to avoid any useless computation

milli/src/documents/builder.rs
irevoire: Done! I sent them a ping in the issue 👍

milli/src/documents/builder.rs
Kerollmops: Hum, yes, but it doesn't take much time to skip whitespace on the front and the back of a string, and, more importantly, I am pretty sure it will be optimized and reordered by the compiler.

Can we put it above this match? 😃

Kerollmops (approved):
I see you @irevoire 😂
I love that PR now. I am happy 🫶

irevoire:
bors merge

PR #3587, opened 2023-03-13
Enable cache again in test suite CI

curquiza:
Following the change in this PR introduced in v1.1: https://github.com/meilisearch/meilisearch/pull/3422

The cache was removed due to failures (lack of space). Now the binary is smaller (from 250Mb to 50Mb) we want to try to enable the cache again.
Indeed, without the cache step, the CIs are wayyyy slower (45min instead of 20-30min).

For later: Rust 1.68 introduced a new way to fetch crates. Updating the rust version might also help in the future!

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -42,9 +42,8 @@ jobs:
         with:
           toolchain: nightly
           override: true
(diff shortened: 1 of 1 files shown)


dureuill (approved):
Thank you @curquiza 👍

PR #3593, opened 2023-03-16
Cluster

irevoire:
- Make sure the followers are really getting their batches from the leader instead of creating their own
- Make sure the leader is really sending its batch to the follower
- Provide a way to join the cluster and be updated via a dump
- Make the consistency configurable at the task level

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -815,6 +815,8 @@ dependencies = [
  "serde",
  "serde_json",
  "thiserror",
+ "time",
+ "uuid 1.3.0",
 ]
 
 [[package]]
@@ -1232,7 +1234,7 @@ dependencies = [
  "tempfile",
  "thiserror",
  "time",
- "uuid 1.2.2",
+ "uuid 1.3.0",
 ]
 
 [[package]]
@@ -1438,7 +1440,7 @@ dependencies = [
  "faux",
  "tempfile",
  "thiserror",
- "uuid 1.2.2",
+ "uuid 1.3.0",
 ]
 
 [[package]]
@@ -1978,7 +1980,7 @@ dependencies = [
  "tempfile",
  "thiserror",
  "time",
- "uuid 1.2.2",
+ "uuid 1.3.0",
 ]
 
 [[package]]
@@ -2605,7 +2607,7 @@ dependencies = [
  "tokio-stream",
  "toml",
  "urlencoding",
- "uuid 1.2.2",
+ "uuid 1.3.0",
  "vergen",
  "walkdir",
  "yaup",
@@ -2628,7 +2630,7 @@ dependencies = [
  "sha2",
  "thiserror",
  "time",
- "uuid 1.2.2",
+ "uuid 1.3.0",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


Kerollmops:
Hey 👋

We investigated this solution and wanted to look at what we can have and fix with [a raft-based implementation](https://github.com/drmingdrmer/openraft).

> 1️⃣ The TCP connection used between the leader and the followers doesn't have the `keepalive` option enabled. Thus the connections are probably going to die often.

The Raft library could fix it, as we are able to use whatever we want as the network connection protocol. [In the basic example](https://github.com/datafuselabs/openraft/blob/add2f7b1cd6754ca566006d67e81bc9f59258403/examples/raft-kv-memstore/src/network/raft_network_impl.rs#L38-L54) they are using an HTTP connection with reqwest.

> * 1️⃣ Add an internal interface

If you are talking about the public and private IPs in a cluster, it should be possible to have two interfaces with Raft.

> * 1️⃣ The tasks received while joining the cluster might be lost

It should be ok with Raft. I hope. We must investigate that.

> * 1️⃣ Handle what happens when a follower doesn’t get the same result the leader got from processing a batch

We must make sure we can express that a node is in the state. Returning that a node is non-healthy from the `/health` route, maybe?

> * 1️⃣ What happens when the tick function need to re-run (cancel + MaxDatabaseSizeReached). For the potential users reading us, I think the whole cluster might get stuck for ever. If you have a « normal » Linux machine that should never happens though

It should be hard to trigger now that the index size is dynamic.

> * 1️⃣ Currently, the only sync we have is made on the index operation

If you mean that the commit is only made on the index level, not the index-scheduler level, e.g., index swap, we can improve it later. If you mean that the API Keys are not supported, it is just a matter of time and effort.

> * 2️⃣ When we send a task (with its update file) or a dump (to let the followers join the cluster), it must be stored entirely in RAM, which definitely won't scale on a small computer

It can be stored on disk in the future.

> * 2️⃣ The followers are still able to receive writes (tasks or API keys), and I don't exactly know what happens in this case, but it's definitely nothing good

We must refuse direct API calls when not the leader.

> * 3️⃣ We spawn like 200 threads that could all be a super small async rust routine that doesn't costs anything

OpenRaft is async and uses tokio. We could even maybe share the same tokio runtime as the one we use with Actix 😍

> * 3️⃣ It doesn't work on Windows.

It should be ok with OpenRaft. That's a matter of Network and disk operations.

> * Be able to compute the same batches of tasks as the other nodes

We must ensure that it is possible with OpenRaft to define the tasks we can process.

> * Only commit when the quorum is ready to commit (direct consistency, not eventual consistency)

It should be possible, but we must look at what we can do about it.

> * Manage the healthiness of a node based on the number of tasks to catch up

We could look at the number of tasks to process to catch up with the number of tasks processed by the leader and change our healthiness based on this number.

Kerollmops:
Thanks for the huge amount of work, @irevoire, but this PR is very far behind, and [we recently released the replicated sharding feature](https://github.com/meilisearch/meilisearch/releases/tag/v1.37.0#v137-replicated-sharding). Closing this then.

PR #3618, opened 2023-03-29
Use bincode to store the tasks in the task queue

irevoire:
Mitigate but does not fix the task queue issue https://github.com/meilisearch/meilisearch/issues/3622

This is one of the multiple « fixes/mitigation » we can work on to reduce the issue of the task queue being full.

Here are some measures of the impact of moving to bincode
| db size/branch              | main                              | bincode                     | Percentage change |
|-----------------------|------------------------|---------------------|--------------------|
| db size for 100k tasks |               80MiB               |        24MiB                |    -70%                    |
| db size for 1M tasks    |              506MiB              |        100MiB              |    -79%                     |
| time to import 1M tasks |           3m36s                 |         3m37s             |         0%                   |
| db size for 10M tasks    |              4.9GiB               |        987MiB             |        -80%                |
| time to import 10M tasks |           1m05s                 |         55s                 |         0%                   |

As we can see, it roughly divides the size of the task database by a factor of 5.

### DIFF
--- a/index-scheduler/Cargo.toml
+++ b/index-scheduler/Cargo.toml
@@ -12,7 +12,7 @@ license.workspace = true
 
 [dependencies]
 anyhow = "1.0.64"
-bincode = "1.3.3"
+bincode = "2.0.0-rc.2"
 csv = "1.1.6"
 derive_builder = "0.11.2"
 dump = { path = "../dump" }
--- a/index-scheduler/src/utils.rs
+++ b/index-scheduler/src/utils.rs
@@ -1,10 +1,12 @@
 //! Utility functions on the DBs. Mainly getter and setters.
 
+use std::borrow::Cow;
 use std::collections::{BTreeSet, HashSet};
 use std::ops::Bound;
 
+pub use bincode::{Decode, Encode};
 use meilisearch_types::heed::types::{DecodeIgnore, OwnedType};
-use meilisearch_types::heed::{Database, RoTxn, RwTxn};
+use meilisearch_types::heed::{BytesDecode, BytesEncode, Database, RoTxn, RwTxn};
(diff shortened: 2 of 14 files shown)


index-scheduler/Cargo.toml
```diff
@@ -12,7 +12,7 @@ license.workspace = true
 
 [dependencies]
 anyhow = "1.0.64"
-bincode = "1.3.3"
+bincode = "2.0.0-rc.2"
```
Kerollmops, line 15: We will merge this PR once the `rc`s are done and released as a final version. What do you think?
They just published a `rc.3` recently on the 30th of March.

index-scheduler/src/utils.rs
```diff
@@ -572,3 +574,30 @@ pub fn dichotomic_search(start_point: usize, mut is_good: impl FnMut(usize) -> b
         }
     }
 }
+
+pub struct Bincode<T>(std::marker::PhantomData<T>);
```
Kerollmops, line 578: Can you add more documentation, please?

milli/Cargo.toml
```diff
@@ -13,7 +13,8 @@ license.workspace = true
 
 [dependencies]
 bimap = { version = "0.6.2", features = ["serde"] }
-bincode = "1.3.3"
+bincodev1 = { package = "bincode", version = "1.3.3" }
```
Kerollmops, line 16: Why do we need `bincodev1`?

Kerollmops (changes requested):

index-scheduler/src/utils.rs
irevoire: Same, we won’t need this if we wait until bincodev2 is officially released

irevoire:
We’ll wait until bincodev2 is officially released before integrating it into the codebase.

PR #3624, opened 2023-03-29
Reduce the time to import a dump

### LINKED ISSUE #3620: Huge latency when importing a dump due to the task queue
I tried to import a dump containing 1M tasks, and that took 3 minutes on main.
It should be faster.

irevoire:
When importing a dump, this PR does multiple things;
- Stops committing the changes between each task import
- Stop deserializing + serializing every bitmap for every task

Pros:
Importing 1M tasks in a dump went from 3m36 on my computer to 6s

Cons: We use slightly more memory, but since we’re using roaring bitmaps, that really shouldn’t be noticeable.

Fixes #3620

### DIFF
--- a/index-scheduler/src/lib.rs
+++ b/index-scheduler/src/lib.rs
@@ -43,7 +43,7 @@ pub use error::Error;
 use file_store::FileStore;
 use meilisearch_types::error::ResponseError;
 use meilisearch_types::heed::types::{OwnedType, SerdeBincode, SerdeJson, Str};
-use meilisearch_types::heed::{self, Database, Env, RoTxn};
+use meilisearch_types::heed::{self, Database, Env, RoTxn, RwTxn};
 use meilisearch_types::milli;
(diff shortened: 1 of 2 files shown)


Kerollmops (approved):
Wow! What you've done is pretty good @irevoire 🦞

irevoire:
bors merge

dureuill:
> Cons: We use slightly more memory,

could this con be mitigated by committing not between each task import, but once in a while (like every 100_000 tasks or so)?

irevoire:
> could this con be mitigated by committing not between each task import, but once in a while (like every 100_000 tasks or so)?

Yes definitely!
But I highly doubt that a dozen of roaring bitmaps have a real impact on the ram consumption. In the worst-case scenario, it’ll maybe consume like ~1MiB, I would say 🤔 
I could try to measure it tomorrow

dureuill:
I know this is something I used to do in the past as a way to put an upper bound on both RAM consumption and the workload of a single IO operation.

PR #3627, opened 2023-03-30
Search refactor exhaustive distinct hits

Kerollmops:
# Pull Request
This PR implements an exhaustive number of hits for when it is requested and a distinct is set. It can be slow as it will iterate over all of the exhaustive candidates, but the user requested it explicitly. However, we would like to work on a faster algorithm in the future, this will involve product and design questions.

## Related issue
It's about the search refactor.


loiclec:
I manually rebased the PR into the `search-refactor` branch :-) Thanks kero!

Issue #3619, opened 2023-03-29
Huge latency when deleting tasks


irevoire:
We want to improve the speed of deleting the tasks from the task queue of Meilisearch. Currently deleting 10M tasks took about 45 minutes, and deleting 250k tasks takes 15 minutes.

That’s huge, considering how little work there is to do.

https://github.com/meilisearch/meilisearch/blob/31bb61ba99cdf419e91c68dc78fdd10d4a418e1c/index-scheduler/src/batch.rs#L1268-L1338

irevoire:
I tried to reproduce the issue locally and on the cloud and could not reproduce the issue.
Thus, the behavior we observed was probably due to something else.

I’m closing for now.

PR #3630, opened 2023-03-30
improve the task deletion slightly.

irevoire:
Related to https://github.com/meilisearch/meilisearch/issues/3619

Deleting 10M tasks went from 19.8s to 16.0s.
Deleting 1M tasks went from 1.6s to 1.5s.

Before merging, I would like to investigate what else we can improve, but to do that, I would need to generate a more realistic dataset (currently, all my tasks share the same date).

-----

Failed attempt:
Caching the `started_at` and `finished_at` affected datetime on ram went from 16.3s to 17.4s.
With the numbers represented as i64 and using a treemap instead of an hashmap, I was able to fall down to 16.0s.

-----

It brings no real gain to my machine.
I'll wait until I can try this PR on a cloud instance, but I’m probably going to close it.

### DIFF
--- a/index-scheduler/src/batch.rs
+++ b/index-scheduler/src/batch.rs
@@ -17,7 +17,7 @@ tasks individally, but should be much faster since we are only performing
 one indexing operation.
 */
 
-use std::collections::{BTreeSet, HashSet};
+use std::collections::BTreeSet;
 use std::ffi::OsStr;
 use std::fs::{self, File};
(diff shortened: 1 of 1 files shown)


irevoire:
I imported + deleted 1M tasks on a cloud machine, and it took 3s.
Thus this PR is useless. I don’t know what happened last time when it took 15 minutes, but it doesn’t seem linked to the actual task deletion code 😩

PR #3645, opened 2023-04-05
Implement `exactness` ranking rule

dureuill:
# Pull Request

## Related issue
Related to the refactor of the search.

## What does this PR do?
- Add 2 new ranking rules:
  1. `ExactAttribute` ranking rule that returns 3 consecutive buckets:
     1. documents that have an attribute with the exact query as value.
     2. documents that are not in (1) and that have an attribute starting with the exact query as value.
     3. documents that are not in (1) or (2)
  4. `ExactnessGraph` graph ranking rule with 2 kinds of conditions:
     1. term is contained exactly in an attribute, cost 0
     2. unconditional, cost 1

Exactly means: "matches, no synonym, no ngram, no prefix"

⚠️ Unlike the existing implementation and what is currently specified, these new ranking rules do not consider single-word (nor multiple-words, for that matter) synonyms as exact matches.

### DIFF
--- a/milli/src/search/new/ranking_rule_graph/exactness/mod.rs
+++ b/milli/src/search/new/ranking_rule_graph/exactness/mod.rs
@@ -58,6 +58,7 @@ impl RankingRuleGraphTrait for ExactnessGraph {
             docids,
             universe_len: universe.len(),
             start_term_subset: None,
+            // TODO/FIXME:  modify `end_term_subset` to signal to the next ranking rules that the term cannot be removed
             end_term_subset: dest_node.clone(),
         })
     }


milli/src/search/new/exact_attribute.rs
```diff
@@ -0,0 +1,253 @@
...
+                    continue 'words;
+                };
+                // Note: Since the position is stored bucketed in word_position_docids, for queries with a lot of
+                // longer phrases we'll be losing on precision here.
+                let bucketed_position = crate::bucketed_position(position + offset);
+                let word_position_docids = CboRoaringBitmapCodec::bytes_decode(
+                    ctx.get_db_word_position_docids(*word, bucketed_position)?.unwrap_or_default(),
+                )
+                .unwrap_or_default();
+                candidates &= word_position_docids;
```
loiclec, line 166: we could add one more check for `candidates.is_empty()` here

milli/src/search/new/exact_attribute.rs
```diff
@@ -0,0 +1,253 @@
...
+                            .unwrap_or_default())
+                    }),
+            )?;
+            intersection &= &candidates;
+            if !intersection.is_empty() {
+                let candidates_with_exact_word_count = ctx
+                    .index
+                    .field_id_word_count_docids
+                    .get(ctx.txn, &(fid, exact_term_position_ids.len() as u8))?
+                    .unwrap_or_default();
```
loiclec, line 202: in terms of performance, it is not really worth putting this in the DB cache, but it would be good for the sake of consistency

milli/src/search/new/exact_attribute.rs
```diff
@@ -0,0 +1,253 @@
...
+                candidates_per_attribute.push(FieldCandidates {
+                    start_with_exact: intersection,
+                    exact_word_count: candidates_with_exact_word_count,
+                });
```
loiclec, line 206: do we have to store them all in an array, or is there a way to perform the union here already?

milli/src/search/new/exact_attribute.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    fn next_bucket(
+        &mut self,
+        _ctx: &mut SearchContext<'ctx>,
+        _logger: &mut dyn SearchLogger<QueryGraph>,
+        universe: &roaring::RoaringBitmap,
+    ) -> Result<Option<RankingRuleOutput<QueryGraph>>> {
+        let state = std::mem::take(&mut self.state);
+        let (state, output) = State::next(state, universe);
+        self.state = state;
```
loiclec, line 50: Could we crash if the state is not initialised here instead?

milli/src/search/new/exact_attribute.rs
```diff
@@ -0,0 +1,253 @@
...
+                        *term.positions.start(),
+                        *term.term_ids.start(),
+                    ))
+                }
+                QueryNodeData::Deleted | QueryNodeData::Start | QueryNodeData::End => continue,
+            }
+        }
+
+        exact_term_position_ids.sort_by_key(|(_, _, id)| *id);
+        // bail if there is a "hole" (missing word) in remaining query graph
```
loiclec, line 116: IMPORTANT: the `exact_term_position_ids` need to be deduplicated here!

milli/src/search/new/query_term.rs
```diff
@@ -138,7 +139,43 @@ pub struct LocatedQueryTermSubset {
...
+impl ExactTerm {
+    pub fn interned_words<'ctx>(
+        &self,
+        ctx: &'ctx SearchContext<'ctx>,
+    ) -> impl Iterator<Item = Option<Interned<String>>> + 'ctx {
+        match *self {
+            ExactTerm::Phrase(phrase) => {
+                let phrase = ctx.phrase_interner.get(phrase);
+                Either::Left(phrase.words.iter().copied())
+            }
```
loiclec, line 157: could we filter out the `None` here already?

milli/src/search/new/ranking_rule_graph/exactness/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+                (compute_docids(ctx, dest_node, universe)?, dest_node)
+            }
+            ExactnessCondition::Skip(dest_node) => (universe.clone(), dest_node),
+        };
+        Ok(ComputedCondition {
+            docids,
+            universe_len: universe.len(),
+            start_term_subset: None,
+            end_term_subset: dest_node.clone(),
+        })
```
loiclec, line 62: IMPORTANT: the `end_term_subset` here should be modified to signal to the next ranking rules that the term cannot be removed

loiclec (approved):
Thanks Louis! I left a couple of comments. Some are just questions and some are marked `IMPORTANT`. I approved the PR and will let you merge it whenever you're ready :) Could you copy the important comments to `TODOs` in the code?

PR #3647, opened 2023-04-06
Improve the health route by ensuring lmdb is not down

### LINKED ISSUE #3644: Fix the health route
We encountered a bug where all the lmdb databases were stuck on a blocking call.
We can’t reproduce the issue, but since it makes the whole engine useless and we know it exists, we should, at the very least, see if we can access the auth controller before answering on the health route.

irevoire:
Fixes #3644

In this PR, I try to make a small read on the `AuthController` and `IndexScheduler` databases.
The idea is not to validate that everything works but just to avoid the bug we had last time when lmdb was stuck forever.

In order to get access to the `AuthController` without going through the extractor, I need to wrap it in the `Data` type from `actix-web`.
And to do that, I had to patch our extractor so it works with the `Data` type as well.


Kerollmops (approved):
Looks good to me 🛹
Thank you ❤️

irevoire:
bors merge

PR #3659, opened 2023-04-11
stops receiving tasks once the task queue is full

irevoire:
Give 20GiB to the task queue + once 50% of the task queue is used, it blocks itself and only receives task deletion requests to ensure we never get in a state where we can’t do anything.

Also, create a new error message when we reach this case:
```
Meilisearch cannot receive write operations because the size limit of the tasks database has been reached. Please delete tasks to continue performing write operations.
```

### DIFF
--- a/index-scheduler/src/error.rs
+++ b/index-scheduler/src/error.rs
@@ -61,6 +61,8 @@ pub enum Error {
     SwapDuplicateIndexesFound(Vec<String>),
     #[error("Index `{0}` not found.")]
     SwapIndexNotFound(String),
+    #[error("No space left in database. Free some space by deleting tasks.")]
+    NoSpaceLeftInTaskQueue,
     #[error(
         "Indexes {} not found.",
         .0.iter().map(|s| format!("`{}`", s)).collect::<Vec<_>>().join(", ")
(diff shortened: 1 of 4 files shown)


Kerollmops (dismissed):
I approve this PR even if the condition is difficult to read. It is only a small quick fix.

gmourier:
> My only issue is that the we should create a new error code + message because « no space left on device » is not good.

Can we do that in the same PR?

curquiza:
Do we want to merge it once approved? If not, can we convert it to draft?

gmourier:
@irevoire I'm summarizing the proposals for the related error. (Only `code` fields are changing between the 2 proposals).

## A

HTTP Code: `422`

```json
{
    "code": "no_space_left_on_tasks_database",
    "message": "Meilisearch cannot receive write operations because the size limit of the tasks database has been reached. Please delete tasks to continue performing write operations.",
    ...
}

## B 

HTTP Code: `422`

```json
{
    "code": "tasks_database_size_limit_reached",
    "message": "Meilisearch cannot receive write operations because the size limit of the tasks database has been reached. Please delete tasks to continue performing write operations.",
    ...
}

Both use existing error code names convention, e.g https://docs.meilisearch.com/reference/errors/error_codes.html#database-size-limit-reached and  https://docs.meilisearch.com/reference/errors/error_codes.html#no-space-left-on-device

irevoire:
After a meeting with @Kerollmops @gmourier @macraig we decided to:
1. Increase the size of the map to 20GiB just in case
2. Keep the limit at 10GiB
3. Don't create a new error_code to avoid a breaking change on the SDK
4. Improve the error msg as stated by @gmourier

Kerollmops (approved):
Thank you for this first fix @irevoire!
Love it ❤️

bors merge

PR #3665, opened 2023-04-12
Lookup non-derived terms in exact DBs

dureuill:
# Pull Request

## Related issue
Related to the search refactor.

## What does this PR do?
- Introduces a `Word` enum that wraps an `Interned<String>` by adding whether the word is an `Original` word from the user query or a `Derived` one.
- `ngram`, prefixes, words with typos, split words, are all `Derived` words. Words from the query or from a phrase are all `Original` words.
- Introduces `SearchContext::word_docids` and `SearchContext::word_prefix_docids` that accept a `Word` and call either `SearchContext::get_db_word_docids` only or also `SearchContext::get_db_exact_word_docids` depending on if the word is original or derived.
- Update call sites, make `SearchContext::get_db_word_docids` and `SearchContext::get_prefix_word_docids` private.

## PR checklist
- [ ] Do at least some tests with `DisableTypoForAttributes` to check it works correctly.
- [ ] The prefix might be unexpectedly `None` when a term is initialized and we should look at the `exact_word_db`.

### DIFF
--- a/milli/src/search/new/db_cache.rs
+++ b/milli/src/search/new/db_cache.rs
@@ -89,7 +89,7 @@ impl<'ctx> SearchContext<'ctx> {
     }
 
     /// Retrieve or insert the given value in the `word_docids` database.
-    pub fn get_db_word_docids(&mut self, word: Interned<String>) -> Result<Option<RoaringBitmap>> {
+    fn get_db_word_docids(&mut self, word: Interned<String>) -> Result<Option<RoaringBitmap>> {
         DatabaseCache::get_value(
             self.txn,
             word,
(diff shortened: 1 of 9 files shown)


milli/src/search/new/db_cache.rs
```diff
@@ -67,8 +69,28 @@ impl<'ctx> SearchContext<'ctx> {
...
+                    (None, None) => None,
+                    (None, Some(tolerant)) => Some(tolerant),
+                    (Some(exact), None) => Some(exact),
+                    (Some(exact), Some(tolerant)) => {
+                        let mut both = exact;
+                        both |= tolerant;
+                        Some(both)
+                    }
+                })
+            }
```
loiclec, line 87: Using an `unwrap_or_default()` here and then taking the union should be almost exactly as fast, as unions of owned roaring bitmap will only keep the largest one of the two in memory (they won't create a new one).

loiclec (approved):
It looks good, thanks a lot Louis! I have merged it manually but GitHub didn't understand that. So I will close the PR here manually as well :)

PR #3666, opened 2023-04-12
Update README to reference new docs website

guimachiavelli:
With the launch of the new website, we need to update the README so it references the correct URLs.

Two minor details:
- we have removed the contact page from the documentation (it had the same links present in this readme and on the community section of the landing page) 
- we have recently separated filtering and faceted search into two separate articles


curquiza (approved):
thanks @guimachiavelli

bors merge

Issue #3440, opened 2023-01-31
Batch document additions and deletions together


loiclec:
When someone sends many document addition tasks interspersed with document deletion tasks, meilisearch is currently forced to process these tasks serially. For example, if the following tasks are sent in this order:
```
1. document addition
2. document addition
3. document addition
4. document deletion
5. document deletion
6. document addition
```
Then they will be batched as follows:
```
a. (1–3) document additions
b. (4–5) document deletions
c. (6) document addition
```
This can cause significant indexing performance problems, as we rely on incremental indexing speed to keep up with the updates. If the updates are sent quicker than meilisearch can process then, then the task queue will keep growing bigger and bigger.

Ideally, we want to batch all of these tasks into:
```
a. (1–6) document additions and deletions
```

---

## Proposed solution
We could add a new function in `milli` which can accumulate document additions and deletions, respecting the order of the operations. The output of this function should be two things:
1. A set (roaring bitmap) of documents to delete
2. A `Transform` or `TransformOutput` containing the documents to add

Then we can use the existing indexing functions to process (1) and (2) serially. 

---

## impacted team

Can impact @meilisearch/docs-team in some part of the docs talking about auto batching

gmourier:
Thank you, @loiclec! 

In the future, (not saying it will be needed), could we leverage that enhancement to provide an endpoint permitting sending hybrid documents ops akka containing additions (updates/replace) and deletions in the same HTTP request? 

I see it as a friction reducer in the case you receive a batch of ops (from a message queue or a file system) being mixed and they are forced to separate them on their hand before sending them to Meilisearch.

loiclec:
@gmourier  Sure, it's technically possible :) The main challenge would be in making a good API for it.

gmourier:
Thank you @loiclec 

It could be related to https://github.com/meilisearch/product/discussions/554

curquiza:
We will try to be investigated and maybe done during v1, no guarantee however

gillian-meilisearch:
Hello everyone 👋 

We have just released the first RC (release candidate) of Meilisearch containing this fix!
You can test it by using
-  the [release assets](https://github.com/meilisearch/meilisearch/releases)
- the [Meilisearch Docker image](https://hub.docker.com/repository/docker/getmeili/meilisearch/general)

```bash
docker run -it --rm -p 7700:7700 -v $(pwd)/meili_data:/meili_data getmeili/meilisearch:v1.3.0-rc.0
```

[...]

PR #3667, opened 2023-04-12
Disable autobatching of additions and deletions

dureuill:
# Pull Request

## Related issue
Fixes #3664

## What does this PR do?
- Modifies the autobatcher to not batch document additions and deletions, as a workaround to the DB corruption in #3664

PR #3670, opened 2023-04-13
Fix addition deletion bug

irevoire:
The first commit of this PR is a revert of https://github.com/meilisearch/meilisearch/pull/3667. It re-enable the auto-batching of addition and deletion of tasks. No new changes have been introduced outside of `milli`. So all the changes you see on the autobatcher have actually already been reviewed.

It fixes https://github.com/meilisearch/meilisearch/issues/3440.

### What was happening?

The issue was that the `external_documents_ids` generated in the `transform` were used in a very strange way that wasn’t compatible with the deletion of documents.
Instead of doing a clear merge between the external document IDs of the DB and the one returned by the transform + writing it on disk, we were doing some weird tricks with the soft-deleted to avoid writing the fst on disk as much as possible.
The new algorithm may be a bit slower but is way more straightforward and doesn’t change depending on if the soft deletion was used or not. Here is a list of the changes introduced:
1. We now do a clear distinction between the `new_external_documents_ids` coming from the transform and only held on RAM and the `external_documents_ids` coming from the DB.
2. The `new_external_documents_ids` (coming out of the transform) are now represented as an `fst`. We don't need to struggle with the hard, soft distinction + the soft_deleted => That's easier to understand
3. When indexing documents, we merge the `external_documents_ids` coming from the DB and the `new_external_documents_ids` coming from the transform.

### Other things introduced in this  PR

Since we constantly have to write small, very specialized fuzzers for this kind of bug, we decided to push the one used to reproduce this bug.
It's not perfect, but it's easy to improve in the future.
It'll also run for as long as possible on every merge on the main branch.


Kerollmops (changes requested):
Could you make that a test and maybe make it always run :)

.github/workflows/fuzzer-indexing.yml
```diff
@@ -0,0 +1,24 @@
...
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  fuzz:
+    name: Setup the action
+    runs-on: ubuntu-latest
+    timeout-minutes: 4320 # 72h
```
curquiza, line 12: Why do you need to put a time-out? How long should last the CI usually?

irevoire: I must admit that I overlooked that part a little bit :joy: 

While the fuzzer can't find any bugs, it runs forever.
I took the same timeout we had on the benchmark CI but:
- Now I remember that since the benchmarks run on our custom machine, it can run way longer than a normal CI
- Running for 3 days per commit doesn't really make sense

I don't know if it really makes sense to run the fuzzer on each commit, but I think we could run it for ~1h, maybe? :thinking: 
Or run it longer, but only when we make a release?

Kerollmops (approved):
Thank you for the changes, @irevoire, and @loiclec!

This PR looks pretty good, but it could be great to improve the CI/CLI to make it possible to stop the fuzzer after q certain amount of time.

curquiza:
@Kerollmops does this fix this issue?
https://github.com/meilisearch/meilisearch/issues/3440

Can you please add it to the PR if yes?

loiclec:
I made a small change to stop the fuzzer after one hour automatically. @Kerollmops @ManyTheFish

ManyTheFish (approved):

PR #3673, opened 2023-04-13
Handle the task queue being full

dureuill:
# Pull Request

## Related issue
Fixes a remaining issue with #3659 where it was not always possible to send tasks back even after deleting some tasks when prompted.

## Tests

- see integration test
- also manually tested with a 1MiB task queue. Was not possible to become unblocked before this PR, is now possible.

## What does this PR do?
- Use the `non_free_pages_size` method to compute the space occupied by the task db instead of the `real_disk_size` which is not always affected by task deletion.
- Expand the test so that it adds a task after the deletion. The test now fails before this PR and succeeds after this PR.

### DIFF
--- a/meilisearch/tests/tasks/mod.rs
+++ b/meilisearch/tests/tasks/mod.rs
@@ -1062,6 +1062,9 @@ async fn test_task_queue_is_full() {
     }
     "###);
 
+    let result = server.wait_task(result["taskUid"].as_u64().unwrap()).await;
+    snapshot!(json_string!(result["status"]), @r###""succeeded""###);
+
     // Now we should be able to register tasks again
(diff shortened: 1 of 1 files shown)


meilisearch/tests/tasks/mod.rs
```diff
@@ -1062,6 +1062,19 @@ async fn test_task_queue_is_full() {
     }
     "###);
 
+    // But we should still be able to register tasks deletion IF they delete something
```
irevoire, line 1065:
```suggestion
    // Now we should be able to register tasks again
```

irevoire (changes requested):

irevoire (dismissed):
bors merge

irevoire (approved):
bors merge

PR #3687, opened 2023-04-23
Allow to disable specialized tokenizations (again)

jirutka:
In PR #2773, I added the `chinese`, `hebrew`, `japanese` and `thai` feature flags to allow melisearch to be built without huge specialed tokenizations that took up 90% of the melisearch binary size. Unfortunately, due to some recent changes, this doesn't work anymore. The problem lies in excessive use of the `default` feature flag, which infects the dependency graph.

Instead of adding `default-features = false` here and there, it's easier and more future-proof to not declare `default` in `milli` and `meilisearch-types`. I've renamed it to `all-tokenizers`, which also makes it a bit clearer what it's about.


jirutka:
```
    Compiling charabia v0.7.1
error[E0432]: unresolved import `crate::normalizer::greek`
  --> /github/home/.cargo/registry/src/github.com-1ecc6299db9ec823/charabia-0.7.1/src/normalizer/mod.rs:14:24
   |
14 | use crate::normalizer::greek::GreekNormalizer;
   |                        ^^^^^ could not find `greek` in `normalizer`
```

This is already fixed in master (https://github.com/meilisearch/charabia/pull/202), you just need to release a new version of charabia.

Kerollmops (commented):
Thank you @jirutka,

Do you think we can test that Meilisearch is compiled with the right dependencies when compiled with the default features and that it also misses the chineses and Japanese when the `all-tokenizations` feature is messing?

We could probably grep the output of `nm`/`otool` when we compile different version of the binary. Unfortunately, it involves compiling Meilisearch one more time. Isn't there a way to be sure that a dependency is here or not, [using the `cargo tree` command](https://doc.rust-lang.org/cargo/commands/cargo-tree.html)?

Awaiting @ManyTheFish to probably release a new version of Charabia to make this PR compile.

jirutka:
> Isn't there a way to be sure that a dependency is here or not, using the `cargo tree` command?

Yes, you can show the dependency tree, including the information about enabled features, using `cargo tree -f '{p} {f}' -e normal --no-default-features` (`normal` means no dev and build dependencies).

Kerollmops:
Thank you, @jirutka,

What do you think about adding a test in the CI to check that Meilisearch is built without `lindera` (or others) when we build it with your command? It will help make sure that we do not break it once again. That's maybe not enough to check for `lindera` as the specialized tokenizer can be added later, but it would be a first step toward it.

If you agree with me, would you be interested in creating an issue on the Meilisearch repo for us to prioritize that, please?

---

Could you rebase your branch, please? I will review it once again and merge it as-is.

Have a nice day 🍐

ManyTheFish:
Hello @jirutka, Charabia has been updated, everything should compile if you rebase your branch on main.

Kerollmops:
Thank you for your time @jirutka and @ManyTheFish 👏
bors merge

Kerollmops (approved):

Issue #3622, opened 2023-03-29
Circumvent reaching the task DB size limit


curquiza:
Related product team resources: [roadmap card](https://www.notion.so/meilisearch/ef7d86c80ef44c549e2361c70a367723?v=d11906bcef8b4b77815f84b801931c00&p=4b0a96687ea64674be7c9f5b186ceb13&pm=s) (_internal only_)
Related product discussion: None
Related spec: WIP

## Motivation

When Meilisearch reaches the maximum size for the task DB (the hard-coded value of 10GB), Meilisearch cannot accept any new tasks, including tasks deletion. It's even impossible to flush the task history.
In this situation, the users are stuck and cannot do anything.
We need to provide a way to by pass this issue.

## Usage

No solution has been decided yet, needs to be discussed with @macraig and @gmourier.

Refer to the final spec to know the details and the final decisions about the usage.

## TODO

- [x] Fix https://github.com/meilisearch/meilisearch/issues/3644
- [ ] Merge changes into `main`
- [ ] Update the spec
- [ ] Make the @meilisearch/cloud-team test the RC

## Impacted teams

@meilisearch/docs-team -> the new behavior should be described, at least, as an engine limitation.
@meilisearch/cloud-team

curquiza:
This issue (https://github.com/meilisearch/meilisearch/issues/3644) should be fixed as well as part of this current issue solving.

curquiza:
Hello everyone following this issue 👋 

We have just released the first RC (release candidate) of Meilisearch containing this new implementation!
You can this feature it by using
-  the [release assets](https://github.com/meilisearch/meilisearch/releases)
- or the [Meilisearch Docker image](https://hub.docker.com/repository/docker/getmeili/meilisearch/general)
```bash
docker run -it --rm -p 7700:7700 -v $(pwd)/meili_data:/meili_data getmeili/meilisearch:v1.2.0-rc.0
```

You are more than welcome to communicate any feedback about this new implementation in this current issue.
If you encountered any bugs, please report them [here](https://github.com/meilisearch/meilisearch/issues/new?assignees=&labels=&template=bug_report.md&title=).

Thanks in advance for your help and your involvement in Meilisearch ❤️ 

🎉 Official and stable release containing this change will be available on 3rd April 2023

⚠️ RC (release candidates) are not recommended for production

PR #3693, opened 2023-04-25
Implement the auto deletion of tasks

irevoire:
Fixes https://github.com/meilisearch/meilisearch/issues/3622

This PR should be the definite fix for #3622.

It adds a limit (1M) to the maximum number of tasks the task queue can hold.
Once the task queue reaches this limit (1M of tasks are in the task queue, whatever their status is), meilisearch will schedule a task deletion that tries to delete the oldest 100k tasks.
If meilisearch can't delete 100k tasks because some of them are not yet finished, it will delete as many tasks as possible.

Once the limit is reached, you're still able to register new tasks. The engine will only stop you from adding new tasks once [the other hard limit](https://github.com/meilisearch/meilisearch/pull/3659) of 10GiB of tasks is reached (that's between 5M and 15M of tasks depending on your workflow).

-------

Technically;
- We only try to schedule our task deletion when calling the tick function but before creating a new batch. This means we never enqueue a task we're not going to process ~right away.
- If our task deletion doesn't delete anything, we don't enqueue it and log a warn the user that the engine is not working properly


dureuill (changes requested):
Thank you for this PR!

The filter doesn't look correct to me as we can't use limit and from in DELETE tasks today.

dureuill (approved):
Tested working, thanks for taking my feedback into account :-) 

bors merge

PR #3694, opened 2023-04-25
Remove Uffizzi because not used by the team

curquiza:
After discussion with the team, we don't really use Uffizzi and even had issues with it recently: the preview build failing randomly leading to unwanted GitHub notifications + issue to reach the container

Capture d’écran 2023-04-25 à 11 55 39

Thanks for the involvement of Uffizzi team anyway, the tool is just not adapted to our team 😊


Kerollmops (approved):
Thank you 🧣
bors merge

PR #3717, opened 2023-05-02
Geo 2

irevoire:
- Only deserialize the rtree once if needed and keep it in memory until the end of the search
- Make the geo descendent sort fast by using the antipodal coordinate of the queried coord + an ascendant sort


dureuill:
Rebased on `search-refactor` + slight refactor to avoid the `unwrap` of the rtree

PR #3731, opened 2023-05-06
Move comments above keys in config.toml

jirutka:
The current style is very unusual, confusing and breaks compatibility with tools for parsing config files including comments. Everyone writes comments above the items to which they refer (maybe except pythonists), so let's stick to that.


curquiza (approved):
Thanks @jirutka for this PR!

bors merge

Issue #3737, opened 2023-05-09
Add analytics on the get / fetch document route


irevoire:
Related spec: WIP

## Motivation

We didn't have the time to implement the analytics before the end of the sprint on the get documents route.

## TODO

- [ ] Merge
- [ ] Update the spec

## Impacted teams

@meilisearch/docs-team

curquiza:
Fixed by #3738 
Will be released this week in rc1

PR #3738, opened 2023-05-09
Add analytics on the get documents resource

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/3737
Related spec https://github.com/meilisearch/specifications/pull/234

## What does this PR do?
Add the analytics for the following routes:
- `GET` - `/indexes/:uid/documents`
- `GET` - `/indexes/:uid/documents/:doc_id`
- `POST` - `/indexes/:uid/documents/fetch`

These analytics are aggregated between two events:
- `Documents Fetched GET`
- `Documents Fetched POST`

That shares the same payload:
 Property name | Description | Example |
|---------------|-------------|---------|
| `requests.total_received` | Total number of request received in this batch | 325 |
| `per_document_id` | `false` | false |
| `per_filter` | `true` if `POST /indexes/:indexUid/documents/fetch` endpoint was used with a filter in this batch, otherwise `false` | false |
| `pagination.max_limit` | Highest value given for the `limit` parameter in this batch | 60 |
| `pagination.max_offset` | Highest value given for the `offset` parameter in this batch | 1000 |

### DIFF
--- a/meilisearch/src/analytics/segment_analytics.rs
+++ b/meilisearch/src/analytics/segment_analytics.rs
@@ -1218,7 +1218,7 @@ impl DocumentsFetchAggregator {
         self.per_filter |= other.per_filter;
 
         self.max_limit = self.max_limit.max(other.max_limit);
-        self.max_offset |= self.max_offset.max(other.max_offset);
+        self.max_offset = self.max_offset.max(other.max_offset);
     }
 
     pub fn into_event(self, user: &User, event_name: &str) -> Option<Track> {
--- a/meilisearch/src/analytics/segment_analytics.rs
+++ b/meilisearch/src/analytics/segment_analytics.rs
@@ -23,7 +23,9 @@ use tokio::select;
 use tokio::sync::mpsc::{self, Receiver, Sender};
 use uuid::Uuid;
 
-use super::{config_user_id_path, DocumentDeletionKind, MEILISEARCH_CONFIG_PATH};
+use super::{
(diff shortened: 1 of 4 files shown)


meilisearch/src/analytics/segment_analytics.rs
```diff
@@ -1135,3 +1165,73 @@ impl HealthAggregator {
...
+        for user_agent in other.user_agents {
+            self.user_agents.insert(user_agent);
+        }
+
+        self.total_received = self.total_received.saturating_add(other.total_received);
+        self.per_document_id |= other.per_document_id;
+        self.per_filter |= other.per_filter;
+
+        self.max_limit = self.max_limit.max(other.max_limit);
+        self.max_offset |= self.max_offset.max(other.max_offset);
```
dureuill, line 1221: I don't think `|=` makes a lot of sense here.

```suggestion
        self.max_offset = self.max_offset.max(other.max_offset);
```

dureuill (changes requested):
LGTM apart one use of bitwise or I don't understand.

meilisearch/src/analytics/segment_analytics.rs
irevoire: Oops, yes, you’re right!

dureuill (dismissed):
Looks Good To Me

curquiza (changes requested):
Needs to be rebased to `release-v1.2.0` 😇

dureuill (approved):
Still good

irevoire:
bors merge

irevoire:
bors merge

dureuill:
bors merge

PR #3739, opened 2023-05-09
fix: update `payload_too_large` error message to include human readable maximum acceptable payload size

### LINKED ISSUE #3736: Include current payload size limit in the payload_too_large error
# Description
Following a discussion we had [in the cloud repository](https://github.com/meilisearch/meilisearch-cloud/issues/2466), we would like to start using the `MEILI_HTTP_PAYLOAD_SIZE_LIMIT` [environment variable](https://www.meilisearch.com/docs/learn/configuration/instance_options#payload-limit-size) and would need a way to provide to the user the current payload size limit [in the error returned by Meilisearch](https://github.com/meilisearch/specifications/blob/main/text/0061-error-format-and-definitions.md#payload_too_large) `payload_too_large`.

# Expected result

It should be a `413 Payload Too Large` from [what is specified in the specs.](https://github.com/meilisearch/specifications/blob/main/text/0061-error-format-and-definitions.md#payload_too_large)

Update the message to specify the current payload size limit. Example:
```json
{
    "message": "The provided payload reached the size limit. The maximum accepted payload size is :payloadSizeInHumanReadableFormat",
    "code": "payload_too_large",
    "type": "invalid_request",
    "link": "https://docs.meilisearch.com/errors#payload_too_large"
}
```

cymruu:
# Pull Request

## Related issue
Fixes #3736 

## What does this PR do?
- update `payload_too_large` error message as requested in ticket

## PR checklist

### DIFF
--- a/meilisearch/src/extractors/payload.rs
+++ b/meilisearch/src/extractors/payload.rs
@@ -59,7 +59,9 @@ impl Stream for Payload {
                         self.limit = new_limit;
                         Poll::Ready(Some(Ok(bytes)))
                     }
-                    None => Poll::Ready(Some(Err(MeilisearchHttpError::PayloadTooLarge))),
+                    None => {
+                        Poll::Ready(Some(Err(MeilisearchHttpError::PayloadTooLarge(self.limit))))
+                    }
                 },
                 x => Poll::Ready(Some(x.map_err(MeilisearchHttpError::from))),
             },
--- a/meilisearch/src/extractors/payload.rs
+++ b/meilisearch/src/extractors/payload.rs
@@ -11,6 +11,7 @@ use crate::error::MeilisearchHttpError;
 pub struct Payload {
     payload: Decompress<dev::Payload>,
     limit: usize,
+    remaining: usize,
 }
 
 pub struct PayloadConfig {
@@ -43,6 +44,7 @@ impl FromRequest for Payload {
         ready(Ok(Payload {
             payload: Decompress::from_headers(payload.take(), req.headers()),
             limit,
+            remaining: limit,
         }))
     }
 }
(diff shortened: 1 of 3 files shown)


meilisearch/src/extractors/payload.rs
```diff
@@ -59,7 +59,9 @@ impl Stream for Payload {
                         self.limit = new_limit;
```
Kerollmops, line 59: Unfortunately, as you can see here we decrease the limit that we previously stored in the `Payload` struct. What you'll have to do is probably to rename the `limit` field into `current_limit` and decrement it. But also introduce a new limit field that is never modified and that you'll have to use in the error variant (as you have already done).

Kerollmops: As an example of something that doesn't work:

```bash
cargo run --release -- --http-payload-size-limit 7MiB

# send the movies.json dataset that is 9.5MiB
# https://github.com/ducaale/xh
cat ~/Documents/data/movies.json | xh 'localhost:7700/indexes/products-blable/documents' 'content-type:application/json'
```

As the client is sending the body into multiple blocks, the error message is not correct:

```json
{
    "message": "The provided payload reached the size limit. The maximum accepted payload size is 42.27 KiB.",
    "code": "payload_too_large",
    "type": "invalid_request",
    "link": "https://docs.meilisearch.com/errors#payload_too_large"
}
```

Kerollmops (changes requested):
Thank you very much for your work. There is a little issue with your changes. If you have time, could you please add a test that sends a body in multiple blocks to ensure we don't reproduce the same mistake?

Have a nice day 🪩

curquiza:
@Kerollmops and @cymruu 
Is it possible for you to rebase this PR to make it point to `release-v1.2.0`? We would love to integrate this into v1.2.0 release, if you agree @Kerollmops, I mean, if it's not too risky

meilisearch/src/extractors/payload.rs
cymruu: Thank you for your feedback and detailed tips on how to improve the code. 

I'd love to help with the test that sends data in blocks but I'll require assistance with that.  

From my observations when the documents are inserted with `curl`/`xh`, the `poll_next` reads the data chunk by chunk, however if test `index.add_documents` is used the stream is consumed immediately. I thought it might be caused by `actix::test` magic, but dunno.
Do you mind explaining me why that happens?

Kerollmops: It looks indeed more complex than I thought. Adding a test like that will require going quite deep into the `actix::test` libraries. Let's keep your PR that way and merge it 😃

Kerollmops:
Thank you very much for your work and time @cymruu, A lot of Meilisearch users will thank you for your work. Let's merge it now. It will be part of v1.2.0 🎉 Have a nice day!

bors merge

Kerollmops (approved):
bors merge

PR #3742, opened 2023-05-10
Compute split words derivations of terms that don't accept typos

loiclec:
Allows looking for the split-word derivation for short words in the user's query (like `the -> "t he"` or `door -> do or`) as well as for 3grams.


curquiza:
Hello @loiclec 
Is it a bug introduced by the search refactoring or a bug that was already present in v1.1?

ManyTheFish:
Hello @curquiza, this bug has been introduced with the refactor, and should be fix before the final release

PR #3757, opened 2023-05-16
Adjust the cost of edges in the `position` ranking rule by bucketing positions more aggressively

loiclec:
This PR significantly improves the performance of the `position` ranking rule when:
1. a query contains many words
2. the `position` ranking rule needs to be called many times
3. the score of the documents according to `position` is high

These conditions greatly increase:
1. the number of edge traversals that are needed to find a valid path from the `start` node to the `end` node
2. the number of edges that need to be deleted from the graph, and therefore the number of times that we need to recompute all the possible costs from START to END

As a result, a majority of the search time is spent in `visit_condition`, `visit_node`, and `update_all_costs_before_node`. This is frustrating because it often happens when the "universe" given to the rule consists of only a handful of document ids.

By limiting the number of possible edges between two nodes from `20` to `10`, we:
1. reduce the number of possible costs from START to END
2. reduce the number of edges that will be deleted 
3. make it faster to update the costs after deleting an edge
4. reduce the number of buckets that need to be computed

In terms of relevancy, I don't think we lose or gain much. We still prefer terms that are in a lower positions, with decreasing precision as we go further. The previous choice of bucketing wasn't chosen in a principled way, and neither is this one. They both "feel" right to me.


dureuill:
to give an idea, with about 500 words on a single page, before we would differentiate up to 16 pages of single-spaced text. Now `position` see everything beyond 2 pages as "very far".

It would be great to benchmark this for relevancy and see if it has a significant impact.

ManyTheFish (approved):
> Might be interesting to have in RC1 so we can benchmark the relevancy changes?

I agree with @dureuill, I'll run a benchmark on rc1 and we will watch the relevancy changes

loiclec:
bors merge

PR #3758, opened 2023-05-16
Test the new introduced routes

irevoire:
I realized we forgot to test the auth of the new fetch & delete documents route.

I also noticed that we forgot to test the `delete-batch` route introduced a long time ago; thus, I added it to the list of routes to test.

------

Since this doesn't impact the code/features but makes us more confident in how the routes work, I would like to merge this in the `release-v1.2` branch. If we have to fix a bug and make a patch release later, I'll be way more confident with these tests in the release.


dureuill:
FYI the commit of this PR was also in https://github.com/meilisearch/meilisearch/pull/3759/commits, that I `bors merge`d

irevoire:
Oooh my bad, well I guess we can close this one then

Issue #3785, opened 2023-05-28
Meilisearch panic at search time


irevoire:
**Describe the bug**
When I send the following totally legit request to meilisearch:
```json
{
  "q":"\u001b[1m"
}
```

It returns:
```json
{
  "message": "task 1983 panicked",
  "code": "internal",
  "type": "internal",
  "link": "https://docs.meilisearch.com/errors#internal"
}
```

Here is what I get in the logs:
```log
thread 'tokio-runtime-worker' panicked at 'attempt to add with overflow', milli/src/search/new/query_term/parse_query.rs:82:25
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9ecda8de85ce893cc3fc748ab06be0b8250147a7/library/std/src/panicking.rs:578:5
   1: core::panicking::panic_fmt
             at /rustc/9ecda8de85ce893cc3fc748ab06be0b8250147a7/library/core/src/panicking.rs:67:14
   2: core::panicking::panic
             at /rustc/9ecda8de85ce893cc3fc748ab06be0b8250147a7/library/core/src/panicking.rs:117:5
   3: milli::search::new::query_term::parse_query::located_query_terms_from_tokens
             at ./milli/src/search/new/query_term/parse_query.rs:82:25
   4: milli::search::new::execute_search
             at ./milli/src/search/new/mod.rs:390:27
   5: milli::search::Search::execute
             at ./milli/src/search/mod.rs:106:13
   6: meilisearch::search::perform_search
             at ./meilisearch/src/search.rs:323:81
   7: meilisearch::routes::indexes::search::search_with_post::{{closure}}::{{closure}}
             at ./meilisearch/src/routes/indexes/search.rs:180:61
   8: <tokio::runtime::blocking::task::BlockingTask<T> as core::future::future::Future>::poll
             at /home/irevoire/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.27.0/src/runtime/blocking/task.rs:42:21
   9: tokio::runtime::task::core::Core<T,S>::poll::{{closure}}
             at /home/irevoire/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.27.0/src/runtime/task/core.rs:223:17
  10: tokio::loom::std::unsafe_cell::UnsafeCell<T>::with_mut
[...]

PR #3786, opened 2023-05-29
Consistently use wrapping add to avoid overflow in debug when query s…

dureuill:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/3785

## What does this PR do?
- Some of the code paths would erroneously use the default addition operator that has the semantics that "overflow is an error, checked at runtime in debug" instead of the intended "overflow is expected" semantics that this code use (this code is using `u16::MAX` as a sentinel). This PR makes it so the wrapping add operator is used everywhere.


milli/src/search/new/query_term/parse_query.rs
```diff
@@ -79,10 +79,10 @@ pub fn located_query_terms_from_tokens(
             TokenKind::Separator(separator_kind) => {
                 match separator_kind {
                     SeparatorKind::Hard => {
-                        position += 1;
+                        position = position.wrapping_add(1);
                     }
                     SeparatorKind::Soft => {
-                        position += 0;
+                        position = position.wrapping_add(0); // TODO: this is a no-op. Consider removing?
```
ManyTheFish, line 85: don't you want to do it in this PR? By replacing it with an `if let ...`?

ManyTheFish (approved):

ManyTheFish:
@dureuill could we add a test ensuring that the bug is fixed?

milli/src/search/new/query_term/parse_query.rs
dureuill: Changed to a `if let ...`

dureuill:
@ManyTheFish 

> could we add a test ensuring that the bug is fixed?

Added a test that panics before this PR with the issue of https://github.com/meilisearch/meilisearch/issues/3785, and passes with the changes introduced by this PR.

ManyTheFish (approved):
LGTM!

dureuill:
Tested on ` "q": "\u001b[1m"`, no longer panics!

bors merge

PR #3789, opened 2023-05-30
Improve the metrics

irevoire:
# Pull Request

## Related issue
Implements https://github.com/meilisearch/meilisearch/issues/3790
Associated specification: https://github.com/meilisearch/specifications/pull/242

## Be cautious; it's DB-breaking :scream: 

While reviewing and after merging this PR, be cautious; if you already have a `data.ms` and run meilisearch with this code on it, it won't work because we need to cache a new information on the index stats (that are backed up on disk). You'll get internal errors.

### About the breaking-change label

We only break the API of the metrics route, which does not pose any problem since it's experimental.

## What does this PR do?
- Create a method to get the « facet distribution » of the task queue.
- Prefix all the metrics by `meilisearch_`
- Add the real database size used by meilisearch
- Add metrics on the task queue
- Update the grafana dashboard to these new changes
- Move the dashboard to the `assets` directory
- Provide a new prometheus file to scrape meilisearch easily


dureuill:
> We only break the API of the metrics route, which does not pose any problem since it's experimental.

Maybe something to keep in mind for the changelogs. Having a "Breaking changes to experimental features" section could be a nice touch.

index-scheduler/src/index_mapper/mod.rs
```diff
@@ -90,6 +90,8 @@ pub struct IndexStats {
     pub number_of_documents: u64,
     /// Size of the index' DB, in bytes.
     pub database_size: u64,
+    /// Size of the index' DB, in bytes.
```
dureuill, line 93: We probably want to update either this comment, the above one, or both, so that we can differentiate between the two values.

```suggestion
    /// Size taken by the used pages of the index' DB, in bytes.
    ///
    /// As the DB backend does not return to the disk the pages that are not currently used by the DB,
    /// this value is typically smaller than `database_size`.
```

meilisearch/src/routes/metrics.rs
```diff
@@ -17,7 +17,7 @@ pub fn configure(config: &mut web::ServiceConfig) {
 
 pub async fn get_metrics(
     index_scheduler: GuardedData<ActionPolicy<{ actions::METRICS_GET }>, Data<IndexScheduler>>,
-    auth_controller: GuardedData<ActionPolicy<{ actions::METRICS_GET }>, Data<AuthController>>,
+    auth_controller: Data<AuthController>,
```
dureuill, line 20: why are we no longer guarding the data behind the authentication? was that redundant with guarding the index scheduler?

dureuill (commented):
good from what I can tell

i have a few comments, mostly some requests for additional documentation

PR #3794, opened 2023-05-30
Add `publish-post-release.yml` CI workflow

### LINKED ISSUE #3146: Homebrew CI: run only after docker image and binaries are published
To avoid this [kind of issue](https://github.com/meilisearch/meilisearch/issues/3139), when we publish a [GitHub release of Meilisearch](https://github.com/meilisearch/meilisearch/issues/3139), we should ensure the [Homebrew workflow](https://github.com/meilisearch/meilisearch/blob/main/.github/workflows/publish-deb-brew-pkg.yml) runs only if the following workflows have passed:
- [docker image workflow](https://github.com/meilisearch/meilisearch/blob/main/.github/workflows/publish-docker-images.yml)
- [binaries workflow](https://github.com/meilisearch/meilisearch/blob/main/.github/workflows/publish-binaries.yml)

The best would be to create dependencies btw worklows.
Looks like it's possible following [this Stackoverflow answer](https://stackoverflow.com/questions/58457140/dependencies-between-workflows-on-github-actions), by using [`workflow_run`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run)

## EDIT
Not really possible to only add `workflow_run` because our Docker and binaries CIs also run when we are publishing pre-release, and the Homebrew CI should not run in this case despite Docker and binaries CIs succeeded
I realized it's not that easy to do, it would involve a whole refactoring of our CIs, if I'm correct.
I would love to have some help/insight on this

To give more information to the community
- [docker image workflow](https://github.com/meilisearch/meilisearch/blob/main/.github/workflows/publish-docker-images.yml) should run when a tag is created (not a GitHub release, a git tag!)
- [binaries workflow](https://github.com/meilisearch/meilisearch/blob/main/.github/workflows/publish-binaries.yml) should run when a GitHub release or a GitHub pre-release is created
- [Homebrew workflow](https://github.com/meilisearch/meilisearch/blob/main/.github/workflows/publish-apt-brew-pkg.yml) should only run when a GitHub release is created

Thanks in advance for any help 🙏

vvv:
New CI workflow is triggered when a release is published, or a pre-release
is changed to a released. It can also be triggered manually.
`publish-post-release.yml` calls `publish-binaries.yml` and
`publish-docker-images.yml` --- they are made reusable for this purpose.
Once both workflows succeed, `publish-apt-brew-pkg.yml` is called.

Closes #3146

### Other changes

- Remove stray `matrix` reference in `.github/workflows/sdks-tests.yml`.

### DIFF
--- a/.github/workflows/latest-git-tag.yml
+++ b/.github/workflows/latest-git-tag.yml
@@ -1,9 +1,9 @@
 # Create or update a latest git tag when releasing a stable version of Meilisearch
 name: Update latest git tag
 on:
-  workflow_dispatch:
   release:
     types: [released]
+  workflow_dispatch:
 
 jobs:
   check-version:
--- a/.github/workflows/publish-apt-brew-pkg.yml
+++ b/.github/workflows/publish-apt-brew-pkg.yml
@@ -14,7 +14,7 @@ jobs:
         run: bash .github/scripts/check-release.sh
 
   debian:
-    name: Publish debian packagge
+    name: Publish debian package
     runs-on: ubuntu-latest
     needs: check-version
     container:
--- a/.github/workflows/publish-apt-brew-pkg.yml
+++ b/.github/workflows/publish-apt-brew-pkg.yml
@@ -1,8 +1,9 @@
 name: Publish to APT & Homebrew
 
 on:
-  release:
-    types: [released]
+  # This workflow is reusable, i.e. it may be called from another workflow.
+  # See https://docs.github.com/en/actions/using-workflows/reusing-workflows
+  workflow_call:
 
 jobs:
   check-version:
@@ -14,7 +15,7 @@ jobs:
         run: bash .github/scripts/check-release.sh
 
   debian:
-    name: Publish debian package
+    name: Publish Debian package
     runs-on: ubuntu-latest
     needs: check-version
     container:
@@ -41,7 +42,7 @@ jobs:
         file: target/debian/meilisearch.deb
         asset_name: meilisearch.deb
         tag: ${{ github.ref }}
-    - name: Upload debian pkg to apt repository
+    - name: Upload Debian package to apt repository
       run: curl -F package=@target/debian/meilisearch.deb https://${{ secrets.GEMFURY_PUSH_TOKEN }}@push.fury.io/meilisearch/
 
   homebrew:
--- a/.github/workflows/publish-binaries.yml
+++ b/.github/workflows/publish-binaries.yml
@@ -1,11 +1,14 @@
 name: Publish binaries to GitHub release
 
 on:
-  workflow_dispatch:
   schedule:
     - cron: '0 2 * * *' # Every day at 2:00am
   release:
     types: [published]
+  # This workflow is reusable, i.e. it may be called from another workflow.
+  # See https://docs.github.com/en/actions/using-workflows/reusing-workflows
+  workflow_call:
+  workflow_dispatch:
 
 jobs:
   check-version:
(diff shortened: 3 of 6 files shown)


.github/scripts/check-release.sh
```diff
@@ -1,24 +1,31 @@
...
 
-lock_file='Cargo.lock'
-lock_tag=$(grep -A 1 'name = "meilisearch-auth"' $lock_file | grep version | cut -d '=' -f 2 | tr -d '"' | tr -d ' ')
-check_tag $current_tag $lock_tag $lock_file
+lock_tag=$(grep -A 1 '^name = "meilisearch-auth"' Cargo.lock | read_version)
+check_tag $current_tag $lock_tag Cargo.lock || ret=1
 
-if [[ "$ret" -eq 0 ]] ; then
-  echo 'OK'
+if (( ret == 0 )); then
```
vvv, line 28: Since we are using Bash anyway, there is no reason not to use its [syntactic conveniences](https://www.man7.org/linux/man-pages/man1/bash.1.html#ARITHMETIC_EVALUATION).

.github/scripts/check-release.sh
```diff
@@ -1,24 +1,31 @@
-#!/bin/bash
+#!/usr/bin/env bash
+set -eu -o pipefail
 
-# check_tag $current_tag $file_tag $file_name
-function check_tag {
-  if [[ "$1" != "$2" ]]; then
-      echo "Error: the current tag does not match the version in Cargo.toml: found $2 - expected $1"
```
vvv, line 6: This function also works with `Cargo.lock`, so the original error message may be misleading.

curquiza (changes requested):
Hello @vvv 

just a quick review before a deeper one: can you put the changes related to the `check-release.sh` script in another PR?

Thanks again for the PR
It's not related to the purpose of this one, if I understand correctly.

vvv:
> can you put the changes related to the `check-release.sh` script in another PR?
> \[...\] It's not related to the purpose of this one, if I understand correctly.

@curquiza Sure. Do I need to create an issue for that, or a PR will suffice?

curquiza:
Hello @vvv 
Thanks for being around 😁
I really appreciate your work but the impact of this PR is huge and not my priority so far (the current CIs are working ok so not at top of my emergencies): I do my best to review and make a decision about it by the end of June but cannot guarantee anything.
Thanks again for your involvement. Our team is small and we try to prioritize and focus on the most impactful work without giving up the community at the same time, but not always easy 😊

vvv:
@curquiza Understood. Thanks for the explanation.

However small Meilisearch team is, you folks are doing sterling work at prioritizing and focusing! The product looks great, the documentation is a beaut, and my contribution experience has been nothing but pleasant so far. ❤️

curquiza:
Hello @vvv 
Thanks again for your PR
I'm going to be transparent: I don't feel confident merging this. Not because of the quality of your work, or anything, but because it will change something that already works, by making it more complex (even if it's just a little bit) to fix a small issue we have.
I'm afraid if it does not work immediately (it never does in the CI world 😂) I have to spend time trying to understand what is happening and fixing this. A time I currently don't have. I realize it's really convenient I currently know how our CIs work, including in edge-case situations. It would not be strategic to lose this, especially because we will probably change our workflow in the future, and I have to master our current CI process to adapt it to our new decisions.

In any case, I keep your solution in mind: I get your suggestions, I just don't know how it works in every situation, and I'm sure I will be able to refer to it in the future.

Sorry to close this PR, and thank you again for your time. It's definitely not lost ❤️

vvv:
@curquiza No problem. I trust your judgment. Cheers!

PR #3799, opened 2023-05-31
Fix error messages in `check-release.sh`

vvv:
- `check_tag`: Report file name correctly. Use named variables.
- Introduce `read_version` helper function. Simplify the implementation.
- Show meaningful error message if `GITHUB_REF` is not set or its format is incorrect.

### DIFF
--- a/.github/scripts/check-release.sh
+++ b/.github/scripts/check-release.sh
@@ -1,24 +1,41 @@
-#!/bin/bash
+#!/usr/bin/env bash
+set -eu -o pipefail
 
-# check_tag $current_tag $file_tag $file_name
-function check_tag {
-  if [[ "$1" != "$2" ]]; then
-      echo "Error: the current tag does not match the version in Cargo.toml: found $2 - expected $1"
-      ret=1
-  fi
+check_tag() {
+    local expected=$1
+    local actual=$2
+    local filename=$3
+
+    if [[ $actual != $expected ]]; then
+        echo >&2 "Error: the current tag does not match the version in $filename: found $actual, expected $expected"
+        return 1
+    fi
 }
 
+read_version() {
+    grep '^version = ' | cut -d \" -f 2
+}
+
+if [[ -z "${GITHUB_REF:-}" ]]; then
+    echo >&2 "Error: GITHUB_REF is not set"
+    exit 1
+fi
+
+if [[ ! "$GITHUB_REF" =~ ^refs/tags/v[0-9]+\.[0-9]+\.[0-9]+(-[a-z0-9]+)?$ ]]; then
+    echo >&2 "Error: GITHUB_REF is not a valid tag: $GITHUB_REF"
+    exit 1
+fi
+
+current_tag=${GITHUB_REF#refs/tags/v}
 ret=0
-current_tag=${GITHUB_REF#'refs/tags/v'}
 
-file_tag="$(grep '^version = ' Cargo.toml | cut -d '=' -f 2 | tr -d '"' | tr -d ' ')"
-check_tag $current_tag $file_tag
+toml_tag="$(cat Cargo.toml | read_version)"
+check_tag "$current_tag" "$toml_tag" Cargo.toml || ret=1
 
-lock_file='Cargo.lock'
-lock_tag=$(grep -A 1 'name = "meilisearch-auth"' $lock_file | grep version | cut -d '=' -f 2 | tr -d '"' | tr -d ' ')
-check_tag $current_tag $lock_tag $lock_file
+lock_tag=$(grep -A 1 '^name = "meilisearch-auth"' Cargo.lock | read_version)
+check_tag "$current_tag" "$lock_tag" Cargo.lock || ret=1
 
-if [[ "$ret" -eq 0 ]] ; then
-  echo 'OK'
+if (( ret == 0 )); then
+    echo 'OK'
 fi
 exit $ret


.github/scripts/check-release.sh
```diff
@@ -1,24 +1,31 @@
...
+    grep '^version = ' | cut -d \" -f 2
 }
 
 ret=0
 current_tag=${GITHUB_REF#'refs/tags/v'}
 
-file_tag="$(grep '^version = ' Cargo.toml | cut -d '=' -f 2 | tr -d '"' | tr -d ' ')"
-check_tag $current_tag $file_tag
+file_tag="$(cat Cargo.toml | read_version)"
+check_tag $current_tag $file_tag Cargo.toml || ret=1
```
vvv, line 23: Explicit is better than implicit.[©](https://peps.python.org/pep-0020/) That's why we set `ret` here instead of letting it be side-affected by `check_tag`.

.github/scripts/check-release.sh
```diff
@@ -1,24 +1,31 @@
...
+}
+
+read_version() {
+    grep '^version = ' | cut -d \" -f 2
 }
 
 ret=0
 current_tag=${GITHUB_REF#'refs/tags/v'}
 
-file_tag="$(grep '^version = ' Cargo.toml | cut -d '=' -f 2 | tr -d '"' | tr -d ' ')"
```
vvv, line 14: `cut -d '=' -f 2 | tr -d '"' | tr -d ' '` can be simplified to `cut -d \" -f 2` in this particular case.

.github/scripts/check-release.sh
```diff
@@ -1,24 +1,31 @@
-#!/bin/bash
+#!/usr/bin/env bash
+set -eu -o pipefail
 
-# check_tag $current_tag $file_tag $file_name
-function check_tag {
-  if [[ "$1" != "$2" ]]; then
-      echo "Error: the current tag does not match the version in Cargo.toml: found $2 - expected $1"
```
vvv, line 6: This function also works with `Cargo.lock`. The original error message would be misleading in this case.

.github/scripts/check-release.sh
```diff
@@ -1,24 +1,31 @@
-#!/bin/bash
+#!/usr/bin/env bash
+set -eu -o pipefail
```
irevoire, line 2: nice idea

irevoire (changes requested):
Hey @vvv.
There are a lot of good ideas here, but it doesn’t work anymore. Try to run the script before and after your PR, and you’ll see.

On my computer here are the results;
```
irevoire@macbook-pro-4 ~/clean/MeiliSearch 🌱 % bash .github/scripts/check-release.sh
Error: the current tag does not match the version in Cargo.toml: found  - expected 1.2.0
Error: the current tag does not match the version in Cargo.toml: found Cargo.lock - expected 1.2.0
1 irevoire@macbook-pro-4 ~/clean/MeiliSearch 🍐 % gh pr checkout 3799
Switched to branch 'fix.check-release-script'
irevoire@macbook-pro-4 ~/clean/MeiliSearch 🍴 % bash .github/scripts/check-release.sh
.github/scripts/check-release.sh: line 7: $3: unbound variable
```

vvv:
Hey @irevoire,

Thanks for looking at the patch. _(And nice shell prompt! 🧑🏼‍🌾 Pretty organic.)_

```
irevoire@macbook-pro-4 ~/clean/MeiliSearch 🌱 % bash .github/scripts/check-release.sh
Error: the current tag does not match the version in Cargo.toml: found  - expected 1.2.0
Error: the current tag does not match the version in Cargo.toml: found Cargo.lock - expected 1.2.0
1 irevoire@macbook-pro-4 ~/clean/MeiliSearch 🍐 % gh pr checkout 3799
Switched to branch 'fix.check-release-script'
irevoire@macbook-pro-4 ~/clean/MeiliSearch 🍴 % bash .github/scripts/check-release.sh
.github/scripts/check-release.sh: line 7: $3: unbound variable
```

`check-release.sh` is supposed to run in CI environment, where `GITHUB_REF` variable is set.

When you ran the script on your laptop, [`GITHUB_REF`](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) variable wasn't set, so `current_tag` [got evaluated](https://github.com/vvv/meilisearch/blob/2a903a0e6e8a61e4d593d958e0caa50dff693102/.github/scripts/check-release.sh#L20-L23) to an empty string.

```bash
current_tag=${GITHUB_REF#'refs/tags/v'}

file_tag="$(cat Cargo.toml | read_version)"
check_tag $current_tag $file_tag Cargo.toml || ret=1
```

`check_tag` command doesn't recognize unquoted empty string as a separate argument due to the way parameter expansion works in bash. `check_tag` sees only 2 arguments instead of 3 and complains:

```
❯ bash .github/scripts/check-release.sh || echo FAIL $?
.github/scripts/check-release.sh: line 7: $3: unbound variable
FAIL 1
```

When CI runs the script, `GITHUB_REF` is set to a string that (supposedly) starts with `refs/tags/` prefix. In this case the script performs its duties properly:

```
# Happy case
❯ GITHUB_REF=refs/tags/v1.2.0 bash .github/scripts/check-release.sh || echo FAIL $?
OK

# Wrong tag
❯ GITHUB_REF=refs/tags/v8.8.8 bash .github/scripts/check-release.sh
Error: the current tag does not match the version in Cargo.toml: found 1.2.0, expected 8.8.8
Error: the current tag does not match the version in Cargo.lock: found 1.2.0, expected 8.8.8

# Invalid tag
❯ GITHUB_REF=refs/tags/bad-tag bash .github/scripts/check-release.sh
Error: the current tag does not match the version in Cargo.toml: found 1.2.0, expected refs/tags/bad-tag
Error: the current tag does not match the version in Cargo.lock: found 1.2.0, expected refs/tags/bad-tag

# Not a tag
❯ GITHUB_REF=refs/heads/main bash .github/scripts/check-release.sh
Error: the current tag does not match the version in Cargo.toml: found 1.2.0, expected refs/heads/main
Error: the current tag does not match the version in Cargo.lock: found 1.2.0, expected refs/heads/main
```

Still, the UX is bad. Let me improve the error reporting...

.github/scripts/check-release.sh
```diff
@@ -1,24 +1,41 @@
...
+fi
+
+if [[ ! "$GITHUB_REF" =~ ^refs/tags/v[0-9]+\.[0-9]+\.[0-9]+(-[a-z0-9]+)?$ ]]; then
+    echo >&2 "Error: GITHUB_REF is not a valid tag: $GITHUB_REF"
+    exit 1
+fi
+
+current_tag=${GITHUB_REF#refs/tags/v}
 ret=0
-current_tag=${GITHUB_REF#'refs/tags/v'}
```
vvv, line 12: The quotes are unnecessary and only add noise.

.github/scripts/check-release.sh
```diff
@@ -1,24 +1,41 @@
...
 
-file_tag="$(grep '^version = ' Cargo.toml | cut -d '=' -f 2 | tr -d '"' | tr -d ' ')"
-check_tag $current_tag $file_tag
+toml_tag="$(cat Cargo.toml | read_version)"
+check_tag "$current_tag" "$toml_tag" Cargo.toml || ret=1
```
vvv, line 33: Technically, the quotes are unnecessary. We've validated `GITHUB_REF`, so we can be sure that `$current_tag` is not an empty string. `$toml_tag` is not empty also.

\<gold-plating\>
And yet, it doesn't hurt to add the double quotes. If `$current_tag` was quoted in the first version of the patch, the error returned by the script would be slightly less opaque.
\</gold-plating\>

curquiza:
Hello @vvv 
Sorry for the delay, but @irevoire is sick and I rely on his review. So, merging PR can takes time but we don't forget you!

vvv:
> Sorry for the delay, but @irevoire is sick and I rely on his review. So, merging PR can takes time but we don't forget you!

@curquiza Sure, no worries Thanks for reaching out to me. 😊

@irevoire Get well soon!

irevoire (approved):
That looks good to me. @curqui I don’t know how to test it on our CI though 😬

vvv:
> @curquiza I don’t know how to test it on our CI though 😬

One way to test a CI job is to noop its dependencies — comment out their implementations, put bash `true`/`false` commands here and there, tweak [`on:`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on) value to ensure the worflow runs when the PR's branch is updated... It's manual work.

Do let me know if I can help with anything.

vvv:
> > @curquiza I don’t know how to test it on our CI though 😬
> 
> One way to test a CI job is to noop its dependencies — comment out their implementations, put bash `true`/`false` commands here and there, tweak [`on:`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on) value to ensure the worflow runs when the PR's branch is updated... It's manual work.
> 
> Do let me know if I can help with anything.

Ah, sorry. I was thinking of another PR (#3794).

curquiza (changes requested):
Hello

Thanks for your PR @vvv 

Th change you made for the current script does not work as expected.

Our tags start with `v`, and if I try to do the following commands, I have an error:

```bash
bash-3.2$ export GITHUB_REF=v1.2.0
bash-3.2$ bash .github/scripts/check-release.sh
Error: GITHUB_REF is not a valid tag: v1.2.0
```

vvv:
> [vvv](https://github.com/vvv) dismissed [irevoire](https://github.com/irevoire)’s [stale review](https://github.com/meilisearch/meilisearch/pull/3799#pullrequestreview-1486327716) via [685422a](https://github.com/meilisearch/meilisearch/commit/685422abe248d4c7a5df1329f9f6a621c34af9fb)

I didn't mean to dismiss the review. I guess the repository is [configured](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews) to auto-dismiss the review once the PR's branch is updated.

> If both required reviews and stale review dismissal are enabled and a code-modifying commit is pushed to the branch of an approved pull request, the approval is dismissed. The pull request must be reviewed and approved again before it can be merged.

curquiza:
> I didn't mean to dismiss the review. I guess the repository is [configured](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews) to auto-dismiss the review once the PR's branch is updated.

Yes don't worry I know 😄 

> @curquiza GitHub adds refs/tags/ prefix to the value of GITHUB_REF.

Oh sorry @vvv, my bad, the worst part is I read what you wrote last week 🙈 I'm not a morning person definitely

@vvv 
Can you remove your last commit? It's not useful for us to add this permission, I really want to avoid useless changes in the repository 😊 We will add it if we need it

vvv:
@curquiza Sure. I've removed the `chmod +x` commit and rebased the branch onto the latest `master`.

curquiza (approved):
thank you @vvv 🙏 for your work, reactivity and involvement 

bors merge

PR #3813, opened 2023-06-06
Fix SDK CI for scheduled jobs

curquiza:
The SDK CI does not run for the scheduled job (`cron`) every day, and only works for manual triggers.

I added a job to define the Docker image we use depending on the event: `worflow_dispatch` = manual triggering, or `scheduled` = cron jobs

### DIFF
--- a/.github/workflows/sdks-tests.yml
+++ b/.github/workflows/sdks-tests.yml
@@ -16,8 +16,23 @@ env:
   MEILI_NO_ANALYTICS: 'true'
 
 jobs:
+  define-docker-image:
+    runs-on: ubuntu-latest
+    outputs:
+      docker-image: ${{ steps.define-image.outputs.docker-image }}
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


Kerollmops (commented):
Basically, it means that when we trigger it by hand, we must define the image, but when it is triggered by the cron job it is running on the latest released version. Am I right?

curquiza:
> Basically, it means that when we trigger it by hand, we must define the image, but when it is triggered by the cron job it is running on the latest released version. Am I right?

Triggered by hand
- either you choose the docker image
- or (by default) it will use the `nightly` image

Scheduled
- use `nightly`

The bug is, previously, when scheduled jobs were running, no images were defined, so it tried to fetch the `getmeili/meilisearch:` image instead of `getmeili/meilisearch:nighly`

Kerollmops (approved):
Thank you for this fix then 👌 
bors merge

curquiza:
bors merge

curquiza:
bors merge

Issue #3816, opened 2023-06-07
Incorrect deletion of documents when using the `disableOnAttributes` typo setting


loiclec:
To reproduce, on Meilisearch v1.2.0 (and earlier versions):

Set the searchable attributes to `[text, exact]`
Set the `disableOnAttributes` to: `[exact]`

Add the following documents:
```
    { id: 0, text: "hello" }
    { id: 1, exact: "hello" }
```

Delete the document with the external id `1` (make sure to disable soft deletion).

The remaining documents are correct:
```
    { id: 0, text: "hello" }
```

But the `words_fst` structure in the index was incorrectly updated. It does not contain `hello` anymore.
As a result, if we now search for `hello`, we get zero results.

curquiza:
Also reduce size of the index https://github.com/meilisearch/meilisearch/pull/3819#issuecomment-1580321320

PR #3817, opened 2023-06-07
Fix deletion bug that sometimes occur when using `disableOnAttributes` typo setting

loiclec:
Fixes https://github.com/meilisearch/meilisearch/issues/3816

The idea of the fix is simple: we keep not only a list of words to delete, but also a list of words to **keep**, which takes precedence.
To get the final, correct list of words to delete from the FST, we look at those that should both be deleted and not kept.

### DIFF
--- a/milli/src/update/delete_documents.rs
+++ b/milli/src/update/delete_documents.rs
@@ -15,8 +15,7 @@ use crate::facet::FacetType;
 use crate::heed_codec::facet::FieldDocIdFacetCodec;
 use crate::heed_codec::CboRoaringBitmapCodec;
 use crate::{
-    ExternalDocumentsIds, FieldId, FieldIdMapMissingEntry, Index, Result, RoaringBitmapCodec,
-    SmallString32, BEU32,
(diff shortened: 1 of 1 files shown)


Kerollmops (dismissed):
Wow! Thank you very much @loiclec 🤩
Thank you for the test too!
bors merge

loiclec:
Oh? 🤔  I saw the same test failures in my tests, but I thought I fixed them already. Sorry about that, will double-check my work!

loiclec:
Should be fixed now! :) let's wait for the full test results!
I initialised the `must_keep` variable to `true`, but it should have been initialised to `false`.

curquiza (changes requested):
Please don't merge this PR without discussing with @gillian-meilisearch. Plus it's not [the process we do usually](https://github.com/meilisearch/engine-team/blob/main/resources/meilisearch-release.md#how-to-do-patched-release-following-hotfixes)
No new branches have been created for the patch + no milestones.
I remind v1.2 has already been released

loiclec:
Closed. The fix will be in v1.3 but not in a patch release of v1.2

PR #3819, opened 2023-06-07
Remove the `docid_word_positions` database

loiclec:
Remove the `docid_word_positions` database, which was only used during deletion operations. In the process, also fixes https://github.com/meilisearch/meilisearch/issues/3816


loiclec:
I ran some quick tests now. On a clean indexing of `movies.json` and `smol_wiki_articles_3_4.csv`, we eliminate about 15% of the index size.

PR #3820, opened 2023-06-07
Reduce the maximum word proximity from 8 to 4

loiclec:
This is an experiment to evaluate the impact of storing fewer word pairs. I am not 100% sure that it is implemented perfectly, but it is good enough to run some initial experiments I think.

It reduces the size of the indexed `smol-wiki-articles-3_4.csv` dataset from 3.49GB to 2.19GB, a reduction of 37.5%. In combination with https://github.com/meilisearch/meilisearch/pull/3819#pullrequestreview-1467155039 , we reduced the index size from 4.19GB to 2.18GB. This means that the index size would be almost halved between v1.2 and v1.3.

For `movies.json`, we go from 212MB to 116MB.

While I haven't launched any benchmark yet, indexing also feels significanty faster to me.

### DIFF
--- a/milli/src/index.rs
+++ b/milli/src/index.rs
@@ -21,10 +21,9 @@ use crate::heed_codec::facet::{
 };
 use crate::heed_codec::{ScriptLanguageCodec, StrBEU16Codec, StrRefCodec};
 use crate::{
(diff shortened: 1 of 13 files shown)


Kerollmops:
Thanks for the experiment @loiclec 👏 We have since changed that in the engine, and we are now using a proximity max distance of 4. Closing this now!

PR #3821, opened 2023-06-07
Add normalized and detailed scores to documents returned by a query

### LINKED ISSUE #3771: Display ranking details at search
Related product team resources: [roadmap card](https://www.notion.so/meilisearch/ef7d86c80ef44c549e2361c70a367723?v=d11906bcef8b4b77815f84b801931c00&p=f09afa99f5ff4c34b82b480aab0ae726&pm=s) (_internal only_)
Related product discussions:
- https://github.com/meilisearch/product/discussions/594
- https://github.com/meilisearch/product/discussions/602
Related spec: WIP

## Motivation

Algolia does it and is requested by the users to migrate to Meilisearch.

## Usage

The purpose of this feature is to return ranking details for each document to understand and tweak the score of the documents more easily.

⚠️ Refer to the final spec to know the details and the final decisions about the usage.

## TODO

- [ ] Release a prototype
- [ ] If prototype validated, merge changes into `main`
- [ ] Update the spec

## Impacted teams

@meilisearch/docs-team and @meilisearch/integration-team

dureuill:
# Pull Request

## Related issue
Fixes #3771 

### User standpoint

- Documents returned by a search now contain an additional field `_rankingScore` that is a float bigger than 0.0 and lower or equal to 1.0. This field represents the relevancy of the document relatively to the search query and the settings of the index, with 1.0 meaning "perfect match" and "0.0" meaning "not matching the query" (Meilisearch should never return documents not matching the query at all). 
  - The `sort` and `geosort` ranking rules do not influence the `_rankingScore`.
- If adding a `showRankingScoreDetails` parameter to the search query, then the returned documents will now contain an additional `_rankingScoreDetails` field that is a JSON object containing one field per ranking rule that was applied, whose value is a JSON object with the following fields:
  - `order`: a number indicating the order this rule was applied (0 is the first applied ranking rule)
  - `score` (except for `sort` and `geosort`): a float indicating how the document matched this particular rule.
  - other fields that are specific to the rule, indicating for example how many words matched for a document and how many typos were counted in a matching document.
- Search queries that are part of a `multi-search` requests are modified in the same way and each of the queries can take the `showRankingScoreDetails` parameter independently. The results are still returned in separate lists and providing a unified list of results between multiple queries is not in the scope of this PR (but is unblocked by this PR and can be done manually by using the scores of the various documents). 

### Implementation standpoint

- Fix difference in how the position of terms were computed at indexing time and query time: this difference meant that a query containing a hard separator would fail the exactness check.
- Fix the id reported by the sort ranking rule (very minor)
- Change how the cost of removing words is computed. After this change the cost no longer works for any other ranking rule than `words`. Also made `words` have a cost of 0 such that the entire cost of `words` is given by the termRemovalStrategy. The new cost computation makes it so the score is computed in a way consistent with the number of words in the query. There's still an unresolved issue here when there are phrases in the query
- Remove optimization where ranking rules are not executed on buckets of a single document: this is important to allow the computation of an accurate score. If computing a score becomes optional in the future we could make this optimization available again when not compute scores.
- add virtual conditions to fid and position to always have the max cost: this ensures that the score is independent from the dataset
- Add a new `milli::score_details` module containing all the types that are involved in score computation.
- Make it so a bucket of result now contains a `ScoreDetails` and changed the ranking rules to produce their `ScoreDetails`.
- Expose the scores in the REST API.


dureuill:
Some changes: 

- Score is now on a scale of 0 to 1000 rather than a unit scale
- `_rankingScore` injection is gated behind a `showRankingScore` parameter

milli/src/score_details.rs
```diff
@@ -0,0 +1,295 @@
...
+        }
+    }
+
+    // max_rank = max_typo + 1
+    // max_typo = max_rank - 1
+    //
+    // rank = max_typo - typo + 1
+    // rank = max_rank - 1 - typo + 1
+    // rank + typo = max_rank
+    // typo = max_rank - rank
```
loiclec, line 210: nice!

milli/src/search/new/graph_based_ranking_rule.rs
```diff
@@ -162,12 +164,16 @@ impl<'ctx, G: RankingRuleGraphTrait> RankingRule<'ctx, QueryGraph> for GraphBase
         // Then pre-compute the cost of all paths from each node to the end node
         let all_costs = graph.find_all_costs_to_end();
 
+        let next_max_cost =
+            all_costs.get(graph.query_graph.root_node).iter().copied().max().unwrap_or(0) + 1;
+
```
loiclec, line 169: why "next" max cost?

Good that it is computed at `start_iteration` 👍

milli/src/search/new/ranking_rule_graph/build.rs
loiclec, line 1: ☀️

milli/src/search/new/ranking_rule_graph/fid/mod.rs
```diff
@@ -80,6 +81,35 @@ impl RankingRuleGraphTrait for FidGraph {
...
+            if let Some(max_fid) = ctx
+                .index
+                .searchable_fields_ids(ctx.txn)?
+                .map(|field_ids| field_ids.into_iter().max())
+            {
+                max_fid
+            } else {
+                ctx.index.fields_ids_map(ctx.txn)?.ids().max()
+            }
+        };
```
loiclec, line 95: I was ready to write a comment to tell you to use `fields_ids_map` in case `searchable_fields_ids` was `None`, but you did it already :D That's a detail of the codebase that is too tricky IMO.

milli/src/search/new/ranking_rule_graph/position/mod.rs
```diff
@@ -105,8 +106,20 @@ impl RankingRuleGraphTrait for PositionGraph {
             ));
         }
 
+        // artificial empty condition for computing max cost
+        let max_cost = term.term_ids.len() as u32 * 10;
+        edges.push((
+            max_cost,
+            conditions_interner
+                .insert(PositionCondition { term: term.clone(), positions: Vec::default() }),
+        ));
```
loiclec, line 115: What happens if an edge already exists with this cost? Generally, I tried to have only one condition associated with each cost. I think everything still works correctly if that's not the case, but still, it seems conceptually simpler to have 1 cost -> 1 condition.

milli/src/search/new/ranking_rules.rs
```diff
@@ -66,4 +67,6 @@ pub struct RankingRuleOutput<Q> {
     pub query: Q,
     /// The allowed candidates for the child ranking rule
     pub candidates: RoaringBitmap,
+    /// The score for the candidates of the current bucket
+    pub score: ScoreDetails,
```
loiclec, line 71: I assume you tested the score details feature with a placeholder search as well, and that it only returns the `Sort` score details? Also worth checking it when multiple `field:asc/desc` ranking rules exist.

milli/src/search/new/sort.rs
```diff
@@ -67,7 +69,7 @@ impl<'ctx, Query> Sort<'ctx, Query> {
 impl<'ctx, Query: RankingRuleQueryTrait> RankingRule<'ctx, Query> for Sort<'ctx, Query> {
     fn id(&self) -> String {
         let Self { field_name, is_ascending, .. } = self;
-        format!("{field_name}:{}", if *is_ascending { "asc" } else { "desc " })
+        format!("{field_name}:{}", if *is_ascending { "asc" } else { "desc" })
```
loiclec, line 72: oops 😶

milli/src/search/new/sort.rs
```diff
@@ -150,7 +183,15 @@ impl<'ctx, Query: RankingRuleQueryTrait> RankingRule<'ctx, Query> for Sort<'ctx,
...
         } else {
             let query = self.original_query.as_ref().unwrap().clone();
-            Ok(Some(RankingRuleOutput { query, candidates: universe.clone() }))
+            Ok(Some(RankingRuleOutput {
+                query,
+                candidates: universe.clone(),
+                score: ScoreDetails::Sort(score_details::Sort {
+                    field_name: self.field_name.clone(),
+                    ascending: self.is_ascending,
+                    value: serde_json::Value::Null,
```
loiclec, line 192: hmm... I don't think this is technically correct. We could return documents whose value for this field is different than `null` (ie field doesn't exist, or contains an empty array, or maybe even an object). It doesn't necessarily make it wrong, but it is tricky.

loiclec (approved):
Thanks Louis! :) It was a surprisingly small and clear PR to add these two features!

milli/src/search/new/bucket_sort.rs
```diff
@@ -104,23 +125,23 @@ pub fn bucket_sort<'ctx, Q: RankingRuleQueryTrait>(
...
 
     while valid_docids.len() < length {
-        // The universe for this bucket is zero or one element, so we don't need to sort
-        // anything, just extend the results and go back to the parent ranking rule.
-        if ranking_rule_universes[cur_ranking_rule_index].len() <= 1 {
-            let bucket = std::mem::take(&mut ranking_rule_universes[cur_ranking_rule_index]);
-            maybe_add_to_results!(bucket);
+        // The universe for this bucket is zero, so we don't need to sort
+        // anything, just go back to the parent ranking rule.
+        if ranking_rule_universes[cur_ranking_rule_index].is_empty() {
```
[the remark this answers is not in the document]
dureuill: I just re-added the optimization when we don't want to compute the score :-)

milli/src/search/new/geo_sort.rs
```diff
@@ -80,7 +81,7 @@ pub struct GeoSort<Q: RankingRuleQueryTrait> {
     field_ids: Option<[u16; 2]>,
     rtree: Option<RTree<GeoPoint>>,
 
-    cached_sorted_docids: VecDeque<u32>,
+    cached_sorted_docids: VecDeque<(u32, [f64; 2])>,
```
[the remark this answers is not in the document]
dureuill: Not in my understanding.

- The first `u32` is the public id of the document
- The 2D point is the geopoint of that document.

I don't think it is expected that consecutive entries in this queue are all at the same point. The only rule is that they are stored in increasing (decreasing?) distance from the target point.

milli/src/score_details.rs
```diff
@@ -0,0 +1,295 @@
...
+    GeoSort(GeoSort),
+}
+
+impl ScoreDetails {
+    pub fn local_score(&self) -> Option<f64> {
+        self.rank().map(Rank::local_score)
+    }
+
+    pub fn rank(&self) -> Option<Rank> {
+        match self {
```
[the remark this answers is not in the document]
dureuill: I hesitated for a long time to make it so the "highest" rank is 0, and the worst rank is "max rank", so that they're ordinals, like in a race. But then I kinda want to start them at 1 instead of 0 😬 

Would maybe `Grade` or `Mark` work better than `Rank` for you?

Today the main distinction between rank and score is that the score is normalized by dividing the rank by its max rank.

Wording is important, so I'm very open to make the necessary changes here, I'm just a bit unsure what these are :-)

milli/src/score_details.rs
```diff
@@ -0,0 +1,295 @@
...
+        }
+    }
+
+    pub fn global_score<'a>(details: impl Iterator<Item = &'a Self>) -> f64 {
+        Rank::global_score(details.filter_map(Self::rank))
+    }
+
+    pub fn global_score_linear_scale<'a>(details: impl Iterator<Item = &'a Self>) -> u64 {
+        (Self::global_score(details) * LINEAR_SCALE_FACTOR).round() as u64
+    }
```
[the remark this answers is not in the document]
dureuill: Unfortunately, the `global_score` is already normalized too.

Perhaps the correct name is `global_score_integer_scale`?

milli/src/score_details.rs
```diff
@@ -0,0 +1,295 @@
...
+                    order += 1;
+                }
+            }
+        }
+        details_map
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Words {
```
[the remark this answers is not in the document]
dureuill: these are meant to be used prefixed by the module name, so I followed the convention of not repeating context that can be inferred from the module name.

Users should use the qualified `score_details::Words` form, or too verbose, `use score_details::Words as WordsScore`.

Do you still feel they should be qualified?

milli/src/search/new/graph_based_ranking_rule.rs
dureuill: Because of the `+1`. It is computed as the max cost of all the cost, + 1.

Initially I just called that `max_cost`, but it felt misleading as it is not one of the costs.

Feel free to suggest something clearer, or maybe I should add some comment here?

Maybe `ceiling_cost` but uhhh

milli/src/search/new/ranking_rule_graph/fid/mod.rs
dureuill: I agree IMO as a caller in `fid` I shouldn't have to worry about this details.

A possible fix could be to change the `searchable_fields_ids` method to return an enum:

```rust
enum SearchableFieldIds {
    UserDefined(Vec<FieldId>),
    Inferred(Vec<FieldId>),
}

impl SearchableFieldIds {
    pub fn into_vec(self) -> Vec<FieldId> { /* ... */ }
}
```

Then my code could become:

```rust
let max_fid = ctx.index.searchable_fields_ids(ctx.txn)?.into_iter().max();
```

while still leaving the possibility for interested callers to match on the enum to know if the searchable fields are inferred or user-defined.

milli/src/search/new/ranking_rule_graph/position/mod.rs
dureuill: Ah, I assumed that since it could be the case for edges created by distinct calls to `build_edges` to have the same cost, it was also OK inside of a single call.

Do you want me to add a special kind of condition then? Or to check that the max cost wasn't already added in the look above?

milli/src/search/new/ranking_rules.rs
dureuill: > I assume you tested the score details feature with a placeholder search as well

Yes, with placeholder search, only the ranking rules that are applied in the placeholder search have details, so basically only sort and geosort.

> Also worth checking it when multiple field:asc/desc ranking rules exist.

can't be sure if I already checked, can check again.

milli/src/search/new/bucket_sort.rs
```diff
@@ -89,11 +105,16 @@ pub fn bucket_sort<'ctx, Q: RankingRuleQueryTrait>(
             } else {
                 cur_ranking_rule_index -= 1;
             }
+            // FIXME: check off by one
```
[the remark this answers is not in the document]
dureuill: from the tests and in particular displaying the detailed scores, it appeared correct.

I'm sorry I don't have a more theoretical proof though.

milli/src/search/new/sort.rs
dureuill: it happens 😅

milli/src/search/new/ranking_rule_graph/build.rs
dureuill: 🌻

milli/src/score_details.rs
dureuill: thanks I needed the computation laid out, anyway 😓

milli/src/score_details.rs
```diff
@@ -0,0 +1,295 @@
...
+pub struct Rank {
+    /// The ordinal rank, such that `max_rank` is the first rank, and 0 is the last rank.
+    ///
+    /// The higher the better. Documents with a rank of 0 have a score of 0 and are typically never returned
+    /// (they don't match the query).
+    pub rank: u32,
+    /// The maximum possible rank. Documents with this rank have a score of 1.
+    ///
+    /// The max rank should not be 0.
+    pub max_rank: u32,
```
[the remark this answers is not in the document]
dureuill: would it make it better to use `grade` or `mark` in the fields (`the ordinal rank` is exactly what it is, IMO)

dureuill:
Hey @loiclec thank you for the insightful review.

I answered your comments and, in the case the score is not computed, reintroduced the optimization we lost

milli/src/search/new/ranking_rules.rs
dureuill: Yeah, it works with multiple rules also.

example:

```json
{
      "title": "Batman: The Dark Knight Returns, Part 1",
      "id": 123025,
      "_rankingScoreDetails": {
        "words": {
          "order": 0,
          "matchingWords": 7,
          "maxMatchingWords": 7,
          "score": 1000
        },
        "typo": {
          "order": 1,
          "typoCount": 0,
          "maxTypoCount": 7,
          "score": 1000
        },
        "proximity": {
          "order": 2,
          "score": 1000
        },
        "attribute": {
          "order": 3,
          "attributes_ranking_order": 1000,
          "attributes_query_word_order": 761
        },
        "title:asc": {
          "order": 4,
          "value": "batman: the dark knight returns, part 1"
        },
        "release_date:desc": {
          "order": 5,
          "value": 1345507200.0
        },
        "exactness": {
          "order": 6,
          "exactIn": "matchesFull",
          "score": 1000
        }
      }
    }
```

loiclec (approved):
Great, thanks again! :)

milli/src/search/new/ranking_rule_graph/position/mod.rs
loiclec: It should work correctly either way, but it does make it easier to debug when no duplicate costs exist :)

milli/src/search/new/ranking_rule_graph/fid/mod.rs
loiclec: That would be great :) but not strictly needed to do it in this PR IMO

milli/src/score_details.rs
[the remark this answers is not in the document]
loiclec: I get the argument for using simply `Words`, do whatever is best for you :)

milli/src/score_details.rs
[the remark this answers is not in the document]
loiclec: `global_score_per_mille`?

milli/src/score_details.rs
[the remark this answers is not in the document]
loiclec: I think the difference is that `Criterion` is used both in the settings and search API, so it needs to be located in a higher module. On the other hand, `ScoreDetails` only exists within the search API.

milli/src/search/new/geo_sort.rs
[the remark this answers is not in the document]
loiclec: oops, sorry, you're completely right!

milli/src/score_details.rs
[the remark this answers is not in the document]
dureuill: I think I prefer keeping `Words` for now, I'll see if some other reviewers have the same comment :-)

dureuill:
Updated PR description

milli/src/search/new/bucket_sort.rs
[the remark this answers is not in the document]
Kerollmops: Can we remove this TODO, then? And all the other ones, please? :)

milli/src/search/new/bucket_sort.rs
```diff
@@ -143,10 +174,12 @@ pub fn bucket_sort<'ctx, Q: RankingRuleQueryTrait>(
         ranking_rule_universes[cur_ranking_rule_index] -= &next_bucket.candidates;
 
         if cur_ranking_rule_index == ranking_rules_len - 1
-            || next_bucket.candidates.len() <= 1
+            || (scoring_strategy == ScoringStrategy::Skip && next_bucket.candidates.len() <= 1)
             || cur_offset + (next_bucket.candidates.len() as usize) < from
         {
             maybe_add_to_results!(next_bucket.candidates);
+            // FIXME: use index based logic like all the other rules so that you don't have to maintain the pop/push?
+            ranking_rule_scores.pop();
```
Kerollmops, line 182: Can we refer this change in an issue or can we do it right now?

Kerollmops (changes requested):
After reviewing with @ManyTheFish and you the logic of the feature in the ranking rules and reviewing this PR commit by commit, I am pretty happy with the result. The amount of added code is particularly important, fortunately, it is due to the test and insta tests!

milli/src/search/new/bucket_sort.rs
dureuill: unfortunately, it is hard to do due to score details not having a meaningful "default" value. The options I see:

1. keep it this way (it works, but is inconsistent)
2. Make the vector a vector of `Option<ScoreDetails>` so we have a default value. I find it sad since everytime the option will be actually accessed it will have to be a Some that we will subsequently unwrap.
3. Use a default `ScoreDetails` value. They all have date, so which one should we use?

Do you see another option?

Kerollmops: Let's keep it this way and remove the comment :)

dureuill: done

milli/src/score_details.rs
```diff
@@ -0,0 +1,314 @@
...
+                    // For now, position is a virtual rule always preceded by the "fid" rule
+                    let attribute_details = details_map
+                        .get_mut("attribute")
+                        .expect("position not preceded by attribute");
+                    let attribute_details = attribute_details
+                        .as_object_mut()
+                        .expect("attribute details was not an object");
+                    let Some(fid_details) = fid_details
+                    else {
+                        panic!("position not preceded by attribute");
```
ManyTheFish, line 102: Curiosity question, in this case, shouldn't we use `unimplemented!`?

ManyTheFish (dismissed):
Approved! Sorry for the delay!

milli/src/score_details.rs
dureuill: 🤔 I guess if the idea is to allow this in the future and that it is not currently implemented, yes this is a good use case for `unimplemented`. I'll make the change!

dureuill: done!

foremtehan:
Does it also have a highlight feature? I'm referring to the ability for html or markdown tags to be used to highlight the matched words in the returned query

Kerollmops (approved):
That looks perfect to me 🙏
A new era is starting for Meilisearch ❤️

bors merge

dureuill:
Somehow bors did not start?

bors merge

dureuill:
Hey @foremtehan :wave:

Thank you for the suggestion. Does the existing [highlighting feature](https://www.meilisearch.com/docs/reference/api/search#highlight-tags) answer your use case? If so, great 😊! If not, I invite you to [open a new discussion](https://github.com/meilisearch/product/discussions/categories/feedback-feature-proposal) in our product repository, as this current thread is meant to discuss the implementation of the scoring feature ☺️, and opening the discussion there will make sure that the right people take a look at it :+1:

PR #3825, opened 2023-06-08
Accept semantic vectors and allow users to query nearest neighbors

Kerollmops:
This Pull Request brings a new feature to the current API. The engine accepts a new `_vectors` field akin to the `_geo` one. This vector is stored in Meilisearch and can be retrieved via search. This work is the first step toward hybrid search, bringing the best of both worlds: keyword and semantic search ❤️‍🔥

## ToDo
 - [x] Make it possible to get the `limit` nearest neighbors from a user-generated vector by using the `vector` field of search route.
 - [x] Delete the documents and vectors from the HNSW-related data structures.
     - [x] Do it the slow and ugly way (we need to be able to iterate over all the values).
     - [ ] Do it the efficient way (Wait for a new method or implement it myself).
 - [ ] ~~Move from the `hnsw` crate to the hgg one~~ The hgg crate is too slow.
   Meilisearch takes approximately 88s to answer a query. It is related to the time it takes to deserialize the `Hgg` data structure or search in it. I didn't take the time to measure precisely. We moved back to the hnsw crate which takes approximately 40ms to answer.
   - [ ] ~~Wait for a fix for https://github.com/rust-cv/hgg/issues/4.~~
 - [x] Fix the current dot product function.
 - [x] Fill in the other `SearchResult` fields.
 - [x] Remove the `hnsw` dependency of the meilisearch crate.
 - [x] Fix the pages by taking the offset into account.
 - [x] Release a first prototype https://github.com/meilisearch/product/discussions/621#discussioncomment-6183647
 - [x] Make the pagination and filtering faster and more correct.
 - [x] Return the original vector in the output search results (like `query`).
 - [x] Return an `_semanticSimilarity` field in the documents (it's a dot product)
   - [x] Return this score even if the `_vectors` field is not displayed
   - [x] Rename the field `_semanticScore`.
   - [ ] Return the `_geoDistance` value even if the `_geo` field is not displayed
 - [x] Store the HNSW on possibly multiple LMDB values.
   - [ ] Measure it and make it faster if needed
   - [ ] Export the `ReadableSlices` type into a small external crate
 - [x] Accept an `_vectors` field instead of the `_vector` one.
 - [x] Normalize all vectors.
 - [ ] Remove the `_vectors` field from the default searchable attributes (as we do with `_geo`?).
 - [ ] Correctly compute the candidates by remembering the documents having a valid `_vectors` field.
 - [ ] Return the right errors:
     - [ ] Return an error when the query vector is not the same length as the vectors in the HNSW.
     - [ ] We must return the user document id that triggered the vector dimension issue.
     - [x] If an indexation error occurs.
     - [ ] Fix the error codes when using the search route.
 - [ ] ~~Introduce some settings:~~
    We currently ensure that the vector length is consistent over the whole set of documents and return an error for when a vector dimension doesn't follow the current number of dimensions.
     - [ ] The length of the vector the user will provide.
     - [ ] The distance function (we only support dot as of now).
 - [ ] Introduce other distance functions
    - [ ] Euclidean
    - [ ] Dot Product
    - [ ] Cosine
    - [ ] Make them SIMD optimized
    - [ ] Give credit to qdrant
 - [ ] Add tests.
 - [ ] Write a mini spec.
 - [ ] Release it in v1.3 as an experimental feature.


meilisearch/src/search.rs
```diff
@@ -523,6 +553,20 @@ fn insert_geo_distance(sorts: &[String], document: &mut Document) {
     }
 }
 
+fn insert_semantic_score(query: &[f32], vectors: Value, document: &mut Document) {
+    let vectors =
+        match serde_json::from_value(vectors).map(VectorOrArrayOfVectors::into_array_of_vectors) {
+            Ok(vectors) => vectors,
+            Err(_) => return,
```
dureuill, line 560: It is unclear to me what error conditions we might be ignoring here. Are they always benign?

meilisearch/src/search.rs
```diff
@@ -650,6 +694,22 @@ fn make_document(
...
 }
 
+/// Extract the JSON value under the field name specified
+/// but doesn't support nested objects.
+fn extract_field(
+    field_name: &str,
+    field_ids_map: &FieldsIdsMap,
+    obkv: obkv::KvReaderU16,
+) -> Result<Option<serde_json::Value>, MeilisearchHttpError> {
+    match field_ids_map.id(field_name) {
```
dureuill, line 704: nitpick, but couldn't the similarity be returned for each document by the underlying search, rather than being recomputed here by guessing that the value that was used to compute it was the one producing the highest similarity?

meilisearch/src/search.rs
```diff
@@ -281,12 +292,17 @@ pub fn add_search_rules(query: &mut SearchQuery, rules: IndexSearchRules) {
...
     let rtxn = index.read_txn()?;
 
     let mut search = index.search(&rtxn);
 
+    if let Some(ref vector) = query.vector {
+        search.vector(vector.clone());
+    }
+
     if let Some(ref query) = query.q {
         search.query(query);
```
dureuill, line 307: shouldn't adding both `"q"` and `"vector"` to a query be an error?

My understanding is that the current behavior is silently ignoring the `"q"` part, but isn't there a risk of a user thinking they are using hybrid search, which is not currently implemented?

not necessarily to be fixed in this PR though.

meilisearch/src/search.rs
```diff
@@ -523,6 +553,20 @@ fn insert_geo_distance(sorts: &[String], document: &mut Document) {
...
+        match serde_json::from_value(vectors).map(VectorOrArrayOfVectors::into_array_of_vectors) {
+            Ok(vectors) => vectors,
+            Err(_) => return,
+        };
+    let similarity = vectors
+        .into_iter()
+        .map(|v| OrderedFloat(dot_product_similarity(query, &v)))
+        .max()
+        .map(OrderedFloat::into_inner);
+    document.insert("_semanticScore".to_string(), json!(similarity));
```
dureuill, line 567: this way of inserting the field will replace any original value from the document for the field `_semanticScore`.

This behavior is consistent with `_geoDistance`, but inconsistent with `_rankingScore`, `_rankingScoreDetails`, `_formatted` and `_matchesPosition`, that just output both fields.

To be consistent with the latter fields, you'd need to edit the `SearchHit` structure to look like:

```rust
#[derive(Debug, Clone, Serialize, PartialEq)]
pub struct SearchHit {
    #[serde(flatten)]
    pub document: Document,
    #[serde(rename = "_formatted", skip_serializing_if = "Document::is_empty")]
    pub formatted: Document,
    #[serde(rename = "_matchesPosition", skip_serializing_if = "Option::is_none")]
    pub matches_position: Option<MatchesPosition>,
    #[serde(rename = "_rankingScore", skip_serializing_if = "Option::is_none")]
    pub ranking_score: Option<f64>,
    #[serde(rename = "_rankingScoreDetails", skip_serializing_if = "Option::is_none")]
    pub ranking_score_details: Option<serde_json::Map<String, serde_json::Value>>,
    #[serde(rename = "_semanticScore", skip_serializing_if = "Option::is_none")]
    pub semantic_score: Option<f32>,
}
```

And then pass the value in `Some` for the `SearchHit` corresponding to each document with a similarity score.

milli/src/search/new/mod.rs
```diff
@@ -442,6 +448,34 @@ pub fn execute_search(
 
     let fields_ids_map = ctx.index.fields_ids_map(ctx.txn)?;
 
+    let docids = match vector {
+        Some(vector) => {
+            // return the nearest documents that are also part of the candidates.
+            let mut searcher = Searcher::new();
+            let hnsw = ctx.index.vector_hnsw(ctx.txn)?.unwrap_or_default();
+            let ef = hnsw.len().min(100);
```
dureuill, line 456: I don't understand the `min(100)`

milli/src/search/new/mod.rs
```diff
@@ -442,6 +448,34 @@ pub fn execute_search(
...
+            let mut dest = vec![Neighbor { index: 0, distance: 0 }; ef];
+            let vector = normalize_vector(vector.clone());
+            let neighbors = hnsw.nearest(&vector, ef, &mut searcher, &mut dest[..]);
+
+            let mut docids = Vec::new();
+            for Neighbor { index, distance: _ } in neighbors.iter() {
+                let index = BEU32::new(*index as u32);
+                let docid = ctx.index.vector_id_docid.get(ctx.txn, &index)?.unwrap().get();
+                if universe.contains(docid) {
```
dureuill, line 465: I don't understand this check :sweat_smile:

dureuill (commented):
A few comments and questions ☺️

milli/src/search/new/mod.rs
Kerollmops: This is a necessary check because we don't want to leak documents hidden by filters (and tenant token filters). The HNSW can't filter when it dives, so we must do it afterward. However, it can return fewer documents than the requested amount, but it's the only quick solution I found. The other solution would be designing a predicate system on the HNSW crate.

milli/src/search/new/mod.rs
Kerollmops: We will search for a hundred documents in the HNSW and filter afterward. I get the length of the HNSW data structure to make sure the array is not bigger than that otherwise the HNSW crate badly panics.

milli/src/update/index_documents/typed_chunk.rs
```diff
@@ -221,6 +229,40 @@ pub(crate) fn write_typed_chunk_into_index(
...
+
+            let mut cursor = vector_points.into_cursor()?;
+            while let Some((key, value)) = cursor.move_on_next()? {
+                // convert the key back to a u32 (4 bytes)
+                let (left, _index) = try_split_array_at(key).unwrap();
+                let docid = DocumentId::from_be_bytes(left);
+                // convert the vector back to a Vec<f32>
+                let vector: Vec<f32> = pod_collect_to_vec(value);
+
+                // TODO Move this error in the vector extractor
```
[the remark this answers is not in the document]
Kerollmops: I changed the TODO and no we will not land with a fix for now 😄

milli/src/update/index_documents/extract/extract_vector_points.rs
```diff
@@ -0,0 +1,69 @@
...
+
+            for (i, vector) in vectors.into_iter().enumerate() {
+                match u16::try_from(i) {
+                    Ok(i) => {
+                        let mut key = docid_bytes.to_vec();
+                        key.extend_from_slice(&i.to_ne_bytes());
+                        let bytes = cast_slice(&vector);
+                        writer.insert(key, bytes)?;
+                    }
+                    Err(_) => continue,
```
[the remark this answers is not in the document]
Kerollmops: I fixed it, and we will silently ignore vectors past `u16::MAX` vectors.

milli/src/update/index_documents/extract/extract_vector_points.rs
```diff
@@ -0,0 +1,69 @@
...
+                    }
+                    .into())
+                }
+            };
+
+            for (i, vector) in vectors.into_iter().enumerate() {
+                match u16::try_from(i) {
+                    Ok(i) => {
+                        let mut key = docid_bytes.to_vec();
+                        key.extend_from_slice(&i.to_ne_bytes());
```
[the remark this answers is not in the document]
Kerollmops: This is an excellent analysis here, I switched to big-endian as we are writing in the `grenad::Writer` which needs lexicographically ordered keys.

meilisearch/src/search.rs
Kerollmops: I just returned an Internal error when this happened. Thank you :)

meilisearch/src/search.rs
Kerollmops: It's much simpler to recompute this afterward and reduce the amount of code to inject and pass through layers between the search and the HTTP server.

meilisearch/src/search.rs
Kerollmops: I just printed a warning and edited the feature discussion https://github.com/meilisearch/product/discussions/677

milli/src/search/new/mod.rs
Kerollmops: Fixed now, thanks Louis ❤️

dureuill (approved):
Great PR, thanks for adding this feature :heart:

Kerollmops:
Thank you for this great review 🪁
bors merge

macraig:
Hi @Adel-Moumen , changing the distance function isn't available in the current version of Vector Search, but we will take it into account for our second iteration of the feature we'll start working on soon. We have an ongoing GitHub discussion with users to chat about their use cases and feature requests, you can check it out here: https://github.com/orgs/meilisearch/discussions/677

Adel-Moumen:
> Hi @Adel-Moumen , changing the distance function isn't available in the current version of Vector Search, but we will take it into account for our second iteration of the feature we'll start working on soon. We have an ongoing GitHub discussion with users to chat about their use cases and feature requests, you can check it out here: https://github.com/orgs/meilisearch/discussions/677

Thanks! Maybe we'll implement it ourselves then. Do you have any guidelines for contributors if we want to open a PR on that?

macraig:
> > Hi @Adel-Moumen , changing the distance function isn't available in the current version of Vector Search, but we will take it into account for our second iteration of the feature we'll start working on soon. We have an ongoing GitHub discussion with users to chat about their use cases and feature requests, you can check it out here: https://github.com/orgs/meilisearch/discussions/677
> 
> Thanks! Maybe we'll implement it ourselves then. Do you have any guidelines for contributors if we want to open a PR on that?

Yes, here's the link: https://github.com/meilisearch/meilisearch/blob/main/CONTRIBUTING.md

PR #3834, opened 2023-06-13
Define searchable fields at runtime

### LINKED ISSUE #3772: Define field to search on at search-time
Related product team resources: [roadmap card](https://www.notion.so/meilisearch/Define-field-to-search-on-at-search-time-fe5adbc9c965415f8335f6fd33b33d66?pvs=4) (_internal only_)

Related product discussions:
- https://github.com/meilisearch/product/discussions/107
- https://github.com/meilisearch/product/discussions/107#discussioncomment-1211737 
- https://github.com/meilisearch/product/discussions/595

Related spec: WIP

## Motivation

Both Algolia and Typesense support this feature and it has been flagged several times as a blocker to adoption.

## Usage

Enables running searches over a subset of searchableAttributes without modifying Meilisearch’s index settings. 

⚠️  Refer to the final spec to know the details and the final decisions about the usage.

## TODO

- [ ] Release a prototype
- [ ] If prototype validated, merge changes into `main`
- [ ] Update the spec

### Technical prototype's TODO (@ManyTheFish addition)

- [x] Milli side:
  - [x] Ensure a database exists discriminating the fields that contain the searched words
  - [x] Add a parameter to the [Search struct](https://github.com/meilisearch/meilisearch/blob/0a7817a0022dfdb0ac3ee2ea97dc33b13af28627/milli/src/search/mod.rs#L23-L36)
  - [x] Modify [resolve_universe](https://github.com/meilisearch/meilisearch/blob/0a7817a0022dfdb0ac3ee2ea97dc33b13af28627/milli/src/search/new/mod.rs#L411-L412)
  - [x] Modify [bucket_sort](https://github.com/meilisearch/meilisearch/blob/0a7817a0022dfdb0ac3ee2ea97dc33b13af28627/milli/src/search/new/mod.rs#L414)
- [x] API side:
  - [x] Define a new Search parameter (`restrictSearchableAttributes`)
  - [x] Make Milli work with the search parameter
- [x] Test side (@ManyTheFish):
  - [x] A query searching in an unsearchable field **MAY** return an error or **MAY** return an empty result (TBD)
  - [x] A query with the terms matching strategy `All` must only return documents containing all the requested words in the searched field
  - [x] Ensure the criteria take into account the filtered field:
    - [x] Word: A document containing query-words in the searched field must be ranked before a document containing more query-words in all fields but fewer in the searched field.
    - [x] Typo: A document containing query-words without typos in the searched field must be ranked before a document containing more query-words without typos in all fields but fewer in the searched field.
    - [x] Proximity: A document containing query-words with a distance of `x` in the searched field must be ranked before a document containing query-words with a perfect distance in another field but a distance `y` worse than `x` in the searched field.
    - [x] Attribute: A document containing query-words in the searched field must be ranked before a document containing more query-words in a higher-ranked field but fewer in the searched field.
    - [x] Exactness: A document containing exact query-words in the searched field must be ranked before a document containing more exact query-words in all fields but fewer in the searched field

### Remaining Uncertainties (@ManyTheFish addition) (poke @macraig)
After making a technical tour of the feature, some uncertainties remain:
- `API`: The dedicated search parameter is not defined, [here is the comment link showing some competitor APIs](https://github.com/meilisearch/product/discussions/107#discussioncomment-1211737)
  - `restrictSearchableAttributes`
- `Undefined behavior`: when the end user search in a field that is not part of the searchable attribute the behavior of Meilisearch is undefined, we could return an error or return an empty response
  - Empty response
- `Expectations of the feature`: Does the order of the field given at search time reorder the searchable attribute order given in the settings 
  - The order doesn't change
- `Expectations of the feature`, the feature could be implemented as an exclusive filter ("I only want the documents matching in **this** field") or as a [ranking priority](https://github.com/meilisearch/product/discussions/595) ("I want the documents matching in **this** field to be ranked before the others")
  - The feature is implemented as an exclusive filter 

## Impacted teams

ManyTheFish:
## Summary
This feature allows the end-user to search in one or multiple attributes using the search parameter `attributesToSearchOn`:

```json
{
  "q": "Captain Marvel",
  "attributesToSearchOn": ["title"]
}
```

This feature act like a filter, forcing Meilisearch to only return the documents containing the requested words in the attributes-to-search-on. Note that, with the matching strategy `last`, Meilisearch will only ensure that the first word is in the attributes-to-search-on, but, the retrieved documents will be ordered taking into account the word contained in the attributes-to-search-on. 

## Trying the prototype

A dedicated docker image has been released for this feature:

#### last prototype version:

```bash
docker pull getmeili/meilisearch:prototype-define-searchable-fields-at-search-time-1
```

#### others prototype versions:

```bash
docker pull getmeili/meilisearch:prototype-define-searchable-fields-at-search-time-0
```

## Technical Detail

The attributes-to-search-on list is given to the search context, then, the search context uses the `fid_word_docids`database using only the allowed field ids instead of the global `word_docids` database. This is the same for the prefix databases.
The database cache is updated with the merged values, meaning that the union of the field-id-database values is only made if the requested key is missing from the cache.

### Relevancy limits

Almost all ranking rules behave as expected when ordering the documents.
Only `proximity` could miss-order documents if all the searched words are in the restricted attribute but a better proximity is found in an ignored attribute in a document that should be ranked lower. I put below a failing test showing it:
```rust
#[actix_rt::test]
async fn proximity_ranking_rule_order() {
    let index = index_with_documents(
        &server,
        &json!([
        {
            "title": "Captain super mega cool. A Marvel story",
            // Perfect distance between words in an ignored attribute
            "desc": "Captain Marvel",
            "id": "1",
        },
        {
            "title": "Captain America from Marvel",
            "desc": "a Shazam ersatz",
            "id": "2",
        }]),
    )
    .await;

    // Document 2 should appear before document 1.
    index
        .search(json!({"q": "Captain Marvel", "attributesToSearchOn": ["title"], "attributesToRetrieve": ["id"]}), |response, code| {
            assert_eq!(code, 200, "{}", response);
            assert_eq!(
                response["hits"],
                json!([
                    {"id": "2"},
                    {"id": "1"},
                ])
            );
        })
        .await;
}
```

Fixing this would force us to create a `fid_word_pair_proximity_docids` and a `fid_word_prefix_pair_proximity_docids` databases which may multiply the keys of `word_pair_proximity_docids` and `word_prefix_pair_proximity_docids` by the number of attributes in the searchable_attributes list. If we think we should fix this test, I'll suggest doing it in another PR.

## Related

Fixes #3772

### DIFF
--- a/milli/src/search/new/db_cache.rs
+++ b/milli/src/search/new/db_cache.rs
@@ -4,12 +4,13 @@ use std::hash::Hash;
 
 use fxhash::FxHashMap;
 use heed::types::ByteSlice;
-use heed::{BytesDecode, BytesEncode, Database, RoTxn};
+use heed::{BytesEncode, Database, RoTxn};
 use roaring::RoaringBitmap;
 
 use super::interner::Interned;
 use super::Word;
-use crate::heed_codec::StrBEU16Codec;
+use crate::heed_codec::{BytesDecodeOwned, StrBEU16Codec};
+use crate::update::MergeFn;
 use crate::{
     CboRoaringBitmapCodec, CboRoaringBitmapLenCodec, Result, RoaringBitmapCodec, SearchContext,
 };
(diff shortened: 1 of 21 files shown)


milli/src/search/new/mod.rs
```diff
@@ -68,8 +69,18 @@ impl<'ctx> SearchContext<'ctx> {
...
             phrase_docids: <_>::default(),
+            restricted_fids: None,
         }
     }
+
+    pub fn searchable_attributes(&mut self, searchable_attributes: &'ctx [String]) -> Result<()> {
+        let fids_map = self.index.fields_ids_map(self.txn)?;
+        let restricted_fids =
+            searchable_attributes.iter().filter_map(|name| fids_map.id(name)).collect();
+        self.restricted_fids = Some(restricted_fids);
```
Kerollmops:
```suggestion
        self.restricted_fids =
            Some(searchable_attributes.iter().filter_map(|name| fids_map.id(name)).collect());
```

milli/src/search/new/db_cache.rs
```diff
@@ -22,50 +23,110 @@ use crate::{
...
+    ) -> Result<Option<DC::DItem>>
     where
         K1: Copy + Eq + Hash,
         KC: BytesEncode<'v>,
+        DC: BytesDecodeOwned,
     {
-        let bitmap_ptr = match cache.entry(cache_key) {
-            Entry::Occupied(bitmap_ptr) => *bitmap_ptr.get(),
+        match cache.entry(cache_key) {
+            Entry::Occupied(_) => {}
```
Kerollmops, line 61:
```suggestion
            Entry::Occupied(_) => (),
```

milli/src/search/new/db_cache.rs
```diff
@@ -22,50 +23,110 @@ use crate::{
...
+        merger: MergeFn,
+    ) -> Result<Option<DC::DItem>>
+    where
+        K1: Copy + Eq + Hash,
+        KC: BytesEncode<'v>,
+        DC: BytesDecodeOwned,
+        KC::EItem: Sized,
+    {
+        match cache.entry(cache_key) {
+            Entry::Occupied(_) => {}
```
Kerollmops, line 94:
```suggestion
            Entry::Occupied(_) => (),
```

Kerollmops (changes requested):
Thank you for your work!

If I understand correctly now, you store a `Cow<'ctx, [u8]>` instead of an `&'ctx [u8]` because you fetch multiple keys from different databases (that are CBO-encoded) and merge them with the `merge_cbo_roaring_bitmaps` function (which gives you a `Cow<'ctx, [u8]>`). You store that in the cache and deserialize the data from the raw bytes (like before).

If I am right, there is now the thing that I don't quite like: the previous version kept the `&'ctx [u8]` because it was a pointer in the LMDB database, so there was nearly nothing to keep in memory at the expense of a deserialization cost. However, storing a `Cow<'ctx, [u8]>::Owned` (so a `Vec<u8>`) isn't interesting. I prefer that we change the type to something like a `Either<&'ctx [u8], RoaringBitmap>` instead.

Additionally, I don't know if we can eliminate the new `BytesDecodeOwned` trait you introduced. I know we talked a bit about it, but maybe we can ask the help of @dureuill. Can't we just for it like: `DC::DItem + 'static`? This way, we are sure that the deserialized data doesn't depend on any lifetime for the DB.

ManyTheFish:
Hello @Kerollmops,
Thank you for your suggestions, after some tries, making the changes you suggested are not as trivial as I expected.
I propose to continue this work in a dedicated PR and merge this one.
Obviously, I will work on this this week. What do you think? Do you see any blockers to this? 😄

Kerollmops:
Hey @ManyTheFish,

You are right, let's merge this right now. I am impressed of the low number of code needed for the logic of the feature. I hope you find a way to improve the cache system for owned value next week 🤞

Bors merge

dureuill:
Could we consider rewriting the tests to use `insta` instead of `assert_eq`? it makes the experience of modifying the tests later if needed much smoother

Kerollmops:
> Could we consider rewriting the tests to use `insta` instead of `assert_eq`? it makes the experience of modifying the tests later if needed much smoother

That's a good idea. @ManyTheFish Could you do that, I'm pretty sure it will not take a lot of time ;)

ManyTheFish:
@Kerollmops @dureuill, yup! no problem!

Kerollmops (approved):
Looks perfect to me, thank you 🐝
bors merge

PR #3844, opened 2023-06-19
Use the input from the previous job instead of the workflow dispatch

curquiza:


.github/workflows/sdks-tests.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       meilisearch:
-        image: getmeili/meilisearch:${{ github.event.inputs.docker_image }}
+        image: getmeili/meilisearch:${{ needs.define-docker-image.outputs.docker-image }}
```
curquiza, line 40: Before, we were using the inputs from the format directly, so it worked for manual triggering but not for scheduled jobs...

Kerollmops (approved):
Thank you @curquiza 🤞

PR #3846, opened 2023-06-21
Gate score details behind an experimental flag

dureuill:
# Pull Request

## What does this PR do?
- Add a new `Option.experimental_score_details` option.
- Add struct `RouteFeatures` that is passed down to the routes as actix application data
- Add struct `NotEnabledFeatureError` that is returned if the `experimental_score_details` option is not set, but `showScoreDetails` is requested in a search query.
- Also return an error on attempts to get the metrics without the corresponding experimental feature


dureuill:
Superseded by #3850 that allows specifying this while the instance is running via a new route rather than by an instance configuration setting.

PR #3866, opened 2023-06-29
Update charabia v0.8.0

### LINKED ISSUE #3815: Phrase search doesn't return all document that matches the query
**Describe the bug**
Phrase search doesn't return all document that matches the query.

**To Reproduce**
1. Index the doc sample in this [discussion](https://github.com/meilisearch/meilisearch/discussions/3704)
2. Normal search returns both documents that match the search term
3. Phrase search doesn't return all documents that match the search term. 

**Expected behavior**
All documents with the matching search term or query should be returned. Also, highlighting is inconsistent for example the arrow-pointed term is even seen as a complete match. In v1.2 it's more inconsistent. 

**Screenshots**
Screenshot 2023-05-24 at 23 10 38

Screenshot 2023-05-24 at 23 16 16

Screenshot 2023-06-06 at 12 51 27

**Meilisearch version:** v1.1.1 and v1.2

ManyTheFish:
# Pull Request

Update Charabia:
- enhance Japanese segmentation
- enhance Latin Tokenization
  - words containing `_` are now properly segmented into several words
  - brackets `{([])}` are no more considered as context separators so word separated by brackets are now considered near together for the proximity ranking rule
- fixes #3815
- fixes #3778
- fixes [product#151](https://github.com/meilisearch/product/discussions/151)

> Important note: now the float numbers are segmented around the `.` so `3.22` is segmented as [`3`, `.`, `22`] but the middle dot isn't considered as a hard separator, which means that if we search `3.22` we find documents containing `3.22`


milli/src/search/new/tests/stop_words.rs
```diff
@@ -259,7 +259,7 @@ fn test_ignore_stop_words() {
             ),
             Proximity(
                 Rank {
-                    rank: 7,
+                    rank: 1,
```
dureuill, line 262: Super surprised at this change. What happened?

ManyTheFish: About this one, I was surprised by the old rank 🤔 the good rank should be `1` because there is only 1 searchable word in the query, no?

dureuill: let me check

dureuill: So there are 2 searchable words in the query, because the query is `to the dragon xyz`, the stop words are `["to", "The", "xyz"]` (note the capital T, stop words are case sensitive), so the possible matching words are "the" and "dragon".

The max_rank of proximity would be 1 if there was only a single word. Since the two matching words are very close ("the" and "dragon"), the expected result should indeed be close to the max (so 7 is more correct than 1)

PR #3889, opened 2023-07-05
Display the total number of tasks matching a filter/query

### LINKED ISSUE #3888: Updates to task endpoint
Related product team resources: 
Related product discussion:
Related spec: https://github.com/meilisearch/specifications/pull/253 

## Motivation

Give people a better sense of what's happening in their environment by being able to easily and accurately tell them how many tasks meet a filter criteria without having to page through the results and count them manually. 

## Usage

Enables easier UI building for task view work. 

* When calling the /tasks endpoint with no filters show the total number of tasks there are in the environment.
* When calling the /tasks endpoint with a filter (e.g. /tasks?statuses=enqueued ) show the total number of tasks that meet that filter criteria.

## TODO

- [x] Update the spec

## Impacted teams

@meilisearch/cloud-team @meilisearch/docs-team @meilisearch/integration-team

Kerollmops:
This PR returns a new field on the `/tasks` routes. The `total` field exposes the total number of tasks that matches the given filter/query. It is useful to display information on a user interface and can help understand when progress is made in processing tasks, i.e., the total number of tasks on `/tasks?statuses=succeeded` will increase over time.

Fixes #3888.

- [ ] Update the specs fo the `/tasks` route.

## How have I implemented it?

I found it much easier to run two times the task filtering system. Once with the original `from` and `limit` parameters and a second time without. The second call will return the total number of tasks that match the query, not only the number of tasks on the current page.

So far, in terms of performance, there doesn't seem to be any issue. I tried different filters with something like 250k tasks. Note that there is a limit of 1M tasks in the queue.

### DIFF
--- a/index-scheduler/src/lib.rs
+++ b/index-scheduler/src/lib.rs
@@ -2797,43 +2797,43 @@ mod tests {
 
         let rtxn = index_scheduler.env.read_txn().unwrap();
         let query = Query { limit: Some(0), ..Default::default() };
-        let tasks = index_scheduler
+        let (tasks, _) = index_scheduler
             .get_task_ids_from_authorized_indexes(&rtxn, &query, &AuthFilter::default())
             .unwrap();
         snapshot!(snapshot_bitmap(&tasks), @"[]");
 
         let query = Query { limit: Some(1), ..Default::default() };
-        let tasks = index_scheduler
+        let (tasks, _) = index_scheduler
             .get_task_ids_from_authorized_indexes(&rtxn, &query, &AuthFilter::default())
             .unwrap();
         snapshot!(snapshot_bitmap(&tasks), @"[2,]");
 
         let query = Query { limit: Some(2), ..Default::default() };
-        let tasks = index_scheduler
+        let (tasks, _) = index_scheduler
             .get_task_ids_from_authorized_indexes(&rtxn, &query, &AuthFilter::default())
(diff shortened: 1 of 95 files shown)


index-scheduler/src/lib.rs
```diff
@@ -835,7 +842,8 @@ impl IndexScheduler {
         rtxn: &RoTxn,
         query: &Query,
         filters: &meilisearch_auth::AuthFilter,
-    ) -> Result<RoaringBitmap> {
+    ) -> Result<(RoaringBitmap, u64)> {
+        let total_tasks = self.get_task_ids(rtxn, &query.clone().without_limits())?;
```
dureuill, line 846:
```suggestion
        // compute all tasks matching the filter by ignoring the limits, to find the number of tasks matching
        // the filter.
        // As this causes us to compute the filter twice it is slightly inefficient, but doing it this way spares
        // us from modifying the underlying implementation, and the performance remains sufficient.
        // Should this change, we would modify `get_task_ids` to directly return the number of matching tasks.
        let total_tasks = self.get_task_ids(rtxn, &query.clone().without_limits())?;
```

dureuill (commented):
Looks good, I think we should add a small comment (see suggestion)

Looks like some tests don't pass (see CI)

dureuill (approved):
bors merge

PR #3890, opened 2023-07-05
Fix the analytics of the sort facet values by count feature

Kerollmops:
This PR ensures we return the right analytics from the settings route.


dureuill (approved):
sure

bors merge

PR #3908, opened 2023-07-10
Allow a comma-separated value to the `vector` argument in GET search

dureuill:
# Pull Request

For request:

```
 curl \
  -X GET 'http://localhost:7700/indexes/movies/search?vector=0.123,1.124,244'
```

Before PR: 

```
{"message":"Invalid value type for parameter `vector`: expected a string, but found a string: `0,1,2`","code":"invalid_search_vector","type":"invalid_request","link":"https://docs.meilisearch.com/errors#invalid_search_vector"}%
```

After PR:

```
{"hits":[],"query":"","vector":[0.123,1.124,244.0],"processingTimeMs":0,"limit":20,"offset":0,"estimatedTotalHits":1000}%
```

cc @gmourier @bidoubiwa


dureuill:
> I am just pretty sure that [this limit](https://stackoverflow.com/a/76337704/1941280) can be reached quite easily 🙈

yay, makes me wonder if we want to provide the get parameter at all ...

PR #3913, opened 2023-07-10
Expose a Puffin server to profile the indexing process

Kerollmops:
This PR exposes a puffin HTTP server to expose the internal timing it takes to index documents, delete documents, or update the settings of an index.

Capture d’écran 2023-07-10 à 18 44 58

## To be done

 - [x] Move the puffin HTTP server under a feature flag.
 - [x] Use [the `puffin::set_scopes_on` function](https://docs.rs/puffin/latest/puffin/fn.set_scopes_on.html) to toggle it (by using the feature directly).
     When this function is called with `false`, [a call to `profile_scope!` talked 1-2ns](https://docs.rs/puffin/latest/puffin/fn.set_scopes_on.html).
 - [x] Create a _PROFILING.md_ file explaining how to use it.
   - [x] Explain that merging scopes on the interface is not always useful.
 - [x] Add more info on the number of batched tasks (using the `puffin::profile_scope!` macro data).
   - I added more info, but that's more continuous work when we consider we need more info here and there.
 - [x] Clean up some scopes, and don't touch too much code to inject puffin.
   - I am not sure that the _index_documents/mod.rs_ function is that complex with the addition of the scope.
 - [x] Think about what we consider frames. One indexation operation or the wall program. When must we stop the frame, then?
   - What we consider a frame is one single `IndexScheduler::tick` execution.
   - We can change that later.


milli/src/update/index_documents/mod.rs
```diff
@@ -322,6 +330,7 @@ where
 
         // Run extraction pipeline in parallel.
         pool.install(|| {
+            puffin::profile_scope!("split_grenad_by_chunks");
```
ManyTheFish, line 333: This profile scope doesn't seem to only measure the `split_grenad_by_chunks` process but`data_from_obkv_documents` too.

ManyTheFish (dismissed):
Hello @Kerollmops, 
This puffin measurement can only be activated via a compile-time feature flag,
what about making it easier to activate via an experimental feature? This would be a handy tool for debugging cloud performance issues.

> poke @macraig on this

ManyTheFish (approved):
Approved

Kerollmops:
bors merge

PR #3929, opened 2023-07-17
Fix a panic when sorting geo fields represented by strings

### LINKED ISSUE #3927: task 577 panicked when using "sort":["_geoPoint(46.4982953,11.3547582):asc"]
Have around 11k documents in my index but in this case i'm event filtering the results by id like so:

`"filter" : ["type=1", "_facet_something=1", "platform_ground=1", "id IN [11137]"]`

but when trying to sort the results by distance I get the following error:

```
{
    "message": "task 577 panicked",
    "code": "internal",
    "type": "internal",
}
```

Note: Not all the documents have a `_geo`attribute (we do not have this information in all the cases)

Kerollmops:
This issue fixes #3927 by retrieving and parsing the original string values into f64s. I also added a test to ensure we don't break it in a future version.


ManyTheFish (approved):
LGTM

Kerollmops:
Thank you for the review 🚴
bors merge

jmarcher:
Nice

PR #3933, opened 2023-07-18
Stop computing the update files size

Kerollmops:
This PR removes the part which computes the total size of the `data.ms/update_files` folder, which can take a lot of time when many updates must be processed.

It is not breaking API-side but is breaking on the result we will show to the user. The `databaseSize` field returned by the `/stats` endpoint will be reduced.


ManyTheFish (approved):
LGTM!
bors merge

PR #3942, opened 2023-07-20
Normalize for the search the facets values

Kerollmops:
This PR improves and fixes the search for facet values feature. Searching for _bre_ wasn't returning facet values like _brévent_ or _brô_.

The issue was related to the fact that facets are normalized but not in the same way as the `searchableAttributes` are. We decided to normalize them further and add another intermediate database where the key is the normalized facet value, and the value is a set of the non-normalized facets. We then use these non-normalized ones to get the correct counts by fetching the associated databases.

### What's missing in this PR?
 - [x] Apply the change to the whole set of `SearchForFacetValue::execute` conditions.
 - [x] Factorize the code that does an intermediate normalized value fetch in a function.
 - [x] Add or modify the search for facet value test.

### DIFF
--- a/meilisearch/tests/search/facet_search.rs
+++ b/meilisearch/tests/search/facet_search.rs
@@ -57,6 +57,54 @@ async fn simple_facet_search() {
 }
 
 #[actix_rt::test]
+async fn advanced_facet_search() {
+    let server = Server::new().await;
+    let index = server.index("test");
+
+    let documents = DOCUMENTS.clone();
+    index.update_settings_filterable_attributes(json!(["genres"])).await;
+    index.update_settings_typo_tolerance(json!({ "enabled": false })).await;
+    index.add_documents(documents, None).await;
+    index.wait_task(2).await;
+
+    let (response, code) =
+        index.facet_search(json!({"facetName": "genres", "facetQuery": "adventre"})).await;
+
+    assert_eq!(code, 200, "{}", response);
+    assert_eq!(dbg!(response)["facetHits"].as_array().unwrap().len(), 0);
+
+    let (response, code) =
+        index.facet_search(json!({"facetName": "genres", "facetQuery": "àdventure"})).await;
+
+    assert_eq!(code, 200, "{}", response);
+    assert_eq!(response["facetHits"].as_array().unwrap().len(), 1);
+}
+
+#[actix_rt::test]
+async fn more_advanced_facet_search() {
+    let server = Server::new().await;
+    let index = server.index("test");
+
+    let documents = DOCUMENTS.clone();
+    index.update_settings_filterable_attributes(json!(["genres"])).await;
+    index.update_settings_typo_tolerance(json!({ "disableOnWords": ["adventre"] })).await;
+    index.add_documents(documents, None).await;
+    index.wait_task(2).await;
+
+    let (response, code) =
+        index.facet_search(json!({"facetName": "genres", "facetQuery": "adventre"})).await;
+
+    assert_eq!(code, 200, "{}", response);
+    assert_eq!(dbg!(response)["facetHits"].as_array().unwrap().len(), 0);
+
+    let (response, code) =
+        index.facet_search(json!({"facetName": "genres", "facetQuery": "adventure"})).await;
+
+    assert_eq!(code, 200, "{}", response);
+    assert_eq!(response["facetHits"].as_array().unwrap().len(), 1);
+}
+
+#[actix_rt::test]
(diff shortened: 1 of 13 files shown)


meilisearch/tests/search/facet_search.rs
```diff
@@ -56,6 +56,54 @@ async fn simple_facet_search() {
...
+    index.update_settings_filterable_attributes(json!(["genres"])).await;
+    index.update_settings_typo_tolerance(json!({ "enabled": false })).await;
+    index.add_documents(documents, None).await;
+    index.wait_task(2).await;
+
+    let (response, code) =
+        index.facet_search(json!({"facetName": "genres", "facetQuery": "adventre"})).await;
+
+    assert_eq!(code, 200, "{}", response);
+    assert_eq!(dbg!(response)["facetHits"].as_array().unwrap().len(), 0);
```
ManyTheFish, line 74:
```suggestion
    snapshot!(code, @"200 OK");
    snapshot!(response["facetHits"].as_array().unwrap().len(), @"0");
```

meilisearch/tests/search/facet_search.rs
```diff
@@ -56,6 +56,54 @@ async fn simple_facet_search() {
...
+
+    let (response, code) =
+        index.facet_search(json!({"facetName": "genres", "facetQuery": "àdventure"})).await;
+
+    assert_eq!(code, 200, "{}", response);
+    assert_eq!(response["facetHits"].as_array().unwrap().len(), 1);
```
ManyTheFish, line 80:
```suggestion
    snapshot!(code, @"200 OK");
    snapshot!(response["facetHits"].as_array().unwrap().len(), @"1");
```

meilisearch/tests/search/facet_search.rs
```diff
@@ -56,6 +56,54 @@ async fn simple_facet_search() {
...
+    index.update_settings_filterable_attributes(json!(["genres"])).await;
+    index.update_settings_typo_tolerance(json!({ "disableOnWords": ["adventre"] })).await;
+    index.add_documents(documents, None).await;
+    index.wait_task(2).await;
+
+    let (response, code) =
+        index.facet_search(json!({"facetName": "genres", "facetQuery": "adventre"})).await;
+
+    assert_eq!(code, 200, "{}", response);
+    assert_eq!(dbg!(response)["facetHits"].as_array().unwrap().len(), 0);
```
ManyTheFish, line 98:
```suggestion
    snapshot!(code, @"200 OK");
    snapshot!(response["facetHits"].as_array().unwrap().len(), @"0");
```

meilisearch/tests/search/facet_search.rs
```diff
@@ -56,6 +56,54 @@ async fn simple_facet_search() {
...
+
+    let (response, code) =
+        index.facet_search(json!({"facetName": "genres", "facetQuery": "adventure"})).await;
+
+    assert_eq!(code, 200, "{}", response);
+    assert_eq!(response["facetHits"].as_array().unwrap().len(), 1);
```
ManyTheFish, line 104:
```suggestion
    snapshot!(code, @"200 OK");
    snapshot!(response["facetHits"].as_array().unwrap().len(), @"1");
```

milli/src/update/index_documents/helpers/merge_functions.rs
```diff
@@ -44,6 +45,27 @@ pub fn merge_roaring_bitmaps<'a>(_key: &[u8], values: &[Cow<'a, [u8]>]) -> Resul
...
+        let strings: BTreeSet<String> = values
+            .iter()
+            .map(AsRef::as_ref)
+            .map(serde_json::from_slice::<BTreeSet<String>>)
+            .map(StdResult::unwrap)
+            .reduce(|mut current, new| {
+                for x in new {
+                    current.insert(x);
+                }
+                current
```
ManyTheFish, line 62: What do you think of [using `union`](https://doc.rust-lang.org/stable/std/collections/struct.BTreeSet.html#method.union) instead of using a for `loop`?

It's not as practical as `RoaringBitmap` unfortunately.

ManyTheFish (changes requested):

milli/src/update/index_documents/helpers/merge_functions.rs
Kerollmops: I will not use the `BitOr` trait as, unfortunately, it will clone the values. Here, we move the values from one set to the other without duplication.

ManyTheFish (approved):
bors merge

PR #3950, opened 2023-07-25
Fix matches positions

### LINKED ISSUE #3938: Wrong matchesPosition 'start' for Cyrillic text.
First of all thank you for this wonderful piece of tech! Such a delight to work with.
And second, I'm only begin to work with it, so it might be incorrect configuration from my part, please point me in right direction if it is so.

The problem example:
title == `Разработка заданий на развитие функциональной грамотности по теме "Африка"`
query == `Африка`
matchesPosition.title.0 == `{start: 125, length: 6}`

So the length is correct, but the start is not. Whole title is 74 characters long. If the matching query at the beginning of the string - all good, start is 0.

And other thing: `_formatted` response is highlighting correct part of the string:
`Разработка заданий на развитие функциональной грамотности по теме "<ais-highlight-0000000000>Африка</ais-highlight-0000000000>"`

I'm using getmeili/meilisearch:latest inside *docker compose* with [symfony bundle](https://github.com/meilisearch/meilisearch-symfony).

And I only assume that the problem with Cyrillic texts, because it is the only kind of texts I have.

Thank you and have a wonderful day

ManyTheFish:
# Pull Request

The match positions object returned in `_matchesPosition` is wrong, the `start` field is in bytes length but the `len` field is in character length, both should be in bytes.

related documentation: https://www.meilisearch.com/docs/reference/api/search#show-matches-position

This PR uses now the `char_len` and the `byte_len` coming from Charabia, but a panic appears when we try to partially highlight a word containing diacritics.
To be completely solved this PR may need a fix on Charabia side, it surely comes from the below part:
https://github.com/meilisearch/charabia/blob/75e5e087767a0850eaab09c2fef237f7cf2ff513/charabia/src/token.rs#L112-L163

## Related issue
Fixes #3938

### DIFF
--- a/meilisearch/tests/search/formatted.rs
+++ b/meilisearch/tests/search/formatted.rs
@@ -1,4 +1,5 @@
 use insta::{allow_duplicates, assert_json_snapshot};
+use meili_snap::snapshot;
 use serde_json::json;
 
 use super::*;
@@ -32,7 +33,7 @@ async fn formatted_contain_wildcard() {
                   "cattos": [
                     {
                       "start": 0,
-                      "length": 5
+                      "length": 6
                     }
                   ]
                 }
@@ -79,7 +80,7 @@ async fn formatted_contain_wildcard() {
                       "cattos": [
                         {
                           "start": 0,
-                          "length": 5
+                          "length": 6
                         }
                       ]
                     }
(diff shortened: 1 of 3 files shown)


gillian-meilisearch:
Removing from 1.3 after a quick chat with Many :) We'll try get it in a later release

PR #3952, opened 2023-07-26
Use the new safe `read-txn-no-tls` heed feature

Kerollmops:
[We recently found out](https://github.com/meilisearch/heed/issues/191#issuecomment-1650280513) that the `read-sync-txn` heed feature was invalid and must be removed from this crate. We were declaring it in milli/meilisearch but, fortunately, not sharing the `RoTxn`s across threads 😮‍💨

[I recently introduced the `read-txn-no-tls` heed feature](https://github.com/meilisearch/heed/pull/194), which implements `RoTxn: Send` and allows multiple read transactions on a single thread (which we use).

This PR replaces the `sync-read-txn` heed feature with the `read-txn-no-tls` one in the _Cargo.toml_ file. I will fix this in heed v0.20.0 and will fill a RustSec advisory in the meantime.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1737,8 +1737,8 @@ checksum = "95505c38b4572b2d910cecb0281560f54b440a19336cbbcb27bf6ce6adc6f5a8"
 
 [[package]]
 name = "heed"
-version = "0.12.5"
-source = "git+https://github.com/meilisearch/heed?tag=v0.12.6#8c5b94225fc949c02bb7b900cc50ffaf6b584b1e"
+version = "0.12.7"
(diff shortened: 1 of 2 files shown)


curquiza (dismissed):
Checked with @Kerollmops, let's merge it!

bors merge

Kerollmops:
It is failing due to the removal of the `sync-read-txn` feature (obviously), which wasn't only implementing `RoTxn: Sync` but also making it possible to have multiple `RoTxn` on the same thread. This is why I renamed the flag and [I will be releasing a new version of heed v0.12.7](https://github.com/meilisearch/heed/pull/194), which introduces a new valid and safe `read-txn-no-tls` feature implementing `RoTxn: Send` and making it possible to have multiple `RoTxn` on the same thread.

ManyTheFish (approved):
bors merge

Issue #3962, opened 2023-07-31
Fix the benchmarks


irevoire:
The benchmarks do not run anymore, see https://github.com/meilisearch/meilisearch/actions/runs/5598028794/job/15163405092

PR #3963, opened 2023-07-31
Fix the milli crate

irevoire:
Milli was using the serde feature of either without enabling it first; thus, it wasn't working.

It was working in meilisearch, though, because `meilisearch-types` was using the feature which enables it globally for all the other crates.

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/3962


ManyTheFish (approved):
Thank you @irevoire,

bors merge

PR #3971, opened 2023-08-02
Zookeeper ha

irevoire:
- [ ] Expose the thing as an experimental feature
- [ ] Should we do something about the encryption of the exchange with the zookeeper or is it ok like that?


Kerollmops:
Thanks @irevoire for the huge work on that. Unfortunately, this PR is very far behind, and we will not land it in main any time soon and we have recently released [replicated sharding in Meilisearch v1.37](https://github.com/meilisearch/meilisearch/releases/tag/v1.37.0#v137-replicated-sharding) 🎉

Issue #3972, opened 2023-08-04
Prometheus `meilisearch_nb_tasks` seems to report nothing


gmourier:
Mentioned by @gabrieltz [here](https://github.com/meilisearch/product/discussions/625#discussioncomment-6612886) 

"I upgraded to version 1.3.0 yesterday and used some of the new metrics, however i noticed that the metric meilisearch_nb_tasks{instance="XXXXXX:7700", job="meilisearch", kind="statuses", value="processing"} is zero all the time, even while importing a few million documents in over 700 batches. The /tasks endpoint values were correct though"

irevoire:
I reproduced it; it’s a bug

curquiza:
Will be fixed in v1.3.1 and release this week or next week

curquiza:
Fixed by #3976

PR #3976, opened 2023-08-07
Fix the get stats method

irevoire:
# Pull Request

- The get stats method of the index-scheduler was not using at all the processing tasks. That was returning a wrong number of enqueued tasks and 0 processing tasks.
- Added a test
- Currently this method was **ONLY** used to compute the `meilisearch_nb_tasks` field of the **experimental feature** metrics.

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/3972

### DIFF
--- a/index-scheduler/src/lib.rs
+++ b/index-scheduler/src/lib.rs
@@ -790,10 +790,19 @@ impl IndexScheduler {
 
         let mut res = BTreeMap::new();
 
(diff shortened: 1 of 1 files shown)


ManyTheFish (approved):
LGTM!

ManyTheFish:
bors merge

Issue #3973, opened 2023-08-04
`_geoBoundingBox` filter returns no results when `lat` and `lng` are indexed as strings


martinmaillard:
**Describe the bug**

When indexing documents with a `_geo` key containing `lat` and `lng` as string values, the `_geoBoundingBox` filter never returns any resuls. Reindexing the same documents with `lat` and `lng` as `float` values solves the issue.

I'm reporting this as a bug because [the documentation states that `string` values are valid for `lat` and `lng`](https://www.meilisearch.com/docs/learn/fine_tuning_results/geosearch#preparing-documents-for-location-based-search).

> When using JSON and NDJSON, _geo must contain an object with two keys: lat and lng. Both fields must contain either a floating point number or a string indicating, respectively, latitude and longitude:

Note that `_geoRadius` works as expected.

**To Reproduce**
1. Index the following documents: `[{"id": 1, "_geo": {"lng": "6", "lat": "46"}}, {"id": 2, "_geo": {"lng": 6, "lat": 46}]`
2. Query the index with the following filter: `{"filter": "_geoBoundingBox([47, 7], [44, 4])"}`
3. Only the document with `id=2` is returned

**Expected behavior**
Both documents should be returned

**Meilisearch version:**
Tested with 1.2 and 1.3

------

## Comment from the team:
This issue was an indexing issue. The fix will be shipped in a patch version which means you don't need a dump to use it.
For all the new documents you send, it'll take effect immediately BUT, if your database already contains geo documents with their coordinates as string, you won't be able to query them until you re-index them by removing and then adding again `_geo` as a faceted attribute (i.e.: removing it from both the sortable and filterable attributes).
You can also delete and send back all the affected documents if that's easier for you.

irevoire:
Hey @martinmaillard.

It seems like you mixed up a bit the order of your latitude/longitude.

[As you can see in the documentation](https://www.meilisearch.com/docs/learn/fine_tuning_results/geosearch#_geoboundingbox), the `_geoBoundingBox` function must receive the latitude and then the longitude, i.e.: `_geoBoundingBox([{lat}, {lng}], [{lat}, {lng}])`.
And as stated later:
> The first array indicates the geographic coordinates of the top right corner of the rectangular area. The second array indicates the coordinates of the bottom left corner of the rectangular area.

Following these indications, this means the top right corner should be expressed as `[47, 7]`, and your bottom left corner as `[44, 6]`.
[...]

PR #3986, opened 2023-08-08
Fix geo bounding box with strings

irevoire:
# Pull Request

When sending a document with one geofield of type string (i.e.: `{ "_geo": { "lat": 12, "lng": "13" }}`), the geobounding box would exclude this document.

This PR fixes this issue by automatically parsing the string value in case we're working on a geofield.

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/3973

## What does this PR do?
- Automatically parse the facet value iif we're working on a geofield.
- Make insta works with snapshots in loops or closure executed multiple times. (you may need to update your cli if it panics after this PR: `cargo install cargo-insta`).
- Add one integration test in milli and in meilisearch to ensure it works forever.
- Add three snapshots for the dump that mysteriously disappeared I don't know how

### DIFF
--- a/dump/src/reader/snapshots/dump__reader__test__import_dump_v1-10.snap
+++ b/dump/src/reader/snapshots/dump__reader__test__import_dump_v1-10.snap
@@ -1,0 +1,24 @@
+---
+source: dump/src/reader/mod.rs
+expression: spells.settings().unwrap()
+---
+{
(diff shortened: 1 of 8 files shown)


ManyTheFish (approved):
LGTM!

ManyTheFish:
bors merge

PR #4062, opened 2023-09-14
Prototype limit the number batched tasks

irevoire:
Hey everyone 👋

Since we don't have a nice solution yet to solve this issue, we’ll expose this new experimental flag that lets you limit the maximum number of tasks meilisearch can [batch together](https://www.meilisearch.com/docs/learn/core_concepts/documents#auto-batching).

### How to use the feature?

- Launch meilisearch using the `--experimental-limit-batched-tasks` CLI flag or the `MEILI_EXPERIMENTAL_LIMIT_BATCHED_TASKS` environment variable by specifying how many tasks you want to batch together.
- Finding a good value is very dependent on your use case, you should try to trigger this error multiple times and see how many tasks were processing at the same time. The [experimental metrics feature](https://github.com/meilisearch/product/discussions/625) could be of great use to determine this value since it shows you in real time how many tasks are processing.

### DIFF
--- a/index-scheduler/src/batch.rs
+++ b/index-scheduler/src/batch.rs
@@ -587,7 +587,9 @@ impl IndexScheduler {
         let index_tasks = self.index_tasks(rtxn, index_name)? & enqueued;
 
         // If autobatching is disabled we only take one task at a time.
-        let tasks_limit = if self.autobatching_enabled { usize::MAX } else { 1 };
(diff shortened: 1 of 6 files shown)


curquiza:
@irevoire can we close this PR since we have released [this experimental feature](https://github.com/orgs/meilisearch/discussions/713) in v1.6.0?

PR #4038, opened 2023-09-06
Fix filter escaping issues

Kerollmops:
This PR fixes #4034 by always escaping the sequences. Users must always put quotes (simple or double) to escape the filter values.

PR #4069, opened 2023-09-19
Revert the fix on the filters about escaping sequences

Kerollmops:
This PR reverts #4038 as it introduces a breaking change. However, the fix will be kept for the v1.4.0 release.


irevoire (approved):
bors merge

PR #4089, opened 2023-09-28
Use a bufreader and bufwriter everytime there is a grenad<file>

### LINKED ISSUE #4087: Make dump import process faster
**Describe the bug**
We found out that some dumps are really slow to import for no clear reason.
A customer sent us a dump that was taking 2h to import on my machine and multiple days on the cloud (it never finished, actually).

On a first investigation, I found out that most of the time was spent on the `read` syscall inside of the grenad operation.
image

**To Reproduce**
1. Create a dump with indexes that contain a lot of documents, each containing a lot of text
2. import the dump

**Expected behavior**
Is fast.

**Meilisearch version:**: all

**Additional context**
Additional information that may be relevant to the issue.
[e.g. architecture, device, OS, browser]

irevoire:
# Pull Request
Wrap all the files we give to a grenad in a `BufReader` or `BufWriter`.

The dump import I tried in the issue went from 2h to 10 minutes on my machine.

I also ran a bunch of benchmarks on my machine, and we're faster by a few seconds everywhere but nothing huge.

-----

The one thing I’m afraid about is if we used to get the inner file in a grenad and then do a read right after without a seek at the beginning of the file or a reopen.
Since we now use a bufreader our read would return the bytes one buffer later and probably completely corrupt what we were supposed to read.

From what I see, it looks like it works, but I may have missed something, I don't know much about this part of the codebase.

This issue should not arise on the bufwriter, though, because if we're not able to write the content of the buffer I ensured that the `into_inner` of the bufwriter should return an internal error.

## Related issue
Fixes #4087

### DIFF
--- a/milli/src/error.rs
+++ b/milli/src/error.rs
@@ -47,6 +47,8 @@ pub enum InternalError {
     IndexingMergingKeys { process: &'static str },
     #[error("{}", HeedError::InvalidDatabaseTyping)]
     InvalidDatabaseTyping,
+    #[error("Could not access the inner of a buf-reader/writer: {0}")]
+    BufIntoInnerError(String),
     #[error(transparent)]
     RayonThreadPool(#[from] ThreadPoolBuildError),
     #[error(transparent)]
--- a/milli/src/error.rs
+++ b/milli/src/error.rs
@@ -47,8 +47,6 @@ pub enum InternalError {
     IndexingMergingKeys { process: &'static str },
     #[error("{}", HeedError::InvalidDatabaseTyping)]
     InvalidDatabaseTyping,
-    #[error("Could not access the inner of a buf-reader/writer: {0}")]
-    BufIntoInnerError(String),
     #[error(transparent)]
     RayonThreadPool(#[from] ThreadPoolBuildError),
     #[error(transparent)]
--- a/milli/src/update/index_documents/helpers/grenad_helpers.rs
+++ b/milli/src/update/index_documents/helpers/grenad_helpers.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 use std::fs::File;
-use std::io::{self, Seek};
+use std::io::{self, BufReader, BufWriter, Seek};
 use std::time::Instant;
 
 use grenad::{CompressionType, Sorter};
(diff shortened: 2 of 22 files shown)


milli/src/update/index_documents/helpers/grenad_helpers.rs
```diff
@@ -64,16 +64,21 @@ pub fn sorter_into_reader(
...
 
-pub fn writer_into_reader(writer: grenad::Writer<File>) -> Result<grenad::Reader<File>> {
-    let mut file = writer.into_inner()?;
+pub fn writer_into_reader(
+    writer: grenad::Writer<BufWriter<File>>,
+) -> Result<grenad::Reader<BufReader<File>>> {
+    let mut file = writer
+        .into_inner()?
+        .into_inner()
+        .map_err(|err| InternalError::BufIntoInnerError(err.to_string()))?;
```
Kerollmops, line 73: Can't we use the [`IntoInnerError::into_error`](https://doc.rust-lang.org/std/io/struct.IntoInnerError.html#method.into_error) method to return a classic `io::Error` instead?

milli/src/error.rs
```diff
@@ -47,6 +47,8 @@ pub enum InternalError {
     IndexingMergingKeys { process: &'static str },
     #[error("{}", HeedError::InvalidDatabaseTyping)]
     InvalidDatabaseTyping,
+    #[error("Could not access the inner of a buf-reader/writer: {0}")]
+    BufIntoInnerError(String),
```
Kerollmops, line 51: I don't know if we can eliminate this variant or simply change this string into an `io::Error`. See my other comment.

Kerollmops (changes requested):
> The one thing I’m afraid about is if we used to get the inner file in a grenad and then do a read right after without a seek at the beginning of the file or a reopen.
> Since we now use a bufreader our read would return the bytes one buffer later and probably completely corrupt what we were supposed to read.

What you describe is likely not to happen as you may have noticed that we always use `File` and never a `&File`. Therefore, need to call `BufRead/BufWriter::into_inner` to get the inner file and do read operations on it. This method forces a flush of the content to disk before we will read. It is, therefore, safe to read in it after extracting it.

curquiza (changes requested):
Can you rebase on `release-v1.4.1`? 😇

milli/src/update/index_documents/helpers/grenad_helpers.rs
irevoire: Oh yes nice, didn't know it existed!

Kerollmops:
bors merge

Kerollmops:
The issue seems related to GitHub as we are not running on our `benchmarks` machines but on the official `Ubuntu` ones...

curquiza:
bors merge

Issue #4086, opened 2023-09-27
Make the experimental route `/metrics` activable via HTTP too


curquiza:
We would like to make the experimental “Prometheus metrics” feature accessible via the HTTP API, as it is for the [score details](https://github.com/meilisearch/product/discussions/674) and [vector search](https://github.com/meilisearch/product/discussions/677) experimental features.

More info regarding this exp feature: https://github.com/meilisearch/product/discussions/625

### Activate the feature:

The experimental feature to activate the route `/metrics` is currently activable using a flag `--experimental-enable-metrics`

```jsx
curl \
  -X PATCH 'http://localhost:7700/experimental-features/' \
  -H 'Content-Type: application/json'  \
--data-binary '{
    "metrics": true
  }'
```

### Getting the information about the feature

```jsx
curl -X GET 'http://localhost:7700/experimental-features/'

{
    "scoreDetails": false,
    "vectorStore": false,
    "metrics": true
}
```

## TODO

- [ ]  Make it available via HTTP as described above
- [ ]  The users can still launch Meilisearch using the `--experimental-enable-metrics` flag.
- [ ]  If the flag `--experimental-enable-metrics` is activated, a call to the `GET /experimental-features` route right after the launch will show `"metrics": true` even if the user has not called the `PATCH /experimental-features` route yet.
- [ ]  Even if the `--experimental-enable-metrics` flag is present at launch, calling the `PATCH /experimental-features` route with `"metrics": false` disables the experimental feature.
- [ ]  Update the spec

## Impacted teams
@meilisearch/cloud-team @meilisearch/docs-team

aneesh9909:
Hello! I would like to pick up this issue but it's my first time touching this codebase so I wanted to confirm if I've understood this correctly.

My understanding of this issue is that the `metrics` experimental feature needs to be added to the [RuntimeTogglableFeatures](https://github.com/meilisearch/meilisearch/blob/86b314626dc846b68972b62a2ec3188be09ff035/meilisearch/src/routes/features.rs#L42C1-L42C1)

curquiza:
Hello @aneesh9909 

For your information, we prefer not assigning people to our issues because sometimes people ask to be assigned and never come back, which discourages the volunteer contributors from opening a PR to fix this issue.
We will accept and merge the first PR that fixes correctly and well implements the issue following our [contributing guidelines](../blob/main/CONTRIBUTING.md).   

We are looking forward to reviewing your PR 😊

[...]

PR #4128, opened 2023-10-16
Route /metrics activable via HTTP

vivek-26:
# Pull Request

## Related issue

Closes [#4086](https://github.com/meilisearch/meilisearch/issues/4086)

## What does this PR do?

- [x]  Make `/metrics` available via HTTP as described in [Make the experimental route `/metrics` activable via HTTP too #4086](https://github.com/meilisearch/meilisearch/issues/4086)
- [x]  The users can still launch Meilisearch using the `--experimental-enable-metrics` flag.
- [x]  If the flag `--experimental-enable-metrics` is activated, a call to the `GET /experimental-features` route right after the launch will show `"metrics": true` even if the user has not called the `PATCH /experimental-features` route yet.
- [x]  Even if the --experimental-enable-metrics flag is present at launch, calling the `PATCH /experimental-features` route with `"metrics": false` disables the experimental feature.
- [ ]  Update the spec
    - Not sure where to update the documentation. Requesting help from Meilisearch team! 

## Approach Taken
In order to solve this problem, I am keeping track of whether the `metrics` feature has been updated using `PATCH /experimental-features`. The reason to do this is explained in the below with the help of a scenario - 
- Let's assume that initially the `metrics` feature is disabled at runtime (which is persisted in DB).
- I start the Meilisearch server with `--experimental-enable-metrics` flag. Starting the server this way, should enable the `metrics` feature only for this server instance (assuming the user does not call `PATCH /experimental-features`).
- Now if I persist the `metrics` feature value `true` to the database when the server starts, it would be wrong - because if the user never updated the `metrics` feature using `PATCH /experimental-features` and restarted the server without `--experimental-enable-metrics` flag, in the next run, `metrics` would be enabled.
- To handle this, it is necessary to maintain whether `metrics` feature was updated by user or not. If yes, then we also keep track of it's current value.
- Since the `metrics` can be toggled at runtime, the [metrics middleware](https://github.com/meilisearch/meilisearch/blob/main/meilisearch/src/middleware.rs#L48-L74) should also run only when its activated.
- The metrics middleware runs before each request's business logic, and, making a call to database to fetch the current status of metrics feature will degrade the performance. That's why we keep track of the current value as well in memory and use a `RWMutex` to allow concurrent reads.

@Kerollmops Let me know if the approach and it's implementation are optimal. Requesting your review.
cc/ @curquiza 

## PR checklist

Please check if your PR fulfils the following requirements:

- [x]  Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [x]  Have you read the contributing guidelines?
- [x]  Have you made sure that the title is accurate and descriptive of the changes?

### DIFF
--- a/index-scheduler/src/features.rs
+++ b/index-scheduler/src/features.rs
@@ -6,7 +6,8 @@ use meilisearch_types::features::{
 use meilisearch_types::heed::types::{SerdeJson, Str};
 use meilisearch_types::heed::{Database, Env, RoTxn, RwTxn};
 
-use crate::error::{Error::RuntimeFeatureToggleError, FeatureNotEnabledError};
+use crate::error::Error::RuntimeFeatureToggleError;
+use crate::error::FeatureNotEnabledError;
 use crate::Result;
 
 const EXPERIMENTAL_FEATURES: &str = "experimental-features";
--- a/meilisearch/tests/auth/authorization.rs
+++ b/meilisearch/tests/auth/authorization.rs
@@ -2,10 +2,11 @@ use std::collections::{HashMap, HashSet};
 
 use ::time::format_description::well_known::Rfc3339;
 use maplit::{hashmap, hashset};
+use meilisearch::Opt;
 use once_cell::sync::Lazy;
 use time::{Duration, OffsetDateTime};
 
-use crate::common::{Server, Value};
+use crate::common::{default_settings, Server, Value};
 use crate::json;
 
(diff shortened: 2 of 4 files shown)


curquiza:
hello @vivek-26 

I saw you just opened your PR.
Someone already created a PR 3 days ago here: https://github.com/meilisearch/meilisearch/pull/4126

Would be awesome to have your review on it 😊

Issue #3983, opened 2023-08-08
Stops sending the `http_response_time_seconds_count` in the metrics because it overwhelms prometheus


irevoire:
The `http_response_time_seconds_count` field sends way too much information, and that slows down Prometheus a lot and makes it use a lot of memory (the same is probably true for meilisearch as well, even if it’s less impactful).
We should get rid of it and group the route behind a common denominator.
Maybe instead of exposing these metrics that updates with every possible route, we should instead send a fixed number of route.
Something like:
- `http_response_time_tasks` for all the tasks
- `http_response_time_search` parametrized by the index name for all the searches.
- Is that enough? Should we share more, like the time to add a document?

Reported by @miiton here: https://github.com/meilisearch/product/discussions/625#discussioncomment-6600928

------

I see that the Prometheus process is consuming a lot of CPU/Memory due to the large number of values related to TASKs.
I think this is a point for improvement.

```
http_response_time_seconds_count{method="GET",path="/tasks/1"} 1
http_response_time_seconds_bucket{method="GET",path="/tasks/1",le="0.0005"} 1
http_response_time_seconds_bucket{method="GET",path="/tasks/1",le="0.0008"} 1
http_response_time_seconds_bucket{method="GET",path="/tasks/1",le="0.00085"} 1
http_response_time_seconds_bucket{method="GET",path="/tasks/1",le="0.0009"} 1
...
...
...
http_response_time_seconds_count{method="GET",path="/tasks/9977"} 1
http_response_time_seconds_bucket{method="GET",path="/tasks/9978",le="0.0005"} 1
http_response_time_seconds_bucket{method="GET",path="/tasks/9978",le="0.0008"} 1
http_response_time_seconds_bucket{method="GET",path="/tasks/9978",le="0.00085"} 1
http_response_time_seconds_bucket{method="GET",path="/tasks/9978",le="0.0009"} 1
...
```

**with scrape meilisearch metrics**
![image](https://github.com/meilisearch/product/assets/468745/56723350-382a-41f9-8518-c1f68fe1fbce)

**without scrape meilisearch metrics**
![image](https://github.com/meilisearch/product/assets/468745/1a21055e-27ae-428b-bfb2-9bc74d0cd41e)

- Meilisearch v1.2.0
- Prometheus v2.45.0

itsajay1029:
I'd love to work on this @irevoire. Can you assign it to me and explain how to resolve this ?

Kerollmops:
[...]

PR #4145, opened 2023-10-22
FIX: Exclude unneeded routes from prometheus metrics

Zarathustra2:
# Pull Request

## Related issue
Fixes [#3983](https://github.com/meilisearch/meilisearch/issues/3983)

## What does this PR do?
Excludes the following routes from being observed for prometheus metrics: 
- `/tasks/{task_id}`
- `/manifest.json`
- Anything starting with `/fonts/`
- Anything starting with `/static/`
- Anything starting with `/favicon`

### DIFF
--- a/meilisearch/src/middleware.rs
+++ b/meilisearch/src/middleware.rs
@@ -48,7 +48,21 @@ where
     fn call(&self, req: ServiceRequest) -> Self::Future {
         let mut histogram_timer: Option<HistogramTimer> = None;
         let request_path = req.path();
-        let is_registered_resource = req.resource_map().has_resource(request_path);
+        let is_registered_resource = match req.resource_map().match_pattern(request_path).as_deref()
+        {
+            None => false,
+            Some("/tasks/{task_id}") => false,
+            Some("/manifest.json") => false,
+            Some(x)
+                if x.starts_with("/fonts/")
+                    || x.starts_with("/static/")
+                    || x.starts_with("/favicon") =>
+            {
+                false
+            }
+            _ => true,
+        };
+
         if is_registered_resource {
             let request_method = req.method().to_string();
             histogram_timer = Some(
--- a/meilisearch/src/middleware.rs
+++ b/meilisearch/src/middleware.rs
@@ -57,30 +57,34 @@ where
 
         if features.check_metrics().is_ok() {
             let request_path = req.path();
-            let is_registered_resource = match req.resource_map().match_pattern(request_path).as_deref()
-            {
-                None => false,
-                Some("/tasks/{task_id}") => false,
-                Some("/manifest.json") => false,
+            let generic_route = req.resource_map().match_pattern(request_path);
+
+            // We group certain routes together as we otherwise would generate too many
+            // prometheus metric with no additional value.
+            // See: https://github.com/meilisearch/meilisearch/issues/3983
+            let opt_metric_path = match generic_route.as_deref() {
+                Some("/tasks/{task_id}") => Some("/tasks/{task_id}"),
+                Some(x) if x.starts_with("/keys") => None,
                 Some(x)
                     if x.starts_with("/fonts/")
                         || x.starts_with("/static/")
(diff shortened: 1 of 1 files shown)


Zarathustra2:
Hi @irevoire 

Hope you had great vacations!

I can add groups, let me try to update this PR soon and will let you know.

Best regards,
Dario

Zarathustra2:
Hi @irevoire,

sorry got taken away by some stuff so took me longer than anticipated to get back to this: 

I added some of your feedback. 

> search requests (GET and POST grouped together 🤔) by index
I am against grouping GET & POST together because then either GET requests will be labeled as POST or the other way around. This would be confusing for the end user. Would rather keep it separate and also this allows the end user to quickly spot when someone does GET requests instead of POST ones for searching. 

> multi-search requests by index
I don't understand this task. Do you want me to parse the payload and then create an event/metric for every index that is being searched? Or make a label `multi_search_request_{index_one}_{index_two}_{index_three}`.

> Registering tasks
I didn't understand this either. Which api route are you talking about? 

Best regards & hope you had a good weekend,
Dario

irevoire:
> I am against grouping GET & POST together because then either GET requests will be labeled as POST or the other way around. This would be confusing for the end user.

Well, maybe I’m wrong, but in my opinion, we shouldn’t even talk about GET or POST requests and simply say we caught a search request that took `n` ms.

What I was saying is that we should stop exposing raw HTTP routes to the end users because most routes do the same thing in meilisearch. Like the GET and POST search.
But also, like all the routes linked to the API keys.

irevoire:
I was thinking of getting rid of the method entirely and making something like this;
```rust
crate::metrics::MEILISEARCH_HTTP_RESPONSE_TIME_SECONDS
                        .with_label_values(&[metric_group])
                        .start_timer(),
```

Where the `metric_group` can be something like `search`, `multi_search`, `static_files`, `get_tasks`, etc

Zarathustra2:
The only issue with this is that you then group routes together that do different things for instance for https://www.meilisearch.com/docs/reference/api/settings#dictionary we have 1 route with 3 different operation and the metric for those should be tracked separately. 

Could also just pre-/postfix the routes with their method for some routes and then something like `search` does not get prefixed so could have as group:

- search
- multi_search
- static_files
- get_tasks
- get_dictionary
- reset_dictionary
- update_dictionary 

Only issue is that when new routes get added we need to add some mapping otherwise we would fallback to just their plain route or not render anything at all.

irevoire:
> we have 1 route with 3 different operation and the metric for those should be tracked separately.

That’s the thing I’m wondering about, why should they be tracked separately?
What is the point of having a lot of routes if they're handled exactly the same by the engine? 🤔

Zarathustra2:
Not every route is handled the same by the engine only the POST & GET for search is, right? 

But for instance getting, updating & resetting a settings re different operations. One is just a read while others are changing data so these routes perform differently. 

If you then do something like median time of requests per minute and you are heavily skewed to the GET metrics (more GET metrics than POST but POST requests are terribly slow) then you won't easily spot it.

irevoire:
Hey, I'm so sorry for the huge delay. I had some issues and then forgot about your PR :sob: 

> Not every route is handled the same by the engine only the POST & GET for search is, right?

No, most routes are handled exactly the same by the engine actually.
All the routes that generate tasks are pretty much doing the same thing:
1. Receive a payload
2. Write it in the DB

The only exception is when you send a document; we have a third step where we write the content of your update in a different file + update its format to something custom.

Then for the rest, most of the time, it's only one simple access to the DB except for the stats, which are computing something a little bit more complex.

And finally, we have two other categories, it's;
1. When you get documents by filter
2. When you get tasks with a filter

And everything else should be ignored I think

irevoire:
Hey @Zarathustra2,
Someone found a better approach to fix this issue: https://github.com/meilisearch/meilisearch/pull/4619

Thanks for your efforts, and sorry, I didn’t know we could do it this way 😔

PR #4155, opened 2023-10-23
Diff Indexing for the geo points

Kerollmops:
This PR is part of [the Diff Indexing work](https://github.com/meilisearch/meilisearch/pull/4090) we are doing and will be merged into this branch.

### DIFF
--- a/milli/src/update/index_documents/extract/extract_geo_points.rs
+++ b/milli/src/update/index_documents/extract/extract_geo_points.rs
@@ -15,7 +15,6 @@ use crate::{FieldId, InternalError, Result};
 /// Returns the generated grenad reader containing the docid as key associated to the (latitude, longitude)
 #[logging_timer::time]
 pub fn extract_geo_points<R: io::Read + io::Seek>(
-    // TODO grenad::Reader<Obkv<FieldId, Obkv<DelAdd, JsonValue>>>
     obkv_documents: grenad::Reader<R>,
     indexer: GrenadParameters,
     primary_key_id: FieldId,
--- a/milli/src/update/index_documents/extract/extract_geo_points.rs
+++ b/milli/src/update/index_documents/extract/extract_geo_points.rs
@@ -60,19 +60,21 @@ pub fn extract_geo_points<R: io::Read + io::Seek>(
                     .map(|(lat, lng)| extract_lat_lng(lat, lng, document_id))
                     .transpose()?;
 
-                let mut obkv = KvWriterDelAdd::memory();
-                if let Some([lat, lng]) = del_lat_lng {
-                    #[allow(clippy::drop_non_drop)]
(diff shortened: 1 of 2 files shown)


milli/src/update/index_documents/extract/extract_geo_points.rs
```diff
@@ -30,39 +31,74 @@ pub fn extract_geo_points<R: io::Read + io::Seek>(
...
         let document_id = || -> Value {
             let document_id = obkv.get(primary_key_id).unwrap();
             serde_json::from_slice(document_id).unwrap()
         };
 
-        // first we get the two fields
-        let lat = obkv.get(lat_fid);
-        let lng = obkv.get(lng_fid);
+        // HELP we will receive two DelAdds here, one for the lat and one for the lng
+        //      what happens if there is a missing Del or Add for one of them?
```
Kerollmops, line 42: I need your attention, here 👀

ManyTheFish: Hello @Kerollmops,
In the previous behavior, if you are **replacing** a document with a missing latitude, Meilisearch would throw an error, I think your current version wouldn't.
let's say I have a document:
`{"id": 1, "_geo": { "lat": 0.0, "lng": 0.0 }}`
that we **replace** by:
`{"id": 1, "_geo": { "lat": 1.0 }}`
 
 The DelAdd version of the document would be:
```
"_geo.lat": {"Del": 0.0, "Add": 1.0 }, 
"_geo.lng": {"Del": 0.0 }
```
Which would be interpreted as:
1) line 45-46: `(Some({"Del": 0.0, "Add": 1.0 }), Some({"Del": 0.0 }))`
2) line 51: `del_lat_lng = Some(<extracted_lat_lng>)` because lat.Del is Some and lng.Del is Some
3) line 56: `del_lat_lng = None` because lat.Add is Some but lng.Add is None
4) line 64: insert only the deletion in the Sorter and continue

Whereas the previous version would have directly jumped to line 81-83 `return Err(GeoError::MissingLongitude { document_id: document_id() }.into())`.

However, the case where lat or lng are missing or are not well formatted is not possible because the document is checked earlier in [enrich](https://github.com/meilisearch/meilisearch/blob/cd40f6b28c4d5167b1e97152a59fcc0333c8f395/milli/src/update/index_documents/enrich.rs#L375C1-L403). Meaning that the error-throwing cases of the current `match` should be unreachable. 🤔

Kerollmops: I will try to make it return the right error message, thank you 😊

Kerollmops: Ok, after thinking more about this, it is impossible to reach this situation. I will keep the code I did and merge right away. It is forbidden to only specify the lat or LNG. Both are required.

PR #4163, opened 2023-10-24
Use value buffer instead of the initial value when writting the final result in the sorter

ManyTheFish:
In the word position extractor, we were using the wrong buffer to write the final values in the sorter.

### DIFF
--- a/milli/src/update/index_documents/extract/extract_docid_word_positions.rs
+++ b/milli/src/update/index_documents/extract/extract_docid_word_positions.rs
@@ -114,6 +114,7 @@ pub fn extract_docid_word_positions<R: io::Read + io::Seek>(
         let (add_obkv, add_script_language_word_count) = add?;
 
         // merge deletions and additions.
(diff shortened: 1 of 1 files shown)


milli/src/update/index_documents/extract/extract_docid_word_positions.rs
```diff
@@ -114,15 +114,16 @@ pub fn extract_docid_word_positions<R: io::Read + io::Seek>(
...
         del_add_from_two_obkvs(
             KvReader::<FieldId>::new(del_obkv),
             KvReader::<FieldId>::new(add_obkv),
             &mut value_buffer,
         )?;
 
-        // write them into the sorter.
-        let obkv = KvReader::<FieldId>::new(value);
+        // write each KV<DelAdd, KV<u16, String>> into the sorter, field by field.
+        let obkv = KvReader::<FieldId>::new(&value_buffer);
```
ManyTheFish, line 126: Here is the mistake 😬

PR #4179, opened 2023-10-31
Diff indexing fix nested primary key

ManyTheFish:


milli/src/update/index_documents/transform.rs
```diff
@@ -796,24 +816,35 @@ impl<'a, 'i> Transform<'a, 'i> {
...
-        for result in self.index.all_documents(wtxn)? {
-            let (docid, obkv) = result?;
+        let mut document_sorter_key_buffer = Vec::new();
+        let mut document_sorter_value_buffer = Vec::new();
+        for result in self.index.external_documents_ids().0.iter(wtxn)? {
+            let (external_id, docid) = result?;
+            let obkv = self.index.documents.get(wtxn, &docid)?.ok_or(
+                InternalError::DatabaseMissingEntry { db_name: db_name::DOCUMENTS, key: None },
+            )?;
+            let docid = docid.get();
```
ManyTheFish, line 847: We are now iterating over the external document ids map because we need to get the external id, which force us to create sorters in order to re-sort in the internal docid order.

milli/src/update/index_documents/typed_chunk.rs
```diff
@@ -124,13 +125,15 @@ pub(crate) fn write_typed_chunk_into_index(
...
-            while let Some((docid, reader)) = cursor.move_on_next()? {
+            while let Some((key, reader)) = cursor.move_on_next()? {
                 let mut writer: KvWriter<_, FieldId> = KvWriter::memory();
                 let reader: KvReader<FieldId> = KvReader::new(reader);
-                let docid = docid.try_into().map(DocumentId::from_be_bytes).unwrap();
+
+                let (document_id_bytes, external_id_bytes) = try_split_array_at(key)
+                    .ok_or(SerializationError::Decoding { db_name: Some(DOCUMENTS) })?;
+                let docid = DocumentId::from_be_bytes(document_id_bytes);
+                let external_id = std::str::from_utf8(external_id_bytes)?;
```
ManyTheFish, line 136: Now we get the external id from the grenad file instead of trying to retrieve it in the documents and potentially fail if the primary key is nested.

milli/src/update/index_documents/typed_chunk.rs
```diff
@@ -190,8 +158,20 @@ pub(crate) fn write_typed_chunk_into_index(
 
                 if !writer.is_empty() {
                     db.put(wtxn, &BEU32::new(docid), &writer.into_inner().unwrap())?;
+                    operations.push(DocumentOperation {
+                        external_id: external_id.to_string(),
+                        internal_id: docid,
+                        kind: DocumentOperationKind::Create,
+                    });
+                    docids.insert(docid);
```
ManyTheFish, line 166: We now always push the external-internal docid pair when we create or update a document which is unnecessary for updates

milli/src/update/index_documents/transform.rs
Kerollmops: I would prefer that you introduce an `iter(&self, rotxn: &RoTxn)` method on the `ExternalDocumentsIds` method instead of making the unit field public and accessing it with `.0` 😃

Kerollmops (changes requested):

PR #4181, opened 2023-11-01
Use rayon to sort entries in parallel

Kerollmops:
Currently, it takes 732.5s to run the `Transform::read_documents` method on a 142M documents dataset on a 12 threads machine (with max threading to 12). When enabling this new rayon sorting feature, the method takes 351.8s, so a gain of 52%.

### DIFF
--- a/milli/src/update/index_documents/transform.rs
+++ b/milli/src/update/index_documents/transform.rs
@@ -114,24 +114,43 @@ impl<'a, 'i> Transform<'a, 'i> {
         };
 
         // We initialize the sorter with the user indexing settings.
-        let original_sorter = create_sorter(
-            grenad::SortAlgorithm::Stable,
-            merge_function,
-            indexer_settings.chunk_compression_type,
-            indexer_settings.chunk_compression_level,
-            indexer_settings.max_nb_chunks,
-            indexer_settings.max_memory.map(|mem| mem / 2),
-        );
+        let original_sorter = {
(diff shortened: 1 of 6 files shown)


milli/src/update/index_documents/transform.rs
```diff
@@ -114,24 +114,43 @@ impl<'a, 'i> Transform<'a, 'i> {
...
+                builder.max_nb_chunks(nb_chunks);
+            }
+            if let Some(memory) = indexer_settings.max_memory.map(|mem| mem / 2) {
+                builder.dump_threshold(memory);
+                builder.allow_realloc(false);
+            }
+            builder.sort_algorithm(grenad::SortAlgorithm::Stable);
+            builder.sort_in_parallel(true);
+            builder.build()
+        };
```
Kerollmops, line 152: I need to create a function for that maybe...

ManyTheFish: WHy don't you modify the create sorter function?

Kerollmops: Will do, good idea :)

Kerollmops:
Hey @ManyTheFish 👋
I modified the `create_sorter` function to take a new setting but do you think it would have been better to _always_ sort in parallel by default, maybe?

ManyTheFish:
let's try and deactivate it if needed 👍

Kerollmops:
Done! What do you think?

PR #4203, opened 2023-11-09
Extract external document docids from docs on deletion by filter

dureuill:
This fixes some of the performance regression observed on `diff-indexing` when doing delete-by-filter with a filter matching many documents.

To delete 19 768 771 documents (hackernews dataset, all documents matching `type = comment`), here are the observed time:

|branch (commit sha1sum)|time|speed-down factor (lower is better)|
|--|--|--|
|`main` (48865470d7aaf42fa5bbfd01cf73423afb77addf)|1212.885536s (~20min)|x1.0 (baseline)|
|`diff-indexing` (523519fdbfd3a28ca15320641cb096f26230a7ca)|5385.550543s (90min)|x4.44|
|**`diff-indexing-extract-primary-key`**(f8289cd974d957d38645ca66c993ca518ec81955)|2582.323324s (43min) | x2.13|

So we're still suffering a speed-down of x2.13, but that's much better than x4.44.

---

Changes:

- Refactor the logic of PrimaryKey extraction to a struct
- Add a trait to abstract the extraction of field id from a name between `DocumentBatch` and `FieldIdMap`.
- Add `Index::external_id_of` to get the external ids of a bitmap of internal ids.
- Use this new method to add new Transform and Batch methods to remove documents that are known to be from the DB.
- Modify delete-by-filter to use the new method

### DIFF
--- a/milli/src/documents/primary_key.rs
+++ b/milli/src/documents/primary_key.rs
@@ -11,7 +11,11 @@ const PRIMARY_KEY_SPLIT_SYMBOL: char = '.';
 /// The default primary that is used when not specified.
 pub const DEFAULT_PRIMARY_KEY: &str = "id";
 
-pub trait FieldDistribution {
+/// Trait for objects that can map the name of a field to its [`FieldId`].
+pub trait FieldIdMapper {
+    /// Attempts to map the passed name to its [`FieldId`].
+    ///
+    /// `None` if the field with this name was not found.
     fn id(&self, name: &str) -> Option<FieldId>;
 }
 
@@ -30,7 +34,7 @@ pub enum DocumentIdExtractionError {
 }
 
 impl<'a> PrimaryKey<'a> {
-    pub fn new(path: &'a str, fields: &impl FieldDistribution) -> Option<Self> {
+    pub fn new(path: &'a str, fields: &impl FieldIdMapper) -> Option<Self> {
         Some(if path.contains(PRIMARY_KEY_SPLIT_SYMBOL) {
             Self::Nested { name: path }
         } else {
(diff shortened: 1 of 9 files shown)


milli/src/documents/primary_key.rs
```diff
@@ -0,0 +1,168 @@
...
+
+/// The symbol used to define levels in a nested primary key.
+const PRIMARY_KEY_SPLIT_SYMBOL: char = '.';
+
+/// The default primary that is used when not specified.
+pub const DEFAULT_PRIMARY_KEY: &str = "id";
+
+pub trait FieldDistribution {
+    fn id(&self, name: &str) -> Option<FieldId>;
+}
```
Kerollmops, line 16: Can we change the name of the trait, please?

Kerollmops (changes requested):

milli/src/update/index_documents/transform.rs
```diff
@@ -481,6 +481,89 @@ impl<'a, 'i> Transform<'a, 'i> {
...
+    #[logging_timer::time]
+    pub fn remove_documents_from_db_no_batch<FA>(
+        &mut self,
+        to_remove: &RoaringBitmap,
+        wtxn: &mut heed::RwTxn,
+        should_abort: FA,
+    ) -> Result<usize>
+    where
+        FA: Fn() -> bool + Sync,
+    {
```
Kerollmops, line 503: Can we factorize a little bit to avoid fixing a bug in one function and forgetting about this part, please?

Kerollmops (changes requested):

milli/src/update/index_documents/transform.rs
dureuill: Factored the removal from DB part to a function

Kerollmops (approved):
bors merge

PR #4205, opened 2023-11-10
Prevent search hang on the processing index

### LINKED ISSUE #4206: Search hangs on indexing indexes when other indexes are also being searched
**Describe the bug**
A search request on an index may hang until the index is finished processing an update task if sufficiently many other indexes also receive search requests during the update task.

**To Reproduce**
- Create 20 empty indexes by sending settings to them
- repeatedly send placeholder search requests to each of the indexes in a loop
- Create another index and send a significant batch of documents to index.
- Attempt to perform a search request on that last index.
  - The search request hangs for as long as the index update task is processing

**Expected behavior**
The search request responds after a typical search duration, even while the index update task is processing.

**Meilisearch version:**
Reported and reproduced on v1.3.0, still reproducible on v1.5.0-rc.2.

**Additional context**
Reported by a user [on Discord](https://discord.com/channels/1006923006964154428/1148983671026618579/1148983671026618579)

dureuill:
Fixes #4206, an issue originally [reported on Discord](https://discord.com/channels/1006923006964154428/1148983671026618579/1148983671026618579) where having parallel search requests on more indexes than the index cache capacity would cause search requests on the currently updating index to hang until the index is done updating.

## Test setup

- Create 20 empty indexes by sending settings to them
- repeatedly send placeholder search requests to each of the indexes in a loop
- Create another index and send a significant batch of documents to index.
- Attempt to perform a search request on that last index.
  - Before this PR, the search request hangs while the index update task is processing
  - After this PR, the search request respond immediately even while the index update task is processing

## Changes

- When getting the handle to an index for some potentially long running batches of tasks, save it in the index scheduler.
- Drop the handle from the index-scheduler when the task is done so that we don't leak indexes.
- When getting an index from outside the task queue processor, check if there is such an handle matching the requested index. If so, skip the cache entirely and clone the handle.

### DIFF
--- a/index-scheduler/src/batch.rs
+++ b/index-scheduler/src/batch.rs
@@ -924,7 +924,8 @@ impl IndexScheduler {
                 };
 
                 // the index operation can take a long time, so save this handle to make it available tothe search for the duration of the tick
-                *self.currently_updating_index.write().unwrap() = Some((index_uid.clone(), index.clone()));
+                *self.currently_updating_index.write().unwrap() =
+                    Some((index_uid.clone(), index.clone()));
 
                 let mut index_wtxn = index.write_txn()?;
                 let tasks = self.apply_index_operation(&mut index_wtxn, &index, op)?;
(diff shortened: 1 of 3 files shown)


Kerollmops (approved):
bors merge

PR #4207, opened 2023-11-13
Diff indexing prefix databases

ManyTheFish:
This PR modifies the prefix databases to make them work with the diff-indexing, but not all the databases are kept, we saw that the proximity-based prefix databases reduced the relevancy of the search queries when they were activated, moreover, the indexing time to compute these databases is huge compared to the other prefix databases.
I chose to try removing them because the gains seemed not worth all the losts.


milli/src/search/new/db_cache.rs
```diff
@@ -295,35 +297,47 @@ impl<'ctx> SearchContext<'ctx> {
...
+                for result in remap_key_type {
+                    let (_, docids) = result?;
+
+                    prefix_docids |= docids;
+                }
+                entry.insert(Some(prefix_docids.clone()));
+                Some(prefix_docids)
+            }
+        };
+        Ok(docids)
```
ManyTheFish, line 330: generate the word pair prefix cache at search time

Kerollmops (approved):
Looks good to me and simpler 😄

PR #4214, opened 2023-11-16
Save panic in memory as reports, accessible through a route

dureuill:
# Pull Request

## Related issue
Fixes TODO: create issue

### User standpoint

- Increase release binary size to 107MB
- Panic payloads and location are now logged as errors in the logs
- Panic payloads, location and backtrace are now available through a new `/reports` endpoint.
  - Users can get all reports by calling `GET /reports`
  - Users can get a specific report by its uid by calling `GET /reports/{report_uid}`. The uid is communicated in the error log, and in internal error messages in tasks.
  - Reports are **not** persisted to disk, restarting the instance deletes all reports.
  - A limited number of reports is kept is memory, older reports are dropped from memory as new reports arrive. As reports are only bug reports, they should be infrequent.

### User workflow in case of an internal error

1. User sees the report uid in the task error message or in the logs
2. Optionally, user goes to the website to see the description of the internal error that invites the user to check the report with the `/reports` endpoint.
3. User executes `GET /reports/{report_uid}`, or if they lost their uid, `GET /reports` and find their report
4. User sends the report on GitHub or another support channel so that a developer can attempt to reproduce their issue.

### Technical standpoint

- Enabled `debug = true` for release builds
- Installed a global panic hook that writes incoming panic to a FIFO queue, evicting older reports after the capacity is exceeded and also logs the panic as an error.
- Added a `ReportReader` to the index-scheduler that can read from the FIFO queue
- Added `GET /reports` and `GET /reports/{report_uid}` routes that use the `ReportReader` to consume reports
- Added backtrace as a dependency of the index-scheduler

### Screenshots

Capture d’écran 2023-11-16 à 09 59 43

Log output

--- 

Capture d’écran 2023-11-16 à 10 01 22

Failing task output

---

Capture d’écran 2023-11-16 à 10 10 59

Report backtrace, filename and line

---

Capture d’écran 2023-11-16 à 10 11 22

Report backtrace, function

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -231,9 +231,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.20.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f4fa78e18c64fce05e902adecd7a5eed15a5e0a3439f7b0e169f0252214865e3"
+checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
 dependencies = [
  "gimli",
 ]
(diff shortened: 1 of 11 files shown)


dureuill:
Superseded by the work on the logs route.

PR #4223, opened 2023-11-22
Update to heed 0.20

Kerollmops:
This PR brings the v0.20-alpha.9 version of heed into Meilisearch 🎉 The main goal is to test it in a real environment to make the necessary changes if needed. We also want to merge it as soon as possible during the pre-release phase to ensure we catch bugs before the release.

Most of the calls to heed are the same as before, except:
 - The `PolyDatabase` has been replaced with a `Database<Unspecified, Unspecified>`. We replaced the `get<T, U>()` by a `remap<T, U>().get()` calls.
 - The `Database` `append(...)` method has been replaced with a `put_with_flags(PutFlags::APPEND, ...)`.
 - The `RwTxn<'e, 'p>` has been simplified into a `RwTxn<'e>`.
 - The `BytesEncode/Decode` traits return a `Result<_, BoxedError>` instead of an `Option<_>`.
 - We no longer need to wrap and unwrap the `BEU32` integer when storing/getting them from heed.

### TODO
 - [x] Create actual, simple error types instead of using strings in the codecs.

### Follow-up work
 - Move the codecs into another member crate (we depend on the uuid one in the meilitool crate).
 - Display the internal decoding error in the `SerializationError` internal error variant.

### DIFF
--- a/meilisearch-auth/src/store.rs
+++ b/meilisearch-auth/src/store.rs
@@ -4,11 +4,13 @@ use std::collections::HashSet;
 use std::convert::{TryFrom, TryInto};
 use std::fs::create_dir_all;
 use std::path::Path;
+use std::result::Result as StdResult;
 use std::str;
 use std::str::FromStr;
 use std::sync::Arc;
 
 use hmac::{Hmac, Mac};
+use meilisearch_types::heed::BoxedError;
 use meilisearch_types::index_uid_pattern::IndexUidPattern;
 use meilisearch_types::keys::KeyId;
 use meilisearch_types::milli;
@@ -294,23 +296,23 @@ pub struct KeyIdActionCodec;
 impl<'a> milli::heed::BytesDecode<'a> for KeyIdActionCodec {
     type DItem = (KeyId, Action, Option<&'a [u8]>);
 
-    fn bytes_decode(bytes: &'a [u8]) -> Option<Self::DItem> {
-        let (key_id_bytes, action_bytes) = try_split_array_at(bytes)?;
-        let (action_bytes, index) = match try_split_array_at(action_bytes)? {
+    fn bytes_decode(bytes: &'a [u8]) -> StdResult<Self::DItem, BoxedError> {
+        let (key_id_bytes, action_bytes) = try_split_array_at(bytes).unwrap();
+        let (action_bytes, index) = match try_split_array_at(action_bytes).unwrap() {
             (action, []) => (action, None),
             (action, index) => (action, Some(index)),
         };
         let key_id = Uuid::from_bytes(*key_id_bytes);
-        let action = Action::from_repr(u8::from_be_bytes(*action_bytes))?;
+        let action = Action::from_repr(u8::from_be_bytes(*action_bytes)).unwrap();
 
-        Some((key_id, action, index))
+        Ok((key_id, action, index))
     }
 }
 
 impl<'a> milli::heed::BytesEncode<'a> for KeyIdActionCodec {
     type EItem = (&'a KeyId, &'a Action, Option<&'a [u8]>);
 
-    fn bytes_encode((key_id, action, index): &Self::EItem) -> Option<Cow<[u8]>> {
+    fn bytes_encode((key_id, action, index): &Self::EItem) -> StdResult<Cow<[u8]>, BoxedError> {
         let mut bytes = Vec::new();
 
         bytes.extend_from_slice(key_id.as_bytes());
@@ -320,7 +322,7 @@ impl<'a> milli::heed::BytesEncode<'a> for KeyIdActionCodec {
             bytes.extend_from_slice(index);
         }
 
-        Some(Cow::Owned(bytes))
+        Ok(Cow::Owned(bytes))
     }
 }
 
--- a/meilisearch-auth/src/store.rs
+++ b/meilisearch-auth/src/store.rs
@@ -14,7 +14,7 @@ use meilisearch_types::heed::BoxedError;
 use meilisearch_types::index_uid_pattern::IndexUidPattern;
 use meilisearch_types::keys::KeyId;
 use meilisearch_types::milli;
-use meilisearch_types::milli::heed::types::{ByteSlice, DecodeIgnore, SerdeJson};
+use meilisearch_types::milli::heed::types::{Bytes, DecodeIgnore, SerdeJson};
 use meilisearch_types::milli::heed::{Database, Env, EnvOpenOptions, RwTxn};
 use sha2::Sha256;
 use time::OffsetDateTime;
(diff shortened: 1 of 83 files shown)


meilisearch-auth/src/store.rs
```diff
@@ -294,23 +296,23 @@ pub struct KeyIdActionCodec;
...
-        let (action_bytes, index) = match try_split_array_at(action_bytes)? {
+    fn bytes_decode(bytes: &'a [u8]) -> StdResult<Self::DItem, BoxedError> {
+        let (key_id_bytes, action_bytes) = try_split_array_at(bytes).unwrap();
+        let (action_bytes, index) = match try_split_array_at(action_bytes).unwrap() {
             (action, []) => (action, None),
             (action, index) => (action, Some(index)),
         };
         let key_id = Uuid::from_bytes(*key_id_bytes);
-        let action = Action::from_repr(u8::from_be_bytes(*action_bytes))?;
+        let action = Action::from_repr(u8::from_be_bytes(*action_bytes)).unwrap();
```
Kerollmops, line 306: Do not unwrap here. Use real errors.

meilisearch-types/src/settings.rs
```diff
@@ -532,7 +532,10 @@ pub fn settings(
 
     let faceting = FacetingSettings {
         max_values_per_facet: Setting::Set(
-            index.max_values_per_facet(rtxn)?.unwrap_or(DEFAULT_VALUES_PER_FACET),
+            index
+                .max_values_per_facet(rtxn)?
+                .map(|x| x as usize)
+                .unwrap_or(DEFAULT_VALUES_PER_FACET),
```
Kerollmops, line 538: Please take a look at this @dureuill

milli/Cargo.toml
```diff
@@ -29,8 +29,8 @@ geoutils = "0.5.1"
 grenad = { version = "0.4.5", default-features = false, features = [
     "rayon", "tempfile"
 ] }
-heed = { git = "https://github.com/meilisearch/heed", tag = "v0.12.7", default-features = false, features = [
-    "lmdb", "read-txn-no-tls"
+heed = { version = "0.20.0-alpha.9", default-features = false, features = [
+    "serde-json", "serde-bincode", "read-txn-no-tls"
```
Kerollmops, line 33: Check features

milli/src/heed_codec/script_language_codec.rs
```diff
@@ -1,30 +1,31 @@
...
 impl<'a> heed::BytesDecode<'a> for ScriptLanguageCodec {
     type DItem = (Script, Language);
 
-    fn bytes_decode(bytes: &'a [u8]) -> Option<Self::DItem> {
-        let sep = bytes.iter().position(|b| *b == 0)?;
-        let (s_bytes, l_bytes) = bytes.split_at(sep);
-        let script = str::from_utf8(s_bytes).ok()?;
+    fn bytes_decode(bytes: &'a [u8]) -> Result<Self::DItem, BoxedError> {
+        let cstr = CStr::from_bytes_until_nul(bytes)?;
+        let script = cstr.to_str()?;
```
Kerollmops, line 15: Create an issue about using heed/LMDB custom comparator instead of storing NUL bytes in the middle of utf8 strings.

milli/src/heed_codec/str_beu32_codec.rs
```diff
@@ -42,33 +46,33 @@ pub struct StrBEU16Codec;
...
         if bytes.len() < footer_len + 1 {
-            return None;
+            return Err(SliceTooShortError.into());
         }
 
         let (word_plus_nul_byte, bytes) = bytes.split_at(bytes.len() - footer_len);
-        let (_, word) = word_plus_nul_byte.split_last()?;
-        let word = str::from_utf8(word).ok()?;
-        let pos = bytes.try_into().map(u16::from_be_bytes).ok()?;
+        let (_, word) = word_plus_nul_byte.split_last().unwrap();
```
Kerollmops, line 57: Add a comment to say that it is not possible

milli/src/index.rs
```diff
@@ -264,20 +262,23 @@ impl Index {
...
         created_at: OffsetDateTime,
         updated_at: OffsetDateTime,
     ) -> heed::Result<()> {
         let mut txn = env.write_txn()?;
         // The db was just created, we update its metadata with the relevant information.
-        if main.get::<_, Str, SerdeJson<OffsetDateTime>>(&txn, main_key::CREATED_AT_KEY)?.is_none()
+        if main
+            .remap_types::<Str, SerdeJson<OffsetDateTime>>()
+            .get(&txn, main_key::CREATED_AT_KEY)?
+            .is_none()
```
Kerollmops, line 274: Do the remap only once.

milli/src/index.rs
```diff
@@ -319,11 +320,11 @@ impl Index {
     /// This value is the maximum between the map size passed during the opening of the index
     /// and the on-disk size of the index at the time of opening.
     pub fn map_size(&self) -> Result<usize> {
-        Ok(self.env.map_size()?)
+        Ok(self.env.info().map_size) // TODO remove Result
```
Kerollmops, line 323: Remove it.

milli/src/index.rs
```diff
@@ -319,11 +320,11 @@ impl Index {
...
     }
 
     pub fn copy_to_path<P: AsRef<Path>>(&self, path: P, option: CompactionOption) -> Result<File> {
```
Kerollmops, line 326: Rename me into `copy_to_file`.

meilisearch-types/src/settings.rs
dureuill: Looks OK to me.

Reasoning below.

---

We're converting a `u64` to a `usize`.

What can go wrong?

If:
1. Ihe system is 32-bit
2. And the value is > u32::max()
Then:
- the value will be truncated

It is unclear how much support we have for (1), and (2) would mean a very large max_values_per_facet, so it is unlikely to happen in practice.

If it were to happen anyway, users would just notice a different max facet value than they specified. Which is a bug, but I think it has priority P182847.

meilisearch/src/routes/multi_search.rs
```diff
@@ -46,7 +46,7 @@ pub async fn multi_search_with_post(
     // Explicitly expect a `(ResponseError, usize)` for the error type rather than `ResponseError` only,
     // so that `?` doesn't work if it doesn't use `with_index`, ensuring that it is not forgotten in case of code
     // changes.
-    let search_results: Result<_, (ResponseError, usize)> = (|| {
+    let search_results: Result<_, (ResponseError, usize)> = {
```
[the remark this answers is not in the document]
dureuill: Looks OK, I think the initial implemetnation wasn't async, and required the immediately executed lambda so that `?` inside of the then-non-async block wouldn't return from the enclosing function.

Since the value is async now, the `?` already doesn't return from the enclosing function, but from the async block itself, which is OK.

Actually it can be further reduced from:

```rust
{
    async {
    // ...
    }
}.await
```

to just
```rust
async { /*...*/}.await
```

milli/Cargo.toml
Kerollmops: It's okay because we don't need the (default) "serde" feature to serialize some types.

milli/src/index.rs
Kerollmops: Thanks Rust Analyzer ❤️

milli/src/heed_codec/script_language_codec.rs
Kerollmops: https://github.com/meilisearch/meilisearch/issues/4237

dureuill (approved):
bors merge

PR #4233, opened 2023-11-27
Fix search on exact attributes using attributeToSearchOn (fix #4232)

### LINKED ISSUE #4232: `typoTolerance.disableOnAttributes` does not seem to be working when `attributesToSearchOn` is used.
**Describe the bug**

The `typoTolerance.disableOnAttributes` seems to be ignored when the `attributesToSearchOn` search parameter is used.

**To Reproduce**
I've not done a separate test other than my current project as described in the **Additional context** section, but these steps should work?

1. create an index with several fields and disable typo tolerance on one of them, lets call it `exact_only`
2. fill the index with data, the `exact_only` field should have similar but not exact values.
3. do a search for a term found in the `exact_only` field and set `attributesToSearchOn=exact_only`
4. see more than 1 result.
5. the same search without `attributesToSearchOn` will only return 1 result as expected.

**Expected behavior**
The search with `attributesToSearchOn=exact_only` should only return 1 result - matching the exact search term, or those with the same term plus extra characters.
For example:

`q=test` should return results for `test` and `tested`, but not `tesd`

**Meilisearch version:**
1.5.0

**Additional context**
I've got an index of `products` that has some fields for the `gtin` number and supplier / manufacture product code's, which I have added to the `disableOnAttributes` array - but doing a search for a sku still returns other skus that are close to the search but not exact.

`/indexes/products/settings/typo-tolerance`
```json
{
  "enabled": true,
  "minWordSizeForTypos": {
    "oneTypo": 5,
    "twoTypos": 9
  },
  "disableOnWords": [],
  "disableOnAttributes": [
    "code",
    "gtin",
    "product",
    "supplier"
  ]
}
```

Simple search returns expected results:

`/indexes/local_products/search?q=BCK88C`
```json
{
  "hits": [
    {
      "title": "Brother LC431C Cyan Ink Cartridge",
      "code": "LC431C",
      "gtin": "4977766810616",
      "product": "DV15817",
      "supplier": "BCK88C",
      "id": 95884,
    }
  ],
  "query": "BCK88C",
  "processingTimeMs": 1,
  "limit": 20,
  "offset": 0,
  "estimatedTotalHits": 1
}
```

But with `attributesToSearchOn`, 14 results are returned - only 1 matches the `supplier` field.

`/indexes/local_products/search?attributesToSearchOn=code,supplier,product&q=BCK88C`
```json
{
  "hits": [
    {
      "title": "Brother LC431C Cyan Ink Cartridge",
      "code": "LC431C",
      "gtin": "4977766810616",
      "product": "DV15817",
      "supplier": "BCK88C",
      "id": 95884
    },
    {
      "title": "Brother LC39C Cyan Ink Cartridge",
      "code": "LC39C",
      "gtin": "4977766676731",
      "product": "dv3622",
      "supplier": "BCK18C",
      "id": 13384
    },
    {
      "title": "Brother LC23EC Cyan Super High Yield Ink Cartridge",
      "code": "LC23EC",
      "gtin": "4977766754958",
      "product": "DV09882",
      "supplier": "BCK82C",
      "id": 52932
    },
    ...
  ],
  "query": "BCK88C",
  "processingTimeMs": 2,
  "limit": 20,
  "offset": 0,
  "estimatedTotalHits": 14
}
```

ManyTheFish:
- add a test reproducing the bug
- fix the bug by creating 2 different restricting lists of attributes, one for the exact attributes, and the other for the tolerant attributes

## Related issue
Fixes #4232

### DIFF
--- a/milli/src/search/new/db_cache.rs
+++ b/milli/src/search/new/db_cache.rs
@@ -154,10 +154,11 @@ impl<'ctx> SearchContext<'ctx> {
 
     /// Retrieve or insert the given value in the `word_docids` database.
     fn get_db_word_docids(&mut self, word: Interned<String>) -> Result<Option<RoaringBitmap>> {
-        match &self.restricted_tolerant_fids {
+        match &self.restricted_fids {
             Some(restricted_fids) => {
                 let interned = self.word_interner.get(word).as_str();
-                let keys: Vec<_> = restricted_fids.iter().map(|fid| (interned, *fid)).collect();
+                let keys: Vec<_> =
+                    restricted_fids.tolerant.iter().map(|fid| (interned, *fid)).collect();
 
                 DatabaseCache::get_value_from_keys::<_, _, CboRoaringBitmapCodec>(
                     self.txn,
@@ -182,10 +183,11 @@ impl<'ctx> SearchContext<'ctx> {
         &mut self,
         word: Interned<String>,
     ) -> Result<Option<RoaringBitmap>> {
-        match &self.restricted_exact_fids {
+        match &self.restricted_fids {
             Some(restricted_fids) => {
                 let interned = self.word_interner.get(word).as_str();
-                let keys: Vec<_> = restricted_fids.iter().map(|fid| (interned, *fid)).collect();
+                let keys: Vec<_> =
+                    restricted_fids.exact.iter().map(|fid| (interned, *fid)).collect();
 
                 DatabaseCache::get_value_from_keys::<_, _, CboRoaringBitmapCodec>(
                     self.txn,
@@ -231,10 +233,11 @@ impl<'ctx> SearchContext<'ctx> {
         &mut self,
         prefix: Interned<String>,
     ) -> Result<Option<RoaringBitmap>> {
-        match &self.restricted_tolerant_fids {
+        match &self.restricted_fids {
             Some(restricted_fids) => {
                 let interned = self.word_interner.get(prefix).as_str();
-                let keys: Vec<_> = restricted_fids.iter().map(|fid| (interned, *fid)).collect();
+                let keys: Vec<_> =
+                    restricted_fids.tolerant.iter().map(|fid| (interned, *fid)).collect();
 
                 DatabaseCache::get_value_from_keys::<_, _, CboRoaringBitmapCodec>(
                     self.txn,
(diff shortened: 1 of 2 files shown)


milli/src/search/new/mod.rs
```diff
@@ -63,7 +63,8 @@ pub struct SearchContext<'ctx> {
     pub phrase_interner: DedupInterner<Phrase>,
     pub term_interner: Interner<QueryTerm>,
     pub phrase_docids: PhraseDocIdsCache,
-    pub restricted_fids: Option<Vec<u16>>,
+    pub restricted_tolerant_fids: Option<Vec<u16>>,
+    pub restricted_exact_fids: Option<Vec<u16>>,
```
ManyTheFish, line 67: We now have 2 different restricting lists used allowing us to retrieve the exact words only or the tolerant words only

milli/src/search/new/db_cache.rs
```diff
@@ -182,13 +182,28 @@ impl<'ctx> SearchContext<'ctx> {
...
+                )
+            }
+            None => DatabaseCache::get_value::<_, _, RoaringBitmapCodec>(
+                self.txn,
+                word,
+                self.word_interner.get(word).as_str(),
+                &mut self.db_cache.exact_word_docids,
+                self.index.exact_word_docids.remap_data_type::<ByteSlice>(),
+            ),
+        }
```
ManyTheFish, line 206: This method now filters on the restricted exact attributes if there are some, and it's the same for the prefix version below.

milli/src/search/new/mod.rs
dureuill: This is a nitpick, but we could consider having a struct here:

```rust
struct RestrictedFids {
    pub tolerant: Vec<FieldId>,
    pub exact: Vec<FieldId>,
}
```

This way, we could have:

```rust
struct SearchContext<'ctx> {
    // ...
    pub restricted_fids: Option<RestrictedFids>,
}
```

This is not about saving a boolean, but rather it would enable two cool things:

1. Make the state where `restricted_exact_fids` is `None`, and `restricted_tolerant_fids` is `Some` irrepresentable. Making impossible states irrepresentable is a good thing, as it spares us from having to handle impossible cases.
2. Allow for methods on the struct, for instance:
```rust
impl RestrictedFieldIds {
    pub fn contains(&self, fid: &FieldId) -> bool {
        self.tolerant.contains(fid) || self.exact.contains(fid)
    }
}
```
(I recommend not over-doing it, and only adding the methods that you actually use)

then the code in DB cache that does:

```rust
        if self.restricted_fids.as_ref().map_or(false, |fids| !fids.contains(&fid)) {
```

does not need to change at all 😎 

again, nitpick, feel free to tell you don't want to do it for now (or at all).

milli/src/search/new/db_cache.rs
dureuill: Thank you for the comment explaining the change 👨‍🍳

milli/src/search/new/mod.rs
ManyTheFish: Done @dureuill :D

dureuill (approved):
Looks good to me :-) 

bors merge

PR #4234, opened 2023-11-27
Fix puffin in the index scheduler

irevoire:
Currently, we can't compile the index scheduler without this feature.

It could be cool to specify the dependencies in the main workspace cargo toml like quickwit does to avoid this kind of error in the future; https://github.com/quickwit-oss/quickwit/blob/main/quickwit/Cargo.toml#L41


dureuill (approved):
bors merge

PR #4273, opened 2023-12-20
Adapt prefix thresholds

ManyTheFish:
This PR changes the prefix database creation thresholds to reduce their impact on the indexing time.

## Explanations

There are two thresholds that we can change in the prefix database:
1) the minimum number of words prefixed by the prefix candidate, named `threshold`
2) the maximum number of characters in a prefix, named `max_prefix_length`

## Prior values and drawback

On Meilisearch v1.5, we have 100 as the `threshold` and 4 as the `max_prefix_length`, meaning that we compute and store every prefix that has at least 100 words prefixed by it and less than 5 characters.

By making some statistics using `movies`, `Wikipedia`, and `e-commerce` datasets, I found out that only 10% of the total count of stored prefixes have 1 character on a small dataset (10MB), and less than 1% on a bigger dataset (2GB).
This is a shame because the 1-character prefixes are the most important regarding search time gain.
On the opposite, the prefixes containing 3 or 4 characters represent more than 30% of the prefixes on a small dataset, which is acceptable, but represent more than 80% of the prefixes on big datasets.
In this configuration for the big datasets, we allocate between 80 and 100 times more for the least important prefixes than for the most important ones, which seems non-optimal to me. 

## New value
This PR changes the `threshold` from 100 to 500 and the `max_prefix_length` from 4 to 3 in order to rebalance the distribution.
This way:
1) on a small dataset:
  - the 1-character prefixes represent more than 40% of the prefixes (10% before)
  - the 3-character prefixes represent around 2% of the prefixes (30% before counting the 4-character prefixes)
2) on a bigger dataset:
  - the 1-character prefixes represent around 5% of the prefixes (less than 1% before)
  - the 3-character prefixes represent around 60% of the prefixes (more than 80% before counting the 4-character prefixes)

This suggested change is conservative, and we could think of raising the thresholds even more, but I'm afraid that it would have a significant impact on the search, and because we already are in the prerelease phase we will not be able to make that many benchmarks.

## Related

google sheet with some stats: https://docs.google.com/spreadsheets/d/1EOL4Bmg_2RW2WGt6DN5Ec7V2hsbIc42eC9iOjy5vzN8/edit#gid=558148218

### DIFF
--- a/milli/src/search/new/tests/proximity.rs
+++ b/milli/src/search/new/tests/proximity.rs
@@ -371,7 +371,7 @@ fn test_proximity_prefix_db() {
     s.scoring_strategy(crate::score_details::ScoringStrategy::Detailed);
     s.query("best s");
     let SearchResult { documents_ids, document_scores, .. } = s.execute().unwrap();
-    insta::assert_snapshot!(format!("{documents_ids:?}"), @"[10, 9, 6, 7, 8, 11, 12, 13, 15]");
+    insta::assert_snapshot!(format!("{documents_ids:?}"), @"[10, 13, 9, 12, 6, 7, 8, 11, 15]");
     insta::assert_snapshot!(format!("{document_scores:#?}"));
     let texts = collect_field_values(&index, &txn, "text", &documents_ids);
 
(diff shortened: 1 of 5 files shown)


Kerollmops (approved):
I like what you did there!

dureuill:
Great thinking @ManyTheFish !

Could you add very visibly in the description of the PR: a table indicating the benefits of this PR for index size, indexing time, and the tradeoff for search time?

curquiza:
After discussing with Many, removing this PR from v1.6.0 since the benchmarks results are not satisfying

Kerollmops:
This PR is very far behind and was not satisfying at the time. Let's open a new PR to reintroduce the changes if needed.

PR #4279, opened 2023-12-21
Check experimental feature on setting update query rather than in the task.

dureuill:
Improve the UX by checking for the vector store feature and returning an error synchronously when sending a setting update, rather than in the indexing task.


ManyTheFish (approved):
Ready to go,
Sorry for the time,
Bors merge

PR #4293, opened 2024-01-02
Update SDK test dependencies

curquiza:
Replace dependabot updates

### DIFF
--- a/.github/workflows/sdks-tests.yml
+++ b/.github/workflows/sdks-tests.yml
@@ -22,7 +22,7 @@ jobs:
     outputs:
       docker-image: ${{ steps.define-image.outputs.docker-image }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Define the Docker image we need to use
         id: define-image
         run: |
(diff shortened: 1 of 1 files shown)


curquiza:
@meilisearch/engine-team 
Sorry for the lack of information about my PR, I updated the PR description

curquiza:
> I'm curious, do we have a way of sync'ing the SDK configuration more automatically? My current understanding is that we would need to change this file if the configuration of any of the SDK changes?

I don't know any, but I have not looked that much (only 5-10min)

I will not do this update often, to be honest. First time I have done it since I created the CI (9 months ago). It's not a big deal if our dependencies are not exactly the same
- because these dependencies are only github-actions oriented, so not related to our test. We do use the tests created and updated in the integration repository directly
- the most important is to test at least with one version of the language (i.e. one version of dart/php...). It is extremely rare you have an issue due to meilisearch you can see with one version of the language and not with another. It even never happened in the past, during my 4 years of integrations background

So I will not spend more time investigate to avoid this, at least until we get more issues due to this github action dependency update.

curquiza (commented):
bors merge

Issue #4187, opened 2023-11-02
[EXP] setting to remove proximity ranking rule


curquiza:
Related product team resources: [PRD](https://www.notion.so/meilisearch/Settings-to-remove-proximity-ranking-rule-518de76c78c948b08f0cf90457d2b45d) (_internal only_)

## Motivation

[[Proximity](https://www.meilisearch.com/docs/learn/core_concepts/relevancy#3-proximity)](https://www.meilisearch.com/docs/learn/core_concepts/relevancy#3-proximity) is one of the ranking rules of Meilisearch and makes Meilisearch return results that are sorted by increasing the distance between matched query terms.

To provide this kind of relevancy, Meilisearch needs to make a lot of computations and create many internal databases. This takes space on the user (or Cloud) disk, and also time when indexing.

Indexing time is a real problem and pain point for the users.

The point is to provide a setting to make the user lose a part of relevancy that probably do not fit their use case by drastically reducing their indexing speed.

## Usage

More information about the usage here: https://www.notion.so/meilisearch/Remove-proximity-settings-usage-e33675cbbd9b48b3af471f3da91c454a

## TODO

- [ ] Release a prototype
- [ ] If prototype validated, merge changes into `main`
- [ ] Create [product discussion](https://github.com/orgs/meilisearch/discussions)

## Impacted teams

@meilisearch/docs-team to let you know about this new experimental feature

curquiza:
Update:
@meilisearch/integration-team @meilisearch/docs-team : the feature is finally not experimental. check ou the API references linked in the issue description

curquiza:
@meilisearch/integration-team as well -> concerned by the feature since it's not experimental anymore

ManyTheFish:
Hello, the spec has been created I put the link in the issue and below:
https://github.com/meilisearch/specifications/pull/261

PR #4303, opened 2024-01-08
Display default value when proximityPrecision is not set

ManyTheFish:
# Pull Request

## Related
Issue: #4187
Spec change requests: https://github.com/meilisearch/specifications/pull/261#discussion_r1441725272

## What does this PR do?
- Display default value when proximityPrecision is not set instead of Null


dureuill (approved):
thank you many :heart:

bors merge

firstcommader (commented):
Help

PR #4304, opened 2024-01-08
Add CUDA GPU support for Hugging Face embedders

dureuill:
Adds a "cuda" feature to `milli`.

Compiling with this feature requires that the CUDA support library be installed (see "with CUDA support" paragraph in https://huggingface.github.io/candle/guide/installation.html), and adds CUDA support to the `huggingFace` embedder.

To enable GPU support, users will need to:

1. Have a compatible NVidia GPU under Linux
2. Follow [the guide](https://huggingface.github.io/candle/guide/installation.html) to install the CUDA dependencies
3. Compile Meilisearch with the `cuda` feature: `cargo build --release --features cuda`

# Impact

Enabling the CUDA feature allows to use an available GPU to compute embeddings with a `huggingFace` embedder. 
On an AWS Graviton 2, this yields a x3 - x5 improvement on indexing time.

# Technical details

- I had to change the CI so that the cuda feature is not included in the `Tests all features` workflow
- To achieve that, I had to add a binary following the `cargo xtask` design pattern, to list all features excepted the cuda one.
- I then changed the workflow accordingly (renamed to "Tests almost all features" 😉)
- A test run of the new feature was done on a temporary version of this PR that had it enabled for PRs: [See the results here](https://github.com/meilisearch/meilisearch/actions/runs/7461331929/job/20301216732)

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -82,7 +82,7 @@ jobs:
           args: --locked --release --all
 
   test-all-features:
-    name: Tests all features
+    name: Tests almost all features
     runs-on: ubuntu-latest
     container:
       # Use ubuntu-18.04 to compile with glibc 2.27, which are the production expectations
@@ -98,16 +98,12 @@ jobs:
         with:
           toolchain: stable
           override: true
-      - name: Run cargo build with all features
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --workspace --locked --release --all-features
-      - name: Run cargo test with all features
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
-          args: --workspace --locked --release --all-features
+      - name: Run cargo build with almost all features
+        run: |
+          cargo build --workspace --locked --release --features "$(cargo xtask list-features --exclude-feature cuda)"
+      - name: Run cargo test with almost all features
+        run: |
+          cargo test --workspace --locked --release --features "$(cargo xtask list-features --exclude-feature cuda)"
 
   test-disabled-tokenization:
     name: Test disabled tokenization
--- a/milli/Cargo.toml
+++ b/milli/Cargo.toml
@@ -137,3 +137,6 @@ greek = ["charabia/greek"]
 
 # allow khmer specialized tokenization
 khmer = ["charabia/khmer"]
+
+# allow CUDA support
+cuda = ["candle-core/cuda"]
--- a/milli/Cargo.toml
+++ b/milli/Cargo.toml
@@ -138,5 +138,5 @@ greek = ["charabia/greek"]
 # allow khmer specialized tokenization
 khmer = ["charabia/khmer"]
 
-# allow CUDA support
+# allow CUDA support, see <https://github.com/meilisearch/meilisearch/issues/4306>
 cuda = ["candle-core/cuda"]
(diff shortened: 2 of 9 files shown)


curquiza:
Hello Louis!
I guess cuda improves with hybrid search and huggingface, but how does it improve exactly?

By reading the PR description I guess it's about GPU. Does it mean GPU support will be able now with huggingface model when computing vectors?

dureuill:
Hello curqui:

I added an `Impact` section to the PR description that hopefully answers your question.

As for the changelog, note that the feature here requires compiling Meilisearch on your own. Other than that, yes, it is adding the capability to generate embeddings using the GPU for `huggingFace` embedders.

curquiza:
@dureuill 
can you change the branch from `release-v1.6.0` to `main` since we decided to release it for v1.7.0 and not v1.6.0? 😊

dureuill:
I need to rebase on main then, I'm not sure it will work.

The easiest would be to wait for the changes of v1.6 to be merged into main

milli/Cargo.toml
```diff
@@ -137,3 +137,6 @@ greek = ["charabia/greek"]
 
 # allow khmer specialized tokenization
 khmer = ["charabia/khmer"]
+
+# allow CUDA support
```
Kerollmops, line 141: Can you link this PR to this feature flag or find another way to link to the documentation you provided, please?

xtask/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "xtask"
+version.workspace = true
+authors.workspace = true
+description.workspace = true
+homepage.workspace = true
```
Kerollmops, line 6: Can you change the description of this crate, please? Using something different from the Meilisearch description.

.github/workflows/test-suite.yml
```diff
@@ -78,7 +78,7 @@ jobs:
           args: --locked --release --all
 
   test-all-features:
-    name: Tests all features
+    name: Tests almost all features
```
Kerollmops, line 81: 😂

milli/src/vector/hf.rs
```diff
@@ -70,7 +70,7 @@ impl std::fmt::Debug for Embedder {
 
 impl Embedder {
     pub fn new(options: EmbedderOptions) -> std::result::Result<Self, NewEmbedderError> {
-        let device = candle_core::Device::Cpu;
+        let device = candle_core::Device::cuda_if_available(0).unwrap();
```
Kerollmops, line 73: Does it panic if CUDA is unavailable?

Kerollmops (changes requested):
I love your changes. Do you plan to work with the doc team to port this PR guide to the documentation?

xtask/Cargo.toml
dureuill: Done!

.github/workflows/test-suite.yml
dureuill: 👍

milli/src/vector/hf.rs
dureuill: 🤩 Good catch, the behavior was a bit subtle here:

1. Without the cuda feature, it would always just return `Ok(Device::Cpu)`, so no risk in packing in this case
2. With the cuda feature, it would panic if the CUDA support library could not be found or if the GPU doesn't support it for some reason.

I modified the code to avoid the panic and fallback to CPU in all cases (with a warning stating the issue), it should feel more robust now 👍

Kerollmops (approved):
I like what you've done here! Thank you! Merging 🍨

bors merge

PR #4314, opened 2024-01-11
Fix proximity precision telemetry

ManyTheFish:
The proximity precision telemetry was partially missing in the global setting route.
This PR adds the missing field and return the default value when the value is not set.


Kerollmops (approved):
bros merge

Kerollmops:
bors merge

PR #4318, opened 2024-01-11
Hide embedders

dureuill:
Hides `embedders` when it is an empty dictionary.

Manual tests:

- getting settings with empty embedders: not displayed
- getting settings with non-empty embedders: displayed like before
- dump with empty embedders: can be imported
- dump with non-empty embedders: can be imported


ManyTheFish (approved):
bors merge

PR #4337, opened 2024-01-17
Remove panic on the geosearch

irevoire:
# Pull Request

## Related issue
Fixes  #4333

## What does this PR do?
- Add tests for the enrich pipeline on malformed documents with `null` value
- Reproduce the issue when updating the settings while there is malformed documents in the DB
- Fix the bug

PR #4364, opened 2024-01-25
Revert "Remove panic on the geosearch"

irevoire:
After more thought about it, we want to fix this bug in a patch release instead of `main`.
I revert this PR for now, but the fix will still land on `main` once we bring back the change of the `v1.6.1` on `main`.

Reverts meilisearch/meilisearch#4337


curquiza (approved):
tutto bene 🇮🇹

bors merge

Issue #4333, opened 2024-01-16
Panic with malformed _geo data


laurentS:
**Describe the bug**
In v1.6, indexing malformed `_geo` data causes the server to panic and crash (without the [error](https://www.meilisearch.com/docs/reference/errors/error_codes#invalid_document_geo_field) listed in the [docs](https://www.meilisearch.com/docs/learn/fine_tuning_results/geosearch#preparing-documents-for-location-based-search)).

**To Reproduce**
1. `docker run -it --rm -p 7700:7700 -e MEILI_ENV='development' -e RUST_BACKTRACE=full -v $(pwd)/meili_data:/meili_data getmeili/meilisearch:v1.6`
2. Upload the following json document:
```json
{"id":25843,"url_name":"mairie_champmillon","name":"Mairie - Champmillon","rip":false,"category":"mairie","addresse":"Route de la Charente, 16290, Champmillon","cp":16290,"_geo":{"lng":null,"lat":null}}
```
4. (I use the python client) `client.index("myindex").update_filterable_attributes(["_geo"])
`
5. See error:
```
[2024-01-16T16:37:17Z INFO  actix_web::middleware::logger] 172.17.0.1 "PUT /indexes/myindex/settings/filterable-attributes HTTP/1.1" 202 126 "-" "Meilisearch Python (v0.28.4)" 0.025347
thread 'indexing-thread:7' panicked at 'called `Result::unwrap()` on an `Err` value: Error("expected value", line: 1, column: 1)', milli/src/update/index_documents/extract/extract_geo_points.rs:38:49
stack backtrace:
   0:     0x562c77d4a411 - std::backtrace_rs::backtrace::libunwind::trace::h2801c3d891641b62
                               at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x562c77d4a411 - std::backtrace_rs::backtrace::trace_unsynchronized::h9d8e6ca1d0a50926
                               at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x562c77d4a411 - std::sys_common::backtrace::_print_fmt::h4ca85c9e04b799d6
                               at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/sys_common/backtrace.rs:65:5
[...]

PR #4366, opened 2024-01-25
Fix geo error message

irevoire:
# Pull Request
I backported #4337 from `main` to the current release so it could make it in the next patch version. 

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4333
Re-implement: #4337 after it was reverted in #4364

## What does this PR do?
- Add tests for the enrich pipeline on malformed documents with `null` value
- Reproduce the issue when updating the settings while there is malformed documents in the DB
- Fix the bug

### DIFF
--- a/meilisearch/tests/common/mod.rs
+++ b/meilisearch/tests/common/mod.rs
@@ -64,7 +64,7 @@ impl Display for Value {
         write!(
             f,
             "{}",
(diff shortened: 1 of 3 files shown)


dureuill (approved):
Thanks tamo 👍

dureuill:
bors merge

PR #4380, opened 2024-01-31
Bring back changes from `release-v1.6.1` into `main`

curquiza:
Following v1.6.1 release of Meilisearch


curquiza:
@dureuill @irevoire I don't know why but the CIs fail 😢

dureuill:
The cargo.lock wasn't good for some reason. The `xtask` crate hadn't be updated to v1.6.1 (it was still v1.6.0).

Without updating this version, cargo would fail.

Normally, cargo would update the version, but due to our use of `--locked` (to prevent silent updates), it wouldn't.

I opened #4381 with a new branch where the Cargo.lock change is applied, it should pass the CI this time.

This PR is superseded by #4381, so closing.

PR #4386, opened 2024-02-01
Issue 4379: correct `download-latest.sh` to make it return error when system is using `MUSL` or `bionic` C library

### LINKED ISSUE #4379: download-latest.sh: return an error when the musl C library is present on the machine
Meilisearch binary attached to the release cannot work with musl C library, like we can have with Alpine Linux.

The issue is, the [download-latest.sh](https://github.com/meilisearch/meilisearch/blob/main/download-latest.sh) script does not return an error when we use it on a machine where musl C library is present.

We might want to return an equivalent of [this error](https://github.com/meilisearch/meilisearch/blob/b968616a99c99c7b3eed9bb4f0c4d05e16b5cf0d/download-latest.sh#L102C1-L106)

mathias-vandaele:
Adapt the script to terminate on error in case the system is using bionic or MUSL

# Pull Request

## Related issue
Fixes #4379 

## What does this PR do?
- It adds a function that retrieve the C library host system is using.
- For now only check with `bionic`,  `glibc` and `musl` , with `glibc` being the only one working

## PR checklist
- [ x] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [x ] Have you read the contributing guidelines?
- [ x] Have you made sure that the title is accurate and descriptive of the changes?

If there is anything you want me to rework on, just let me know!

### DIFF
--- a/download-latest.sh
+++ b/download-latest.sh
@@ -65,6 +65,21 @@ get_os() {
     return 0
 }
 
+# Gets the C lib the system is using by setting the $c_lib variable.
+# Returns 0 in case of success, 1 otherwise.
+get_C_library(){
+  if ldd --version | grep -i glibc;then
+    c_lib='glibc'
+  elif ldd --version | grep -i musl;then
+    c_library_failure_usage
+  elif ldd --version | grep -i bionic; then
+    c_library_failure_usage
+  else
+    return 1
+  fi
+  return 0
+}
+
 # Gets the architecture by setting the $archi variable.
 # Returns 0 in case of success, 1 otherwise.
 get_archi() {
@@ -106,6 +121,12 @@ not_available_failure_usage() {
     echo 'Follow the steps at the page ("Source" tab): https://www.meilisearch.com/docs/learn/getting_started/installation'
 }
 
+c_library_failure_usage() {
+    printf "$RED%s\n$DEFAULT" 'ERROR: Meilisearch binary is not compatible with the current C library your system is using.'
+    echo ''
+    echo 'Meilisearch is designed to work with certain C libraries, such as glibc. Ensure that your system uses a compatible C library environment.'
+}
+
 fetch_release_failure_usage() {
     echo ''
     printf "$RED%s\n$DEFAULT" 'ERROR: Impossible to get the latest stable version of Meilisearch.'
@@ -134,6 +155,11 @@ fill_release_variables() {
         not_available_failure_usage
         exit 1
      fi
+     # Fill $c_lib variable.
+     if ! get_C_library; then
+       c_library_failure_usage
+     fi
+
 }
 
 download_binary() {


download-latest.sh
```diff
@@ -65,6 +65,21 @@ get_os() {
     return 0
 }
 
+# Gets the C lib the system is using by setting the $c_lib variable.
+# Returns 0 in case of success, 1 otherwise.
+get_C_library(){
+  if ldd --version | grep -i glibc;then
+    c_lib='glibc'
+  elif ldd --version | grep -i musl;then
+    c_library_failure_usage
```
curquiza, line 74:
```suggestion
  if ldd --version | grep -i glibc; then
    c_lib='glibc'
  elif ldd --version | grep -i musl; then
    c_library_failure_usage
```

curquiza (changes requested):
Hello @mathias-vandaele 

thank you very much for this PR, I will test it soon

Before I test it, can you ensure we display the workarounds to users when getting the error?
The possible workarounds are compiling from the source files or using the Docker image (https://www.meilisearch.com/docs/learn/getting_started/installation#local-installation)

mathias-vandaele:
Hi @curquiza, 

I will do it later this evening

curquiza (changes requested):
Hello @mathias-vandaele 

On my mac I got the following errors (before downloading correctly the binary)

Capture d’écran 2024-02-14 à 18 38 35

<s>And on alpine linux I don't get any error, and it downloads correctly the binary 😅 

Capture d’écran 2024-02-14 à 18 43 36</s>

-> retrying, I might have done a mistake... ⌛

curquiza:
And on Alpine I get a lot of necessary logs + I get a message saying the binary is downloaded + the binary is downloaded (but does not work, of course, as expected)

Capture d’écran 2024-02-15 à 12 38 07

mathias-vandaele:
@curquiza I will correct it.

curquiza:
Hello @mathias-vandaele 
I try to review this next week. But I don't forget you 😊

download-latest.sh
```diff
@@ -65,6 +65,30 @@ get_os() {
...
+  elif ldd --version | grep -qi musl; then
+    c_library_failure_usage
+    return 1
+  elif ldd --version | grep -qi bionic; then
+    c_library_failure_usage
+    return 1
+  else
+    return 1
+  fi
+  return 0
```
curquiza, line 89: we will never reach the final `return 0` because of the `else return 1`, so let's replace it by `return 1`

curquiza:
```suggestion
  fi
  return 1
```

download-latest.sh
```diff
@@ -65,6 +65,30 @@ get_os() {
...
+    return 0
+  fi
+  if ldd --version | grep -qi eglibc; then
+    c_lib='eglibc'
+    return 0
+  elif ldd --version | grep -qi glibc; then
+    c_lib='glibc'
+    return 0
+  elif ldd --version | grep -qi musl; then
+    c_library_failure_usage
```
curquiza, line 81: you already display the failure usage in the function in `main`, so don't display it here.

curquiza, line 81:
```suggestion
```

download-latest.sh
```diff
@@ -65,6 +65,30 @@ get_os() {
...
+    return 1
+  elif ldd --version | grep -qi bionic; then
+    c_library_failure_usage
```
curquiza, line 84:
```suggestion
```

download-latest.sh
```diff
@@ -65,6 +65,30 @@ get_os() {
     return 0
 }
 
+# Gets the C lib the system is using by setting the $c_lib variable.
```
curquiza, line 68: we don't use `$c_lib` in the script, so no point in storing it.
Can you remove it?

download-latest.sh
```diff
@@ -164,4 +201,4 @@ download_binary() {
 main() {
     download_binary
 }
-main
+main
```
curquiza, line 204:
```suggestion
main

```

curquiza (changes requested):
Thank you for the improvement

Can you remove the useless commands output? 

Capture d’écran 2024-03-07 à 14 18 29

You can use `&> /dev/null`

mathias-vandaele:
Thank you for coming back to me,
getting on it

Kerollmops:
This PR is very old, and we no longer provide MUSL binaries anyway. We will probably reintroduce them later, but I prefer to close this stalled PR. Thank you for the help, @mathias-vandaele, and I'm happy to accept another contribution in the future. Have a nice week.

PR #4350, opened 2024-01-22
Make several indexing optimizations

ManyTheFish:
# Summary

Implement several enhancements to reduce the indexing time.

# Steps

- Compute the indexing chunk size dynamically based on the available threads and the data size
- Remove the merging step before the writing step and merge at the writing time
- Remove append function
- Make Facet search indexing incremental

# Running Indexing process

## `main`
Each type of data is written after a merging phase:
Capture d’écran 2024-01-23 à 10 18 08

> Highlighted parts are the writings

## `remove-merging-phase-from-indexing`
When the extraction of a chunk is finished, the data is written:
Capture d’écran 2024-01-23 à 10 18 18

> Highlighted parts are the writings

## Related

This PR removes the appending writes on several indexing parts, which may fix https://github.com/meilisearch/meilisearch/issues/4300. However, all of the appending writes are not removed. There are 2 remaining calls that could trigger this bug:
- When [putting embedders in the settings](https://github.com/meilisearch/meilisearch/blob/b6fc18199380e902fcf4a0d634fb03c6c783614d/milli/src/update/settings.rs#L996)
- when [bulk indexing the facets](https://github.com/meilisearch/meilisearch/blob/b6fc18199380e902fcf4a0d634fb03c6c783614d/milli/src/update/facet/bulk.rs#L150)

PR #4388, opened 2024-02-06
Cap the maximum memory of the grenad sorters

### LINKED ISSUE #4152: Cap the size of the grenad Sorter to a defined value
**Describe the bug**
See: https://github.com/db-benchmarks/db-benchmarks/issues/27#issuecomment-1536355078

This person has 128GiB of RAM, and by default, in the transform, meilisearch will allocate _all_ the available memory (of the 2/3) in the transform for two grenads: https://github.com/meilisearch/meilisearch/blob/eae9eab181555364513ade91f0a5cefaa0ffc817/milli/src/update/index_documents/transform.rs#L114-L131

When we reduce this value to zero, we don't see much of a difference, so I think by default, we should heavily cap this value to, like, 1Ko max to leave the maximum amount of RAM to the other process going on at the same time like:
- A big hashmap with all the encountered primary key
- Multiple buffers
- The current documents being deserialized to json + its old version + its old flattened version + its current flattened version
- A bunch of other stuff here and there
- Searches

------

And that’s only for the `transform`. But the same thing is happening here: https://github.com/meilisearch/meilisearch/blob/eae9eab181555364513ade91f0a5cefaa0ffc817/milli/src/update/index_documents/extract/mod.rs#L156-L230

Where all of these spawned tasks are trying to use 100% of the available memory.

--------

I expect this change to have pretty much no impact on performances and to greatly reduce our chance of getting OOM-killed by sharing more RAM with LMDB.
That’s also one of the things that loic told us to do before leaving.

Kerollmops:
This PR clamps the memory usage of the grenad sorters to a reasonable maximum. Grenad sorters are opened on multiple threads at a time. This can result in higher memory usage than expected, even though it shouldn't consume more than the memory available.

Fixes #4152.

### DIFF
--- a/milli/src/update/index_documents/helpers/grenad_helpers.rs
+++ b/milli/src/update/index_documents/helpers/grenad_helpers.rs
@@ -9,6 +9,10 @@ use super::{ClonableMmap, MergeFn};
 use crate::update::index_documents::valid_lmdb_key;
 use crate::Result;
 
+/// This is something reasonable given the fact
+/// that there is one grenad sorter by thread.
+const MAX_GRENAD_SORTER_USAGE: usize = 200 * 1024 * 1024; // 200 MiB
+
 pub type CursorClonableMmap = io::Cursor<ClonableMmap>;
 
 pub fn create_writer<R: io::Write>(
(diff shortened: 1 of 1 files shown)


curquiza:
@Kerollmops does this PR fix https://github.com/meilisearch/meilisearch/issues/4152?

irevoire (approved):
Codewise it looks ok for me if we’re happy with the benchmarks

irevoire:
IMO, considering the RAM we give to our big client, I would not be shocked if we went as up as 1GiB 👀

Kerollmops:
Seems ok to me, as #4350 will probably change many things. This PR is here to reduce the memory crashes not to speed up the engine. We can work on improving the performances in a lot of other domains.

bors merge

irevoire (approved):
Yep definitely!

bors merge

curquiza:
bors merge

PR #4414, opened 2024-02-14
Add index name to sort error message

### LINKED ISSUE #4392: Error message could be clearer
**Describe the bug**

     "Attribute `id` is not sortable. This index does not have configured sortable attributes.",

**Expected behavior**

I would LOVE to see the name of the index in this message.

     "Attribute `id` is not sortable. index 'Author' does not have configured sortable attributes.",

In particular, I _do_ have sortable attributes:

http://127.0.0.1:7700/indexes/Author/settings

image

Perhaps the real issue is that the id is the primary key?  More likely I'm using the wrong index, but I can't see what that is.  A link to the settings would be even more awesome!

cisco877:
# Pull Request

## Related issue
Fixes #4392

## What does this PR do?
- I take the index name from the URI, and add it to the name property of the Index struct

## PR checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -536,6 +536,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "bindgen"
+version = "0.68.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "726e4313eb6ec35d2730258ad4e15b547ee75d6afaa1361a922e78e59b7d8078"
+dependencies = [
+ "bitflags 2.4.1",
+ "cexpr",
+ "clang-sys",
+ "lazy_static",
+ "lazycell",
+ "peeking_take_while",
+ "proc-macro2",
+ "quote",
+ "regex",
+ "rustc-hash",
+ "shlex",
+ "syn 2.0.48",
+]
+
+[[package]]
 name = "bit-set"
 version = "0.5.3"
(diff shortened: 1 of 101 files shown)


curquiza:
@cisco877 let us know when you think the PR is ready for review (in draft currently so we suppose it's not over)

cisco877:
Hi @curquiza sorry, I had some errors here that I wasn't able to get locally about fmt and clippy and was investigating my local setup, now seems to be all green for automated tests, thanks!

irevoire (changes requested):
Hey, it's nice your solution fixes the problem.
But it's not exactly the implementation we were thinking of.
We would prefer if you added the index name to _every_ milli error message by putting it directly in the `index_scheduler` that knows the index name around here : https://github.com/meilisearch/meilisearch/blob/82b43e9a7f09de51ea723817cc0878a77fccf049/index-scheduler/src/error.rs#L118-L119

I think you then won't be able to implement the `#[from] milli::Error` and will need a bunch of `.map_err(|err| Error::from_milli(err, index_uid))` or something like that.

Let me know if that's ok for you, and don't hesitate to reach me if something seems wrong with this approach or if you're struggling with something!

Thanks again for working on this issue ❤️

cisco877:
Hi @irevoire no problem, thanks for the review, I am going to apply the requested changes!

PR #4416, opened 2024-02-14
Create automation when creating Milestone to create update-version issue

curquiza:
Following our discussion @irevoire -> we miss reminder to update cargo version BEFORE rc0

Issue template [here](https://github.com/meilisearch/engine-team/blob/main/issue-templates/update-version-issue.md)

### DIFF
--- a/.github/workflows/milestone-workflow.yml
+++ b/.github/workflows/milestone-workflow.yml
@@ -110,6 +110,25 @@ jobs:
             --milestone $MILESTONE_VERSION \
             --assignee curquiza
 
+  create-update-version-issue:
+    needs: get-release-version
+    # Create the changelog issue if the release is not only a patch release
+    if: github.event.action == 'created'
+    runs-on: ubuntu-latest
+    env:
+      ISSUE_TEMPLATE: issue-template.md
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/milestone-workflow.yml
```diff
@@ -110,6 +110,25 @@ jobs:
             --milestone $MILESTONE_VERSION \
             --assignee curquiza
 
+  create-update-version-issue:
+    needs: get-release-version
+    # Create the changelog issue if the release is not only a patch release
+    if: github.event.action == 'created'
```
ManyTheFish, line 116: Why do these lines differ from the upper action?

```
 github.event.action == 'created' && needs.get-release-version.outputs.is-patch == 'false'
 ```
 
 > BTW the comment speaks about the changelog :)

curquiza: Because I want this issue to be created even if we do a patch release, which is not the case for this other issues

ManyTheFish (approved):

irevoire (approved):
Looks good!
Hopefully, it’ll eases our release process 😁

PR #4433, opened 2024-02-21
Enhance facet incremental

### LINKED ISSUE #4367: Speed up the `FacetsUpdateIncremental` for big databases
When [debugging a very big client](https://github.com/meilisearch/meilisearch-support/issues/95), we noticed that executing the differential indexing on 148M documents with nine filterable and three sortable attributes. When disabling the incremental indexing and only enabling the facet bulk indexing, we take 12 minutes to proceed with 488k document updates, while more than 2 hours and 30 minutes when using the incremental part of the code.

https://github.com/meilisearch/meilisearch/blob/b968616a99c99c7b3eed9bb4f0c4d05e16b5cf0d/milli/src/update/facet/mod.rs#L141-L171

### LINKED ISSUE #4409: Inconsistent string and number faceted documents order
When modifying a small number of documents, they appear in the wrong order.

Related to #4374, #4339, and #4390.

**To Reproduce**
It happens when a few documents are modified in a bigger DB, so the incremental facet update kicks in.
The real fix is to rework the incremental facet update; many have started (or will start) work in this area.
A workaround is to do a 1.6.2 with incremental facet update disabled and only bulk facet update enabled.

**Expected behavior**
Faceted documents, when their facets are modified, must be ordered correctly, whether they are strings or numbers, which doesn't matter either.

**Meilisearch version:**
All versions with the incremental indexing: v1.6.0 and v1.6.1.

ManyTheFish:
# Pull Request

## Related issue
Fixes #4367
Fixes #4409

## What does this PR do?

- Add a test reproducing #4409
- Fix #4409 by removing a document from a level only if it is no more present in all the linked sub-level nodes
- Optimize facet Incremental indexing by creating or deleting a complete level once per field id instead of for each facet value
- Optimize facet Incremental indexing by doing the additions and the deletions in the same process instead of doing them separately

### DIFF
--- a/milli/src/update/facet/incremental.rs
+++ b/milli/src/update/facet/incremental.rs
@@ -201,7 +201,7 @@ impl FacetsUpdateIncrementalInner {
         match (old_value, add_docids, del_docids) {
             // Addition + deletion on an existing value
             (Some(FacetGroupValue { bitmap, .. }), Some(add_docids), Some(del_docids)) => {
-                let value = FacetGroupValue { bitmap: bitmap - del_docids | add_docids, size: 1 };
+                let value = FacetGroupValue { bitmap: (bitmap - del_docids) | add_docids, size: 1 };
                 self.db.put(txn, &key, &value)?;
                 Ok(ModificationResult::InPlace)
             }
@@ -575,7 +575,7 @@ impl FacetsUpdateIncrementalInner {
     fn delete_level(&self, txn: &mut RwTxn, highest_level_prefix: &[u8]) -> Result<()> {
         let mut to_delete = vec![];
         let mut iter =
-            self.db.remap_types::<Bytes, Bytes>().prefix_iter(txn, &highest_level_prefix)?;
+            self.db.remap_types::<Bytes, Bytes>().prefix_iter(txn, highest_level_prefix)?;
         for el in iter.by_ref() {
             let (k, _) = el?;
             to_delete.push(
@@ -603,7 +603,7 @@ impl FacetsUpdateIncrementalInner {
         let mut groups_iter = self
             .db
             .remap_types::<Bytes, FacetGroupValueCodec>()
-            .prefix_iter(txn, &highest_level_prefix)?;
+            .prefix_iter(txn, highest_level_prefix)?;
 
         let nbr_new_groups = size_highest_level / self.group_size as usize;
         let nbr_leftover_elements = size_highest_level % self.group_size as usize;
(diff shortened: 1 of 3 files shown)


milli/src/update/facet/incremental.rs
```diff
@@ -351,49 +312,298 @@ impl FacetsUpdateIncrementalInner {
...
+    /// - `ModificationResult::Reduce/Expand` means that modifying the `facet_value` into the `level` resulted in a change in the
+    /// bounds of the keys of the level. For example, removing a document id from the facet value
+    /// `3` might have caused the facet value `3` to have no corresponding document in level 0. Therefore,
+    /// in level 1, the key with the left bound `3` had to be changed to the next facet value (e.g. 4).
+    /// In that case `ModificationResult::Reduce` is returned. The parent of the reduced key may need to adjust
+    /// its left bound as well.
+    ///
+    /// - `ModificationResult::Nothing` means that modifying the `facet_value` didn't have any impact into the `level`.
+    /// This case is reachable when a document id is removed from a sub-level node but is still present in another one.
+    /// For example, removing `2` from a document containing `2` and `3`, the document id will removed form the `level 0` but should remain in the group node [1..4] in `level 1`.
```
Kerollmops, line 378: It could be better to move that on the enum itself.

milli/src/update/facet/incremental.rs
```diff
@@ -351,49 +312,298 @@ impl FacetsUpdateIncrementalInner {
         self.db.put(txn, &group_left.0.as_ref(), &group_left.1)?;
         self.db.put(txn, &group_right.0.as_ref(), &group_right.1)?;
 
-        Ok(InsertionResult::Insert)
+        Ok(ModificationResult::Insert)
+    }
+
+    fn trim_del_docids<'a>(
```
Kerollmops, line 318: A small sentence to explain this function could be great :)

milli/src/update/facet/incremental.rs
```diff
@@ -351,49 +312,298 @@ impl FacetsUpdateIncrementalInner {
...
+                        insertion_value_size,
+                        ids,
+                    )
+                })
+                .transpose()?
+                .filter(|ids| !ids.is_empty());
+            if let Some(del_docids) = del_docids {
+                updated_value.bitmap -= &*del_docids;
+                insertion_value_was_modified = true;
+            }
```
Kerollmops, line 484: Can you merge this in this `if let Some`, please?

Kerollmops (changes requested):

Kerollmops (approved):
That's very good!
bors merge

Issue #4432, opened 2024-02-21
Make the processing of batches atomic 


irevoire:
**Describe the bug**
Currently, when meilisearch processes the tasks, there is three phases:
1. Find the tasks you need to process in your batch
2. Actually, process the batch, for example, index documents
3. Write the results in the task queue and delete the associated update files.

We just realized that there was a bug in step 3️⃣:
4. We take a write transaction on the task store
5. Loop over all the processed tasks
6. For each task: Write its new status and delete its associated update file
7. Commit the content of the task store

Here, if a crash/restart happens during the loop of 6️⃣, Meilisearch will restart, recreate a batch that contains at least the same tasks, and in step 2️⃣ realize that some of the update files have been deleted.
=> We just lost some data.

**To Reproduce**
1. Index a ton of thing
2. Restart meilisearch in the middle of step 6️⃣ 

**Expected behavior**
We don't lose any tasks

**Meilisearch version:**
Everything above and including v0.30.0

curquiza:
Fixed by #4435, will be release in v1.7.0-rc.2

PR #4435, opened 2024-02-22
Make update file deletion atomic

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4432

## What does this PR do?
- Adds a bunch of logs to help debug this kind of issue in the future
- Delete the update files AFTER committing the update in the `index-scheduler` (thus, if a restart happens, we are able to re-process the batch successfully)
- Multi-thread the deletion of all update files.

### DIFF
--- a/index-scheduler/src/lib.rs
+++ b/index-scheduler/src/lib.rs
@@ -60,6 +60,7 @@ use meilisearch_types::milli::{self, CboRoaringBitmapCodec, Index, RoaringBitmap
 use meilisearch_types::task_view::TaskView;
 use meilisearch_types::tasks::{Kind, KindWithContent, Status, Task};
 use puffin::FrameView;
+use rayon::prelude::{IntoParallelIterator, ParallelIterator};
 use roaring::RoaringBitmap;
 use synchronoise::SignalEvent;
 use time::format_description::well_known::Rfc3339;
@@ -1175,6 +1176,9 @@ impl IndexScheduler {
                 #[cfg(test)]
                 self.breakpoint(Breakpoint::ProcessBatchSucceeded);
 
+                let mut success = 0;
+                let mut failure = 0;
+
                 #[allow(unused_variables)]
                 for (i, mut task) in tasks.into_iter().enumerate() {
                     task.started_at = Some(started_at);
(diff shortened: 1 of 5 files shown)


index-scheduler/src/lib.rs
```diff
@@ -1262,6 +1270,24 @@ impl IndexScheduler {
 
         wtxn.commit().map_err(Error::HeedTransaction)?;
 
+        // Once the tasks are commited, we should delete all the update files associated ASAP to avoid leaking files in case of a restart
+        tracing::debug!("Deleting the upadate files");
+
+        ids.into_par_iter().try_for_each(|id| -> Result<()> {
+            let rtxn = self.read_txn()?;
```
Kerollmops, line 1277: it would be better to create the read txn once (using the appropriate feature). Opening a lot of transactions in parallel can be slower than using the same one on different threads. What do you think?

index-scheduler/src/lib.rs
```diff
@@ -1225,16 +1232,18 @@ impl IndexScheduler {
...
+
                 return Ok(TickOutcome::TickAgain(0));
             }
             // In case of a failure we must get back and patch all the tasks with the error.
             Err(err) => {
                 #[cfg(test)]
                 self.breakpoint(Breakpoint::ProcessBatchFailed);
                 let error: ResponseError = err.into();
-                for id in ids {
+                for id in ids.iter() {
```
Kerollmops, line 1244:
```suggestion
                for id in ids {
```

index-scheduler/src/lib.rs
irevoire: How would I do that? By using the `sync-read-txn` heed feature?

Because yeah, I wasn’t happy about this solution, I thought of using [`for_each_init`](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.for_each_init) with this code:
```rust
...
        ids.into_par_iter().try_for_each_init(
            || self.read_txn(),
            |rtxn, id| -> Result<()> {
                // let rtxn = self.read_txn()?;
                let rtxn = rtxn.as_ref().unwrap();
                let task = self
                    .get_task(rtxn, id)
...
```
But from what I understand, it pretty much ends up calling my closure for every rayon job (thus every iteration, right?).

irevoire: The other easy idea could be to split up the roaring bitmap into as many parts as we have threads and then let each thread open **one** `rtxn` and delete the batch of files on his side.
That’s probably better, but if one finishes early, the other will be waiting.

We could maybe fix that with an AtomicU64 + the `.select` method on the roaring bitmap as we did in Arroy. That seems a bit overkill, but tbh, it’s not that complex to write, and it works.

But let me know if you have anything else in mind

irevoire: I implemented the last solution

Kerollmops (approved):
It's an absolutely wonderful PR ❤️
bors merge

PR #4462, opened 2024-03-06
Divide threshold by ten

ManyTheFish:
Change the facet incremental vs bulk indexing threshold to better fit our user needs, it might be changed in the future if we have more insights


dureuill (approved):
bors merge

PR #4463, opened 2024-03-06
Add tests when the field limit is reached

irevoire:
# Pull Request

## Related issue
Related to https://github.com/meilisearch/meilisearch/discussions/4429#discussioncomment-8689101

This user found out that the error message we’re supposed to return when the maximum number of attributes is reached is _not_ returned in some cases

## What does this PR do?
- This PR adds four tests around the maximum number of attributes:
  1. Add a document with u16::MAX + 1 fields - Meilisearch panics
  2. Add two documents which together adds up to u16::MAX + 1 fields - Meilisearch returns the expected error 
  3. Add a document with u16::MAX + 1 **nested fields** - No error message but the document isn’t indexed
  4. Add two documents which together add up to u16::MAX + 1 nested fields - Meilisearch doesn’t return any error but doesn’t index the document

## PR checklist

### DIFF
--- a/meilisearch/tests/documents/add_documents.rs
+++ b/meilisearch/tests/documents/add_documents.rs
@@ -1237,8 +1237,8 @@ async fn error_add_documents_missing_document_id() {
 }
 
 #[actix_rt::test]
-#[ignore] // // TODO: Fix in an other PR: this does not provoke any error.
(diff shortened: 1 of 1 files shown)


Kerollmops (approved):
bors merge

PR #4466, opened 2024-03-07
Implements the search cutoff

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4488

## What does this PR do?
- Adds a cutoff to the bucket sort after 150ms has been spent
- Adds a new setting to customize the default value of 150ms
- When the time is exceeded, we exit early with what we had the time to sort
- If the cutoff has been reached, the search details are updated with a new `Skip` ranking details for the ranking rules that were skipped
- Adds analytics to measure the total number of degraded search requests
- Adds the number of degraded search requests to the Prometheus metrics and Grafana dashboard
- The cutoff **must not** skip the filters; otherwise, we would leak documents to people who don’t have the right to see them

### DIFF
--- a/milli/src/score_details.rs
+++ b/milli/src/score_details.rs
@@ -17,6 +17,9 @@ pub enum ScoreDetails {
     Sort(Sort),
     Vector(Vector),
     GeoSort(GeoSort),
+
+    /// Returned when we don't have the time to finish applying all the subsequent ranking-rules
+    Skipped,
 }
 
 #[derive(Clone, Copy)]
@@ -50,6 +53,7 @@ impl ScoreDetails {
             ScoreDetails::Sort(_) => None,
             ScoreDetails::GeoSort(_) => None,
             ScoreDetails::Vector(_) => None,
+            ScoreDetails::Skipped => Some(Rank { rank: 0, max_rank: 1 }),
         }
     }
 
@@ -97,6 +101,7 @@ impl ScoreDetails {
             ScoreDetails::Vector(vector) => RankOrValue::Score(
                 vector.value_similarity.as_ref().map(|(_, s)| *s as f64).unwrap_or(0.0f64),
             ),
+            ScoreDetails::Skipped => RankOrValue::Score(0.),
         }
     }
 
@@ -254,6 +259,13 @@ impl ScoreDetails {
                         "similarity": similarity,
                     });
                     details_map.insert(vector, details);
+                    order += 1;
+                }
+                ScoreDetails::Skipped => {
+                    details_map.insert(
+                        "skipped".to_string(),
+                        serde_json::Number::from_f64(0.).unwrap().into(),
+                    );
                     order += 1;
                 }
             }
--- a/milli/src/score_details.rs
+++ b/milli/src/score_details.rs
@@ -101,7 +101,7 @@ impl ScoreDetails {
             ScoreDetails::Vector(vector) => RankOrValue::Score(
                 vector.value_similarity.as_ref().map(|(_, s)| *s as f64).unwrap_or(0.0f64),
             ),
-            ScoreDetails::Skipped => RankOrValue::Score(0.),
+            ScoreDetails::Skipped => RankOrValue::Rank(Rank { rank: 0, max_rank: 1 }),
         }
     }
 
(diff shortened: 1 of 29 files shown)


milli/src/score_details.rs
```diff
@@ -97,6 +101,7 @@ impl ScoreDetails {
             ScoreDetails::Vector(vector) => RankOrValue::Score(
                 vector.value_similarity.as_ref().map(|(_, s)| *s as f64).unwrap_or(0.0f64),
             ),
+            ScoreDetails::Skipped => RankOrValue::Score(0.),
```
dureuill, line 104: This should probably be a Rank

milli/src/score_details.rs
```diff
@@ -256,6 +261,13 @@ impl ScoreDetails {
                     details_map.insert(vector, details);
                     order += 1;
                 }
+                ScoreDetails::Skipped => {
+                    details_map.insert(
+                        "skipped".to_string(),
+                        serde_json::Number::from_f64(0.).unwrap().into(),
```
dureuill, line 267: - the value of a detail must be an object (here it is a float)
- it must contain the "order" field.
- for "skipped", no more details are necessary (we don't need a 0).

So for skipped, the details should look:

```suggestion
                        serde_json::json!({"order": order})
```

milli/src/search/mod.rs
```diff
@@ -244,6 +260,7 @@ impl fmt::Debug for Search<'_> {
             .field("words_limit", words_limit)
             .field("distribution_shift", distribution_shift)
             .field("embedder_name", embedder_name)
+            .field("time_bduget", time_budget)
```
dureuill, line 263:
```suggestion
            .field("time_budget", time_budget)
```

milli/src/score_details.rs
```diff
@@ -256,6 +261,13 @@ impl ScoreDetails {
                     details_map.insert(vector, details);
                     order += 1;
                 }
+                ScoreDetails::Skipped => {
+                    details_map.insert(
```
dureuill, line 265: This will need updating in the scoreDetails spec

milli/src/search/new/bucket_sort.rs
```diff
@@ -154,6 +160,28 @@ pub fn bucket_sort<'ctx, Q: RankingRuleQueryTrait>(
...
 
     while valid_docids.len() < length {
+        if time_budget.exceeded() {
+            loop {
+                let bucket = std::mem::take(&mut ranking_rule_universes[cur_ranking_rule_index]);
+                ranking_rule_scores.push(ScoreDetails::Skipped);
+                maybe_add_to_results!(bucket);
+                ranking_rule_scores.pop();
+
+                if cur_ranking_rule_index == 0 {
```
dureuill, line 170: This is already handled inside of `back!`, although allegedly this is a bit dangerous as the `break` in `back!` is not intended to break the same loop.

milli/src/search/new/tests/cutoff.rs
```diff
@@ -0,0 +1,419 @@
...
+    index
+}
+
+#[test]
+fn basic_degraded_search() {
+    let index = create_index();
+    let rtxn = index.read_txn().unwrap();
+
+    let mut search = Search::new(&rtxn, &index);
+    search.query("hello puppy kefir");
```
dureuill, line 61: ❤️

dureuill (changes requested):
Thank you for implementing this!

We need to change the score details

milli/src/search/hybrid.rs
```diff
@@ -106,6 +106,7 @@ impl ScoreWithRatioResult {
             candidates: left.candidates | right.candidates,
             documents_ids,
             document_scores,
+            degraded: false,
```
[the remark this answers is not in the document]
irevoire: yep definitely

milli/src/search/new/bucket_sort.rs
irevoire: Yes, I was not super comfortable relying on this mechanism (or even on the `back!` macro at all tbh)

dureuill: Agreed

dureuill (approved):
Looking good, and with good tests ☀️ 

bors merge

irevoire:
bors merge

PR #4475, opened 2024-03-12
Allow running benchmarks without sending results to the dashboard

dureuill:
Adds a `--no-dashboard` option to avoid sending results to the dashboard.


irevoire (approved):
Nice, that’s clean

bors merge

PR #4476, opened 2024-03-12
Make the `/facet-search` route use the `sortFacetValuesBy` setting

### LINKED ISSUE #4423: Make `sortFacetValuesBy` work with `/facet-search` endpoint
Related product team resources: [PRD](https://www.notion.so/meilisearch/Faceted-search-improvements-7c4387305e4540a29d2f31ef6720dc7d) (_internal only_)
Related spec: WIP

## Motivation

Bring consistency in our API, so removes confusion. Also, requested by users.

## Usage

`sortFacetValuesBy` only works on the `/search` endpoint. Should work for `/facet-search` too.

Note for the @meilisearch/docs-team: Using the new `sortFacetValuesBy` setting can slow down the `/facet-search` requests as Meilisearch iterates over the whole list of facet values and computes the count of documents on every entry. That is hardly or even impossible to optimize correctly.

## TODO

- [ ] Merge changes into `main`
- [ ] Update the spec

## Impacted teams

@meilisearch/docs-team

Kerollmops:
This PR fixes #4423 by ensuring that the `/facet-search` route uses the `sortFacetValuesBy` setting.

Note for the documentation team (to be moved in the tracking issue): Using the new `sortFacetValuesBy` setting can slow down the facet-search requests as Meilisearch iterates over the whole list of facet values and computes the count of documents on every entry. That is hardly or even impossible to optimize correctly.

### TODO
 - [x] Create a custom HashMap wrapper for the facet `OrderBy` settings.
         This wrapper will return the `OrderBy` setting of the facet, if not defined will use the default `*` one, and if not there either (strange) will fall back on the lexicographic one.
- [x] Create a `ValuesCollection` wrapper that implements the logic for the lexicographic and count order by.
  - [x] Use it when there is no search query.
  - [x] Use it when there is a search query with and without allowed typos.
  - [x] Do not change the original logic, only use a wrapper.
- [x] Add tests


milli/src/order_by_map.rs
```diff
@@ -0,0 +1,57 @@
...
+        self.0.into_iter()
+    }
+}
+
+impl<'de> Deserialize<'de> for OrderByMap {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let mut map = Deserialize::deserialize(deserializer).map(OrderByMap)?;
```
irevoire, line 52: Nice

milli/src/search/facet/search.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    pub fn query(&mut self, query: impl Into<String>) -> &mut Self {
+        self.query = Some(query.into());
+        self
+    }
+
+    pub fn max_values(&mut self, max: usize) -> &mut Self {
+        self.max_values = max;
+        self
+    }
```
irevoire, line 51: Is there a true maximum in meilisearch? We can’t let users ask for as many values as they want, right?

milli/src/search/facet/search.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+impl ValuesCollection {
+    pub fn by_lexicographic(max: usize) -> Self {
+        ValuesCollection::Lexicographic { max, content: Vec::with_capacity(max) }
+    }
+
+    pub fn by_count(max: usize) -> Self {
+        ValuesCollection::Count { max, content: BinaryHeap::with_capacity(max) }
+    }
```
irevoire, line 282: Here if the max hasn’t been checked prior, we could consume a ton of ram

irevoire (approved):

milli/src/search/facet/search.rs
Kerollmops: You are right. I will replace the `with_capacity` with `new` in this case. There is not important point in optimizing this small part.

milli/src/search/facet/search.rs
Kerollmops: There is no actual limit, but there is a warning in the documentation.

irevoire (approved):
bors merge

PR #4477, opened 2024-03-12
Add documentation for benchmarks

dureuill:
See [CONTRIBUTING.md](https://github.com/meilisearch/meilisearch/blob/benchmark-docs/CONTRIBUTING.md#logging)

### DIFF
--- a/BENCHMARKS.md
+++ b/BENCHMARKS.md
@@ -317,6 +317,14 @@ They are JSON files with the following structure (comments are not actually supp
 }
 ```
 
+### Adding new assets
+
+Assets reside in our DigitalOcean S3 space. Assuming you have team access to the DigitalOcean S3 space:
+
+1. go to <https://cloud.digitalocean.com/spaces/milli-benchmarks?i=d1c552&path=bench%2Fdatasets%2F>
+2. upload your dataset:
+   1. if your dataset is a single file, upload that single file using the "upload" button,
+   2. otherwise, create a folder using the "create folder" button, then inside that folder upload your individual files.
 
 ## Upgrading `https://bench.meilisearch.dev`
 
--- a/BENCHMARKS.md
+++ b/BENCHMARKS.md
@@ -1,0 +1,354 @@
+# Benchmarks
+
+Currently this repository hosts two kinds of benchmarks:
+
+1. The older "milli benchmarks", that use [criterion](https://github.com/bheisler/criterion.rs) and live in the "benchmarks" directory.
+2. The newer "bench" that are workload-based and so split between the [`workloads`](./workloads/) directory and the [`xtask::bench`](./xtask/src/bench/) module.
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -81,6 +81,30 @@ Meilisearch follows the [cargo xtask](https://github.com/matklad/cargo-xtask) wo
 
 Run `cargo xtask --help` from the root of the repository to find out what is available.
 
+### Logging
```
irevoire, line 84: Nice section

BENCHMARKS.md
```diff
@@ -0,0 +1,354 @@
...
+3. Download the binary locally, send it to the server:
+  ```
+  scp -6 ~/Downloads/benchboard root@\[<ipv6-address>\]:/bench/new-benchboard
+  ```
+  Note that the ipv6 must be between escaped square brackets for SCP.
+4. SSH to the server:
+  ```
+  ssh root@<ipv6-address>
+  ```
+  Note the the ipv6 must **NOT** be between escaped square brackets for SSH 🥲
```
irevoire, line 336: Why can’t we use the `bench.meilisearch.dev` url instead of the ipv6?

irevoire (approved):

BENCHMARKS.md
dureuill: we're only serving port 443 and port 80.

We could add port 22, but also I think it is not necessary since we'll be using the raw ipv6 only for maintenance operations.

irevoire: Ah ok, true

BENCHMARKS.md
```diff
@@ -0,0 +1,354 @@
...
+
+You can explicitly request a detailed comparison by span with the `main` branch, the branch or origin, or any previous commit, by clicking the links at the bottom of the workload invocation.
+
+In the detailed comparison view, the spans are sorted by improvements, regressions, stable (no statistically significant change) and unstable (the span runtime is comparable to its standard deviation).
+
+You can click on the name of any span to get a box plot comparing the target commit with multiple commits of the selected branch.
+
+### Without dashboard
+
+After the workloads are done running, the reports will live in the Meilisearch repository, in the `bench/reports` directory (by default).
```
curquiza, line 84: Let me know if I understand correctly. `bench/` sub-folder content will never be pushed on `main`. It will only be created and we run bench locally. Am I right?

dureuill: yes, that's correct. It is in the `.gitignore`.

ManyTheFish (approved):

curquiza (dismissed):

ManyTheFish (approved):
Thank you @dureuill, for this addition 😄

Kerollmops (approved):
It looks perfect to me! There is a lot of helpful information to ensure we don't lose knowledge and that the engine is reliable 😃

dureuill:
bors merge

PR #4503, opened 2024-03-19
Add settings diff indexing benchmarks

ManyTheFish:
Add several benchmarks targetting settings diff-indexing enhancements


dureuill:
Looking good
Capture d’écran 2024-03-19 à 11 20 47


I'll wait for the workloads to complete, then we'll accept and merge

dureuill (approved):
bors merge

Issue #4506, opened 2024-03-19
Adds a timeout to the webhook disconnection


irevoire:
**Describe the bug**
Currently, when sending tasks through our webhook, we have a timeout of 30s while establishing the connection.
But we don't have a timeout when the connection hangs.

**Meilisearch version:**
v1.7+

curquiza:
Fixed by #4508

PR #4508, opened 2024-03-19
Add a timeout to the webhook

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4506

## What does this PR do?
- Adds a timeout of 30s (the value we were already using while trying to establish the connection)
- When the timeout is reached, a log is emitted, and the engine continues its work:
```
2024-03-19T13:52:57.676316Z ERROR index_scheduler: While sending data to the webhook: http://localhost:3001/: Network Error: Network Error: Error encountered in the status line: timed out reading response
```


Kerollmops (approved):
Ok, thank you, Tamo. I just checked, and there, indeed, [are no timeout by default](https://docs.rs/ureq/latest/ureq/struct.AgentBuilder.html#method.timeout_write) 🤔
bors merge

PR #4516, opened 2024-03-20
Update sprint_issue.md

curquiza:
Following decision made about specification

Also
- removed useless parts of the template
- add automatic labels -> better to forget to remove them rather than forgetting to add them (some mistakes happened in the past)


Kerollmops (approved):
I love the changes!
bors merge

PR #4535, opened 2024-03-26
Support Negative Keywords

### LINKED ISSUE #4422: Negative operator
Related product team resources: [PRD](https://www.notion.so/meilisearch/Negative-Keyword-Search-8769a40012cc430296b9a63a107af2e6?pvs=18) (_internal only_)
Related spec: WIP

## Motivation

Required by a lot of users on ProductBoard. Also, easy to do.

## Usage

https://meilisearch.notion.site/API-usage-of-negative-operator-5279c6aaeaac4ab9a365cab273970f6b?pvs=74

## TODO

- [ ] Merge changes into `main`
- [ ] Update the spec

## Impacted teams

@meilisearch/docs-team

Kerollmops:
This PR fixes #4422 by supporting `-` before any word in the query.

The minus symbol `-`, from the ASCII table, is not the only character that can be considered the negative operator. You can see the two other matching characters under the `Based on "-" (U+002D)` section on [this unicode reference website](https://www.compart.com/en/unicode/U+002D).

It's important to notice the strange behavior when a query includes and excludes the same word; only the derivative ( synonyms and split) will be kept:
 - If you input `progamer -progamer`, the engine will still search for `pro gamer`.
 - If you have the synonym `like = love` and you input `like -like`, it will still search for `love`.

## TODO
 - [x] Add analytics
 - [x] Add support to the `-` operator
 - [x] Make sure to support spaces around `-` well
 - [x] Support phrase negation
 - [x] Add tests

### DIFF
--- a/meilisearch/src/analytics/segment_analytics.rs
+++ b/meilisearch/src/analytics/segment_analytics.rs
@@ -580,6 +580,7 @@ pub struct SearchAggregator {
     total_received: usize,
     total_succeeded: usize,
     total_degraded: usize,
+    total_used_negative_operator: usize,
     time_spent: BinaryHeap<usize>,
 
     // sort
@@ -760,11 +761,15 @@ impl SearchAggregator {
             facet_distribution: _,
             facet_stats: _,
             degraded,
+            used_negative_operator,
         } = result;
 
         self.total_succeeded = self.total_succeeded.saturating_add(1);
         if *degraded {
             self.total_degraded = self.total_degraded.saturating_add(1);
+        }
+        if *used_negative_operator {
+            self.total_used_negative_operator = self.total_used_negative_operator.saturating_add(1);
         }
         self.time_spent.push(*processing_time_ms as usize);
     }
@@ -808,6 +813,7 @@ impl SearchAggregator {
             embedder,
             hybrid,
             total_degraded,
+            total_used_negative_operator,
         } = other;
 
         if self.timestamp.is_none() {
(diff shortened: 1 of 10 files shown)


ManyTheFish (dismissed):

ManyTheFish (approved):
bors merge

Foddy:
Is there a way to deactivate this behavior or make it usable with a different character? Unfortunately, I have the problem that I have a list of order IDs that contain a "-". When I want to search for the full order ID, I don't get any results because of this?

Or does the search request need to be structured differently? Unfortunately, I can't find anything about this in the documentation :( ...

ManyTheFish:
Hello @Foddy,
Unfortunately, It's not possible to deactivate this feature directly.
However, you could try to apply the following settings to your index:
`PATCH /settings`
```json
{
  "separatorTokens": [" -"]
}
``` 

This will add ` -` (space + dash) to the list of separators, making any sequence of "space + dash" one unique token and not `[" ", "-"]`, which is the pattern detected by the engine for the negative keyword search. Changing this setting will virtually deactivate the feature.
Additionaly, if you never want to consider `-` as a separator, you could apply the following settings:
`PATCH /settings`
```json
{
  "separatorTokens": [" -"]
  "nonSeparatorTokens": ["-"]
}
```

Foddy:
Hi @ManyTheFish,

Wow, thank you so much! This trick works perfectly. It would be great if in the future we could somehow adjust these special characters in the settings, for example with an exclamation mark (!) instead of a dash for negative search queries. Dashes appear too often in my datasets unfortunately.

PR #4548, opened 2024-04-03
v1.8 hybrid search changes

dureuill:
Implements the search changes from the [usage page](https://meilisearch.notion.site/v1-8-AI-search-API-usage-135552d6e85a4a52bc7109be82aeca42#40f24df3da694428a39cc8043c9cfc64)

### ⚠️ Breaking changes in an experimental feature:

- Removed the `_semanticScore`. Use the `_rankingScore` instead.
- Removed `vector` in the response of the search (output was too big).
- Removed all the vectors from the `vectorSort` ranking score details
  - target vector appearing in the name of the rule
  - matched vector appearing in the details of the rule

### Other user-facing changes

- Added `semanticHitCount`, indicating how many hits were returned from the semantic search. This is especially useful in the hybrid search.
- Embed lazily: Meilisearch no longer generates an embedding when the keyword results are "good enough".
- Graceful embedding failure in hybrid search: when doing hybrid search (`semanticRatio in ]0.0, 1.0[`), an embedding failure no longer causes the search request to fail. Instead, only the keyword search is performed. When doing a full vector search (`semanticRatio==1.0`), a failure to embed will still result in failing that search.

### DIFF
--- a/meilisearch/src/routes/indexes/search.rs
+++ b/meilisearch/src/routes/indexes/search.rs
@@ -277,7 +277,7 @@ pub fn search_kind(
         features.check_vector("Passing `hybrid` as a query parameter")?;
     }
 
-    // regardless of anything, always do a semantic search when we don't have a vector and the query is whitespace or missing
+    // regardless of anything, always do a keyword search when we don't have a vector and the query is whitespace or missing
     if query.vector.is_none() {
         match &query.q {
             Some(q) if q.trim().is_empty() => return Ok(SearchKind::KeywordOnly),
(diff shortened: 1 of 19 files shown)


meilisearch/src/routes/indexes/search.rs
```diff
@@ -265,76 +263,58 @@ pub async fn search_with_post(
...
+) -> Result<SearchKind, ResponseError> {
+    if query.vector.is_some() {
+        features.check_vector("Passing `vector` as a query parameter")?;
+    }
+
+    if query.hybrid.is_some() {
+        features.check_vector("Passing `hybrid` as a query parameter")?;
+    }
+
+    // regardless of anything, always do a semantic search when we don't have a vector and the query is whitespace or missing
```
irevoire, line 280:
```suggestion
    // regardless of anything, always do a keyword search when we don't have a vector and the query is whitespace or missing
```

irevoire (changes requested):

irevoire (approved):
Thanks,
bors merge

dureuill:
bors merge

PR #4557, opened 2024-04-08
Bring back change from v1.7.5 to main

irevoire:


Cargo.toml
```diff
@@ -21,7 +21,7 @@ members = [
 ]
 
 [workspace.package]
-version = "1.7.4"
+version = "1.7.5"
```
curquiza, line 24: Why does it change the package version from v1.7.4 to v1.7.5?
https://github.com/meilisearch/meilisearch/blob/d6b6cd322c26768180758ac9936b9eca9e4db987/Cargo.toml#L24

irevoire: Oh yes, that’s super strange! When I try to do the rebase myself on my side, I have a conflict on the `Cargo.toml` and `Cargo.lock`; I don’t know what GitHub is going to do if we bors merge this PR 🤔 

Also, maybe that was the reason why we were updating the version at the very last moment in our previous release process? 🤔 

I’ll create a new tmp-PR that sets the right version

irevoire:
Closing in favor of https://github.com/meilisearch/meilisearch/pull/4560

PR #4576, opened 2024-04-16
increase the default search time budget from 150ms to 1.5s

### LINKED ISSUE #4575: Increase the default search cutoff
Right now, the search cutoff is set at 150ms, which is perfect for a good search-as-you-type experience except... when we do anything related to AI.
Everything related to AI is just so slow that keeping a normal cutoff would stop the search requests before doing anything meaningful.

As a first fix for v1.8, we’re simply going to increase the search cutoff from 150ms to 1.5s; this should still be enough to protect the instance against malicious requests without blocking the AI stuff.
It won’t improve the search-as-you-type experience.

irevoire:
# Pull Request

## Related issue
Fixes #4575

## What does this PR do?
- increase the default search time budget from 150ms to 1.5s

### DIFF
--- a/milli/src/lib.rs
+++ b/milli/src/lib.rs
@@ -128,7 +128,7 @@ impl fmt::Debug for TimeBudget {
 
 impl Default for TimeBudget {
     fn default() -> Self {
-        Self::new(std::time::Duration::from_millis(150))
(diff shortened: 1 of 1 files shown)


ManyTheFish (approved):
LGTM
bors merge

PR #4581, opened 2024-04-17
Always show facet numbers in alpha order in the facet distribution

### LINKED ISSUE #4559: sortFacetValuesBy does not work with integers - facet values are sorted randomly
**Describe the bug**

I have a field `release_year` in my documents of type `integer`.

```
{
    "objectID": "01be6693-797c-4ec2-a9cb-5618da8d9c88",
    "release_date": "2018-03-23",
    "release_year": 2018,
    "release_decade": 2010
}
```

The field is configured as filterable. Returned facets in `facetDistribution` are **not**  sorted based in the `sortFacetValuesBy`-setting.

**To Reproduce**
1. Add documents like the one above to an index.
2. Configure the index settings to have the desired field to be filterable.
3. Search for documents with configured filters.
5. See wrong sorted facets in `facetDistribution`.

**Expected behavior**
The facets are sorted correctly according to the `sortFacetValuesBy`-setting.

**Meilisearch version:**
Meilisearch 1.7.4

Kerollmops:
This PR fixes #4559 by making sure that the number facets (facets that come from numbers from the documents) are always displayed in alpha order, even when there is a small amount to display.

The issue was due to some algorithms executed when the number of facet values to display was small. We can see that now, facet values are always displayed correctly.

```json
"facetDistribution": {
    "release_year": {
        "2010": 1,
        "2011": 1,
        "2012": 1,
        "2013": 1,
        "2014": 1,
        "2015": 1,
        "2016": 1,
        "2017": 1,
        "2018": 1,
        "2019": 19,
        "2020": 1,
        "2021": 1,
        "2022": 1,
        "2023": 1,
        "2024": 1,
        "2025": 1
    }
}
```

### DIFF
--- a/milli/src/search/facet/facet_distribution.rs
+++ b/milli/src/search/facet/facet_distribution.rs
@@ -97,6 +97,7 @@ impl<'a> FacetDistribution<'a> {
     ) -> heed::Result<()> {
         match facet_type {
             FacetType::Number => {
+                let mut lexicographic_distribution = BTreeMap::new();
                 let mut key_buffer: Vec<_> = field_id.to_be_bytes().to_vec();
 
                 let distribution_prelength = distribution.len();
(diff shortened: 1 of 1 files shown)


ManyTheFish (approved):
letsgo!

bors merge

PR #4593, opened 2024-04-22
Stop crashing when panic occurs in thread pool

### LINKED ISSUE #4362: Stops crashing the whole instance when a bug occurs in the indexing process
**Describe the bug**
Currently, in milli, we use `rayon` quite a lot but don't configure the thread pool much.
If something panics in the thread pool, rayon calls `std::process::exit` and kills the whole instance instead of being caught in the index scheduler like we're supposed to do. 

This issue happened to us twice on this release (v1.6) with:
- https://github.com/meilisearch/meilisearch/issues/4361
- https://github.com/meilisearch/meilisearch/issues/4333

**Expected behavior**
Meilisearch continues its life, and the crash gets reported to us in an error message or in the logs, at the very least.

**Screenshots**

**Meilisearch version:**
All versions are impacted since the v0.21, but when there are no bugs, there is no crash.
To my knowledge, currently only the v1.6 contains bugs.

Kerollmops:
This PR fixes #4362 by introducing a new boolean to catch panics in the rayon thread pool. The boolean is read after performing the operations in rayon, and the indexation process is stopped. This first version doesn't expose the panic message but marks the task as failed.

The current implementation exposes a `ThreadPoolNoAbort` wrapper. The `rayon::ThreadPool` has been wrapped to check that nothing went wrong after running the `ThreadPool::install` function. An atomic boolean and some `store/load` logic make the system work efficiently.

Before, Meilisearch was completely crashing...

Capture d’écran 2024-04-22 à 15 49 02

Now, it handles the panics correctly and marks the task as failed.

Capture d’écran 2024-04-22 à 15 42 14

### DIFF
--- a/meilisearch/src/option.rs
+++ b/meilisearch/src/option.rs
@@ -6,6 +6,7 @@ use std::num::ParseIntError;
 use std::ops::Deref;
 use std::path::PathBuf;
 use std::str::FromStr;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::{env, fmt, fs};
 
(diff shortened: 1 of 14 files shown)


curquiza:
@Kerollmops if you want to make it available for v1.8, you should not merge into `main`

ManyTheFish (approved):
bors merge

PR #4622, opened 2024-05-07
Bump Rustls to non-vulnerable versions

### LINKED ISSUE #4599: Vulnerabilities to fix for SOC2
Dependabot reported the following vulnerabilities and they have to be fixed for SOC2

## More context

When we will officially be SOC2 compliant, we will have
- 15 days to fix a critical vulnerability
- 30 days to fix a high vulnerability
- 45 days to fix a medium vulnerability
- 90 days to fix a low vulnerability

In the meantime, let's do it during v1.9.0 without rushing. And you are now aware of what will be expected from Q3 😊 

## TODO

Both of them are high security vulnerabilities
- https://github.com/meilisearch/meilisearch/security/dependabot/23
- https://github.com/meilisearch/meilisearch/security/dependabot/24

Kerollmops:
This PR Fixes #4599 by bumping the Rustls dependency to v0.21.12 and [ureq to v2.9.7](https://github.com/algesten/ureq/blob/main/CHANGELOG.md#297) (which bump rustls to v0.22.4).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -487,6 +487,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
 
 [[package]]
+name = "base64"
+version = "0.22.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
+
+[[package]]
 name = "base64ct"
 version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


irevoire (approved):

irevoire:
to be 100% sure, you can run a `cargo update -p rustls` so it’ll update even if we don’t update ureq

Kerollmops:
Note: I am in the process of moving meilisearch to rustls v0.22.4, this way we only have a single rustls dependency.

Kerollmops:
> to be 100% sure, you can run a `cargo update -p rustls` so it’ll update even if we don’t update ureq

I did that but I found out that there are **two** vulnerable dependencies and ureq is requesting one.

> Note: I am in the process of moving meilisearch to rustls v0.22.4, this way we only have a single rustls dependency.

There is an issue with this, segment is using an old version of reqwest and therefore the version 0.21.x of rustls. @irevoire do you mind updating?

Kerollmops:
bors merge

PR #4630, opened 2024-05-13
update segment and reqwest to the latest version

irevoire:
Continue the work of https://github.com/meilisearch/meilisearch/pull/4622


Kerollmops (changes requested):
What saddens me is that we are increasing the number of dependencies by updating reqwest and segment. We should look further into this to shrink the Cargo.lock, instead.

irevoire:
Superseeded by #4786

Issue #4640, opened 2024-05-19
v1.8.0 _geoRadius filter returns no results


bcicen:
Following the exact [example steps in the geosearch guide](https://www.meilisearch.com/docs/learn/fine_tuning_results/geosearch), a search using only `_geoRadius` filter will consistently return no results: 

```bash
curl \
  -X POST 'http://localhost:7700/indexes/restaurants/search' \
  -H 'Content-type:application/json' \
  --data-binary '{ "filter": "_geoRadius(45.472735, 9.184019, 2000)" }'
```

The same request returns results as expected on v1.7.6.

In both versions, `_geoBoundingBox` works as expected.

ManyTheFish:
fixed by https://github.com/meilisearch/meilisearch/pull/4642

Issue #4628, opened 2024-05-10
_geoPoint and _geoRadius not working


skalra90:
Here is my code:

```
const results = await global.meiliClient.index("airports").search("", {
                sort: [`_geoPoint(${topLatitude}, ${topLongitude}):asc`],
                showRankingScore: true,
                showRankingScoreDetails: true
            });
```

results contains all airports not sorted by distance.

```
Index details:
{
    "displayedAttributes": [
        "airport_id",
        "name",
        "alias",
        "city_name",
        "city_alias",
        "country_name",
        "iata",
        "country_code",
        "type",
        "capital",
        "_geo"
    ],
    "searchableAttributes": [
        "name",
        "alias",
        "city_name",
        "city_alias",
        "country_name",
        "iata",
        "country_code",
        "capital"
    ],
    "filterableAttributes": [
        "_geo",
        "country_code",
        "type"
    ],
    "sortableAttributes": [
        "_geo",
        "type"
    ],
    "rankingRules": [
        "sort",
        "words",
        "typo",
        "proximity",
        "attribute",
        "exactness"
    ],
    "stopWords": [],
    "nonSeparatorTokens": [],
    "separatorTokens": [],
    "dictionary": [],
    "synonyms": {},
    "distinctAttribute": "airport_id",
    "proximityPrecision": "byWord",
    "typoTolerance": {
        "enabled": true,
        "minWordSizeForTypos": {
            "oneTypo": 5,
            "twoTypos": 9
        },
        "disableOnWords": [],
        "disableOnAttributes": []
    },
    "faceting": {
        "maxValuesPerFacet": 100,
        "sortFacetValuesBy": {
            "*": "alpha"
        }
    },
    "pagination": {
        "maxTotalHits": 1000
    },
    "searchCutoffMs": null
}
```

Checked all documents and found no invalid entry:
```
function checkGeoData(documents) {
    const invalidEntries = [];

    documents.forEach((doc, index) => {
        // Check if _geo exists and has both lat and lng
        if (!doc._geo || typeof doc._geo.lat !== "number" || typeof doc._geo.lng !== "number") {
            invalidEntries.push({ index, doc });
        }
    });
    return invalidEntries;
}
```

First document always with airport_id 1:
```
{
  name: "Goroka Airport",
  alias: null,
  city_name: "Goroka",
  city_alias: "Goronka",
  country_name: "Papua New Guinea",
  iata: "GKA",
  country_code: "PG",
[...]

PR #4642, opened 2024-05-20
Index the _geo fields when changing the setting while there is already documents in the DB

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4640
Fixes https://github.com/meilisearch/meilisearch/issues/4628

## What does this PR do?
- Add an integration test that first indexes the document and then changes the settings
- Fix `extract_geo_point` by detecting if the `_geo` field has been faceted in this setting change and index all documents

### DIFF
--- a/milli/src/update/index_documents/extract/extract_geo_points.rs
+++ b/milli/src/update/index_documents/extract/extract_geo_points.rs
@@ -8,6 +8,7 @@ use super::helpers::{create_writer, writer_into_reader, GrenadParameters};
 use crate::error::GeoError;
 use crate::update::del_add::{DelAdd, KvReaderDelAdd, KvWriterDelAdd};
 use crate::update::index_documents::extract_finite_float_from_value;
+use crate::update::settings::{InnerIndexSettings, InnerIndexSettingsDiff};
 use crate::{FieldId, InternalError, Result};
 
 /// Extracts the geographical coordinates contained in each document under the `_geo` field.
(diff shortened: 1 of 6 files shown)


milli/src/update/settings.rs
```diff
@@ -1161,6 +1161,11 @@ impl InnerIndexSettingsDiff {
     pub fn settings_update_only(&self) -> bool {
         self.settings_update_only
     }
+
+    pub fn run_geo_indexing(&self) -> bool {
+        self.old.geo_fields_ids != self.new.geo_fields_ids
+            || (!self.settings_update_only && self.new.geo_fields_ids.is_some())
+    }
```
irevoire, line 1168: Shouldn’t we ensure that we went from « not faceted » to « faceted » instead of checking if we're only updating the settings?
It seems more future-proof, but maybe I’m missing something? 🤔 

```suggestion
    pub fn run_geo_indexing(&self) -> bool {
        self.old.geo_fields_ids != self.new.geo_fields_ids
            || (!self.old.geo_fields_ids.is_none() && self.new.geo_fields_ids.is_some())
    }
```

milli/src/update/index_documents/extract/extract_geo_points.rs
```diff
@@ -40,64 +41,65 @@ pub fn extract_geo_points<R: io::Read + io::Seek>(
...
-    Ok([lat, lng])
+            let lng = extract_finite_float_from_value(
+                serde_json::from_slice(lng).map_err(InternalError::SerdeJson)?,
+            )
+            .map_err(|lng| GeoError::BadLongitude { document_id: document_id(), value: lng })?;
+            Ok(Some([lat, lng]))
+        }
+        None => Ok(None),
+    }
 }
```
irevoire, line 105: Nice!

ManyTheFish (approved):
bors merge

Issue #4607, opened 2024-04-30
Only ever store vectors in the vector store + hide embeddings


curquiza:
Related product team resources: [PRD](https://www.notion.so/meilisearch/AI-Search-cd5b8d6ce94944618914ab8d958aceb0) (_internal only_)
Related product discussion: https://github.com/orgs/meilisearch/discussions/677

## Motivation

Improve AI search performance

## Usage

[Only ever store vectors in the vector store and hide embeddings
](https://meilisearch.notion.site/v1-9-AI-search-changes-e90d6803eca8417aa70a1ac5d0225697?pvs=25#38e6d3adf40e4ef1be14a3c4be39df94)

### Reminders when modifying the Setting API

- [ ] Ensure the new setting route is at least tested by the [`test_setting_routes` macro](https://github.com/meilisearch/meilisearch/blob/5204c0b60b384cbc79621b6b2176fca086069e8e/meilisearch/tests/settings/get_settings.rs#L276)
- [ ] Ensure Analytics are fully implemented
  - [ ] `/settings/my-new-setting` configurated in the [`make_setting_routes` macro](https://github.com/meilisearch/meilisearch/blob/5204c0b60b384cbc79621b6b2176fca086069e8e/meilisearch/src/routes/indexes/settings.rs#L141-L165)
  - [ ] global `/settings` route configurated in the [`update_all` function](https://github.com/meilisearch/meilisearch/blob/5204c0b60b384cbc79621b6b2176fca086069e8e/meilisearch/src/routes/indexes/settings.rs#L655-L751)
- [ ] Ensure the dump serializing is consistent with the `/settings` route serializing, e.g., enums case can be different (`camelCase` in route and `PascalCase` in the dump)

#### Special cases when adding a setting for an experimental feature

- [ ] ⚠️ API stability: The setting does not appear on the main settings route when the feature has never been enabled (e.g. mark it `Unset` when returned from the index in this situation. See [an example](https://github.com/meilisearch/meilisearch/blob/7a89abd2a025606a42f8b219e539117eb2eb029f/meilisearch-types/src/settings.rs#L608))
- [ ] The setting cannot be set when the feature is disabled, either by the main settings route or the subroute (see [`validate_settings` function](https://github.com/meilisearch/meilisearch/blob/7a89abd2a025606a42f8b219e539117eb2eb029f/meilisearch/src/routes/indexes/settings.rs#L811))
[...]

PR #4649, opened 2024-05-21
Don't store the vectors in the documents database

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4607

## What does this PR do?
- Ensure that anything falling under `_vectors` is NOT searchable, filterable or sortable
- [x] per embedder, add a roaring bitmap of documents that provide "userProvided" embeddings
- [x] in the indexing process in extract_vector_points, set the bit corresponding to the document depending on the "userProvided" subfield in the _vectors field.
- [x] in the document DB in typed chunks, when writing the _vectors field, remove all keys corresponding to an embedder


dureuill:
/bench workloads/*embeddings.json


shouldn't change drastically

index-scheduler/src/snapshots/index_scheduler__tests__import_vectors-15.snap
```diff
@@ -6,10 +6,6 @@ expression: doc
   "doggo": "Intel",
   "breed": "beagle",
   "_vectors": {
-    "A_fakerest": {
```
dureuill, line 9: I would expect the vector here to still be present?

maybe let's use the new boolean to force displaying the vectors?

index-scheduler/src/snapshots/index_scheduler__tests__import_vectors-22.snap
```diff
@@ -6,10 +6,6 @@ expression: doc
   "doggo": "kefir",
   "breed": "patou",
   "_vectors": {
-    "A_fakerest": {
```
dureuill, line 9: same as above

meilisearch/src/analytics/segment_analytics.rs
```diff
@@ -1811,6 +1825,9 @@ impl SimilarAggregator {
                    "avg_criteria_number": format!("{:.2}", filter_sum_of_criteria_terms as f64 / filter_total_number_of_criteria as f64),
                    "most_used_syntax": used_syntax.iter().max_by_key(|(_, v)| *v).map(|(k, _)| json!(k)).unwrap_or_else(|| json!(null)),
                 },
+                "vector": {
+                    "retrieve_vectors": retrieve_vectors,
```
dureuill, line 1829: I would expect this to be under `hybrid` as well.

`hybrid` in the context of `similar` sounds illogical, but was chosen to avoid multiplying the number of properties

meilisearch/src/search.rs
```diff
@@ -1045,6 +1064,26 @@ fn make_hits(
...
+        if retrieve_vectors {
+            let mut vectors = serde_json::Map::new();
+            for (name, mut vector) in index.embeddings(rtxn, id)? {
+                let user_defined = embedding_configs
+                    .iter()
+                    .find(|conf| conf.name == name)
+                    .is_some_and(|conf| conf.user_defined.contains(id));
+                let mut embedding = serde_json::Map::new();
+                embedding.insert("userDefined".to_string(), user_defined.into());
+                if vector.len() == 1 {
```
dureuill, line 1076: rather than checking len and unwrap, you can probably match on the mut slice and take the vec here: see [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=99af8ab191d2743fea5c9c414662808a)

meilisearch/tests/dumps/mod.rs
```diff
@@ -1938,3 +1938,209 @@ async fn import_dump_v6_containing_experimental_features() {
...
+    snapshot!(code, @"202 Accepted");
+    let response = index.wait_task(response.uid()).await;
+    snapshot!(response);
+
+    let (response, code) = server.create_dump().await;
+    snapshot!(code, @"202 Accepted");
+    let response = index.wait_task(response.uid()).await;
+    snapshot!(response["status"], @r###""succeeded""###);
+
+    // ========= We made a dump, now we should clear the DB except and try to import our dump
```
dureuill, line 1996: not sure I understand "clear the DB except"

index-scheduler/src/snapshots/index_scheduler__tests__import_vectors-15.snap
irevoire: Here, we can't because we're not in meilisearch but in the index-scheduler directly.
We need to look into arroy to find the vector back, but since we were ignoring its value anyway, I don’t think it’s an issue if I don't try to retrieve it, right?

meilisearch/src/analytics/segment_analytics.rs
irevoire: Huum, but we already have a `vector` for the search. Shouldn’t we keep the sequence `vector.retrieve_vectors` everywhere for consistency when retrieving these properties through amplitude?
image
Or should I move both to `hybrid.retrieve_vectors`?
But then, it seems less logical since it has nothing to do with the hybrid search.

Maybe @macraig, if you have 2s to check this out, you have a preference as the consumer of these analytics?

meilisearch/tests/dumps/mod.rs
irevoire: Oops yes my bad, `except`  was part of an old comment, removing it now

milli/src/fields_ids_map.rs
```diff
@@ -41,6 +41,16 @@ impl FieldsIdsMap {
...
 
+    /// Get the ids of a field and all its nested fields based on its name.
+    pub fn nested_ids(&self, name: &str) -> Vec<FieldId> {
+        self.names_ids
+            .range(name.to_string()..)
+            .take_while(|(key, _)| key.starts_with(name))
+            .filter(|(key, _)| crate::is_faceted_by(key, name))
+            .map(|(_name, id)| *id)
+            .collect()
+    }
```
dureuill, line 52: meta note but all this string manipulation makes me wonder if we wouldn't want to store nested fields as arrays rather than as dot separated strings?

meilisearch/src/search.rs
irevoire: It’s way better thank you!

milli/src/search/new/tests/snapshots/milli__search__new__tests__attribute_fid__attribute_fid_ngrams.snap
```diff
@@ -1,244 +0,0 @@
----
```
dureuill, line 1: why are we deleting this file?

milli/src/snapshots/index.rs/initial_field_distribution/1/field_distribution.snap
dureuill, line 1: why are we deleting this file?

milli/src/snapshots/index.rs/initial_field_distribution/field_distribution.snap
dureuill, line 1: why are we deleting this file?

index-scheduler/src/snapshots/index_scheduler__tests__import_vectors-15.snap
dureuill: hmm, we're ignoring its value, but that there is still a user-provided vector is important I guess. Now I need to read the test again to check the context

meilisearch/src/analytics/segment_analytics.rs
dureuill: ah I didn't see it was under vector for the search, I had seen it under hybrid. Please disregard my comment then

milli/src/update/settings.rs
```diff
@@ -933,15 +943,19 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
...
                     match joined {
                         // updated config
-                        EitherOrBoth::Both((name, mut old), (_, new)) => {
+                        EitherOrBoth::Both((name, (mut old, user_defined)), (_, new)) => {
                             changed |= EmbeddingSettings::apply_and_need_reindex(&mut old, new);
                             if changed {
-                                tracing::debug!(embedder = name, "need reindex");
+                                tracing::debug!(
+                                    embedder = name,
+                                    documents = user_defined.len(),
```
dureuill, line 951: nitpick, but shouldn't `documents` equal the number of documents that are **not** user defined?

milli/src/update/settings.rs
```diff
@@ -1223,16 +1249,19 @@ impl InnerIndexSettings {
             embedding_configs,
             existing_fields,
             geo_fields_ids,
+            non_searchable_fields_ids: vectors_fids.clone(),
+            non_faceted_fields_ids: vectors_fids.clone(),
```
dureuill, line 1253: I don't really get why you need *both* `non_faceted_fields_ids` and `faceted_fields_ids`, can you expand on this design? (maybe as comment where the `non_*_fields_ids` are defined)

milli/src/vector/parsed_vectors.rs
```diff
@@ -61,7 +71,13 @@ impl ParsedVectorsDiff {
                 return Err(error);
             }
         }
-        .flatten();
+        .flatten().map_or(BTreeMap::default(), |del| del.into_iter().map(|(name, vec)| (name, Some(vec))).collect());
+        for embedding_config in embedders_configs {
+            if embedding_config.user_defined.contains(docid) {
+                old.entry(embedding_config.name.to_string()).or_insert(None);
```
dureuill, line 77: It is not clear to me what it means to have None here.

- When does it happen?
- what should calling code do when receiving None as value for an embedder?

I'm guessing it is when we had vectors for the embedder but can't get them because no db.

Can we indicate this somewhere as a comment?

milli/src/vector/rest.rs
```diff
@@ -230,6 +230,7 @@ where
             input_value
         }
         [input] => {
+            dbg!(&options);
```
dureuill, line 233: stray debug

meilisearch/src/search.rs
```diff
@@ -1045,6 +1064,24 @@ fn make_hits(
...
 
+        if retrieve_vectors {
+            let mut vectors = serde_json::Map::new();
+            for (name, mut vector) in index.embeddings(rtxn, id)? {
+                let user_defined = embedding_configs
+                    .iter()
+                    .find(|conf| conf.name == name)
+                    .is_some_and(|conf| conf.user_defined.contains(id));
+                let mut embedding = serde_json::Map::new();
+                embedding.insert("userDefined".to_string(), user_defined.into());
```
dureuill, line 1075: wasn't the string "userProvided" though?

meilisearch/src/search.rs
```diff
@@ -1045,6 +1064,24 @@ fn make_hits(
...
+                match vector.as_mut_slice() {
+                    [one] => embedding.insert("embedding".to_string(), std::mem::take(one).into()),
```
dureuill, line 1077: I think it is `embeddings` here (with an s).

Should we maybe use an `ExplicitVector` object that we serialize like in the dumps?

milli/src/update/index_documents/extract/extract_vector_points.rs
```diff
@@ -35,6 +35,8 @@ pub struct ExtractedVectorPoints {
     // embedder
     pub embedder_name: String,
     pub embedder: Arc<Embedder>,
+    pub user_defined: RoaringBitmap,
+    pub remove_from_user_defined: RoaringBitmap,
```
dureuill, line 39: nitpick: for consistency, user_provided and remove_from_user_provided?

milli/src/update/index_documents/extract/extract_vector_points.rs
```diff
@@ -308,44 +344,25 @@ fn push_vectors_diff(
...
     // We sort and dedup the vectors
-    del_vectors.sort_unstable_by(|a, b| compare_vectors(a, b));
     add_vectors.sort_unstable_by(|a, b| compare_vectors(a, b));
-    del_vectors.dedup_by(|a, b| compare_vectors(a, b).is_eq());
     add_vectors.dedup_by(|a, b| compare_vectors(a, b).is_eq());
 
-    let merged_vectors_iter =
-        itertools::merge_join_by(del_vectors, add_vectors, |del, add| compare_vectors(del, add));
+    // let merged_vectors_iter =
+    //     itertools::merge_join_by(del_vectors, add_vectors, |del, add| compare_vectors(del, add));
```
dureuill, line 351:
```suggestion
```

milli/src/update/index_documents/extract/mod.rs
```diff
@@ -226,18 +230,26 @@ fn send_original_documents_data(
...
+            ) {
                 Ok(extracted_vectors) => {
                     for ExtractedVectorPoints {
                         manual_vectors,
                         remove_vectors,
                         prompts,
                         embedder_name,
                         embedder,
+                        user_defined,
+                        remove_from_user_defined: auto_generated,
```
dureuill, line 252:
```suggestion
                        remove_from_user_defined,
```

milli/src/update/index_documents/extract/mod.rs
```diff
@@ -262,6 +274,8 @@ fn send_original_documents_data(
                                 expected_dimension: embedder.dimensions(),
                                 manual_vectors,
                                 embedder_name,
+                                user_defined,
+                                remove_from_user_defined: auto_generated,
```
dureuill, line 278:
```suggestion
                                remove_from_user_defined,
```

milli/src/update/index_documents/typed_chunk.rs
dureuill, line 1: I feel like I'm missing something here. How does manual addition/deletion still works now that we only pass the add part?

How does one remove one manually provided vector from a document now?

index-scheduler/src/snapshots/index_scheduler__tests__import_vectors-15.snap
irevoire: Yes, we do that!
```rust
            // for consistency with the below
            #[allow(clippy::get_first)]
            let IndexEmbeddingConfig { name, config: fakerest_config, user_defined } =
                configs.get(0).unwrap();
            insta::assert_snapshot!(name, @"A_fakerest");
            insta::assert_debug_snapshot!(user_defined, @"RoaringBitmap<[]>");
            insta::assert_json_snapshot!(fakerest_config.embedder_options);
```

meilisearch/tests/dumps/mod.rs
```diff
@@ -1938,3 +1938,209 @@ async fn import_dump_v6_containing_experimental_features() {
...
+    drop(server);
+    tokio::fs::remove_dir_all(&opt.db_path).await.unwrap();
+    let dump_name = format!("{}.dump", response["details"]["dumpUid"].as_str().unwrap());
+    let dump_path = opt.dump_dir.join(dump_name);
+    assert!(dump_path.exists(), "path: `{}`", dump_path.display());
+
+    opt.import_dump = Some(dump_path);
+    // NOTE: We shouldn't have to change the database path but I lost one hour
+    // because of a « bad path » error and that fixed it.
```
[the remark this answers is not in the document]
irevoire: let’s not talk about it 😭

milli/Cargo.toml
```diff
@@ -44,7 +44,7 @@ once_cell = "1.19.0"
 ordered-float = "4.2.0"
 rand_pcg = { version = "0.3.1", features = ["serde1"] }
 rayon = "1.8.0"
-roaring = "0.10.2"
+roaring = { version = "0.10.2", features = ["serde"] }
```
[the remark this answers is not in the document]
irevoire: Hey, it's because I write the `IndexEmbeddingConfig` in the database with serde;
```rust
        self.main.remap_types::<Str, SerdeJson<Vec<IndexEmbeddingConfig>>>().put(
            wtxn,
            main_key::EMBEDDING_CONFIGS,
            &configs,
        )
```

And since the structure contains a roaring bitmap I need it to implement Serialize and Deserialize;
```rust
#[derive(Debug, Deserialize, Serialize)]
pub struct IndexEmbeddingConfig {
    pub name: String,
    pub config: EmbeddingConfig,
    pub user_defined: RoaringBitmap,
}
```

milli/src/fields_ids_map.rs
irevoire: Can you give an example? I’m not sure to understand what you would put in it?

milli/src/snapshots/index.rs/initial_field_distribution/1/field_distribution.snap
irevoire: Same

milli/src/update/settings.rs
irevoire: technically for this PR I don’t need it, but I would like to do a second PR where we also remove the `_geo` fields from the searchable (but not from the faceted).

I didn’t do it here since the PR is already quite big

milli/src/vector/rest.rs
irevoire: Oops thanks 🙏

index-scheduler/src/snapshots/index_scheduler__tests__import_vectors-15.snap
dureuill: OK then

milli/src/update/settings.rs
irevoire: Yes but we don't have the info, I changed `documents` to `user_provided` to make it clear 👍

milli/src/search/new/tests/snapshots/milli__search__new__tests__attribute_fid__attribute_fid_ngrams.snap
dureuill: nice

milli/src/fields_ids_map.rs
dureuill: instead of storing nested fields as `"root.grandparent.parent.field"`, we would store them as `["root", "grandparent", "parent", "field"]` (or probably, more accurately, with an array of the field ids).

Then `is_faceted_by(field)` becomes `path[..path.len()-1].contains(field)` and `nested_ids(name)` becomes `names_ids.filter(|path| path[0] == name)`

meilisearch/src/search.rs
irevoire: Yes you’re absolutely right, that was buggy good catch

milli/src/fields_ids_map.rs
dureuill: not saying we should change this in this PR of course!

milli/src/update/index_documents/typed_chunk.rs
irevoire: True, that’s fixed in the last commit

dureuill (changes requested):
Hello @irevoire 👋

Thank you for your tremendous work!

I did some tests, unfortunately some of them don't demonstrate the expected behavior.

- `_vectors` should not be duplicated when requesting it on documents and search routes
  - ✅ OK: no duplication on either search or documents routes
- A `_vectors.noise` field that is not linked to an embedder configuration should always display regardless of `retrieveVectors`
  - ❌ KO: `retrieveVector` overrides the `_vectors.noise`
- Clearing all documents should reset the roaring of userProvided
  - ❌ KO: clearing all documents allows to get one document with 0 `userProvided` vectors without specifying the `_vectors` field
- Updating a document without touching the `vectors` field should leave its value unchanged 
  - ✅ OK
- changing the settings of an embedder without removing the configuration should retain the userProvided vectors
  - ❌ KO: updated an embedder => lost userProvided vector
test:
- updating a document containing noise + userProvided vec to remove noise should not remove the userprovided part
  - ✅ OK: this worked
- delete one document with userProvided vectors should not have effects on documents created afterwards (it should reinitialize the `user_provided` bit for that document)
  - ❌ KO: after delete one document, can then add a document without a vector, will be handled as if it had a vector
- retrieveVectors=true returns an error if the `vectorStore` feature is not enabled
  - ⚠️ untested due to macOS error
- when the feature was never enabled, users can still see the content of `_vectors`
  - ⚠️ untested due to macOS error
- `_vectors` doesn't have to be an object when the feature was never enabled
  - ⚠️ untested due to macOS error 
- adding a new embedder "new" should import the embeddings declared in `_vectors.new` in existing documents
  - ⚠️ untested due to macOS error

dureuill:
Discussed with @irevoire 

- A `_vectors.noise` field that is not linked to an embedder configuration should always display regardless of `retrieveVectors`
  - ❌ KO: `retrieveVector` overrides the `_vectors.noise`
-> is OK as-is, should be documented in the usage page
- Clearing all documents should reset the roaring of userProvided
  - ❌ KO: clearing all documents allows to get one document with 0 `userProvided` vectors without specifying the `_vectors` field
-> bug in clear documents, @dureuill 
- changing the settings of an embedder without removing the configuration should retain the userProvided vectors
  - ❌ KO: updated an embedder => lost userProvided vector
-> @dureuill to check
- delete one document with userProvided vectors should not have effects on documents created afterwards (it should reinitialize the `user_provided` bit for that document)
  - ❌ KO: after delete one document, can then add a document without a vector, will be handled as if it had a vector
-> approve suggestion above
- retrieveVectors=true returns an error if the `vectorStore` feature is not enabled
  - ❌ KO
-> @dureuill or @irevoire

dureuill:
Update:

Discussed with @irevoire 

- A `_vectors.noise` field that is not linked to an embedder configuration should always display regardless of `retrieveVectors`
  - KO: `retrieveVector` overrides the `_vectors.noise`
-> ✅ fixed by @dureuill 
- Clearing all documents should reset the roaring of userProvided
  - KO: clearing all documents allows to get one document with 0 `userProvided` vectors without specifying the `_vectors` field
-> ✅  fixed by @dureuill 
- changing the settings of an embedder without removing the configuration should retain the userProvided vectors
  - KO: updated an embedder => lost userProvided vector
-> ✅  fixed by @dureuill 
- delete one document with userProvided vectors should not have effects on documents created afterwards (it should reinitialize the `user_provided` bit for that document)
  - KO: after delete one document, can then add a document without a vector, will be handled as if it had a vector
-> ✅ fixed by @dureuill 
- retrieveVectors=true returns an error if the `vectorStore` feature is not enabled
-> ✅ fixed by @irevoire 

@dureuill also changed the behavior and name of the `userProvided` parameter: it is now a `regenerate` parameter that is true when an update to the document should regenerate vectors

irevoire (commented):
I approve my PR @dureuill

PR #4652, opened 2024-05-23
Remove and optimize dependencies to speedup compilation

Kerollmops:
This PR removes the longest-to-compile dependencies:
 - [x] Remove the zstd crate (required by the zip crate), but uncompress the mini-dashboard release.
 - [x] Replace the git2 crate with the gitcl one, but we require the git CLI to compile and retrieve the git info.
   - [ ] Make sure the Docker build image has the git CLI.
 - [ ] Move to the latest version of Lindera, where @dureuill reworked the build scripts to avoid downloading and decompressing dictionaries at every build.
   - [ ] Update the CONTRIBUTING.md to inform the users to set the `LINDERA_CACHE` env var.

[before-cargo-timing.html.zip](https://github.com/meilisearch/meilisearch/files/15418679/cargo-timing-20240523T144211Z.html.zip)
[after-cargo-timing.html.zip](https://github.com/meilisearch/meilisearch/files/15418738/cargo-timing-20240523T145857Z.html.zip)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -632,7 +632,7 @@ version = "1.8.0"
 dependencies = [
  "anyhow",
  "time",
- "vergen-git2",
+ "vergen-gitcl",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


dureuill:
The lindera work landed so the sole remaining item is

> Make sure the Docker build image has the git CLI.

PR #4679, opened 2024-06-05
DO NOT MERGE Revert mimalloc to 0.1.37

dureuill:
revert mimalloc to the version it was in v1.6.0 (before 5869ca7716e30a9f48ecb3182bc41ef3229328eb)


dureuill:
Not fixing the issue 😞

PR #4681, opened 2024-06-05
Fix concurrency issue

### LINKED ISSUE #4654: Meilisearch stops responding when getting "enough" concurrent traffic
Issue:
When running tests with k6 and when going over specific number of VUs (virtual users) Meilisearch just stops responding to HTTP requests.

How to replicate:
1. Start meilisearch with empty database
2. Create some test index there
```
curl --request POST \
  --url http://localhost:7700/indexes \
  --data '{
  "uid": "test",
  "primaryKey": "id"
}'
```
3. Install k6 (https://k6.io/docs/get-started/installation/)
4. Save content below as `meilisearch-test.js`
5. Run 
```
k6 run --vus 10 --duration 10s meilisearch-test.js
```
It probably passes nicely, here it accomplishes something like 200k reqs.

Here with 50 VUs it also completes, about 170k reqs.

But then, with 100 VUs:
```
k6 run --vus 100 --duration 10s meilisearch-test.js
```

It completes 2700 reqs and then meilisearch http 7700 does not answer anymore.

Having checked TRACE and DEBUG logs, I cannot see anything which sheds more light to it. So no any idea what actually crashes there. Actually, nothing crashes, it just stops responding.

Machine should have enough memory and CPU but CPU/RAM usage is not even spiking.

Tried also to play with `--experimental-search-queue-size` -setting, but no mentionable difference with that.

Edit: seems that setting that to `0` keeps Meilisearch alive.

meilisearch-test.js:
```
import http from "k6/http";
import {
  randomString,
  randomIntBetween,
} from "https://jslib.k6.io/k6-utils/1.4.0/index.js";

export default function () {
  const params = {
    headers: {
      "content-type": "application/json",
    },
  };
  const url = "http://localhost:7700/indexes/test/search";
  const post = {
    q: randomString(randomIntBetween(2, 6)),
    limit: 10,
  };
  http.post(url, JSON.stringify(post), params);
}
```

dureuill:
# Pull Request

## Related issue
Fixes #4654 

## What does this PR do?
- Asynchronously drop permits

### DIFF
--- a/meilisearch/src/search_queue.rs
+++ b/meilisearch/src/search_queue.rs
@@ -40,8 +40,9 @@ pub struct Permit {
 
 impl Drop for Permit {
     fn drop(&mut self) {
+        let sender = self.sender.clone();
         // if the channel is closed then the whole instance is down
-        let _ = futures::executor::block_on(self.sender.send(()));
+        std::mem::drop(tokio::spawn(async move { sender.send(()).await }));
     }
 }
 


irevoire (approved):
bors merge

PR #4682, opened 2024-06-05
Speed Up Filter ANDs operations

### LINKED ISSUE #4659: Slow Filtering when Indexes Contains many Documents
I noticed that our biggest client, which has more than 250M documents can see slow requests when there are many filters involved. Basically the less operations the filter performs the faster it executes.

Here are timings for the different filters executed:
 - A placeholder search takes around 130ms.
 - A basic "Minecraft" query between 563ms and 380ms (when hot).
 - A placeholder search with the following _big_ filter takes around 2843ms.
```sql
view_count > 2500 AND subscriber_count >= 0 AND subscriber_count < 9223372036854775807 AND xf90 >= 2 AND xf90 < 1000000000 AND upload_date>= 1709269200000 AND language EXISTS AND language IN [en] AND made_for_kids=FALSE AND shorts=false AND upload_date > 1706763600000 AND upload_date < 1710561600000
```
 - A placeholder search with the following, smaller filter takes around 1847ms.
```sql
xf90 >= 2 AND xf90 < 1000000000 AND upload_date>= 1709269200000 AND language EXISTS AND language IN [en] AND made_for_kids=FALSE AND shorts=false AND upload_date > 1706763600000 AND upload_date < 1710561600000
```
 - A placeholder search with the following, even smaller filter takes around 928ms.
```sql
xf90 >= 2 AND xf90 < 1000000000 AND upload_date>= 1709269200000 AND language EXISTS AND language IN [en] AND made_for_kids=FALSE AND shorts=false
```

By rewriting the original filter to one that does not do useless checks it know takes 1492ms:
```sql
view_count > 2500 AND xf90 >= 2 AND language IN [en] AND made_for_kids=FALSE AND shorts=false AND upload_date > 1706763600000 AND upload_date < 1710561600000
```

### Solution

It seems pretty trivial to provide the universe that sub-filters should always consider when returning documents. t's especially useful when the filter has a lot of `AND` closes which drastically reduces the candidates. We could also benefit from a method to do an intersection with a bitmap directly on disk ([akin to this one](https://github.com/RoaringBitmap/roaring-rs/pull/263) and only consider the useful internal containers).

https://github.com/meilisearch/meilisearch/blob/e248d2a1e620f2dee6cc141de58b3f78cfaae97c/milli/src/search/facet/facet_range_search.rs#L11-L62

### Meilisearch Version
v1.2.0+

### Additional context
Running on a dedicated machine on Digital Ocean.

Kerollmops:
This PR fixes #4659 and improves the way we do AND operations by using the latest [RoaringBitmap feature to do intersections with serialized bitmaps](https://github.com/RoaringBitmap/roaring-rs/pull/281). Doing so drastically reduces the time spent reading, copying bytes in memory to use and keep a subset of the containers in the bitmap.

### Some Example Results

With a 45M documents dataset running on a good NVMe. This example filter was taking 77ms and with this PR only 13ms (6x speedup):

```sql
artist = 'The Beatles' AND (duration 150 TO 500 OR duration NOT EXISTS) AND genres IN [Rock, 'Rock and Roll'] AND rating > 4 AND released_year 1960 TO 1990
```

By reordering the filter AND clauses we can reach a constant 8ms execution time. However, note that it is a manual operation. On the other side the previous filter pipeline is still at a constant 45ms execution time with this filter. (6x speedup)

```sql
artist = 'The Beatles' AND genres IN [Rock, 'Rock and Roll'] AND released_year 1960 TO 1990 AND (duration 150 TO 500 OR duration NOT EXISTS)
```

### To Do
- [x] Rebase on `release-v1.9.0`.
- [ ] ~Skip branches of the facet/filter tree when nothing is in common with the universe~ slower this way.
- [x] When the universe is required use the universe given in parameter if possible.

### DIFF
--- a/milli/src/search/facet/filter.rs
+++ b/milli/src/search/facet/filter.rs
@@ -224,14 +224,14 @@ impl<'a> Filter<'a> {
     pub fn evaluate(&self, rtxn: &heed::RoTxn, index: &Index) -> Result<RoaringBitmap> {
         // to avoid doing this for each recursive call we're going to do it ONCE ahead of time
         let filterable_fields = index.filterable_fields(rtxn)?;
-
-        self.inner_evaluate(rtxn, index, &filterable_fields)
+        self.inner_evaluate(rtxn, index, &filterable_fields, None)
     }
 
     fn evaluate_operator(
         rtxn: &heed::RoTxn,
         index: &Index,
         field_id: FieldId,
+        universe: Option<&RoaringBitmap>,
         operator: &Condition<'a>,
     ) -> Result<RoaringBitmap> {
         let numbers_db = index.facet_id_f64_docids;
@@ -291,14 +291,22 @@ impl<'a> Filter<'a> {
             }
             Condition::NotEqual(val) => {
                 let operator = Condition::Equal(val.clone());
-                let docids = Self::evaluate_operator(rtxn, index, field_id, &operator)?;
+                let docids = Self::evaluate_operator(rtxn, index, field_id, None, &operator)?;
                 let all_ids = index.documents_ids(rtxn)?;
                 return Ok(all_ids - docids);
             }
         };
 
         let mut output = RoaringBitmap::new();
-        Self::explore_facet_number_levels(rtxn, numbers_db, field_id, left, right, &mut output)?;
+        Self::explore_facet_number_levels(
+            rtxn,
+            numbers_db,
+            field_id,
+            left,
+            right,
+            universe,
+            &mut output,
+        )?;
         Ok(output)
     }
 
@@ -310,6 +318,7 @@ impl<'a> Filter<'a> {
         field_id: FieldId,
         left: Bound<f64>,
         right: Bound<f64>,
+        universe: Option<&RoaringBitmap>,
         output: &mut RoaringBitmap,
     ) -> Result<()> {
         match (left, right) {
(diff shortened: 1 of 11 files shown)


milli/src/search/facet/filter.rs
```diff
@@ -332,15 +341,18 @@ impl<'a> Filter<'a> {
...
         match &self.condition {
             FilterCondition::Not(f) => {
+                // TODO improve the documents_ids to also support intersections at deserialize time.
                 let all_ids = index.documents_ids(rtxn)?;
                 let selected = Self::inner_evaluate(
                     &(f.as_ref().clone()).into(),
                     rtxn,
                     index,
                     filterable_fields,
+                    universe,
```
Kerollmops, line 355: I think `all_ids` here is simply the universe or this call to `documents_ids` if no universe has been given.

milli/src/search/facet/filter.rs
```diff
@@ -398,16 +408,21 @@ impl<'a> Filter<'a> {
...
+                        universe,
                     )?;
                     for f in subfilters_iter {
                         if bitmap.is_empty() {
                             return Ok(bitmap);
                         }
+                        // TODO We are doing the intersections two times,
+                        //      it could be more efficient
+                        //      Can't I just replace this `&=` by an `=`?
                         bitmap &= Self::inner_evaluate(
```
Kerollmops, line 420: Pretty sure this is simply an `=` here. All the intersections were already made. I prefer to keep this extra `&=` in case we forgot an intersection when deserializing somewhere.

Kerollmops:
bors merge

PR #4693, opened 2024-06-11
Introduce distinct attributes at search time

### LINKED ISSUE #4611: `distinctAttribute` at search
Related product team resources: [PRD](https://www.notion.so/meilisearch/Set-Distinct-Attribute-at-Search-ebaaa195ac2149a1a592c9243dde6f5b?pvs=25) (_internal only_)

## Motivation

Required by users

## Usage

Being able to set up distinct attribute at search and not only on indexing side

https://meilisearch.notion.site/Set-distinctAttribute-at-search-usage-947187b312c04616a8ca8ef5bcc5c63c?pvs=74

## Impacted teams

Kerollmops:
This PR fixes #4611.

### To Do
- [x] Remove the `distinguishableAttributes` settings (not even a commit about that).
- [x] Use the `filterableAttributes` to be able to use the `distinct` parameter at search.
- [x] Work on the errors and make tests.

### DIFF
--- a/meilisearch/src/analytics/segment_analytics.rs
+++ b/meilisearch/src/analytics/segment_analytics.rs
@@ -597,6 +597,9 @@ pub struct SearchAggregator {
     // every time a request has a filter, this field must be incremented by one
     sort_total_number_of_criteria: usize,
 
+    // distinct
+    distinct: bool,
+
     // filter
     filter_with_geo_radius: bool,
     filter_with_geo_bounding_box: bool,
@@ -670,6 +673,7 @@ impl SearchAggregator {
             show_ranking_score_details,
             filter,
             sort,
+            distinct,
             facets: _,
             highlight_pre_tag,
             highlight_post_tag,
@@ -691,6 +695,8 @@ impl SearchAggregator {
             ret.sort_with_geo_point = sort.iter().any(|s| s.contains("_geoPoint("));
             ret.sort_sum_of_criteria_terms = sort.len();
         }
+
+        ret.distinct = distinct.is_some();
 
         if let Some(ref filter) = filter {
             static RE: Lazy<Regex> = Lazy::new(|| Regex::new("AND | OR").unwrap());
@@ -795,6 +801,7 @@ impl SearchAggregator {
             sort_with_geo_point,
             sort_sum_of_criteria_terms,
             sort_total_number_of_criteria,
+            distinct,
             filter_with_geo_radius,
             filter_with_geo_bounding_box,
             filter_sum_of_criteria_terms,
@@ -850,6 +857,9 @@ impl SearchAggregator {
             self.sort_sum_of_criteria_terms.saturating_add(sort_sum_of_criteria_terms);
         self.sort_total_number_of_criteria =
             self.sort_total_number_of_criteria.saturating_add(sort_total_number_of_criteria);
+
+        // distinct
+        self.distinct |= distinct;
 
         // filter
         self.filter_with_geo_radius |= filter_with_geo_radius;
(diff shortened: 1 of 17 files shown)


meilisearch/src/analytics/segment_analytics.rs
```diff
@@ -977,6 +988,8 @@ impl SearchAggregator {
                     "with_geoPoint": sort_with_geo_point,
                     "avg_criteria_number": format!("{:.2}", sort_sum_of_criteria_terms as f64 / sort_total_number_of_criteria as f64),
                 },
+                // TODO ask help from María
+                "distinct": distinct,
```
Kerollmops, line 992: Can I have some help on this part @macraig? What I want to know is the format I should employ for the analytics of the distinct field on the search endpoint.

irevoire, line 992:
```suggestion
                "distinct": distinct,
```

I guess we can remove this comment now?

milli/src/search/mod.rs
```diff
@@ -169,6 +176,19 @@ impl<'a> Search<'a> {
...
+        if let Some(distinct) = &self.distinct {
+            let filterable_fields = ctx.index.filterable_fields(ctx.txn)?;
+            if !filterable_fields.contains(distinct) {
+                let (valid_fields, hidden_fields) =
+                    ctx.index.remove_hidden_fields(ctx.txn, filterable_fields)?;
+                return Err(Error::UserError(UserError::InvalidDistinctAttribute {
+                    field: distinct.clone(),
+                    valid_fields,
+                    hidden_fields,
+                }));
```
irevoire, line 188: Hey, I think we're missing something here.
I pushed a test highlighting that it doesn't work with nested fields which I don't think is expected? 🤔 

Shouldn't we check if the filterable fields are being faceted by the filterable fields instead of a simple `contains`?

milli/src/error.rs
```diff
@@ -134,6 +134,17 @@ only composed of alphanumeric characters (a-z A-Z 0-9), hyphens (-) and undersco
...
     )]
     InvalidSortableAttribute { field: String, valid_fields: BTreeSet<String>, hidden_fields: bool },
+    #[error("Attribute `{}` is not filterable and thus, cannot be used as distinct attribute. {}",
+        .field,
+        match .valid_fields.is_empty() {
+            true => "This index does not have configured filterable attributes.".to_string(),
+            false => format!("Available filterable attributes are: `{}{}`.",
+                    valid_fields.iter().map(AsRef::as_ref).collect::<Vec<&str>>().join(", "),
+                    .hidden_fields.then_some(", <..hidden-attributes>").unwrap_or(""),
+                ),
```
irevoire, line 144: I don't understand what are the hidden fields?
Are they some kind of nested fields? If yes I don't think we should add the `<..hidden-attributes>` 🤔

irevoire (changes requested):

milli/src/search/new/bucket_sort.rs
```diff
@@ -34,7 +35,12 @@ pub fn bucket_sort<'ctx, Q: RankingRuleQueryTrait>(
     logger.ranking_rules(&ranking_rules);
     logger.initial_universe(universe);
 
-    let distinct_fid = if let Some(field) = ctx.index.distinct_field(ctx.txn)? {
+    let distinct_field = match distinct {
+        Some(distinct) => Some(distinct),
+        None => ctx.index.distinct_field(ctx.txn)?,
+    };
```
ManyTheFish, line 41: So, the search parameter overrides the settings, is it expected?

milli/src/error.rs
irevoire: Got an explanation from louis & kero

irevoire (dismissed):
bors merge

irevoire (approved):
bors merge

mmachatschek:
@Kerollmops just for my understanding. is there any performance penalty using this new feature over using the `distinctAttribute` on the index settting?

irevoire:
Hello @mmachatschek, no there is no performance penalty 👌

PR #4703, opened 2024-06-17
Update yaup

irevoire:
There was a bug in `yaup` where serializing a structure with an array would give you a wrong query parameter.

Now, yaup is also in charge of sending the initial `?` before the query parameters.


ManyTheFish (approved):
LGTM
bors merge

PR #4706, opened 2024-06-19
specify the rust toolchain

irevoire:
The action we were using was not working with the `rust-toolchain.toml` file.
But the repository is not maintained anymore.
While looking for a solution, I found out that [helix](https://github.com/helix-editor/rust-toolchain) solved the issue on their side by forking the repo and adding a few fixes. That's what I use currently, but I don't know if it's a sustainable solution in the long term

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -31,7 +31,7 @@ jobs:
           apt-get update && apt-get install -y curl
           apt-get install build-essential -y
       - name: Setup test with Rust stable
-        uses: actions-rs/toolchain@v1
+        uses: helix-editor/rust-toolchain@v1
       - name: Cache dependencies
         uses: Swatinem/rust-cache@v2.7.1
       - name: Run cargo check without any default features
@@ -56,7 +56,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Cache dependencies
         uses: Swatinem/rust-cache@v2.7.1
-      - uses: actions-rs/toolchain@v1
+      - uses: helix-editor/rust-toolchain@v1
       - name: Run cargo check without any default features
         uses: actions-rs/cargo@v1
         with:
@@ -81,7 +81,7 @@ jobs:
         run: |
           apt-get update
           apt-get install --assume-yes build-essential curl
-      - uses: actions-rs/toolchain@v1
+      - uses: helix-editor/rust-toolchain@v1
       - name: Run cargo build with almost all features
         run: |
           cargo build --workspace --locked --release --features "$(cargo xtask list-features --exclude-feature cuda)"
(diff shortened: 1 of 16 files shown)


.github/workflows/test-suite.yml
```diff
@@ -154,7 +139,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
+      - uses: helix-editor/rust-toolchain@v1
         with:
           profile: minimal
           toolchain: 1.75.0
```
dureuill, line 145: I think we should probably remove the toolchain override here and just use the same as described in the toolchain.

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.78.0"
```
dureuill, line 2: As far as I know, we currently support Rust v1.75, so I think the version here should be v1.75.0

dureuill (changes requested):
This is looking super great ❤️ 

1. please harmonize the clippy version with the rest, no sense in keeping it separate IMO
2. current supported version is v1.75, not v1.78. We'll upgrade it when the next "upgrade Rust" autogenerated issue will pop

.github/workflows/test-suite.yml
```diff
@@ -173,10 +156,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
+      - uses: helix-editor/rust-toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly
+          toolchain: nightly-2024-06-25
```
curquiza, line 162: why do we fix nightly? the purpose of this test is to check our compatibility with coming version of meilisearch right. @Kerollmops who required for this test

dureuill: I think the purpose of this check is to check if the code is correctly formatted.

Because the stable version of `rustfmt` is very limited in options, we are using a nightly version.
I think tamo is right to be fixing it, to avoid that a regression on some nightly wrecks havoc in our CI (especially for formatting!)

dureuill: Maybe we should add to our "update Rust" template a reminder to also refresh the pinned Rust version, though.

Otherwise the nightly version will eventually become older than the stable version we're using, and that is probably not a great idea.

dureuill (approved):
Looks great to me, thanks for this addition! :+1:

.github/workflows/test-suite.yml
irevoire: > I think the purpose of this check is to check if the code is correctly formatted.

Yes, Louis is 100% right; we're using nightly options in rustfmt and thus need a nightly version of rustfmt, nothing else; we were not trying to catch any future bugs

> Maybe we should add to our "update Rust" template a reminder to also refresh the pinned Rust version, though.

Already done 😎

curquiza: Following you on this

irevoire:
Thanks both of you

bors merge

Issue #4588, opened 2024-04-18
[ARROY] Corruption: panic using hybrid search


sam-ulrich1:
I've confirmed that there is a valid embedding being sent and a valid query

```
2024-04-18T22:38:19.654577Z ERROR meilisearch: info=panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/arroy-0.2.0/src/reader.rs:250:83: called `Option::unwrap()` on a `None` value
2024-04-18T22:38:19.654906Z ERROR HTTP request{method=POST host="meili-meilisearch.gigo-core.svc.cluster.local:7700" route=/indexes/bytes/search query_parameters= user_agent=Meilisearch Go (v0.26.2) status_code=500 error=task 595 panicked}: tracing_actix_web::middleware: Error encountered while processing the incoming HTTP request: ResponseError { code: 500, message: "task 595 panicked", error_code: "internal", error_type: "internal", error_link: "https://docs.meilisearch.com/errors#internal" }
2024-04-18T22:38:19.655012294Z 2024-04-18T22:38:19.654962Z  INFO HTTP request{method=POST host="meili-meilisearch.gigo-core.svc.cluster.local:7700" route=/indexes/bytes/search query_parameters= user_agent=Meilisearch Go (v0.26.2) status_code=500 error=task 595 panicked}: meilisearch: close time.busy=669µs time.idle=7.79ms
2024-04-18T22:38:19.658533Z ERROR meilisearch: info=panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/arroy-0.2.0/src/reader.rs:250:83: called `Option::unwrap()` on a `None` value
2024-04-18T22:38:19.658727629Z 2024-04-18T22:38:19.658687Z ERROR HTTP request{method=POST host="meili-meilisearch.gigo-core.svc.cluster.local:7700" route=/indexes/bytes/search query_parameters= user_agent=Meilisearch Go (v0.26.2) status_code=500 error=task 597 panicked}: tracing_actix_web::middleware: Error encountered while processing the incoming HTTP request: ResponseError { code: 500, message: "task 597 panicked", error_code: "internal", error_type: "internal", error_link: "https://docs.meilisearch.com/errors#internal" }
```

sam-ulrich1:
I dropped the whole index and re-created it and now it doesn't happens so it looks like some form of data corruption issue

sam-ulrich1:
[...]

PR #4715, opened 2024-06-20
Build all arroy indexes that need to be built

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4588

## What does this PR do?
- Update arroy
- Ensure we always rebuild the arroy indexes that need to be built

### DIFF
--- a/milli/src/update/index_documents/mod.rs
+++ b/milli/src/update/index_documents/mod.rs
@@ -547,10 +547,11 @@ where
             pool.install(|| {
                 for k in crate::vector::arroy_db_range_for_embedder(embedder_index) {
                     let writer = arroy::Writer::new(vector_arroy, k, dimension);
-                    if writer.is_empty(wtxn)? {
+                    if writer.need_build(wtxn)? {
+                        writer.build(wtxn, &mut rng, None)?;
+                    } else if writer.is_empty(wtxn)? {
                         break;
                     }
-                    writer.build(wtxn, &mut rng, None)?;
                 }
                 Result::Ok(())
             })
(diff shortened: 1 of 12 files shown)


milli/src/update/index_documents/mod.rs
```diff
@@ -547,7 +547,7 @@ where
             pool.install(|| {
                 for k in crate::vector::arroy_db_range_for_embedder(embedder_index) {
                     let writer = arroy::Writer::new(vector_arroy, k, dimension);
-                    if writer.is_empty(wtxn)? {
+                    if !writer.need_build(wtxn)? && writer.is_empty(wtxn)? {
```
dureuill, line 550: hello, do we still need the `writer.is_empty` part? I would expect `need_build` to suffice?

irevoire: Huum I thought it was possible that you end up with a writer that doesn't need to be built even though the next one needs to be built?
To me, we are looking specifically for the last writer that doesn't need to be built AND that is not used by any document.
But I could be wrong. Let me know!

> it sounds strange to me to have no writer doesn't need build and then do it anyway...

```
                    if writer.need_build(wtxn)? {
                        writer.build(...)?;
                    } else if writer.is_empty(wtxn)? {
                        break;
                    }
```

dureuill: alright good catch. The revised version you suggest expresses the intention better

dureuill:
Hello again,

tests don't pass

irevoire:
We should be good now

dureuill (approved):
bors merge

PR #4722, opened 2024-06-25
Grow by 1TB instead of 1MB

dureuill:
When an index reaches 1TB, increases its size by 1TB rather than 1MB


Kerollmops (approved):
Doesn't pass the CI but I like your changes 😢

dureuill:
> Doesn't pass the CI but I like your changes 😢

@Kerollmops  Nothing will pass on main until we either fix the Rust version of fetch back the commit from v1.8/v1.9 that fixes the new warnings.

It has nothing to do with this PR in particular

dureuill:
bors merge

PR #4725, opened 2024-06-25
Store primary key as String when Number exceeds i64 range

### LINKED ISSUE #4696: Valid primary key is refused
**Describe the bug**
When sending a document with a valid primary key I get a message saying my primary key is invalid.

**To Reproduce**
1. `echo '{ "id": 14630868576586246730, "name": "tamo" }' | mieli d a`
2. Document identifier `14630868576586246730` is invalid. A document identifier can be of type integer or string, only composed of alphanumeric characters (a-z A-Z 0-9), hyphens (-) and underscores (_).

**Expected behavior**
It works or returns a better error message

JWSong:
# Pull Request

## Related issue
Fixes #4696 

## What does this PR do?
- When a Number value exceeding the range of i64 is received as a primary key, it will be stored as a String.

## PR checklist

### DIFF
--- a/meilisearch/tests/documents/add_documents.rs
+++ b/meilisearch/tests/documents/add_documents.rs
@@ -1041,6 +1041,77 @@ async fn document_addition_with_primary_key() {
 }
 
 #[actix_rt::test]
+async fn document_addition_with_huge_int_primary_key() {
+    let server = Server::new().await;
+    let index = server.index("test");
+
+    let documents = json!([
+        {
+            "primary": 14630868576586246730u64,
+            "content": "foo",
+        }
+    ]);
+    let (response, code) = index.add_documents(documents, Some("primary")).await;
+    snapshot!(code, @"202 Accepted");
+    snapshot!(json_string!(response, { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
+        @r###"
+    {
+      "taskUid": 0,
+      "indexUid": "test",
+      "status": "enqueued",
+      "type": "documentAdditionOrUpdate",
+      "enqueuedAt": "[date]"
+    }
+    "###);
+
+    index.wait_task(0).await;
+
+    let (response, code) = index.get_task(0).await;
+    snapshot!(code, @"200 OK");
+    snapshot!(json_string!(response, { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
+        @r###"
+    {
+      "uid": 0,
+      "indexUid": "test",
+      "status": "succeeded",
+      "type": "documentAdditionOrUpdate",
+      "canceledBy": null,
+      "details": {
+        "receivedDocuments": 1,
+        "indexedDocuments": 1
+      },
+      "error": null,
+      "duration": "[duration]",
+      "enqueuedAt": "[date]",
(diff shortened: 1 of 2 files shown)


meilisearch/tests/documents/add_documents.rs
```diff
@@ -1040,6 +1040,77 @@ async fn document_addition_with_primary_key() {
...
+      "enqueuedAt": "[date]"
+    }
+    "###);
+
+    index.wait_task(0).await;
+
+    let (response, code) = index.get_task(0).await;
+    snapshot!(code, @"200 OK");
+    snapshot!(json_string!(response, { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
+        @r###"
```
irevoire, line 1072: I introduced some methods to avoid typos and generally reduce the number of snapshots we do.

```suggestion
    let (response, code) = index.add_documents(documents, Some("primary")).await;
    snapshot!(code, @"202 Accepted");

    let response = index.wait_task(response.uid()).await;
    snapshot!(response,
        @r###"
```

meilisearch/tests/documents/add_documents.rs
```diff
@@ -1040,6 +1040,77 @@ async fn document_addition_with_primary_key() {
...
+    snapshot!(code, @"200 OK");
+    snapshot!(json_string!(response, { ".createdAt" => "[date]", ".updatedAt" => "[date]" }),
+        @r###"
+    {
+      "uid": "test",
+      "createdAt": "[date]",
+      "updatedAt": "[date]",
+      "primaryKey": "primary"
+    }
+    "###);
```
irevoire, line 1101: That doesn't give us much information, we can get rid of it I think

```suggestion
```

milli/src/documents/primary_key.rs
```diff
@@ -166,7 +166,7 @@ pub fn validate_document_id_value(document_id: Value) -> StdResult<String, UserE
             Some(s) => Ok(s.to_string()),
             None => Err(UserError::InvalidDocumentId { document_id: Value::String(string) }),
         },
-        Value::Number(number) if number.is_i64() => Ok(number.to_string()),
+        Value::Number(number) if !number.is_f64() => Ok(number.to_string()),
```
irevoire, line 169: Why did you change the condition from `is_i64` to `is_f64`?

irevoire (changes requested):
Hey @JWSong I think we'll be able to merge quite fast, thanks for the investigation!
I left a few comments

milli/src/documents/primary_key.rs
JWSong: The `Number` type in JSON can be a float, so without the above guard clause, the document id would allow the character `.`.

irevoire: Oh yes I initially missed the new `!`, that’s a neat addition thanks

meilisearch/tests/documents/add_documents.rs
JWSong: applied the suggestion on 6fb36ed30e9bd3768b9a9686606198029d6172b9

irevoire (approved):
Awesome, thanks a lot @JWSong 

bors merge

JWSong:
Tests on macOS 12 failed. Seems like this isn't related to my change. Could you check it please? @irevoire

irevoire:
Yep, it looks like a flaky test; I’ll try again with bors ;

bors merge

PR #4730, opened 2024-06-26
fix a 100% flaky test

irevoire:
On slow CI, it was possible for a document addition to _not_ to be processed and then get autobatched with an index deletion, which changed their task summary details in the end.
Now, I wait for the task to finish, and the result will always be the same


dureuill (dismissed):
sure why not

dureuill (approved):
still OK

irevoire:
Thanks,

bors merge

PR #4746, opened 2024-06-27
Fix hybrid search limit offset

### LINKED ISSUE #4745: Hybrid search don't skip offset results when the keyword results are returned early
**Describe the bug**
@inventor123 is reporting that the hybrid search is returning more documents than `limit` when `offset` is not 0.

Looking at the implementation, this can happen when the keyword results are returned early (because the results have a good enough ranking score).

**To Reproduce**
1. Make a hybrid search request with a low semanticRatio, and a non null offset
2. Observe that the returned number of hits is not equal to `limit` and does not skip any result.

**Expected behavior**
The returned hits respect limit and offset

**Meilisearch version:**
v1.6 to v1.8

**Additional context**
https://github.com/meilisearch/meilisearch/discussions/4742

dureuill:
# Pull Request

## Related issue
Fixes #4745

## What does this PR do?
- Apply offset and limit to the keyword search results when they are returned early.
- Add a test that is initially failing, and then passes

### DIFF
--- a/meilisearch/tests/search/hybrid.rs
+++ b/meilisearch/tests/search/hybrid.rs
@@ -161,7 +161,7 @@ async fn limit_offset() {
         )
         .await;
     snapshot!(code, @"200 OK");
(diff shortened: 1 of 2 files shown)


irevoire (approved):
Perfect, than you!

bors merge

inventor123:
i noticed that when response.semanticHitCount = 0 then it Returning All Hits Up from 0 to Offset as i hav ereported.
if semanticHitCount > 0 then it return the requested Offset,

dureuill:
@inventor123 Thank you for the precision. It is consistent with the issue happening only when we are returning the keyword results early, which is what this PR fixes.

PR #4755, opened 2024-07-01
Add tests on the rest embedder

irevoire:
To be merged after we bring back changes on main.

This PR adds a bunch of tests that highlight a lot of bad behavior from the embedding configurations that we should fix in future PRs


dureuill:
Superseded by https://github.com/meilisearch/meilisearch/pull/4815 which contains https://github.com/meilisearch/meilisearch/pull/4815/commits/988552e1783786b0b065d25a0060529fdf81db07

PR #4760, opened 2024-07-02
Document dictionary compression

### LINKED ISSUE #4750: Document DB compression
Related product team resources: [PRD]() (_internal only_)
Related product discussion:

## Motivation

Reduce the size of the document DB, so the size of the Meilisearch DB.

## Usage

By default

## Impacted teams

None directly.

Kerollmops:
This PR fixes #4750 by introducing document compression to Meilisearch.

I had to use the _zstd_ library directly instead of the _lz4_flex_ because the latter doesn't provide an option to specify the compression level. [According to the documentation](https://raw.githack.com/facebook/zstd/release/doc/zstd_manual.html#Chapter1), I used:
 - A compression level of 19,
 - A dictionary size of 64000 bytes,
 - The first 10k documents in the LMDB database as the sample to generate the dictionary,
 - The dictionary is generated if there is no dictionary yet and we have reached 10k documents,
 - The dictionary is deleted when all the documents are deleted (index clear or by hand).

Note that the benchmarks only represent the first couple hours of real usage: When the user uploads some documents and settings, the number of documents reaches 10k+. This PR will compress the documents when there are 10k+ with a dictionary generated from them and then **never** changes the dictionary (but if no documents are left).

The first results show between 2x and 3x compression of the documents database (from a 25GiB to a 16GiB data.ms and an average document size going from 305B to 126B). Still, we can see performance regression due to the compression done on a single thread (👇).

## To do
 - [x] Resolve merge conflicts by rebasing on main
 - [x] Make sure the tests are passing
 - [x] Measure performance impact with #4762
 - [x] Simplify the usage of document decoding
 - [x] Specify a const for document sample size, compression level, and dictionary size
 - [ ] Compress the documents in parallel instead of in the write loop
 - [x] See if we can use the `experimental` _zstd_ feature to [avoid copying 64k bytes in memory](https://docs.rs/zstd/latest/zstd/dict/struct.EncoderDictionary.html#method.new).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3269,6 +3269,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a7a70ba024b9dc04c27ea2f0c0548feb474ec5c54bba33a7f72f873a39d07b24"
 
 [[package]]
+name = "lz4_flex"
+version = "0.11.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "75761162ae2b0e580d7e7c390558127e5f01b4194debd6221fd8c207fc80e3f5"
+dependencies = [
+ "twox-hash",
+]
+
+[[package]]
 name = "lzma-rs"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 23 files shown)


Kerollmops:
Closed in favor of #5157.

PR #4769, opened 2024-07-03
Federated search

### LINKED ISSUE #4747: Federated search
Related product team resources: [PRD](https://www.notion.so/meilisearch/Federated-Search-PRD-07ff4017f65f4a42b577a7925e94e92d) (_internal only_)

## Motivation

Really required by users. API facilitation to merge results from different indexes into the same array of results.

## Usage

https://meilisearch.notion.site/v1-10-federated-search-698dfe36ab6b4668b044f735fb40f0b2

## Impacted teams

dureuill:
# Pull Request

## Related issue
Fixes #4747 

[Usage](https://meilisearch.notion.site/v1-10-federated-search-698dfe36ab6b4668b044f735fb40f0b2)

## What does this PR do?
- multi-search now allows a top-level federation object. When not `null`, the results of multi-search are modified to be a single list of results rather than a list of a list of results
- changed lifetimes around tokenizer et al. to be able to make hits one by one rather than using a vector
- adds `roaring` to Meilisearch itself. As the federated search happens at the Meilisearch level (reuses the search functions declared at the Meilisearch level + merge happens after the hits were created), `RoaringBitmap`s are needed to track the candidates: hits that were seen,  all candidates.
- Refactor `make_hits` to allow for an individual, optimized `make_hit` 
- Score details comparison no longer fail when sorting on different field names or target point (for geo)


meilisearch/src/search/federated.rs
```diff
@@ -0,0 +1,633 @@
...
+    pub weight: Weight,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, deserr::Deserr)]
+#[deserr(try_from(f64) = TryFrom::try_from -> InvalidMultiSearchWeight)]
+pub struct Weight(f64);
+
+impl Default for Weight {
+    fn default() -> Self {
+        Weight(DEFAULT_FEDERATED_WEIGHT())
```
dureuill, line 45: Remove `DEFAULT_FEDERATED_WEIGHT`

meilisearch/src/search/federated.rs
```diff
@@ -0,0 +1,633 @@
...
+    }
+}
+
+impl std::convert::TryFrom<f64> for Weight {
+    type Error = InvalidMultiSearchWeight;
+
+    fn try_from(f: f64) -> Result<Self, Self::Error> {
+        // the suggested "fix" is: `!(0.0..=1.0).contains(&f)`` which is allegedly less readable
+        #[allow(clippy::manual_range_contains)]
```
dureuill, line 54:
```suggestion
```

meilisearch/src/search/federated/ranking_rules.rs
```diff
@@ -0,0 +1,823 @@
...
+            | (RankingRuleKind::AscendingSort, RankingRuleKind::Relevancy)
+            | (RankingRuleKind::DescendingSort, RankingRuleKind::Relevancy) => {
+                "cannot compare a relevancy rule with a sort rule"
+            }
+
+            (RankingRuleKind::Relevancy, RankingRuleKind::AscendingGeoSort)
+            | (RankingRuleKind::Relevancy, RankingRuleKind::DescendingGeoSort)
+            | (RankingRuleKind::AscendingGeoSort, RankingRuleKind::Relevancy)
+            | (RankingRuleKind::DescendingGeoSort, RankingRuleKind::Relevancy) => {
+                "cannot compare a relevancy rule with a sort rule"
```
dureuill:
```suggestion
                "cannot compare a relevancy rule with a geosort rule"
```

meilisearch/src/search/federated.rs
dureuill: removed

meilisearch/src/search/federated.rs
dureuill: done

meilisearch/tests/search/multi.rs
```diff
@@ -376,3 +1097,3227 @@ async fn search_multiple_query_errors() {
...
... `typo` at position #2 in ranking rules because the query is a placeholder search (`q`: \"\")\n    3. Removed relevancy rule `proximity` at position #3 in ranking rules because the query is a placeholder search (`q`: \"\")\n    4. Removed relevancy rule `attribute` at position #4 in ranking rules because the query is a placeholder search (`q`: \"\")\n    5. Removed relevancy rule `exactness` at position #5 in ranking rules because the query is a placeholder search (`q`: \"\")\n  - note: The ranking rules of query #1 were modified during canonicalization:\n    1. Removed relevancy rule `words` at position #1 in ranking rules because the query is a placeholder search (`q`: \"\")\n    2. Removed relevancy rule `typo` at position #2 in ranking rules because the query is a placeholder search (`q`: \"\")\n    3. Removed relevancy rule `proximity` at position #3 in ranking rules because the query is a placeholder search (`q`: \"\")\n    4. Removed relevancy rule `attribute` at position #4 in ranking rules because the query is a placeholder search (`q`: \"\")\n    5. Removed relevancy rule `exactness` at position #5 in ranking rules because the query is a placeholder search (`q`: \"\")\n",
```
ManyTheFish, line 1597: Well, it's a bit long, but it's clear. 
I think the `note` parts are a bit too much, but on the other hand, there is nowhere else to put it.
Maybe a log::info or debug?

ManyTheFish (changes requested):

meilisearch/tests/search/multi.rs
dureuill: Hello, thanks for your comment.

I think the note parts are very important to understand what is happening in "nonstandard" cases.
Putting them in the logs is a bad location, because users don't look at logs. Currently, Cloud users cannot even look at logs. I think it is best if users can troubleshoot their issues themselves, and so we need to give them the tools to do so.

Now, you've got a point that the output is verbose. The thing could be made less verbose by displaying only notes that are responsible for a rule mismatch 🤔 I'm not 100% sure how to do that, however

ManyTheFish: Yes, I agree it can be useful, let's keep it and see

ManyTheFish (approved):
bors merge

hcbaxiom:
Does it supports federating the facets?

dureuill:
Hello, I answered in [your issue](https://github.com/meilisearch/meilisearch/issues/4913#issuecomment-2337324553).

Short answer: no.

PR #4774, opened 2024-07-04
Rename the sortable into the filterable movies workload

Kerollmops:
Fixes the workload name of one movie searchable.


dureuill (approved):
bors merge

PR #4794, opened 2024-07-12
Optimize GET /indexes route

### LINKED ISSUE #4694: Meilisearch takes an eternity to answer to `GET /indexes` when there is a few thousands indexes and timeout
I made 8000 indexes and while I was pushing documents to them I tried to list the indexes. It timed out AND it stopped all requests from reaching the server for a good minute.
The indexing part also seems to have stopped.
Then, it came back to normal automatically.

**To Reproduce**
Run this script;
```rust
BASELINE=$RANDOM
echo "Sending repeated updates..."
LOREM="Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum."
for i in {0..8}; do LOREM="$(echo "$i" | shasum) $LOREM $(echo "$LOREM" | shasum) $LOREM $(echo "$LOREM $i" | shasum)"; done
echo "With content as large as $(echo "${LOREM}" | wc -c)B"
for iteration in {0..100000};
do
  for index in {0..8000};
  do
    SHASUM=$(echo "$iteration" | shasum)$(echo "$iteration$index" | shasum)
    for pk in {0..100}; do
      echo "{ \"id\": \"${BASELINE}_${pk}\", \"doggo\": \"$iteration $LOREM $iteration\", \"bad for perf\": \"$SHASUM\" }" | mieli -i "$index" document add --async > /dev/null
    done
  done
done
```

And then while it indexes run `mieli indexes list`.

**Expected behavior**
It works

**Meilisearch version:**
v1.8.2

**Additional context**
It seems like the index scheduler is opening and closing all the indexes; it shouldn’t need to do that to list the indexes.
Also, the `stats` routes work, which is even stranger since it's doing way more work.

ThalusA:
# Pull Request

## Related issue
Fixes #4694

## What does this PR do?
- Optimize indexes listings by not processing all indexes and only processing the ones that are used as the result of the pagination
- Don't open the index if they are already cached

## PR checklist

### DIFF
--- a/index-scheduler/src/lib.rs
+++ b/index-scheduler/src/lib.rs
@@ -51,7 +51,7 @@ use meilisearch_types::error::ResponseError;
 use meilisearch_types::features::{InstanceTogglableFeatures, RuntimeTogglableFeatures};
 use meilisearch_types::heed::byteorder::BE;
 use meilisearch_types::heed::types::{SerdeBincode, SerdeJson, Str, I128};
-use meilisearch_types::heed::{self, Database, Env, PutFlags, RoTxn, RwTxn};
+use meilisearch_types::heed::{self, Database, Env, PutFlags, RoIter, RoTxn, RwTxn};
 use meilisearch_types::milli::documents::DocumentsBatchBuilder;
 use meilisearch_types::milli::index::IndexEmbeddingConfig;
 use meilisearch_types::milli::update::IndexerConfig;
@@ -70,6 +70,7 @@ use uuid::Uuid;
 
 use crate::index_mapper::IndexMapper;
 use crate::utils::{check_index_swap_validity, clamp_to_page_size};
+use crate::uuid_codec::UuidCodec;
 
 pub(crate) type BEI128 = I128<BE>;
 
@@ -415,6 +416,23 @@ impl IndexScheduler {
     }
 }
 
+/// An owned type for database entries iterator and its transaction.
+/// To get the inner iterator you should call .iter() on it.
+pub struct IndexIterator<'txn> {
+    rtxn: RoTxn<'txn>,
+    index_mapper: &'txn IndexMapper,
+}
+
+impl<'txn> IndexIterator<'txn> {
+    pub fn new(rtxn: RoTxn<'txn>, index_mapper: &'txn IndexMapper) -> IndexIterator<'txn> {
+        Self { rtxn, index_mapper }
+    }
+
+    pub fn iter(&'txn self) -> Result<RoIter<'txn, Str, UuidCodec>> {
+        self.index_mapper.iter(&self.rtxn)
+    }
+}
+
 impl IndexScheduler {
     /// Create an index scheduler and start its run loop.
     pub fn new(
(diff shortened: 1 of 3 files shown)


meilisearch/src/routes/indexes/mod.rs
```diff
@@ -99,16 +99,33 @@ pub async fn list_indexes(
...
-    let ret = paginate.as_pagination().auto_paginate_sized(indexes.into_iter());
+    let index_iterator = index_scheduler.iter()?;
+    let database_iterator = index_iterator.iter()?;
+    let indexes = database_iterator
+        .filter(|res| {
+            res.as_ref().map(|(name, _)| filters.is_index_authorized(name)).unwrap_or(false)
+        })
+        .flat_map(|res| {
+            res.ok().and_then(|(name, _)| {
+                index_scheduler.index_stats(name).ok().map(|index| IndexView {
```
irevoire, line 110: Hey, I’m wondering, why are we not simply calling the `index_stats` directly on the index-name instead of doing the double iterator stuff you implemented 🤔

index-scheduler/src/lib.rs
```diff
@@ -415,6 +416,23 @@ impl IndexScheduler {
...
+    rtxn: RoTxn<'txn>,
+    index_mapper: &'txn IndexMapper,
+}
+
+impl<'txn> IndexIterator<'txn> {
+    pub fn new(rtxn: RoTxn<'txn>, index_mapper: &'txn IndexMapper) -> IndexIterator<'txn> {
+        Self { rtxn, index_mapper }
+    }
+
+    pub fn iter(&'txn self) -> Result<RoIter<'txn, Str, UuidCodec>> {
```
irevoire, line 431: I’ll be honest, I’m not a fan of leaking the internal `Uuid`, that’s not really supposed to come out of the index-scheduler.
I think we should only return the index_name here since we don't need anything else outside of the index-scheduler.

meilisearch/src/routes/indexes/mod.rs
```diff
@@ -99,16 +99,33 @@ pub async fn list_indexes(
...
+    // In order to find how many elements there is we must create a new iterator that will only
+    // filter the total authorized indexes that are valid, consume it and return the number of elements.
+    let index_iterator = index_scheduler.iter()?;
+    let database_iterator = index_iterator.iter()?;
+    let count = database_iterator
+        .filter(|res| {
+            res.as_ref().ok().map(|(name, _)| filters.is_index_authorized(name)).unwrap_or(false)
+        })
+        .count();
+    let ret = paginate.as_pagination().auto_paginate_unsized(count, indexes);
```
irevoire, line 128: Not a fan of this part but I don't see any other easy way to do it

irevoire (changes requested):

meilisearch/src/routes/indexes/mod.rs
ThalusA: I tried size_hint but it just show me a size of 0 in both way

irevoire:
Superseded by https://github.com/meilisearch/meilisearch/issues/4694#event-15789897476
Closing

PR #4795, opened 2024-07-14
Catch the MDB_TXN_FULL error and reduce the batch size to eventually succeed

### LINKED ISSUE #4791: internal: MDB_TXN_FULL: Transaction has too many dirty pages - transaction too big
**Describe the bug**
Sometimes, Meilisearch fails to index documents or update the settings and returns `MDB_TXN_FULL` errors.

**To Reproduce**
If you send too many documents, force Meilisearch to batch them together. This will result in an `MDB_TXN_FULL` error in the task queue.

**Expected behavior**
Meilisearch should try again, like when we get an `MDB_MAP_FULL` error. In this latter situation, we catch the error instead of returning it to the user. We increase the LMDB map size by 1TiB and start again.

We might encounter the `MDB_TXN_FULL` error. In such a case, we can take the following actions: reduce the batch size, return the error to the user if the batch size is already one (1), and start indexing again by splitting the batch size in half.

**Meilisearch version:**
Every version can auto-batch tasks.

**Additional context**
I was on an iMac M3, but it should be roughly the same on all machines.

Karribalu:
# Pull Request

## Related issue
Fixes #4791

## What does this PR do?
- Catches the MDB_TXN_FULL when batches are executed and reduces the max_number_of_batched_tasks and tries again.

## PR checklist
- [ x] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [x ] Have you read the contributing guidelines?
- [x ] Have you made sure that the title is accurate and descriptive of the changes?

### DIFF
--- a/index-scheduler/src/lib.rs
+++ b/index-scheduler/src/lib.rs
@@ -47,7 +47,7 @@ pub use features::RoFeatures;
 use file_store::FileStore;
 use flate2::bufread::GzEncoder;
 use flate2::Compression;
-use meilisearch_types::error::ResponseError;
+use meilisearch_types::error::{Code, ResponseError};
 use meilisearch_types::features::{InstanceTogglableFeatures, RuntimeTogglableFeatures};
 use meilisearch_types::heed::byteorder::BE;
 use meilisearch_types::heed::types::{SerdeBincode, SerdeJson, Str, I128};
@@ -605,7 +605,7 @@ impl IndexScheduler {
     /// This function will execute in a different thread and must be called
     /// only once per index scheduler.
     fn run(&self) {
-        let run = self.private_clone();
+        let mut run = self.private_clone();
         std::thread::Builder::new()
             .name(String::from("scheduler"))
             .spawn(move || {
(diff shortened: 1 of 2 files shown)


Karribalu:
I tried to replicate the issue by sending lot of documents from the utility which I have created, But surprisingly Meilisearch did a splendid work and processed all without any issues.
So to replicate the issue, I added a random boolean which intermittently fails batches by throwing the MDB_TXN_FULL error.
I tested with this change and it worked! 

I didn't tackle the MDB_MAP_FULL problem, As I don't have complete idea on the LMDB Map size, Happy to do the changes if someone can help me understanding it.

Screenshot 2024-07-14 at 20 48 16

Kerollmops:
Hey @Karribalu 👋

I am sorry for the very, very late reply. This PR seems more difficult than expected, and as we are rewriting the indexer and a part of the indexer scheduler, we will rewrite this part on our own.

I am sorry for this but I prefer closing this PR,
Have a nice week 🙂

PR #4806, opened 2024-07-17
Update rustls as much as possible

irevoire:
# Pull Request

## Related issue
Part of https://github.com/meilisearch/meilisearch/issues/4753

## What does this PR do?
- Update rustls as much as possible

## What is missing

In rustls-0.22.0 two structures we were using have been removed with no explanation or workaround
image

### DIFF
--- a/meilisearch/src/option.rs
+++ b/meilisearch/src/option.rs
@@ -14,11 +14,9 @@ use clap::Parser;
 use meilisearch_types::features::InstanceTogglableFeatures;
 use meilisearch_types::milli::update::IndexerConfig;
 use meilisearch_types::milli::ThreadPoolNoAbortBuilder;
-use rustls::server::{
-    AllowAnyAnonymousOrAuthenticatedClient, AllowAnyAuthenticatedClient, ServerSessionMemoryCache,
-};
+use rustls::server::{ServerSessionMemoryCache, WebPkiClientVerifier};
 use rustls::RootCertStore;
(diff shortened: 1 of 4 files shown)


meilisearch/src/option.rs
```diff
@@ -769,41 +767,48 @@ impl Deref for MaxThreads {
...
         fs::File::open(filename).map_err(|_| anyhow::anyhow!("cannot open certificate file"))?;
     let mut reader = BufReader::new(certfile);
     certs(&mut reader)
-        .map(|certs| certs.into_iter().map(rustls::Certificate).collect())
+        .collect::<Result<Vec<_>, _>>()
         .map_err(|_| anyhow::anyhow!("cannot read certificate file"))
 }
 
-fn load_private_key(filename: PathBuf) -> anyhow::Result<rustls::PrivateKey> {
+fn load_private_key(
```
djc, line 781: I think you can probably use `rustls_pemfile::private_key()` instead of `rsa_private_keys()` and `pkcs8_private_keys()`.

irevoire: Huuum, I don't think it's equivalent because currently, we read both the rsa and pkcs8 and then pick the first pkcs8 if there is one.
`private_key` would return the first key, whatever it is, so we may pick an RSA key where we would choose the pkcs8 key in the past.

I’m not confident enough to make this change tbh 
I don't know what it could break and if people are relying on it

Kerollmops (approved):
Changes look good to me! Thank you again!
bors merge

PR #4808, opened 2024-07-18
Make the tests run faster

irevoire:
### Only check the consistency of the index-scheduler on snapshots when running in release mode

This saves 12s on the tests, and since the tests run in release mode in the CI, we don't lose any information.
From 28s to 16s

### We were snapshotting the index for no reason in `advance_till`, I removed this call

This saved an additional 8s on the tests, going from 16s to 8s.

----

After these two optimizations, the test suite as a whole executes 14% quicker

## Meilisearch integration tests

While profiling this test suite, nothing stands out. The only noticeable thing is that we're losing most of our time creating and dropping threads.
I made the theory that by sharing a single common instance between all integrations tests I would gain some time again.

In https://github.com/meilisearch/meilisearch/pull/4808/commits/355a7acd1c85a1f728f31fb7f4e9c17233ea1a95 I saved another 15s by only testing this theory on the module that tests the error messages. 
But we can do it on many more tests. **We must take care of not making any test flaky, though**.

## Use two indexing threads

By moving from one to two indexing threads, we gain an additional 30% in performance.

# Conclusion

## Before

The execution of the test suite was taking around:
- 4m40s on my computer
- 15 minutes on the debug CI with cache
- 29 minutes on the Windows CI with cache

## After

The execution of the test suite is taking around:
- 2m20 on my computer
- 8 minutes on the debug CI with cache
- 29 minutes on the Windows CI with cache


meilisearch/tests/common/index.rs
```diff
@@ -327,6 +248,146 @@ impl Index<'_> {
...
+#[allow(dead_code)]
+impl<State> Index<'_, State> {
+    pub async fn get(&self) -> (Value, StatusCode) {
+        let url = format!("/indexes/{}", urlencode(self.uid.as_ref()));
+        self.service.get(url).await
+    }
+
+    /// add_documents is not allowed on shared index but we need to use it to initialize
+    /// a bunch of very common indexes in `common/mod.rs`.
+    pub(super) async fn _add_documents(
```
dureuill, line 286: this is test code so I don't really mind, but for the record I'm not a fan of having underscore functions here.

underscore has the specific meaning of "maybe unused" in Rust, so if these are internal functions I would either have made them free functions of the module, or if self contains a lot of data, I would have made an internal index object exposing the functions

dureuill (approved):
thanks!

meilisearch/tests/common/index.rs
irevoire: With the `pub(super)` honestly, it could simply be in the module, and without the `_`.
I’m merging as-is because the PR is already huge, but if you want, I can do either:
- Remove the `_` and let the method in the module
- Free functions of the module

irevoire:
bors merge

PR #4812, opened 2024-07-18
Allow `MEILI_NO_VERGEN` env var to skip vergen

dureuill:
- vergen checks the state of the `.git` directory to embed commit information into the `meilisearch` binary and the `cargo xtask bench` invocations.
- This check unfortunately results in too many recompilation of the `meilisearch` binary.
- This PR allows skipping vergen when the `MEILI_NO_VERGEN` variable is present in the environment


irevoire (approved):
awesome!

bors merge

Issue #4756, opened 2024-07-01
Improve usage of the rest embedder


irevoire:
Here's a list of stuff I found hard to understand and use while trying to make the rest embedder work:
- [ ] [REST API parameter names and behavior are unclear](https://github.com/meilisearch/documentation/pull/2824#issuecomment-2124073720)
- [ ] [Bad error message or bad default value when we don't specify the `query` parameter](https://github.com/meilisearch/meilisearch/blob/85d8455c11bfa0a5c2ba70d15554c974b003d666/meilisearch/tests/vector/rest.rs#L105-L140)
- [ ] [Bad error message when both `pathToEmbeddings` and `embeddingObject` are missing](https://github.com/meilisearch/meilisearch/blob/2141cb3b69f52ca6fcf18d0b89df1d8601a22f85/meilisearch/tests/vector/rest.rs#L142-L178)
- [ ] [Unexpected error when we don't specify both `pathToEmbeddings` and `embeddingObject` (only once should be required)](https://github.com/meilisearch/meilisearch/blob/2141cb3b69f52ca6fcf18d0b89df1d8601a22f85/meilisearch/tests/vector/rest.rs#L180-L260)
- [ ] [Should not panic when the dimensions specified do not work with the model](https://github.com/meilisearch/meilisearch/blob/2141cb3b69f52ca6fcf18d0b89df1d8601a22f85/meilisearch/tests/vector/rest.rs#L262-L299)
- [ ] [Be more flexible on the type of data that is accepted](https://github.com/meilisearch/meilisearch/issues/4757#issuecomment-2201948531)
  - [ ] Always accept arrays of embeddings even if `inputType` is set to `text`
  - [ ] Accept arrays of result at the root level and texts/array of text at the root level.

Karribalu:
Good Day @irevoire,
I found the culprit for internal server error when dimensions is missed and also bad error message when either query field is missed or bad json payload is passed.
I want to understand if the query field is mandatory?
It would be helpful if you can point me to some API Documentation for this.

dureuill:
Hello @Karribalu, thanks for looking into this :heart:

I plan on discussing the REST embedder API at a more global scale, which could significantly change the API. As a result, I would not suggest working on the items of this task at the moment.

We'll keep this task updated.

Karribalu:
Thanks, for letting me know @dureuill !

Issue #4757, opened 2024-07-01
Vector store feature: improve test coverage


dureuill:
# Hybrid search

- [ ] one basic test with every embedder
  - [x] user provided
  - [x] Hugging Face
  - [ ] OpenAI
    - [ ] basic test
    - [ ] search with wrong API key
    - [ ] search with very long query
  - [ ] REST
    - [ ] basic test
    - [ ] search with wrong API key
    - [ ] search with wrong URL
  - [ ] ollama
    - [ ] basic test
    - [ ] search with wrong API key
    - [ ] search with wrong URL
- [x] passing wrong vector dimensions 
- [ ] test all `q`, `vector`, `hybrid` configurations:

| Tested/Configuration | `q` | `vector` |  `hybrid` |
|-----------------------|---|----------|-----------|
| ❌ | missing | missing | missing |
| ✅ | missing | missing | present |
| ❌ | missing | present | missing |
| ✅ | missing | present | present |
| ❌ | present | missing | missing |
| ✅ | present | missing | present |
| ✅ | present | present| missing |
| ✅ | present | present| present |
- [x] test `offset` and `limit`
- [ ] test `page` and `hitsPerPage`
- [ ] test `attributesToRetrieve`
- [x] test `retrieveVectors`
- [ ] test `attributeToCrop`
- [ ] test `cropLength`
- [x] test `attributesToHighlight`
- [ ] test `filter` 
- [ ]  test `showMatchesPosition`
- [x] test `showRankingScore`
- [x] test `showRankingScoreDetails`
- [ ] test `sort`
- [ ] test `distinct`
- [ ] test `facets`
- [x] test `highlight_pre_tag` 
- [x] test `highlight_post_tag`
- [x] test `crop_marker` 
- [ ] test `matchingStrategy`
- [ ] test `attributesToSearchOn`
- [x] test `rankingScoreThreshold`
- [ ] test hybrid
  - [x] test `semanticRatio`
  - [ ] test `embedder`
    - [x]  basic test
    - [ ] passing wrong embedder
    - [ ] without any embedder
- [x] test experimental feature check

# Embedder configuration

- [ ] source doesn't exist
- [ ] `model`
  - [ ] model is not in the list of available models for OpenAI
  - [ ] model is not available on the ollama side
  - [ ] model does not exist for a Hugging Face model
  - [ ] model is not a BERT model for a Hugging Face model
  - [ ] model was specified for a user provided embedder
[...]

PR #4815, opened 2024-07-22
Rest embedder api mk2

dureuill:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4756

- [x] [REST API parameter names and behavior are unclear](https://github.com/meilisearch/documentation/pull/2824#issuecomment-2124073720)
  - unclear names are removed. There remain only two parameters: `request`, a template of what Meilisearch's request to the embedding server should be, and `response`, a template of what the embedding server's response to Meilisearch should look like
- [x] [Bad error message or bad default value when we don't specify the `query` parameter](https://github.com/meilisearch/meilisearch/blob/85d8455c11bfa0a5c2ba70d15554c974b003d666/meilisearch/tests/vector/rest.rs#L105-L140)
  - The replacement for `query`, which is `request`, is now a mandatory parameter. Omitting it will result in the following error message : "`.embedders.rest`: Missing field `request` (note: this field is mandatory for source rest)", which is clear
- [x] [Bad error message when both `pathToEmbeddings` and `embeddingObject` are missing](https://github.com/meilisearch/meilisearch/blob/2141cb3b69f52ca6fcf18d0b89df1d8601a22f85/meilisearch/tests/vector/rest.rs#L142-L178)
  - These parameters no longer exist. Now, the point of extraction is given directly by the location of an `{{embedding}}` placeholder in the `response` parameter.
- [x] [Unexpected error when we don't specify both `pathToEmbeddings` and `embeddingObject` (only once should be required)](https://github.com/meilisearch/meilisearch/blob/2141cb3b69f52ca6fcf18d0b89df1d8601a22f85/meilisearch/tests/vector/rest.rs#L180-L260)
  - These parameters no longer exist. Now, the point of extraction is given directly by the location of an `{{embedding}}` placeholder in the `response` parameter.
- [x] [Should not panic when the dimensions specified do not work with the model](https://github.com/meilisearch/meilisearch/blob/2141cb3b69f52ca6fcf18d0b89df1d8601a22f85/meilisearch/tests/vector/rest.rs#L262-L299)
  - This no longer panics, instead returns "While embedding documents for embedder `rest`: runtime error: was expecting embeddings of dimension `2`, got embeddings of dimensions `3`"
- [x] [Be more flexible on the type of data that is accepted](https://github.com/meilisearch/meilisearch/issues/4757#issuecomment-2201948531)
  - [x] Always accept arrays of embeddings even if `inputType` is set to `text`
    - This is controlled by the repeat placeholder `"{..}"`, an array of embeddings can be configured even if the input is not in an array.
  - [x] Accept arrays of result at the root level and texts/array of text at the root level.
    -  doable with `request: "{{text}}"` and `response: "{{embedding}}"` or `response: ["{{embedding}}"]` (see test `vector::rest::server_raw`)

## What does this PR do?
- [See public usage](https://meilisearch.notion.site/v1-10-AI-search-changes-737c9d7d010d4dd685582bf5dab579e2#8de842673ffa4a139210094a89c1ec3e)
- Add new `milli::vector::json_template` module to parse JSON templates with an injection placeholder and a repeat placeholder
- Change rest embedder to use two JSON templates
- Change ollama and openai embedders to use the new rest embedder
- Update settings
- Update and add tests

## Breaking change

> [!CAUTION]
> This PR is a breaking change to the REST embedder.
> Importing a dump containing a REST embedder configuration will fail in v1.10 with an error: "Error: unknown field `query`, expected one of `source`, `model`, `revision`, `apiKey`, `dimensions`, `documentTemplate`, `url`, `request`, `response`, `distribution` at line 1 column 752".

Upgrade procedure:

1. Remove any embedder with source "rest"
2. Create a dump
3. Import that dump in a v1.10
4. Re-add any removed embedder, using the new settings.


milli/src/vector/json_template.rs
```diff
@@ -0,0 +1,1013 @@
...
+                    }
+                    *current_value = injected_value;
+                    array.push(repeated_value);
+                }
+
+                if array.is_empty() {
+                    return Err(MissingValue);
+                }
+                // 2. inject at the injection point in the rendered value
+                let mut current_value = &mut rendered;
```
dureuill, line 525: Factor these 3 repeated § to a function

milli/src/vector/json_template.rs
```diff
@@ -0,0 +1,1013 @@
...
+    ///   extracted from.
+    /// - `repeat_string`: Sentinel value that can be placed as the second value in an array to indicate that the first value can be repeated
+    ///   any number of times. The first value should contain exactly one placeholder string.
+    ///
+    /// # Errors
+    ///
+    /// - [`TemplateParsingError`]: refer to the documentation of this type
+    pub fn new(
+        template: Value,
+        placeholder_string: String,
```
dureuill, line 443: &str

dureuill: done

milli/src/vector/json_template.rs
dureuill: done

ManyTheFish (approved):
LGTM,
bors merge

PR #4819, opened 2024-07-23
Language settings

### LINKED ISSUE #4749: Language settings
Related product team resources: [PRD](https://www.notion.so/meilisearch/Define-language-settings-to-impact-relevancy-bee62e18b7584c4f87d18a7654855329?pvs=31)

## Motivation

- Users are forced to use an alternative image created by the engine team
- Leads to multiple maintenance work on engineering team side
- Multiple issues and bug for Japanese and Chinese users we cannot start to solve without this settings 

## Usage

https://meilisearch.notion.site/v1-10-Language-settings-usage-26c5d98b553349d9abacbe7aff698e4e

## Impacted teams

ManyTheFish:
# Pull Request

## Related issue
Fixes #4749 

## What does this PR do?
- [Implement localized search](https://github.com/meilisearch/meilisearch/pull/4819/commits/c0c6955c0df98fdb7122f27e47664cf541836334)
- [Implement localized attributes settings](https://github.com/meilisearch/meilisearch/pull/4819/commits/bde827b05512bf50dae10bd9402d4a3600d15061)

## Related PRD

- [PRD](https://www.notion.so/meilisearch/Define-language-settings-to-impact-relevancy-bee62e18b7584c4f87d18a7654855329)
- [Public usage](https://www.notion.so/meilisearch/v1-10-Language-settings-usage-26c5d98b553349d9abacbe7aff698e4e)


meilisearch-types/src/locales.rs
```diff
@@ -0,0 +1,132 @@
...
+        impl std::str::FromStr for Locale {
+            type Err = LocaleFormatError;
+
+            fn from_str(s: &str) -> Result<Self, Self::Err> {
+                milli::tokenizer::Language::from_code(s).map(Self::from).ok_or(LocaleFormatError {
+                    invalid_locale: s.to_string(),
+                })
+            }
+        }
+    };
```
ManyTheFish, line 58: consider moving outside of macro

milli/src/search/facet/search.rs
```diff
@@ -330,3 +336,15 @@ impl ValuesCollection {
...
+    let options = NormalizerOption { lossy: true, ..Default::default() };
+    let mut detection = StrDetection::new(facet_string, locales);
+    let token = Token {
+        lemma: std::borrow::Cow::Borrowed(facet_string),
+        script: detection.script(),
+        language: detection.language(),
+        ..Default::default()
+    };
+
+    token.normalize(&options).lemma.to_string()
```
ManyTheFish, line 349: prefer using co::into_owned

meilisearch-types/src/locales.rs
```diff
@@ -0,0 +1,157 @@
...
+
+        #[derive(Debug)]
+        pub struct LocaleFormatError {
+            pub invalid_locale: String,
+        }
+
+        impl std::fmt::Display for LocaleFormatError {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                let valid_locales = [$(Locale::$language),+].iter().map(|l| format!("`{}`", json!(l).as_str().unwrap())).collect::<Vec<_>>().join(", ");
+                write!(f, "Unknown value `{}`, expected one of {}", self.invalid_locale, valid_locales)
```
dureuill, line 43: Should we change the phrasing here a little bit?

```suggestion
                write!(f, "Unsupported locale `{}`, expected one of {}", self.invalid_locale, valid_locales)
```

?

meilisearch/src/routes/indexes/settings.rs
```diff
@@ -786,6 +808,7 @@ pub async fn update_all(
             },
             "embedders": crate::routes::indexes::settings::embedder_analytics(new_settings.embedders.as_ref().set()),
             "search_cutoff_ms": new_settings.search_cutoff_ms.as_ref().set(),
+            "locales": new_settings.localized_attributes.as_ref().set().map(|rules| rules.iter().flat_map(|rule| rule.locales.iter().cloned()).collect::<std::collections::BTreeSet<_>>()),
```
dureuill, line 811: is not clear to me why the key here is `"locales"` rather than `localized_attributes`?

is this what was decided for analytics?

milli/src/index.rs
```diff
@@ -1562,46 +1553,32 @@ impl Index {
...
-    pub fn script_language_documents_ids(
+    pub fn localized_attributes_rules(
         &self,
         rtxn: &RoTxn<'_>,
-        key: &(Script, Language),
-    ) -> heed::Result<Option<RoaringBitmap>> {
-        self.script_language_docids.get(rtxn, key)
+    ) -> heed::Result<Option<Vec<LocalizedAttributesRule>>> {
+        self.main
+            .remap_types::<Str, SerdeBincode<Vec<LocalizedAttributesRule>>>()
```
dureuill, line 1561: is using Bincode necessary here? Bincode is a format that isn't reflexive, so if we ever need to modify the format of what is being stored, it might be complicated to do manually.

serde_json is going to be longer, but I don't think it is necessarily an issue here?

meilisearch/tests/search/locales.rs
```diff
@@ -0,0 +1,662 @@
...
+              "estimatedTotalHits": 2
+            }
+            "###);
+            snapshot!(code, @"200 OK");
+        })
+        .await;
+}
+
+#[actix_rt::test]
+async fn force_locales() {
```
dureuill, line 198: I think it would be great to have an additional test that has two distinct rules

```
                "localizedAttributes": [
                    {"attributePatterns": ["*_ja"], "locales": ["jpn"]},
                    {"attributePatterns": ["*_zh"], "locales": ["cmn"]}
                ]

File: meilisearch/tests/search/locales.rs [hunk 7]
```diff
@@ -0,0 +1,662 @@
...
+                "###);
+                snapshot!(code, @"200 OK");
+            },
+        )
+        .await;
+
+    // force japanese
+    index
+        .search(
+            json!({"q": "\"进击的巨人\"", "attributesToRetrieve": ["id"], "locales": ["jpn"]}),
```
dureuill, line 249: I am a bit surprised that adding a locale changes the result. From the search code I was under the impression that when no locale is provided, the query is parsed with the locale from the index.

milli/src/localized_attributes_rules.rs
```diff
@@ -0,0 +1,112 @@
...
+            let fields = fields_ids.filter_map(|field_id| {
+                fields_ids_map.name(field_id).map(|field_name| (field_id, field_name))
+            });
+
+            for (field_id, field_name) in fields {
+                let mut locales = Vec::new();
+                for rule in rules {
+                    if rule.match_str(field_name) {
+                        locales.extend(rule.locales.iter());
+                    }
```
dureuill, line 74: I'm a bit surprised here. I would have thought that the locales that would apply would be the locales of the first matching rule, as per the public usage:

> if an attribute matches several rules, only the first rule in the list will be applied

so I would expect:

```suggestion
                        break;
                    }
```

milli/src/update/index_documents/extract/extract_facet_string_docids.rs
```diff
@@ -72,29 +68,66 @@ pub fn extract_facet_string_docids<R: io::Read + io::Seek>(
...
+                    let val = SerdeJson::bytes_encode(&set).map_err(heed::Error::Encoding)?;
+                    obkv.insert(deladd_key, val)?;
+                }
+                obkv.finish()?;
+
+                let key: (u16, &str) = (field_id, new_hyper_normalized_value.as_ref());
+                let key_bytes = BEU16StrCodec::bytes_encode(&key).map_err(heed::Error::Encoding)?;
+                normalized_facet_string_docids_sorter.insert(key_bytes, &buffer)?;
+            } else {
+                // if the facet string is different, we need to insert the deletion and addition in different obkv because the related key is different.
```
dureuill, line 97: it would be interesting to have a test where a change in settings would trigger this case

milli/src/update/settings.rs
```diff
@@ -1118,6 +1128,24 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
...
+                } else {
+                    self.index.put_localized_attributes_rules(self.wtxn, new.clone())?;
+                    true
+                }
+            }
+            Setting::Reset => self.index.delete_localized_attributes_rules(self.wtxn)?,
+            Setting::NotSet => false,
+        };
+
+        Ok(changed)
```
dureuill, line 1146: This returns a boolean but it doesn't seem like it is used?

dureuill (commented):
I have a few comments (see inline comments), but it is exciting to have this feature landing soon 🚀

meilisearch-types/src/error.rs
```diff
@@ -256,6 +256,7 @@ InvalidSearchCropLength               , InvalidRequest       , BAD_REQUEST ;
 InvalidSearchCropMarker               , InvalidRequest       , BAD_REQUEST ;
 InvalidSearchFacets                   , InvalidRequest       , BAD_REQUEST ;
 InvalidSearchSemanticRatio            , InvalidRequest       , BAD_REQUEST ;
+InvalidSearchLocales                  , InvalidRequest       , BAD_REQUEST ;
```
dureuill, line 259: Consider implementing `Display` for this error code (see how it is done e.g. for `deserr_codes::InvalidSearchSemanticRatio` below in this file)

dureuill (approved):
🎉

bors merge

PR #4830, opened 2024-07-29
Use the dtolnay's Rust Toolchain

Kerollmops:
Fixes the CI by using another rust-toolchain GitHub repo.

Note: the [helix-editor/rust-toolchain repository](https://github.com/helix-editor/rust-toolchain) has been deleted so we moved to the [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) one. However, the dtolnay's one doesn't support `rust-toolchain.toml` and the version is directly in the rust-toolchain@version. We keep the `rust-toolchain.toml` for local builds only.

### DIFF
--- a/.github/workflows/bench-manual.yml
+++ b/.github/workflows/bench-manual.yml
@@ -18,7 +18,7 @@ jobs:
         timeout-minutes: 180 # 3h
         steps:
             - uses: actions/checkout@v3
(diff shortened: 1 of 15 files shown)


dureuill:
We'll need to update the issue template about upgrading the rust version to specify that we should update the `@rust-version` in the github CI.

dureuill (approved):
bors merge

PR #4835, opened 2024-07-29
Log error from main using tracing

dureuill:
Engine follow-up to https://github.com/meilisearch/meilisearch-support/issues/252#issuecomment-2251288276 (private link)

> @meilisearch/engine-team we need to open a PR to tracing::error! when an error occurs in the Meilisearch main. It would be nice to have it included in the second RC

Error logged when launching Meilisearch to import dump on path where the dump doesn't exist

---

Error logges when launching Meilisearch with a db path that is not writeable


irevoire (approved):
Awesome thank you!
bors merge

Issue #4825, opened 2024-07-25
Prometheus metrics overload


Malgefor:
Hey! I have a question related to #3983. I was also experiencing an overload of metrics on my Prometheus server which caused it to go down. After #4619 it is partly fixed. However, I am still getting **82.000** metrics just to see that a certain task returned a 200 OK status code (sample below). That amount of metrics still occasionally kills my Prometheus server. 

Can this metric, maybe specific to **/tasks/***, receive the same kind of treatment as in #4619? 

```
meilisearch_http_requests_total{method="GET",path="/tasks/9989",status="200"} 31
meilisearch_http_requests_total{method="GET",path="/tasks/999",status="200"} 36
meilisearch_http_requests_total{method="GET",path="/tasks/9990",status="200"} 28
meilisearch_http_requests_total{method="GET",path="/tasks/9991",status="200"} 1
meilisearch_http_requests_total{method="GET",path="/tasks/9992",status="200"} 1
meilisearch_http_requests_total{method="GET",path="/tasks/9993",status="200"} 2
meilisearch_http_requests_total{method="GET",path="/tasks/9994",status="200"} 2
meilisearch_http_requests_total{method="GET",path="/tasks/9995",status="200"} 14
meilisearch_http_requests_total{method="GET",path="/tasks/9996",status="200"} 30
meilisearch_http_requests_total{method="GET",path="/tasks/9997",status="200"} 31
meilisearch_http_requests_total{method="GET",path="/tasks/9998",status="200"} 31
meilisearch_http_requests_total{method="GET",path="/tasks/9999",status="200"} 36
```

_Originally posted by @Malgefor in https://github.com/meilisearch/meilisearch/issues/3983#issuecomment-2238651849_

Malgefor:
Awesome! Thanks for the fix 😃

PR #4839, opened 2024-07-30
In prometheus metrics return the route pattern instead of the real route when returning the HTTP requests total

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4825

## What does this PR do?
- return the route pattern instead of the real route when returning the HTTP requests total


dureuill (approved):
Cool, thank you :+1:

irevoire:
bors merge

PR #4850, opened 2024-07-31
Use a fixed date format regardless of features

### LINKED ISSUE #4844: Docker image errors with `OffsetDateTime` on Meilisearch 1.7+ when index created on host machine
**Describe the bug**
When pre-building the database on a host machine by adding an index, then copying the database into the Docker image and running it, I get these errors:

> ERROR HTTP request{method=GET host="0.0.0.0:7700" route=/indexes query_parameters= user_agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.5 Safari/605.1.15 status_code=500 error=internal: decoding failed.}: tracing_actix_web::middleware: Error encountered while processing the incoming HTTP request: ResponseError { code: 500, message: "internal: decoding failed.", error_code: "internal", error_type: "internal", error_link: "https://docs.meilisearch.com/errors#internal" }

> 2024-07-30T14:35:43.602926Z ERROR HTTP request{method=GET host="0.0.0.0:7700" route=/stats query_parameters= user_agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.5 Safari/605.1.15 status_code=500 error=error while decoding: invalid type: string "2024-07-30 14:35:19.634762 +00:00:00", expected an `OffsetDateTime` at line 1 column 231}: tracing_actix_web::middleware: Error encountered while processing the incoming HTTP request: ResponseError { code: 500, message: "error while decoding: invalid type: string \"2024-07-30 14:35:19.634762 +00:00:00\", expected an `OffsetDateTime` at line 1 column 231", error_code: "internal", error_type: "internal", error_link: "https://docs.meilisearch.com/errors#internal" }

The web UI also shows the error:

Screenshot 2024-07-30 at 3 41 48 PM

There is a possibly related log:

> 2024-07-30T14:44:30.209399Z  WARN HTTP request{method=POST host="0.0.0.0:7700" route=/multi-search query_parameters= user_agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.5 Safari/605.1.15 status_code=400 error=Invalid value at `.queries[0].indexUid`: `` is not a valid index uid. Index uid can be an integer or a string containing only alphanumeric characters, hyphens (-) and underscores (_).}: tracing_actix_web::middleware: Error encountered while processing the incoming HTTP request: DeserrError { msg: "Invalid value at `.queries[0].indexUid`: `` is not a valid index uid. Index uid can be an integer or a string containing only alphanumeric characters, hyphens (-) and underscores (_).", code: InvalidIndexUid }

I do not have issues when running Meilisearch directly on the host machine (macOS). I believe this is a problem with building the database on a host machine, then copying it into the Docker image.

**To Reproduce**

See https://github.com/Eric-Arellano/meilisearch-internal-error for a script that automates this reproduction.

```Dockerfile
# Dockerfile
FROM getmeili/meilisearch:v1.9.0
COPY .out/search-index.ms ./data.ms/
```

1. `curl -L https://install.meilisearch.com | sh`
2. `./meilisearch --db-path .out/search-index.ms --no-analytics`
4. (new terminal) `curl -s -X POST http://127.0.0.1:7700/indexes -H 'Content-Type: application/json' --data-binary '{"uid": "movies", "primaryKey": "id"}'`
5. kill Meilisearch process
6. `docker build -t meilisearch-debug .`
7. `docker run -p 7700:7700 --init --rm meilisearch-debug`

At the time of writing this issue, the minimum reproduction was using 1.9.0. I also tested with 1.7.5. This bug does not reproduce with 1.3.5 and 1.6.2, so this seems to be a regression.

Things I tested and found it made no difference:
- using JS client vs `curl`
- uploading documents. Simply creating the index is sufficient to reproduce
- setting up an auth key

I have not yet tested out building the database with the Docker image, exporting the contents to host with `docker cp`, then re-running the Docker image with the saved contents

You have to create an index on the host machine. It works fine if you initialize Meilisearch on a host machine and copy over the database, without having created an index beforehand.

**Expected behavior**

I'm not sure if you intended to support this behavior? Based on https://www.meilisearch.com/docs/guides/misc/docker, I'm unclear if you expect all initial indexing to happen inside the Docker image rather than on a host machine.

It's useful to pre-build the database because it simplifies deployment of the Docker image. When we deploy an image, we know that the index will already be built without any delay, i.e. build-time vs run-time.

If this behavior is not expected, it would be helpful to document it and/or have a better runtime error.

**Meilisearch version:**
1.9.0

**Additional context**
I primarily tested on macOS Sonoma 14.5, Apple Silicon. My coworker also tested on macOS x86 and had the same issue.

I tried it in CI with Ubuntu 20.4 as the host machine. I don't have access to the Meilisearch logs, but we're getting a 500 error from the search integration so I believe that this same problem happens when using a Linux host.

dureuill:
# Pull Request

## Related issue
Fixes #4844 

## What does this PR do?

Given the following script: 
```
cargo run -- --db-path meili.ms
sleep 3
curl -s -X POST http://127.0.0.1:7700/indexes -H 'Content-Type: application/json' --data-binary '{"uid": "movies", "primaryKey": "id"}'
sleep 3
cargo run  -p meilisearch --db-path meili.ms
sleep 3
curl -s -X POST http://127.0.0.1:7700/indexes/movies/search -H 'Content-Type: application/json' --data-binary '{}'
```

- Before this PR, the final search returns a decoding error.
- After this PR, the search completes successfully

### Technical standpoint

This PR fixes two locations where the formatting of dates were dependent on the feature set of the `time` crate.

1. The `IndexStats` had two fields without the serialization format specified
2. More subtly, the index dates (`createdAt,` `updatedAt`) were using value remapping in the main DB to `SerdeJson<OffsetDateTime>`, which was using whatever default format was available. This was fixed by creating a local `OffsetDateTime` wrapper that would specify the serialization format

### DIFF
--- a/index-scheduler/src/index_mapper/mod.rs
+++ b/index-scheduler/src/index_mapper/mod.rs
@@ -108,8 +108,10 @@ pub struct IndexStats {
     /// Association of every field name with the number of times it occurs in the documents.
     pub field_distribution: FieldDistribution,
     /// Creation date of the index.
+    #[serde(with = "time::serde::rfc3339")]
     pub created_at: OffsetDateTime,
     /// Date of the last update of the index.
+    #[serde(with = "time::serde::rfc3339")]
     pub updated_at: OffsetDateTime,
 }
 
--- a/milli/src/index.rs
+++ b/milli/src/index.rs
@@ -9,7 +9,6 @@ use heed::{CompactionOption, Database, RoTxn, RwTxn, Unspecified};
 use roaring::RoaringBitmap;
 use rstar::RTree;
 use serde::{Deserialize, Serialize};
-use time::OffsetDateTime;
 
 use crate::documents::PrimaryKey;
 use crate::error::{InternalError, UserError};
@@ -173,8 +172,8 @@ impl Index {
     pub fn new_with_creation_dates<P: AsRef<Path>>(
         mut options: heed::EnvOpenOptions,
         path: P,
-        created_at: OffsetDateTime,
-        updated_at: OffsetDateTime,
+        created_at: time::OffsetDateTime,
+        updated_at: time::OffsetDateTime,
     ) -> Result<Index> {
         use db_name::*;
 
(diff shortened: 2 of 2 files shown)


irevoire (approved):
Awesome, thank you!

bors merge

PR #4853, opened 2024-08-01
Fix rhai deletion

### LINKED ISSUE #4849: Edit document by function do not delete all the selected documents
**Describe the bug**
While deleting documents with the experimental `POST /indexes/:index_uid/documents/edit` route some documents are not deleted.

**To Reproduce**
1. `for i in {0..100}; do echo "{ \"id\": $i }" | mieli d a --async; done`
2. `echo '{ "function": "if doc.id % 3 != 0 { doc = (); } else { doc.title = \"kefir\"; }"}' | http :7700/indexes/mieli/documents/edit`
3. `mieli search`

You should see something looking like that:
```json
{
  "hits": [
    {
      "id": 0,
      "title": "kefir"
    },
    {
      "id": 1
    },
    {
      "id": 3,
      "title": "kefir"
    },
    {
      "id": 4
    },
    {
      "id": 5
    },
    {
      "id": 6,
      "title": "kefir"
    },
    ...
}
```

All the documents without a `title` were supposed to be deleted, but that's not the case.

dureuill:
# Pull Request

## Related issue
Fixes #4849 

## What does this PR do?
- insert inside of the bitmap instead of pushing into it.

### DIFF
--- a/milli/src/update/index_documents/mod.rs
+++ b/milli/src/update/index_documents/mod.rs
@@ -290,7 +290,7 @@ where
 
                 match result? {
                     DocumentEdition::Deleted(docid) => {
-                        documents_to_remove.push(docid);
+                        documents_to_remove.insert(docid);
                     }
                     DocumentEdition::Edited(new_document) => {
                         documents_batch_builder.append_json_object(&new_document)?;


irevoire (approved):
Thanks
bors merge

dureuill:
Thanks to you for the initial issue + the collaborative debug session ❤️

milli/src/update/index_documents/mod.rs
```diff
@@ -290,7 +290,7 @@ where
 
                 match result? {
                     DocumentEdition::Deleted(docid) => {
-                        documents_to_remove.push(docid);
+                        documents_to_remove.insert(docid);
```
Kerollmops, line 293: MDR (as we say in french)

PR #4864, opened 2024-08-08
Don't remove facet value when multiple original values map to the same normalized value

### LINKED ISSUE #4860: Filter doesn't return valid documents
**Describe the bug**
"Complementary" issue to #4848 😢 

Some documents are not returned, despite having a value that the filter should react to.

**To Reproduce**
1. Create an index with primary key `id`
2. Add sortable attribute `foo`
3. Add a document `{ "id": 0, "foo": ["bar", "Bar"]}'
4. Check that `{"filter": "foo = bar"}` correctly returns the document.
5. Update the document `{"id": 0, "foo": "bar"}'
6. `{"filter": "foo = bar"}` no longer returns the document

**Expected behavior**
`{"filter": "foo = bar"}` still returns the document

**Screenshots**

**Meilisearch version:**
Tested in Meilisearch v1.9 and Meilisearch v1.10, but probably affects a larger range: since the differential indexing has been added to the time of writing this issue

**Additional context**
Discovered while investigating #4848.

The technical cause, which could also be an element of #4848, is that in `insert_strings_diff,` pairs of `(normalized_value, value)` are compared between the previous value of the field in the document and the latest value of the field in the document.

As the entire pair is compared lexicographically, but ultimately only one deladd obkv can be written per normalized value, what happens here is that we erroneously record that normalized("Bar")=normalized("bar") needs removal (despite there still being "bar" in the field).

Btw this also affects the facet search since only only exact value is kept among all normalized values, and if one value is added while a different value that normalize to the same thing is removed, it can be that we erroneously remove the facet value from the candidate search result.

dureuill:
# Pull Request

## Related issue

Fixes #4860 

> [!WARNING]  
> This PR contains a fix to the immediate issue, but it looks like the underlying data model is faulty: there is only one possible "original" value for each normalized value in a facet of a document, while because of array values (or manually written nested fields, if you're evil), it is technically possible to have multiple, distinct original values mapping to the same normalized value.

### DIFF
--- a/milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
+++ b/milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
@@ -317,11 +317,15 @@ fn deladd_obkv_cbo_roaring_bitmaps(
 }
 
 /// Truncates a string to the biggest valid LMDB key size.
-fn truncate_string(s: String) -> String {
-    s.char_indices()
-        .take_while(|(idx, _)| idx + 4 < MAX_FACET_VALUE_LENGTH)
-        .map(|(_, c)| c)
-        .collect()
+fn truncate_str(s: &str) -> &str {
+    let index = s
+        .char_indices()
+        .map(|(idx, _)| idx)
+        .chain(std::iter::once(s.len()))
+        .take_while(|idx| idx <= &MAX_FACET_VALUE_LENGTH)
+        .last();
+
+    &s[..index.unwrap_or(0)]
 }
 
 /// Computes the diff between both Del and Add numbers and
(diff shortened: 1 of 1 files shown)


ManyTheFish (approved):
bors merge

ManyTheFish:
Hey @dureuill,
It's ok to keep only one original value. We already do it when merging the values from different documents; this original value is used to display the facet in the facet distribution structure.

dureuill:
> this original value is used to display the facet in the facet distribution structure.

I find this potentially very confusing. It means that the documents with facet values "Maçon" and "Mâcon" will both be sent to "macon", and Meilisearch might display either "Maçon" or "Mâcon" in the facet distribution structure.

I guess I would prefer we keep all the list of original values.

PR #4867, opened 2024-08-08
Autogenerate the openAPI spec

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/5073

## What does this PR do?
- Introduce utoipa and the auto-generation of the openAPI file
- Introduce the scalar swagger when the `swagger` feature flag is enabled.

Generating the openAPI file takes between 15 and 20ms at startup time on my computer. That could be an issue if we plan to stabilize the feature.


crates/meilisearch-types/src/settings.rs
```diff
@@ -134,79 +144,142 @@ impl MergeWithError<milli::CriterionError> for DeserrJsonError<InvalidSettingsRa
     }
 }
 
+#[derive(Default, Serialize, Deserialize, PartialEq, Eq, Clone, ToSchema)]
+#[repr(transparent)]
+#[serde(transparent)]
+pub struct SettingEmbeddingSettings {
```
dureuill, line 150: can you elaborate as a response to this comment why the intermediate struct is needed?

crates/meilisearch-types/src/settings.rs
```diff
@@ -134,79 +144,142 @@ impl MergeWithError<milli::CriterionError> for DeserrJsonError<InvalidSettingsRa
...
     #[deserr(default, error = DeserrJsonError<InvalidSettingsFaceting>)]
+    #[schema(value_type = Option<FacetingSettings>, example = json!({ "maxValuesPerFacet": 10, "sortFacetValuesBy": { "genre": FacetValuesSort::Count }}))]
     pub faceting: Setting<FacetingSettings>,
+    /// Pagination settings.
     #[serde(default, skip_serializing_if = "Setting::is_not_set")]
     #[deserr(default, error = DeserrJsonError<InvalidSettingsPagination>)]
+    #[schema(value_type = Option<PaginationSettings>, example = json!({ "maxValuesPerFacet": 10, "sortFacetValuesBy": { "genre": FacetValuesSort::Count }}))]
     pub pagination: Setting<PaginationSettings>,
 
+    /// Embedder required for performing meaning-based search queries.
```
dureuill, line 262:
```suggestion
    /// Embedder required for performing semantic search queries.
```

crates/meilisearch/src/search/mod.rs
```diff
@@ -648,15 +663,17 @@ pub struct SearchHit {
...
 #[serde(rename_all = "camelCase")]
+#[schema(rename_all = "camelCase")]
 pub struct SearchResult {
     pub hits: Vec<SearchHit>,
     pub query: String,
     pub processing_time_ms: u128,
     #[serde(flatten)]
     pub hits_info: HitsInfo,
     #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(value_type = HashMap<String, Value>)]
```
dureuill, line 676: small question, but other types have BTreeMap and this one uses HashMap, is there any specific reason why?

Also should we put the option here, or is the "optionality" of the field already handled by the actual type being an option?

crates/meilisearch/src/search/mod.rs
```diff
@@ -1021,8 +1041,9 @@ pub fn perform_search(
     Ok(result)
 }
 
-#[derive(Debug, Clone, Default, Serialize)]
+#[derive(Debug, Clone, Default, Serialize, ToSchema)]
 pub struct ComputedFacets {
+    #[schema(value_type = Option<BTreeMap<String, BTreeMap<String, u64>>>)]
```
dureuill, line 1046: I'm not sure if I get why the type here is an option now 😅

dureuill (approved):
I have a few minor questions (see inline comments above), but a big thank you for leading to completion this huge undertaking 🫶 

I love this new era of self-documenting Meilisearch :heart:

I'm missing a change to the CONTRIBUTORS file and to the PR template to remind us all to update any changed route or type and guide us on how to do this, can you open a new issue so we don't forget to do that in the future? Then I think we're good to land.

dureuill (approved):

irevoire:
Thanks
bors merge

PR #4899, opened 2024-08-28
stop trying to process searches after one minute

irevoire:
# Pull Request

## Related issue
May be related to #4654 and https://github.com/meilisearch/meilisearch-support/issues/350

## What does this PR do?
- If we've been waiting for one whole minute for a search to process, we cancel it
- Ideally we should check if the connection was closed instead but that’s not possible currently: https://github.com/actix/actix-web/issues/3462


ManyTheFish (approved):
LGTM

bors merge

ManyTheFish:
@irevoire, tests seem to fail

PR #4901, opened 2024-08-29
Autobatch document deletion by filter

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4897

## What does this PR do?
- Enable autobatching of document deletion by filter with:
  - Document deletion by filter
  - Document deletion
  - Document clear
  - Index deletion

PR #4905, opened 2024-09-02
Do not fail the whole batch when a single document deletion by filter fails

irevoire:
# Pull Request

## Related issue
Fixes a small bug introduced by https://github.com/meilisearch/meilisearch/pull/4901 where a document deletion by filter could fail a whole batch of document deletion task.

## What does this PR do?
- When a document deletion by filter contains an invalid filter, only fails this task instead of the whole batch
- Adds a big test with multiple document deletions batched together ensuring everything works well


dureuill (dismissed):
bors merge

dureuill (approved):
bors merge

PR #4912, opened 2024-09-05
Allow Meilitool to dumplessly, offline upgrade v1.9 -> v1.10 in some conditions

dureuill:
- bail early if the DB contains at least 1 REST embedder, providing the list of detected REST embedders, and without modifying the DB
- Might depend on the feature set that meilitool was compiled with and the featureset that the Meilisearch that created the DB was compiled with 💀. In case of runtime error, try again with a different feature set (passing or not passing `-p meilitool` when building after a `cargo clean`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3520,6 +3520,7 @@ dependencies = [
  "file-store",
  "meilisearch-auth",
  "meilisearch-types",
+ "serde",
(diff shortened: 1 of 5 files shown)


Kerollmops (approved):
That is very nice, thank you again!
bors merge

PR #4928, opened 2024-09-12
Make matches consider phrases as a single `Match`

### LINKED ISSUE #4732: Confusing matches in `_matchesPosition` when using phrase search
> **This issue is open to contribution, for more information see the below comment:**
> https://github.com/meilisearch/meilisearch/issues/4732#issuecomment-2202282648

**Describe the bug**
Given this document:

```json
{
    "id": "123",
    "text": "Lorem ipsum dolor sit amet"
}
```

Searching for `"Lorem ipsum"` (phrase search!) and setting `showMatchesPosition: true`, returns this:

```json
{
  "id": "123",
  "text": "Lorem ipsum dolor sit amet",
  "_matchesPosition": {
    "text": [
      { "start": 0, "length": 5 },
      { "start": 6, "length": 5 }
    ]
  }
}
```

This is unexpected for phrase search, I think. I would have expected one match range that encloses all of "Lorum ipsum". 

**To Reproduce**
1. `curl -H 'Content-Type: application/json' 'localhost:7700/indexes' --data '{"uid":"test","primaryKey":"id"}'`
2. `curl -H 'Content-Type: application/json' 'localhost:7700/indexes/test/documents' --data '[{"id":"123","text":"Lorem ipsum dolor sit amet"}]'`
3. `curl -H 'Content-Type: application/json' 'localhost:7700/indexes/test/search' --data '{"q":"\"Lorem Ipsum\"","showMatchesPosition":true}'`

**Expected behavior**
Getting one match position back that encloses the full phrase: `{ start: 0, length: 11 }`

**Meilisearch version:**
v1.8

flevi29:
# Pull Request

## Related issue
Fixes #4732

## PR checklist


flevi29:
I have found that when we want to crop and have a single phrase match, the first match gets chosen every time as the best interval in the current code. This is because their ids will always match but the distance score will always be worse as we widen our match interval. This means that we're measuring words only after and before the first word of a phrase match. Is this intentional?

If so, it's a little tricky to make it work the same way, as we only have one match when it comes to a phrase, so it won't be cut the same way.

https://github.com/meilisearch/meilisearch/blob/02c2b660f883403a4149aea09eff6529c50a18d4/milli/src/search/new/matches/mod.rs#L354-L380

milli/src/search/new/matches/mod.rs
```diff
@@ -821,30 +880,32 @@ mod tests {
...
         let format_options = FormatOptions { highlight: true, crop: Some(10) };
-        let text = "The groundbreaking invention had the power to split the world between those who embraced progress and those who resisted change!";
 
         let builder = MatcherBuilder::new_test(&rtxn, &temp_index, "\"the world\"");
         let mut matcher = builder.build(text, None);
         // should return 10 words with a marker at the start as well the end, and the highlighted matches.
         insta::assert_snapshot!(
             matcher.format(format_options),
-            @"…had the power to split <em>the</em> <em>world</em> between those who…"
+            @"…had the power to split <em>the world</em> between those who…"
```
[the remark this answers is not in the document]
ManyTheFish: Well, it's weird, but their are still 10 words and the match is centered in the crop window so I feel this behavior is better 🤔

The question is, why the previous behavior was wrong in some ways, maybe the center of the window were computed from the first match 🤔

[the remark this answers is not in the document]
ManyTheFish: The crop algorithm start from the match, then, using a bidirectional iterator, adds a word switching side growing the window until the word count is reached, like:
```
         <em>the world</em>
   SPLIT <em>the world</em>
   split <em>the world</em> BETWEEN
TO split <em>the world</em> between
to split <em>the world</em> between THOSE
```

unless we encounter an hard-separator like `.` or `;` in this case we will prefer add a word in the same "sentence" to have a contextualized crop window.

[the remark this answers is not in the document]
flevi29: > The question is, why the previous behavior was wrong in some ways, maybe the center of the window were computed from the first match 🤔

Yeah, I think as I detailed [here](https://github.com/meilisearch/meilisearch/pull/4928#issuecomment-2350899477), somewhat poorly worded, it's because it counts from the first match of a phrase, while my solution counts from the beginning and the end of the phrase the positions. I'm not entirely sure yet that everything will work as it should, I might want to add more tests, but for now I'm pushing the changes with the tests adjusted.

milli/src/search/new/matches/mod.rs
```diff
@@ -821,30 +902,68 @@ mod tests {
...
+            &rtxn,
+            &temp_index,
+            "\"The groundbreaking invention\" \"embraced progress and those who resisted change\"",
+        );
+        let mut matcher = builder.build(text, None);
+        // should highlight "those" and the phrase "and those".
+        insta::assert_snapshot!(
+            matcher.format(format_options),
+            @"…between those who <em>embraced progress and those who resisted change</em>…"
         );
```
flevi29, line 967: Seems like it prioritizes the last match when cropping. Is this fine?

ManyTheFish: I think in this case their is no more words to highlight after the the match, so we take the words before instead

ManyTheFish: We could incorporate the separator in the crop window making clearer that we are at the end of a sentence. 🤔 
But lets not change this behavior in this PR 😄

ManyTheFish (approved):
I think we can merge this into main, @flevi29 do you want to change anything?

flevi29:
I'm still working on it. There are some bugs as well, for instance if the match is greater than the crop length it will panic.

milli/src/search/new/matches/mod.rs
```diff
@@ -258,59 +451,107 @@ impl<'t, 'tokenizer> Matcher<'t, 'tokenizer, '_, '_> {
...
+            Either::Left(tokens[after_tokens_starting_index..].iter().peekable())
+        } else {
+            // ... starting from the last match token position and going towards the start of the text.
+            Either::Right(tokens[..=after_tokens_starting_index].iter().rev().peekable())
+        };
 
         // grows the crop window peeking in both directions
         // until the window contains the good number of words:
         while remaining_words > 0 {
-            let before_token = before_tokens.peek().map(|t| t.separator_kind());
```
flevi29, line 278: image
What is `TokenKind::Unknown`? We are also counting that as "words", is this intentional?

milli/src/search/new/matches/mod.rs
```diff
@@ -821,30 +778,90 @@ mod tests {
...
+        let builder = MatcherBuilder::new_test(
+            &rtxn,
+            &temp_index,
+            "\"The groundbreaking invention had the power to split the world between those\"",
+        );
+        let mut matcher = builder.build(text, None);
+        insta::assert_snapshot!(
+            matcher.format(format_options),
+            // @TODO: Should probably highlight it all, even if it didn't fit the whole phrase
+            @"The groundbreaking invention had the power to split the world…"
```
flevi29, line 829: Should I make it so that phrases that didn't fit the crop size should be partially highlighted?

ManyTheFish: Yes I think it's the best solution, it's not a big deal if the window is too small.

flevi29: Alright, going to do it this week.

milli/src/search/new/matches/matching_words.rs
```diff
@@ -149,7 +152,7 @@ pub type WordId = u16;
 /// In these cases we need to match consecutively several tokens to consider that the match is full.
 #[derive(Debug, PartialEq)]
 pub enum MatchType<'a> {
-    Full { char_len: usize, ids: &'a RangeInclusive<WordId> },
+    Full { char_count: usize, byte_len: usize, ids: &'a RangeInclusive<WordId> },
```
flevi29, line 155: I added `byte_len` here so I can use it on prefix matches, but it seems that it doesn't work as it should, `token.byte_end - token.byte_start` still returns the normalized lemma length, at least it seemed so when I was testing it. That seems like a bug to me.

milli/src/search/new/matches/mod.rs
```diff
@@ -434,69 +381,87 @@ impl<'t, 'tokenizer> Matcher<'t, 'tokenizer, '_, '_> {
...
-                            formatted.push(&self.text[token.byte_start..highlight_byte_index]);
+
+                            // TODO: This is additional work done, charabia::token::Token byte_len
+                            // should already get us the original byte length, however, that doesn't work as
+                            // it's supposed to, investigate why
+                            let highlight_byte_index = self.text[*m_byte_start..]
+                                .char_indices()
+                                .nth(m.char_count)
+                                .map_or(*m_byte_end, |(i, _)| min(i + *m_byte_start, *m_byte_end));
+                            formatted.push(&self.text[*m_byte_start..highlight_byte_index]);
```
flevi29, line 439: https://github.com/meilisearch/meilisearch/pull/4928/files#r1807316376

milli/src/search/new/matches/matching_words.rs
ManyTheFish: Indeed yes, I will check in charabia if a normalizer didn't change the lemma content 🤔

ManyTheFish:
Hey @flevi29, do you want to change anything or can I review?
About Charabia's bug, I'll investigate on my side and fix it in another PR. :)

flevi29:
@ManyTheFish Hello! I feel like I'm finished with this PR. Let me know if anything needs to be changed.

ManyTheFish (approved):
LGTM

bors merge

thank you for your contribution! :D

flevi29:
@ManyTheFish I have merged with main, there were conflicts.

ManyTheFish (approved):
bors merge

PR #4929, opened 2024-09-12
Add facets support to federated

### LINKED ISSUE #4913: `multi-search` with federation not returning facets
I am using `multi-search` endpoint and having issue with getting facets when I use `federation` to combine the results. 

**To Reproduce**
1. Create a colleciton with filterableAttributes as usual
2. Try search with `multi-search` endpoint with `facets` in queries and use `federation` to get a combined result set for the queries
3. It does not return facets info in the response
4. Now, if we remove the `federation` key from request, it will return `facets` data in every query response object 

**Expected behavior**
It should be returning facets data combined together from both/all of the queries

**Meilisearch version:**
[e.g. v1.10.1]

### LINKED ISSUE #4932: Support facet distribution for federated search
Related product team resources: [PRD](https://www.notion.so/meilisearch/Federated-Search-PRD-07ff4017f65f4a42b577a7925e94e92d) (_internal only_)
Related product discussion: None, but see [open issue](https://github.com/meilisearch/meilisearch/issues/4913)

## Motivation & Usage

[Public usage](https://meilisearch.notion.site/v1-11-Federated-search-59b30e03383c40729d7541a3dffb0069)

## Impacted teams

- @meilisearch/integration-team: add new settings to federated search
- @meilisearch/docs-team: add new settings to federated search

dureuill:
# Pull Request

## Related issue 

- Fixes #4932 (sprint issue)
- Fixes  #4913 (user-opened issue)

## What does this PR do?

See [public usage](https://meilisearch.notion.site/v1-11-Federated-search-59b30e03383c40729d7541a3dffb0069)

> [!CAUTION]
> This PR introduces a 🚨**breaking change**🚨: `queries.facets` when `federation` is present and non-`null` is now **an error**

### Implementation standpoint:

- Facet distribution: fix issue where truncated facet distribution would have a wrong order
- facet distribution: implement Display for OrderBy


meilisearch/src/search/federated.rs
```diff
@@ -313,16 +347,106 @@ struct SearchHitByIndex {
...
+                    std::collections::btree_map::Entry::Vacant(entry) => {
+                        entry.insert(index_stats);
+                    }
+                    std::collections::btree_map::Entry::Occupied(mut entry) => {
+                        let stats = entry.get_mut();
+
+                        stats.min =
+                            if stats.min <= index_stats.min { stats.min } else { index_stats.min };
+                        stats.max =
+                            if stats.max >= index_stats.max { stats.max } else { index_stats.max };
```
Kerollmops, line 414: You should [use `f32::min/max` instead](https://doc.rust-lang.org/stable/std/primitive.f32.html#method.min).

Kerollmops (changes requested):

meilisearch/src/search/federated.rs
dureuill: done

PR #4938, opened 2024-09-17
Remove default embedder

### LINKED ISSUE #4738: Always require `embedder` in semantic search and similar search
# Motivation & Usage

[See public usage](https://meilisearch.notion.site/v1-11-AI-search-changes-0e37727193884a70999f254fa953ce6e#1044b06b651f80edb9d4ef6dc367bad0)

# Rejected solutions

1. Explicit `default` boolean in the settings. Allows to explicitly points which embedder is the default embedder with a boolean set to true in the settings of that embedder.
    - pros: 
      - explicit and not surprising
    - cons:
      - must handle the case where users put `default` to multiple embedders simultaneously
      - somewhat complex implementation
      - what should the behavior be when there is a single embedder? Defaults to `default: true`?
2. Suppress the special role of the embedder named "default". Request `embedder`, to disambiguate which embedder is used, whenever there are multiple embedders. 
    - pros:
      - less surprising
    - cons:
      - samples often use "default" as a name for the embedder, it could be surprising that the "default"-named embedder is not the default.
4. ~~Always requires `embedder` to be specified~~ (chosen solution above)
    - pros:
      - simplest option, no default embedder at all
    - cons:
      - forces the user to add a parameter even if its value is trivial
5. Don't change the behavior, address the confusing error message with documentation and clearer error message
    - pros:
      - almost no implementation work
      - keeps the magic
    - cons:
      - keeps the magic
6. Do nothing
    - pros:
      - nothing to do
      - we got no user feedback that this required change, only internal feedback and occasionally, support request from confused users
    - cons:
      - keeps the magic 
      - this is our last chance to correct this behavior if it needs correction

dureuill:
# Pull Request

## Related issue
Fixes #4738 

## What does this PR do?

[See public usage](https://meilisearch.notion.site/v1-11-AI-search-changes-0e37727193884a70999f254fa953ce6e#1044b06b651f80edb9d4ef6dc367bad0)

- Remove `hybrid.embedder` boolean from analytics because embedder is now mandatory and so the boolean would always be `true`
- Rework search kind so that a search without query but with vector is a vector search regardless of (non-zero) semantic ratio

### DIFF
--- a/meilisearch/src/analytics/segment_analytics.rs
+++ b/meilisearch/src/analytics/segment_analytics.rs
@@ -646,8 +646,6 @@ pub struct SearchAggregator {
     max_vector_size: usize,
     // Whether the semantic ratio passed to a hybrid search equals the default ratio.
     semantic_ratio: bool,
-    // Whether a non-default embedder was specified
-    embedder: bool,
     hybrid: bool,
     retrieve_vectors: bool,
 
@@ -795,7 +793,6 @@ impl SearchAggregator {
 
         if let Some(hybrid) = hybrid {
             ret.semantic_ratio = hybrid.semantic_ratio != DEFAULT_SEMANTIC_RATIO();
-            ret.embedder = hybrid.embedder.is_some();
             ret.hybrid = true;
         }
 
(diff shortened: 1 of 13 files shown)


ManyTheFish (approved):

ManyTheFish:
bors merge

PR #4940, opened 2024-09-17
Make keys `Action` use bitflags crate

### LINKED ISSUE #4645: Change key actions to be bitflags
Currently, the [Meilisearch action keys are single numbers](https://github.com/meilisearch/meilisearch/blob/abe29772dbebbfe5251cfeece2addd4133d3b93b/meilisearch-types/src/keys.rs) and therefore cannot be combined. It's not possible to OR them NOT them. I needed to do a basic `documents.add | documents.delete` for the edit documents by using a function feature.

We must [use bitflags to define those actions](https://lib.rs/crates/bitflags) and be able to combine them. We could have 32 or 64 different combinable actions. We already use this library in heed and it works very well.

flevi29:
# Pull Request

## Related issue
Fixes #4645

## PR checklist

### DIFF
--- a/meilisearch-types/src/keys.rs
+++ b/meilisearch-types/src/keys.rs
@@ -222,7 +222,8 @@ bitflags! {
 }
 
 impl Action {
-    const SERIALIZATION_MAP: [(&'static str, Self); 34] = [
+    // @TODO: Consider using https://github.com/rust-phf/rust-phf
+    const SERDE_MAP_ARR: [(&'static str, Self); 34] = [
         ("*", Self::All),
         ("search", Self::Search),
         ("documents.*", Self::DocumentsAll),
@@ -311,7 +312,7 @@ impl<E: DeserializeError> Deserr<E> for Action {
     ) -> Result<Self, E> {
         match value {
             deserr::Value::String(s) => {
-                match Self::SERIALIZATION_MAP.iter().find(|(serialized, _)| s == *serialized) {
+                match Self::SERDE_MAP_ARR.iter().find(|(serialized, _)| s == *serialized) {
                     Some((_, action)) => Ok(*action),
                     None => Err(deserr::take_cf_content(E::error::<std::convert::Infallible>(
                         None,
@@ -332,6 +333,20 @@ impl<E: DeserializeError> Deserr<E> for Action {
     }
 }
 
+impl Serialize for Action {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        Self::SERDE_MAP_ARR
+            .iter()
+            .find(|(_, action)| self == action)
+            .map(|(serialized, _)| serializer.serialize_str(serialized))
+            // should always be found, so unwrap is safe to use
+            .unwrap()
+    }
+}
+
 impl<'de> Deserialize<'de> for Action {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
@@ -349,9 +364,8 @@ impl<'de> Deserialize<'de> for Action {
             where
                 E: serde::de::Error,
             {
-                // @TODO: Make a to_serialized and to_desiralized on Action
-                match Self::Value::SERIALIZATION_MAP.iter().find(|(serialized, _)| s == *serialized)
-                {
+                // @TODO: Make a to_serialized and to_desiralized on Action perhaps
+                match Self::Value::SERDE_MAP_ARR.iter().find(|(serialized, _)| s == *serialized) {
                     Some((_, action)) => Ok(*action),
                     None => Err(E::invalid_value(serde::de::Unexpected::Str(s), &"a valid action")),
                 }
@@ -362,17 +376,32 @@ impl<'de> Deserialize<'de> for Action {
     }
 }
 
-impl Serialize for Action {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        Self::SERIALIZATION_MAP
-            .iter()
-            .find(|(_, action)| self == action)
-            .map(|(serialized, _)| serializer.serialize_str(serialized))
-            // should always be found, so unwrap is safe to use
-            .unwrap()
+impl Sequence for Action {
+    const CARDINALITY: usize = Self::SERDE_MAP_ARR.len();
+
+    fn next(&self) -> Option<Self> {
+        let next_index = self.bits() as usize + 1;
+        if next_index == Self::CARDINALITY {
+            None
+        } else {
+            Some(Self::SERDE_MAP_ARR[next_index].1)
+        }
+    }
+
+    fn previous(&self) -> Option<Self> {
+        if self.bits() == 0 {
+            None
+        } else {
+            Some(Self::SERDE_MAP_ARR[self.bits() as usize - 1].1)
+        }
+    }
+
+    fn first() -> Option<Self> {
+        Some(Self::SERDE_MAP_ARR[0].1)
+    }
+
+    fn last() -> Option<Self> {
+        Some(Self::SERDE_MAP_ARR[Self::CARDINALITY - 1].1)
     }
 }
 
--- a/meilisearch-types/src/keys.rs
+++ b/meilisearch-types/src/keys.rs
@@ -4,8 +4,9 @@ use std::str::FromStr;
 
 use bitflags::bitflags;
 use deserr::{take_cf_content, DeserializeError, Deserr, MergeWithError, ValuePointerRef};
+use enum_iterator::Sequence;
 use milli::update::Setting;
-use serde::{Deserialize, Deserializer, Serialize};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use time::format_description::well_known::Rfc3339;
 use time::macros::{format_description, time};
 use time::{Date, OffsetDateTime, PrimitiveDateTime};
@@ -180,7 +181,7 @@ fn parse_expiration_date(
 }
 
 bitflags! {
-    #[derive(Copy, Clone, Serialize, Debug, Eq, PartialEq, Hash, PartialOrd, Ord)]
+    #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord)]
     #[repr(transparent)]
     pub struct Action: u8 {
     const All = 0;
@@ -358,6 +359,20 @@ impl<'de> Deserialize<'de> for Action {
         }
 
         deserializer.deserialize_str(Visitor)
+    }
+}
+
+impl Serialize for Action {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        Self::SERIALIZATION_MAP
+            .iter()
+            .find(|(_, action)| self == action)
+            .map(|(serialized, _)| serializer.serialize_str(serialized))
+            // should always be found, so unwrap is safe to use
+            .unwrap()
     }
 }
 
--- a/meilisearch-types/src/keys.rs
+++ b/meilisearch-types/src/keys.rs
@@ -4,9 +4,8 @@ use std::str::FromStr;
 
 use bitflags::bitflags;
 use deserr::{take_cf_content, DeserializeError, Deserr, MergeWithError, ValuePointerRef};
-use enum_iterator::Sequence;
 use milli::update::Setting;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use time::format_description::well_known::Rfc3339;
 use time::macros::{format_description, time};
 use time::{Date, OffsetDateTime, PrimitiveDateTime};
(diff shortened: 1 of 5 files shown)


meilisearch-types/src/keys.rs
```diff
@@ -179,114 +180,101 @@ fn parse_expiration_date(
...
+            .map(|(_, action)| *action)
+    }
+
+    fn action_to_ser_action(v: &Action) -> &'static str {
+        Self::SERDE_MAP_ARR
+            .iter()
+            .find(|(_, ref action)| v == action)
+            .map(|(ser_action, _)| ser_action)
+            .expect("`action_wanted` should always have a matching serialized value")
+    }
```
flevi29, line 276: I'm not a big fan of these, how could I improve it? Maybe use https://github.com/rust-phf/rust-phf ?

meilisearch-types/src/keys.rs
```diff
@@ -328,44 +316,138 @@ impl Action {
...
+
+    fn first() -> Option<Self> {
+        Some(Self::SERDE_MAP_ARR[0].1)
+    }
+
+    fn last() -> Option<Self> {
+        Some(Self::SERDE_MAP_ARR[Self::CARDINALITY - 1].1)
     }
 }
```
flevi29, line 415: I'm also not liking this very much, how could I improve it?

meilisearch-types/src/keys.rs
Kerollmops: I am pretty sure that it's ok. You are comparing at max 33 values in this array. Yes, those are strings but it's fast enough.

meilisearch-types/src/keys.rs
```diff
@@ -179,114 +180,101 @@ fn parse_expiration_date(
...
+    ];
+
+    fn ser_action_to_action(v: &str) -> Option<Action> {
+        Self::SERDE_MAP_ARR
+            .iter()
+            .find(|(ser_action, _)| &v == ser_action)
+            .map(|(_, action)| *action)
+    }
+
+    fn action_to_ser_action(v: &Action) -> &'static str {
```
Kerollmops, line 270: Can you rename those functions with more explicit and complete word names? The following is maybe sufficient. What do you think?

```suggestion
    fn get_action(v: &str) -> Option<Action> {
        Self::SERDE_MAP_ARR
            .iter()
            .find(|(ser_action, _)| &v == ser_action)
            .map(|(_, action)| *action)
    }

    fn get_action_serde_name(v: &Action) -> &'static str {
```

milli/src/vector/mod.rs
```diff
@@ -352,7 +352,7 @@ where
                 None,
                 deserr::ErrorKind::Unexpected {
                     msg: format!(
-                        "the distribution sigma must be in the range ]0, 1], got {}",
+                        "the distribution sigma must be in the range [0, 1], got {}",
```
Kerollmops, line 355: Why did you changed this?

```suggestion
                        "the distribution sigma must be in the range ]0, 1], got {}",
```

meilisearch-types/src/keys.rs
Kerollmops: What are you talking about? This comment is floating around for me 😂

Kerollmops (changes requested):
Hey @flevi29 👋

Thank you VERY MUCH for your work. It seems very good already.

The point of converting Actions to bitflags was mostly to be able to be able to change the following `actions::DOCUMENTS_ALL` to something like `actions::DOCUMENTS_ADD | actions::DOCUMENTS_DELETE`. So maybe you can take that in consideration. But don't change it in this PR. Just try with your changes if possible.

https://github.com/meilisearch/meilisearch/blob/ff523a2357767f5c64296524a01b0ec35bdd33ca/meilisearch/src/routes/indexes/documents.rs#L602-L604

milli/src/vector/mod.rs
flevi29: Oh, didn't quite understand what's going on there. But now I think I do, reverting it.

meilisearch-types/src/keys.rs
flevi29: Well, I'm just not entirely sure if this is the best way to implement `Sequence` for this. Thought I'd get some input.

flevi29:
> The point of converting Actions to bitflags was mostly to be able to be able to change the following `actions::DOCUMENTS_ALL` to something like `actions::DOCUMENTS_ADD | actions::DOCUMENTS_DELETE`. So maybe you can take that in consideration. But don't change it in this PR. Just try with your changes if possible.

I'm not sure I understand, should I leave things as is for now and in a separate PR make use of bitflags where possible, get rid of "all" flags?

Kerollmops:
> I'm not sure I understand, should I leave things as is for now and in a separate PR make use of bitflags where possible, get rid of "all" flags?

No, no. Continue your PR, but ultimately, we would like to be able to use OR (`|`) operands on action rights on some routes. So, just try to make it possible with the bitflags rewrite 😃

flevi29:
Uh oh.
image

@Kerollmops I'm not sure how to fix this, I'll just push the code as is for now so you can see my progress.

meilisearch-types/src/keys.rs
```diff
@@ -179,142 +180,149 @@ fn parse_expiration_date(
...
+            TASKS_ALL => Some(Self::TasksAll),
             SETTINGS_GET => Some(Self::SettingsGet),
             SETTINGS_UPDATE => Some(Self::SettingsUpdate),
-            STATS_ALL => Some(Self::StatsAll),
+            SETTINGS_ALL => Some(Self::SettingsAll),
             STATS_GET => Some(Self::StatsGet),
-            METRICS_ALL => Some(Self::MetricsAll),
+            // @TODO: Issue: Since stats has only one element, all is the same as the one single element
+            //               so this will never match all, because it matches that one and only element first
+            STATS_ALL => Some(Self::StatsAll),
```
flevi29, line 319: I'm not sure if this will results in problems.

Kerollmops: Maybe you'll have to replace entirely this art 🤔

Kerollmops:
> I'm not sure how to fix this

The `BytesDecode::bytes_decode` trait takes a slice of bytes and returns a real type, in this case, `Flags`.

flevi29:
https://github.com/meilisearch/meilisearch/actions/runs/11428773944/job/31794407562
This action is stuck, there should be a reasonable timeout, but there doesn't seem to be.
I couldn't figure out why it's getting stuck yet on `error_json_bad_content_type`.

meilisearch-types/src/keys.rs
```diff
@@ -179,193 +180,271 @@ fn parse_expiration_date(
...
+
+        deserializer.deserialize_str(Visitor)
+    }
+}
+
+impl Sequence for Action {
+    const CARDINALITY: usize = Self::SERDE_MAP_ARR.len();
+
+    fn next(&self) -> Option<Self> {
+        let mut iter = Self::FLAGS.iter();
```
flevi29, line 403: Possibly a linked list solution would be better implemented on private properties of `self`, but it's probably not worth it.

meilisearch-types/src/keys.rs
```diff
@@ -179,193 +180,256 @@ fn parse_expiration_date(
...
-    DumpsAll,
-    #[serde(rename = "dumps.create")]
-    #[deserr(rename = "dumps.create")]
-    DumpsCreate,
-    #[serde(rename = "snapshots.*")]
-    #[deserr(rename = "snapshots.*")]
-    SnapshotsAll,
-    #[serde(rename = "snapshots.create")]
-    #[deserr(rename = "snapshots.create")]
-    SnapshotsCreate,
```
flevi29, line 265: I got rid of "all" flags that were only referring to a single flag. This caused an issue with implementing `Sequence`, as we couldn't differentiate between an "all" flag and its single referred element, and so we ended up in an infinite loop.

dump/src/reader/compat/v5_to_v6.rs
```diff
@@ -405,11 +405,11 @@ impl From<v5::Action> for v6::Action {
             v5::Action::SettingsAll => v6::Action::SettingsAll,
             v5::Action::SettingsGet => v6::Action::SettingsGet,
             v5::Action::SettingsUpdate => v6::Action::SettingsUpdate,
-            v5::Action::StatsAll => v6::Action::StatsAll,
+            v5::Action::StatsAll => v6::Action::StatsGet,
             v5::Action::StatsGet => v6::Action::StatsGet,
-            v5::Action::MetricsAll => v6::Action::MetricsAll,
+            v5::Action::MetricsAll => v6::Action::MetricsGet,
             v5::Action::MetricsGet => v6::Action::MetricsGet,
-            v5::Action::DumpsAll => v6::Action::DumpsAll,
```
flevi29, line 412: @Kerollmops I'm not sure whether this will cause issues, we'd need a v6 to v7 if we do this?

Kerollmops: Hum... You are probably right. However, if the actions can be converted to string it's ok I think. @ManyTheFish told me that in the dumps everything was converted to strings.

ManyTheFish: yes, but it may be possible to create a dump v6 containing `v6::Action::MetricsAll` or `v6::Action::StatsAll`, which is removed in this branch; it may crash, no?

flevi29: Yeah, I think so too. If I'm reading it right, currently it uses `serde_json` to write the dumps, and with `metrics.*` and the others removed, if we try and import an older dump it should crash.

There are two options:
1. Create new dump version
2. Somehow make `Sequence` impl work without removing any flags

I'll try to make 2. work.

meilisearch-types/src/keys.rs
flevi29: As long as `Action` flags are ordered ascending by value manually, we're more efficient the current way.

meilisearch-types/src/keys.rs
flevi29: It does, read more below.

curquiza:
@flevi29 can you fix the tests before we review your PR?

crates/meilisearch-types/src/keys.rs
```diff
@@ -195,206 +196,294 @@ fn parse_expiration_date(
...
+        const KeysUpdate = 1 << 21;
+        const KeysDelete = 1 << 22;
+        // Experimental Features
+        const ExperimentalFeaturesGet = 1 << 23;
+        const ExperimentalFeaturesUpdate = 1 << 24;
+        // Network
+        const NetworkGet = 1 << 25;
+        const NetworkUpdate = 1 << 26;
+        // All
+        const All = 0xFFFFFFFF >> (32 - 1 - 26);
```
Kerollmops, line 253: Shouldn't we remove this variant and use [this `all` method](https://docs.rs/bitflags/latest/bitflags/trait.Flags.html#method.all), instead?

Kerollmops: Because I don't quite like having to track `26` by hand 😞

crates/meilisearch-types/src/keys.rs
```diff
@@ -195,206 +196,294 @@ fn parse_expiration_date(
...
+                }
+            }
+        }
+
+        deserializer.deserialize_str(Visitor)
+    }
+}
+
+// TODO: Once "all" type flags are removed, simplify
+//       Essentially `get_potential_index` will give the exact index, +1 the exact next, -1 the exact previous
```
Kerollmops, line 434: What is this Todo?

crates/meilisearch-types/src/keys.rs
```diff
@@ -195,206 +196,294 @@ fn parse_expiration_date(
...
+
+                potential_next_index += 1;
+            } else {
+                return None;
+            }
+        }
+    }
+
+    fn previous(&self) -> Option<Self> {
+        // -2 because of "all" type flags that represent a single flag, otherwise -1 would suffice
```
Kerollmops, line 455: By removing the `ALL` flag and using the `Flags::all` method we can get rid of this conplexity.

crates/meilisearch/tests/auth/authorization.rs
```diff
@@ -53,14 +53,14 @@ pub static AUTHORIZATIONS: Lazy<HashMap<(&'static str, &'static str), HashSet<&'
...
+            ("GET",     "/stats") =>                                           hashset!{"stats.get", "*"},
+            ("POST",    "/dumps") =>                                           hashset!{"dumps.create", "*"},
+            ("POST",    "/snapshots") =>                                       hashset!{"snapshots.create", "*"},
             ("GET",     "/version") =>                                         hashset!{"version", "*"},
-            ("GET",     "/metrics") =>                                         hashset!{"metrics.get", "metrics.*", "*"},
-            ("POST",    "/logs/stream") =>                                     hashset!{"metrics.get", "metrics.*", "*"},
-            ("DELETE",  "/logs/stream") =>                                     hashset!{"metrics.get", "metrics.*", "*"},
+            ("GET",     "/metrics") =>                                         hashset!{"metrics.get", "*"},
+            ("POST",    "/logs/stream") =>                                     hashset!{"metrics.get", "*"},
+            ("DELETE",  "/logs/stream") =>                                     hashset!{"metrics.get", "*"},
```
Kerollmops, line 63: What does that change means? We cannot use the route with `stats.*` anymore?

Kerollmops: Hum... I think I understand, the `stats.*`, `dumps.*`, `snapshots.*` and `metrics.*` no longer exists as there is only a single variant for each of them. But is it something we want? Don't we want to keep them for backward compatibility?

crates/meilisearch/src/extractors/authentication/mod.rs
```diff
@@ -214,14 +214,14 @@ pub mod policies {
...
+    fn is_keys_action(action: u32) -> bool {
         use actions::*;
         matches!(action, KEYS_GET | KEYS_CREATE | KEYS_UPDATE | KEYS_DELETE)
     }
 
-    pub struct ActionPolicy<const A: u8>;
+    pub struct ActionPolicy<const A: u32>;
 
-    impl<const A: u8> Policy for ActionPolicy<A> {
+    impl<const A: u32> Policy for ActionPolicy<A> {
```
Kerollmops, line 224: Wouldn't it be better to be explicit about what we expect, here, and use the `Action` type instead of the `u32` one?

crates/meilisearch-types/src/keys.rs
```diff
@@ -195,206 +196,294 @@ fn parse_expiration_date(
...
+    pub const VERSION: u32 = A::Version.bits();
+    pub const KEYS_CREATE: u32 = A::KeysAdd.bits();
+    pub const KEYS_GET: u32 = A::KeysGet.bits();
+    pub const KEYS_UPDATE: u32 = A::KeysUpdate.bits();
+    pub const KEYS_DELETE: u32 = A::KeysDelete.bits();
+    pub const EXPERIMENTAL_FEATURES_GET: u32 = A::ExperimentalFeaturesGet.bits();
+    pub const EXPERIMENTAL_FEATURES_UPDATE: u32 = A::ExperimentalFeaturesUpdate.bits();
+    pub const NETWORK_GET: u32 = A::NetworkGet.bits();
+    pub const NETWORK_UPDATE: u32 = A::NetworkUpdate.bits();
+    pub const ALL: u32 = A::All.bits();
```
Kerollmops, line 364: To increase readability, we should add more space between categories.

```suggestion
    pub const SEARCH: u32 = A::Search.bits();

    pub const DOCUMENTS_ADD: u32 = A::DocumentsAdd.bits();
    pub const DOCUMENTS_GET: u32 = A::DocumentsGet.bits();
    pub const DOCUMENTS_DELETE: u32 = A::DocumentsDelete.bits();
    pub const DOCUMENTS_ALL: u32 = A::DocumentsAll.bits();

    pub const INDEXES_CREATE: u32 = A::IndexesAdd.bits();
    pub const INDEXES_GET: u32 = A::IndexesGet.bits();
    pub const INDEXES_UPDATE: u32 = A::IndexesUpdate.bits();
    pub const INDEXES_DELETE: u32 = A::IndexesDelete.bits();
    pub const INDEXES_SWAP: u32 = A::IndexesSwap.bits();
    pub const INDEXES_ALL: u32 = A::IndexesAll.bits();

    pub const TASKS_CANCEL: u32 = A::TasksCancel.bits();
    pub const TASKS_DELETE: u32 = A::TasksDelete.bits();
    pub const TASKS_GET: u32 = A::TasksGet.bits();
    pub const TASKS_ALL: u32 = A::TasksAll.bits();

    pub const SETTINGS_GET: u32 = A::SettingsGet.bits();
    pub const SETTINGS_UPDATE: u32 = A::SettingsUpdate.bits();
    pub const SETTINGS_ALL: u32 = A::SettingsAll.bits();

    pub const STATS_GET: u32 = A::StatsGet.bits();
    pub const STATS_ALL: u32 = A::StatsAll.bits();

    pub const METRICS_GET: u32 = A::MetricsGet.bits();
    pub const METRICS_ALL: u32 = A::MetricsAll.bits();

    pub const DUMPS_CREATE: u32 = A::DumpsCreate.bits();
    pub const DUMPS_ALL: u32 = A::DumpsAll.bits();

    pub const SNAPSHOTS_CREATE: u32 = A::SnapshotsCreate.bits();
    pub const SNAPSHOTS_ALL: u32 = A::SnapshotsAll.bits();

    pub const VERSION: u32 = A::Version.bits();

    pub const KEYS_CREATE: u32 = A::KeysAdd.bits();
    pub const KEYS_GET: u32 = A::KeysGet.bits();
    pub const KEYS_UPDATE: u32 = A::KeysUpdate.bits();
    pub const KEYS_DELETE: u32 = A::KeysDelete.bits();

    pub const EXPERIMENTAL_FEATURES_GET: u32 = A::ExperimentalFeaturesGet.bits();
    pub const EXPERIMENTAL_FEATURES_UPDATE: u32 = A::ExperimentalFeaturesUpdate.bits();

    pub const NETWORK_GET: u32 = A::NetworkGet.bits();
    pub const NETWORK_UPDATE: u32 = A::NetworkUpdate.bits();

    pub const ALL: u32 = A::All.bits();
```

ManyTheFish:
> Can we see proof that these Action flags can be defined in the signature of a route function with a combination, i.e., Action::DumpsAll | Action::SettingsGet

@Kerollmops, there is no need for that, a route never accepts several actions.
For example, the dump creation route can't accept anything else than a key that has at least the `DumpsCreate` action.

ManyTheFish:
@Kerollmops,
My bad, you could do that, but it would not work in this implementation, when defining `documents.delete | documents.set` for the route, what do you expect?
I think the user must have both actions allowed, am I right?

crates/meilisearch-types/src/keys.rs
ManyTheFish: It is not possible without changing the database

crates/meilisearch-types/src/keys.rs
ManyTheFish: to be able to remove the "all" action flags, the keys database must be changed:
https://github.com/meilisearch/meilisearch/issues/5489

crates/meilisearch-types/src/keys.rs
ManyTheFish: related to:
https://github.com/meilisearch/meilisearch/issues/5489

Issue #4873, opened 2024-08-13
Binary quantization


curquiza:
Related product team resources: [PRD](https://www.notion.so/meilisearch/Handle-a-larger-vector-dataset-50f2d10618bf480bba0d994e6571be0c) (_internal only_)

## Motivation

Handle bigger datasets with embedders

## Usage

https://meilisearch.notion.site/Feature-usage-2a9c9559461a4a9d9fa3e0ea5149ad68?pvs=4

## Impacted teams

Maybe @meilisearch/docs-team and @meilisearch/integration-team  if we have user facing changes

PR #4941, opened 2024-09-18
Implement the binary quantization in meilisearch

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4873

## What does this PR do?
- Add a settings for the binary quantization
- Once enabled, the bq cannot be disabled

TODO:
- [ ] Missing a bunch of tests


Cargo.lock
```diff
@@ -384,6 +384,24 @@ version = "0.7.4"
...
+ "rand",
+ "rayon",
+ "roaring",
+ "tempfile",
+ "thiserror",
+]
+
 [[package]]
 name = "arroy"
 version = "0.4.0"
```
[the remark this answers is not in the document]
irevoire: Oups yes, this one was used by the index-scheduler integration tests

milli/src/update/settings.rs
```diff
@@ -1082,19 +1095,13 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
...
-                EmbedderAction::Reindex(ReindexAction::FullReindex) => {
-                    if self.index.embedder_category_id.get(self.wtxn, name)?.is_none() {
-                        let id = find_free_index()
-                            .ok_or(UserError::TooManyEmbedders(updated_configs.len()))?;
-                        tracing::debug!(embedder = name, id, "assigning free id to new embedder");
-                        self.index.embedder_category_id.put(self.wtxn, name, &id)?;
-                    }
-                }
-                EmbedderAction::WriteBackToDocuments(_) => { /* already removed */ }
+            if matches!(action.reindex(), Some(ReindexAction::FullReindex))
```
dureuill, line 1098:
```suggestion
            // ignore actions that are not possible for a new embedder
            if matches!(action.reindex(), Some(ReindexAction::FullReindex))
```

milli/src/update/settings.rs
```diff
@@ -1287,16 +1298,19 @@ impl InnerIndexSettingsDiff {
...
-                        entry.insert(EmbedderAction::Reindex(ReindexAction::RegeneratePrompts));
+                        entry.insert(EmbedderAction::with_reindex(
+                            ReindexAction::RegeneratePrompts,
+                            was_quantized,
+                        ));
+                    }
+                    std::collections::btree_map::Entry::Occupied(entry) => {
+                        let EmbedderAction {
+                            was_quantized: _,
+                            is_being_quantized: _, // We are deleting this embedder, so no point in regeneration
```
dureuill, line 1309: the comments don't seem to fit their line?

milli/src/vector/mod.rs
```diff
@@ -26,6 +30,171 @@ pub type Embedding = Vec<f32>;
...
+impl ArroyWrapper {
+    pub fn new(database: arroy::Database<Unspecified>, index: u16, quantized: bool) -> Self {
+        Self { database, index, quantized }
+    }
+
+    pub fn index(&self) -> u16 {
+        self.index
+    }
+
+    pub fn dimensions(&self, rtxn: &RoTxn) -> Result<usize, arroy::Error> {
```
dureuill, line 48: Will this work on an empty db?

milli/src/vector/mod.rs
```diff
@@ -26,6 +30,171 @@ pub type Embedding = Vec<f32>;
 
 pub const REQUEST_PARALLELISM: usize = 40;
 
+pub struct ArroyWrapper {
```
dureuill, line 33: If we're going to have a struct that abstracts over the quantization, should we also abstract about multiple vectors?

i.e. this could take an embedding_id instead of an index and expose an interface to get/put multiple vectors, encapsulating the arroy_dbs iterators from the Index.

milli/src/vector/mod.rs
```diff
@@ -26,6 +30,171 @@ pub type Embedding = Vec<f32>;
...
+    ) -> Result<(), arroy::Error> {
+        if !self.quantized {
+            let writer = arroy::Writer::new(self.angular_db(), index, dimension);
+            writer.prepare_changing_distance::<BinaryQuantizedAngular>(wtxn)?;
+            self.quantized = true;
+        }
+        Ok(())
+    }
+
+    pub fn need_build(&self, rtxn: &RoTxn, dimension: usize) -> Result<bool, arroy::Error> {
```
dureuill, line 70: Is a build needed when `quantize` was called, or is the following sequence of uses valid?

1. Self::quantize
2. Self::needs_build -> returns false
3. Self::nns_by_item()

dureuill (commented):
I find a bit dangerous to have displaced the metadata checks at usage time rather than at new time

milli/src/vector/mod.rs
irevoire: > Is a build needed when quantize was called

Yes! And we now have a test on this in arroy 😎

meilisearch/tests/vector/binary_quantized.rs
```diff
@@ -0,0 +1,380 @@
...
+                  "source": "userProvided",
+                  "dimensions": 3,
+                  "binaryQuantized": false,
+              }
+          },
+        }))
+        .await;
+    snapshot!(code, @"202 Accepted");
+    let ret = server.wait_task(response.uid()).await;
+    snapshot!(ret, @r###"
```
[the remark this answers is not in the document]
irevoire: Huum from what I understand we ended up saying that if the binary quantization was enabled for an index then you cannot disable it period 🤔 

The discussion wasn't super clear, but that’s what I ended up understanding

milli/src/search/new/vector_sort.rs
```diff
@@ -49,16 +52,15 @@ impl<Q: RankingRuleQueryTrait> VectorSort<Q> {
...
+            ctx.index.arroy_readers(ctx.txn, self.embedder_index, self.quantized).collect();
 
         let target = &self.target;
         let mut results = Vec::new();
 
-        for reader in readers.iter() {
+        for reader in readers?.iter() {
             let nns_by_vector =
-                reader.nns_by_vector(ctx.txn, target, self.limit, None, Some(vector_candidates))?;
+                reader.nns_by_vector(ctx.txn, target, self.limit, Some(vector_candidates))?;
```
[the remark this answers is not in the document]
irevoire: This none was to control the number of node to inspect in arroy, and a second None appeared after the binary quantization to control how much overflow we want to do.
Since Meilisearch will never use any of these two options, I got rids of it in my `ArroyWrapper`.
Also, in the near future I would like to rewrite the way we make query in arroy because there is way too many options here 🥶

milli/src/vector/mod.rs
irevoire: If the DB has never been built, this’ll throw the missing metadata error.
If it has been built but contains no vectors, it should work

meilisearch/tests/vector/binary_quantized.rs
```diff
@@ -0,0 +1,380 @@
...
+        "embedders": {
+          "manual": {
+            "source": "userProvided",
+            "dimensions": 3,
+            "binaryQuantized": false
+          }
+        }
+      },
+      "error": {
+        "message": "`.embedders.manual.binaryQuantized`: Cannot disable the binary quantization",
```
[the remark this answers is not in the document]
irevoire: True, a custom link in the documentation could help as well for this one, I guess

milli/src/vector/mod.rs
dureuill: will the db be built by meilisearch if it contains no vector?

milli/src/search/similar.rs
```diff
@@ -67,9 +70,8 @@ impl<'a> Similar<'a> {
                 .get(self.rtxn, &self.embedder_name)?
                 .ok_or_else(|| crate::UserError::InvalidEmbedder(self.embedder_name.to_owned()))?;
 
-        let readers: std::result::Result<Vec<_>, _> =
-            self.index.arroy_readers(self.rtxn, embedder_index).collect();
-
+        let readers: Result<Vec<_>> =
+            self.index.arroy_readers(self.rtxn, embedder_index, self.quantized).collect();
```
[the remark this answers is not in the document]
irevoire: No, fixed in b8fd85a46d4f5034b32805a0e59653cffbe2dd58

milli/src/search/new/vector_sort.rs
```diff
@@ -49,16 +52,15 @@ impl<Q: RankingRuleQueryTrait> VectorSort<Q> {
         ctx: &mut SearchContext<'_>,
         vector_candidates: &RoaringBitmap,
     ) -> Result<()> {
-        let readers: std::result::Result<Vec<_>, _> =
-            ctx.index.arroy_readers(ctx.txn, self.embedder_index).collect();
-        let readers = readers?;
+        let readers: Result<Vec<_>> =
+            ctx.index.arroy_readers(ctx.txn, self.embedder_index, self.quantized).collect();
```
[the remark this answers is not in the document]
irevoire: No, fixed in b8fd85a46d4f5034b32805a0e59653cffbe2dd58

milli/src/vector/mod.rs
irevoire: No

irevoire:
thanks
bors merge

Issue #4948, opened 2024-09-19
Make the ArroyReader abstraction handle multiple vectors


irevoire:
If we're going to have a struct that abstracts over the quantization, should we also abstract about multiple vectors?

i.e. this could take an embedding_id instead of an index and expose an interface to get/put multiple vectors, encapsulating the arroy_dbs iterators from the Index.

_Originally posted by @dureuill in https://github.com/meilisearch/meilisearch/pull/4941#discussion_r1766807420_

irevoire:
Closed by #4953

PR #4953, opened 2024-09-23
Move the multi arroy index logic to the arroy wrapper

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4948

## What does this PR do?
- Make the `ArroyWrapper` we introduced in the last PR handle all the embedded for a specific docid itself.

### DIFF
--- a/milli/src/update/index_documents/typed_chunk.rs
+++ b/milli/src/update/index_documents/typed_chunk.rs
@@ -680,7 +680,7 @@ pub(crate) fn write_typed_chunk_into_index(
             let mut iter = merger.into_stream_merger_iter()?;
             while let Some((key, _)) = iter.next()? {
                 let docid = key.try_into().map(DocumentId::from_be_bytes).unwrap();
-                writer.del_item_raw(wtxn, expected_dimension, docid)?;
+                writer.del_items(wtxn, expected_dimension, docid)?;
             }
 
             // add generated embeddings
--- a/milli/src/update/index_documents/typed_chunk.rs
+++ b/milli/src/update/index_documents/typed_chunk.rs
@@ -673,22 +673,14 @@ pub(crate) fn write_typed_chunk_into_index(
                 .get(&embedder_name)
                 .map_or(false, |conf| conf.2);
             // FIXME: allow customizing distance
-            let writers: Vec<_> = crate::vector::arroy_db_range_for_embedder(embedder_index)
-                .map(|k| ArroyWrapper::new(index.vector_arroy, k, binary_quantized))
-                .collect();
+            let writer = ArroyWrapper::new(index.vector_arroy, embedder_index, binary_quantized);
 
             // remove vectors for docids we want them removed
             let merger = remove_vectors_builder.build();
             let mut iter = merger.into_stream_merger_iter()?;
             while let Some((key, _)) = iter.next()? {
                 let docid = key.try_into().map(DocumentId::from_be_bytes).unwrap();
-
-                for writer in &writers {
-                    // Uses invariant: vectors are packed in the first writers.
-                    if !writer.del_item(wtxn, expected_dimension, docid)? {
-                        break;
-                    }
-                }
+                writer.del_item(wtxn, expected_dimension, docid)?;
             }
 
             // add generated embeddings
@@ -716,9 +708,7 @@ pub(crate) fn write_typed_chunk_into_index(
                         embeddings.embedding_count(),
                     )));
                 }
-                for (embedding, writer) in embeddings.iter().zip(&writers) {
-                    writer.add_item(wtxn, expected_dimension, docid, embedding)?;
-                }
+                writer.add_items(wtxn, expected_dimension, docid, embeddings)?;
             }
 
             // perform the manual diff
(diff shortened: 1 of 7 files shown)


milli/src/vector/mod.rs
```diff
@@ -32,105 +32,243 @@ pub const REQUEST_PARALLELISM: usize = 40;
...
+            if self.quantized {
+                arroy::Writer::new(self.quantized_db(), index, dimension).clear(wtxn)?;
+            } else {
+                arroy::Writer::new(self.angular_db(), index, dimension).clear(wtxn)?;
+            }
         }
+        Ok(())
     }
 
     pub fn is_empty(&self, rtxn: &RoTxn, dimension: usize) -> Result<bool, arroy::Error> {
```
dureuill, line 260: It looks like the changes around building makes this function unused, and unnecessary

milli/src/vector/mod.rs
```diff
@@ -139,11 +277,18 @@ impl ArroyWrapper {
         dimension: usize,
         item: arroy::ItemId,
     ) -> Result<bool, arroy::Error> {
-        if self.quantized {
-            arroy::Writer::new(self.quantized_db(), self.index, dimension).contains_item(rtxn, item)
-        } else {
-            arroy::Writer::new(self.angular_db(), self.index, dimension).contains_item(rtxn, item)
+        for index in arroy_db_range_for_embedder(self.index) {
```
dureuill, line 280: This should exit as soon as a writer is empty, due to the "no hole" property

dureuill (commented):
Be careful around the "no hole" property: it should always be preserved when writing, and used when reading.

The code looks so much cleaner now, wow!

milli/src/update/index_documents/typed_chunk.rs
```diff
@@ -673,22 +673,14 @@ pub(crate) fn write_typed_chunk_into_index(
...
             while let Some((key, _)) = iter.next()? {
                 let docid = key.try_into().map(DocumentId::from_be_bytes).unwrap();
-
-                for writer in &writers {
-                    // Uses invariant: vectors are packed in the first writers.
-                    if !writer.del_item(wtxn, expected_dimension, docid)? {
-                        break;
-                    }
-                }
+                writer.del_item_raw(wtxn, expected_dimension, docid)?;
```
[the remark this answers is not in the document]
irevoire: Huuum, I would vastly prefer this approach because I didn’t really understand why what we were doing previously was working.

> What we want here (and were doing previously) is to remove all vectors for that docid.

But are you sure about that?
Because from what I see, the previous code was doing the same thing as me:
```rust
            let writers: Vec<_> = crate::vector::arroy_db_range_for_embedder(embedder_index)
                .map(|k| ArroyWrapper::new(index.vector_arroy, k, binary_quantized))
                .collect();
...
               for writer in &writers {
                    // Uses invariant: vectors are packed in the first writers.
                    if !writer.del_item(wtxn, expected_dimension, docid)? {
                        break;
                    }
                }
```

It goes through all the writers from and `break` when it finds something matching 🤔 

Or are you saying that we could get rids of the entire while loop?

milli/src/vector/mod.rs
```diff
@@ -32,105 +32,243 @@ pub const REQUEST_PARALLELISM: usize = 40;
...
-        wtxn: &mut RwTxn,
-        index: u16,
-        dimension: usize,
-    ) -> Result<(), arroy::Error> {
+    pub fn quantize(&mut self, wtxn: &mut RwTxn, dimension: usize) -> Result<(), arroy::Error> {
         if !self.quantized {
-            let writer = arroy::Writer::new(self.angular_db(), index, dimension);
-            writer.prepare_changing_distance::<BinaryQuantizedAngular>(wtxn)?;
+            for index in arroy_db_range_for_embedder(self.index) {
+                let writer = arroy::Writer::new(self.angular_db(), index, dimension);
```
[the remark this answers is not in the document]
irevoire: Yeah so about that, I would feel better if we just quantized everything, because there is a world where we built an empty DB in arroy (after deleting a vector for example) but the metadata are still here and contains the distance used to create this database.

milli/src/vector/mod.rs
```diff
@@ -32,105 +32,239 @@ pub const REQUEST_PARALLELISM: usize = 40;
...
+        &'a self,
+        rtxn: &'a RoTxn<'a>,
+        db: arroy::Database<D>,
+    ) -> impl Iterator<Item = Result<arroy::Reader<D>, arroy::Error>> + 'a {
+        arroy_db_range_for_embedder(self.embedder_index).map_while(move |index| {
+            match arroy::Reader::open(rtxn, index, db) {
+                Ok(reader) => Some(Ok(reader)),
+                Err(arroy::Error::MissingMetadata(_)) => None,
+                Err(e) => Some(Err(e)),
+            }
```
irevoire, line 62: @dureuill, do you think we could early exit here as well when the database is empty?

dureuill (commented):
I agree we should break when `reader.is_empty()` in `readers`. Otherwise looking good :+1:

dureuill (approved):
thank you!

PR #4954, opened 2024-09-24
Fix bench by adding embedder

dureuill:
Fix benchmark workloads following breaking change on embedders


ManyTheFish (approved):
bors merge

PR #4955, opened 2024-09-24
Upgrade "batch failed" log to error level

### LINKED ISSUE #4916: Serious errors sometimes logged at "INFO" level
When a batch task is requested and the disk is full, a log message is written, but it's at "INFO" level so is easily missed if someone's trying to use logs to trigger alerts.

The code in this specific case is: https://github.com/meilisearch/meilisearch/blob/db0cf3b2ed0ca6cf6549510bf373681e83b775dd/index-scheduler/src/lib.rs#L1266 . There may be other places in the code where this can happen.

The error that was logged for me: `{"timestamp":"2024-09-07T18:06:37.387047Z","level":"INFO","fields":{"message":"Batch failed No space left on device (os error 28)"},"target":"index_scheduler"}`

dureuill:
# Pull Request

## Related issue
Fixes #4916

### DIFF
--- a/index-scheduler/src/lib.rs
+++ b/index-scheduler/src/lib.rs
@@ -1263,7 +1263,7 @@ impl IndexScheduler {
                     #[cfg(test)]
                     self.maybe_fail(tests::FailureLocation::UpdatingTaskAfterProcessBatchFailure)?;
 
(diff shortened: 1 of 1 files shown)


irevoire (approved):
thanks,
bors merge

Issue #4840, opened 2024-07-30
Improve the meilisearch integration tests


irevoire:
Hello, fellow contributors! Thanks for coming here 🌟 

In https://github.com/meilisearch/meilisearch/pull/4808, I started some work to make the test faster but didn't apply the techniques implemented to every test since there are so many of them, and that's where you can help us.

There are two tasks to do that I’ll highlight right below.

### Patching the `.wait_task(:task_id)`

A common pattern you'll see in the integration tests is that we create a task and then immediately wait for it by writing `.wait_task(0).await;` or any other number.
Here's an example: https://github.com/meilisearch/meilisearch/blob/abe128476fca60d683725f2598a365027ef5fbec/meilisearch/tests/tasks/mod.rs#L111-L134

This is really easy to get wrong and caused us a lot of flaky tests in the past; that's why we would like to get rid of all these calls and replace them with the actual task ID returned by Meilisearch.

For example, instead of doing:
```rust
    let index = server.index("test");
    index.create(None).await;
    index.wait_task(0).await;
```

We should write:
```rust
    let index = server.index("test");
    let (task, _code) = index.create(None).await;
    index.wait_task(task.uid()).await;
```

Also, did you notice how we're waiting for the task to finish, but then we don't check if it succeeded?
When the task was supposed to succeed in the tests, you should also call `.succeeded()` after waiting for the task:

```rust
    let index = server.index("test");
    let (task, _code) = index.create(None).await;
    index.wait_task(task.uid()).await.succeeded();
```

### Using a shared server when possible

A lot of tests are creating a new Meilisearch instance when they could instead share one with all the other tests.
The rules to know if a test can use a shared server are:
- It should use the default configuration; the server is initialized with `let server = Server::new().await;`
- It must **not** modify the global state of the index:
  - Don't add an API-key
  - Don't modify the features
  - Don't access indexes that are not yours
- ❗ An exception, though, is if the request is expected to fail in the test. For example, if the test try to delete an arbitrary index and then checks for its failure, in this case, we can implement a new method on the shared index, send the task, and ensure it fails; see this method for an example: https://github.com/meilisearch/meilisearch/blob/c45706936732aef16d5a07545e47534d59901a91/meilisearch/tests/common/server.rs#L218-L225

https://github.com/meilisearch/meilisearch/blob/abe128476fca60d683725f2598a365027ef5fbec/meilisearch/tests/search/geo.rs#L41-L63

[...]

PR #4962, opened 2024-09-28
test: improve performance of create_index.rs

DerTimonius:
# Pull Request

## Related issue
related to #4840 

## What does this PR do?
This PR follows the instructions in #4840 and improves the performance of `meilisearch/tests/index/create_index.rs`. The tests run locally, if they fail in the CI I'll try to fix them

## PR checklist

### DIFF
--- a/meilisearch/tests/index/create_index.rs
+++ b/meilisearch/tests/index/create_index.rs
@@ -17,7 +17,7 @@ async fn create_index_no_primary_key() {
 
     assert_eq!(response["status"], "enqueued");
 
-    let response = index.wait_task(0).await;
+    let response = index.wait_task(response.uid()).await;
 
     assert_eq!(response["status"], "succeeded");
     assert_eq!(response["type"], "indexCreation");
@@ -34,7 +34,7 @@ async fn create_index_with_gzip_encoded_request() {
 
     assert_eq!(response["status"], "enqueued");
 
-    let response = index.wait_task(0).await;
+    let response = index.wait_task(response.uid()).await;
 
     assert_eq!(response["status"], "succeeded");
     assert_eq!(response["type"], "indexCreation");
@@ -82,7 +82,7 @@ async fn create_index_with_zlib_encoded_request() {
 
     assert_eq!(response["status"], "enqueued");
 
-    let response = index.wait_task(0).await;
+    let response = index.wait_task(response.uid()).await;
 
     assert_eq!(response["status"], "succeeded");
     assert_eq!(response["type"], "indexCreation");
@@ -99,7 +99,7 @@ async fn create_index_with_brotli_encoded_request() {
 
     assert_eq!(response["status"], "enqueued");
 
-    let response = index.wait_task(0).await;
+    let response = index.wait_task(response.uid()).await;
 
     assert_eq!(response["status"], "succeeded");
     assert_eq!(response["type"], "indexCreation");
@@ -116,7 +116,7 @@ async fn create_index_with_primary_key() {
 
     assert_eq!(response["status"], "enqueued");
 
-    let response = index.wait_task(0).await;
+    let response = index.wait_task(response.uid()).await;
 
     assert_eq!(response["status"], "succeeded");
     assert_eq!(response["type"], "indexCreation");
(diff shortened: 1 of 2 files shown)


meilisearch/tests/index/create_index.rs
```diff
@@ -17,7 +17,7 @@ async fn create_index_no_primary_key() {
 
     assert_eq!(response["status"], "enqueued");
 
-    let response = index.wait_task(0).await;
+    let response = index.wait_task(response.uid()).await;
```
irevoire, line 20: Not related to this line, but above there is a:
```rust
```

That can be replaced with a:
```rust
    let index = server.unique_index();
```

meilisearch/tests/index/create_index.rs
```diff
@@ -34,7 +34,7 @@ async fn create_index_with_gzip_encoded_request() {
...
```
irevoire, line 37: Same as above, we could use a shared server.

But this time we don’t have the right function to do it with an encoder.
Could you create a new method on the `Server<Shared>` called `unique_index_with_encoder` that takes the encoder by parameter?

Here’s the current one you need to duplicate:
https://github.com/meilisearch/meilisearch/blob/7f20c13f3fc374faa0b812d4e21de541ed95746a/meilisearch/tests/common/server.rs#L301-L310

meilisearch/tests/index/create_index.rs
```diff
@@ -129,10 +129,10 @@ async fn create_index_with_invalid_primary_key() {
 
     let server = Server::new().await;
     let index = server.index("movies");
-    let (_response, code) = index.add_documents(document, Some("title")).await;
+    let (response, code) = index.add_documents(document, Some("title")).await;
```
irevoire, line 132: Same as above, we can use a shared server + unique index:
```rust
    let index = server.unique_index();
    let (response, code) = index.add_documents(document, Some("title")).await;
```

meilisearch/tests/index/create_index.rs
```diff
@@ -147,13 +147,13 @@ async fn test_create_multiple_indexes() {
     let index3 = server.index("test3");
     let index4 = server.index("test4");
 
-    index1.create(None).await;
-    index2.create(None).await;
-    index3.create(None).await;
+    let (task1, _) = index1.create(None).await;
+    let (task2, _) = index2.create(None).await;
+    let (task3, _) = index3.create(None).await;
```
irevoire, line 152: I cannot select the right place but it’s the same as above, we can use a shared server + 4 unique index

meilisearch/tests/index/create_index.rs
```diff
@@ -169,9 +169,9 @@ async fn error_create_existing_index() {
 
     assert_eq!(code, 202);
 
-    index.create(Some("primary")).await;
+    let (task, _) = index.create(Some("primary")).await;
```
irevoire, line 172: Same as above, we can use a shared server + two unique index

irevoire (changes requested):
Hey @DerTimonius,

Pretty much the same comment as the other PR.
Everything you did is top notch but you missed the two other part of the issue about using shared server and unique index as much as possible.

Thanks for working on this!

meilisearch/tests/index/create_index.rs
irevoire: Here you can still uses a shared server

meilisearch/tests/index/create_index.rs
irevoire: Same here, you can use a shared server

meilisearch/tests/index/create_index.rs
```diff
@@ -68,21 +68,20 @@ async fn create_index_with_gzip_encoded_request_and_receiving_brotli_encoded_res
     let parsed_response =
         serde_json::from_slice::<Value>(decoded.into().as_ref()).expect("Expecting valid json");
 
-    assert_eq!(parsed_response["taskUid"], 0);
```
DerTimonius, line 71: I had to remove this as it appears to be different locally and in the CI, so it's hard to know the actual taskUid here...

meilisearch/tests/index/create_index.rs
DerTimonius: sorry, must have overlooked that one in my revision 🙈

meilisearch/tests/index/create_index.rs
irevoire: Makes sense since the server is shared, thanks!

meilisearch/tests/index/create_index.rs
irevoire: No problem, that’s perfect 👌

DerTimonius:
thanks for the help @irevoire

PR #4963, opened 2024-09-29
test: improve performance of delete_index.rs

DerTimonius:
# Pull Request

## Related issue
related to #4840

## What does this PR do?
This PR follows the instructions in #4840 and improves the performance of `meilisearch/tests/index/delete_index.rs`. The tests run locally, if they fail in the CI I'll try to fix them

## PR checklist

### DIFF
--- a/meilisearch/tests/common/index.rs
+++ b/meilisearch/tests/common/index.rs
@@ -272,6 +272,20 @@ impl<'a> Index<'a, Shared> {
         }
         (task, code)
     }
+
+    pub async fn delete_index_fail(&self) -> (Value, StatusCode) {
+        let (mut task, code) = self._delete().await;
+        if code.is_success() {
+            task = self.wait_task(task.uid()).await;
+            if task.is_success() {
+                panic!(
+                    "`delete_index_fail` succeeded: {}",
+                    serde_json::to_string_pretty(&task).unwrap()
+                );
+            }
+        }
+        (task, code)
+    }
 }
 
 #[allow(dead_code)]
@@ -314,6 +328,12 @@ impl<State> Index<'_, State> {
         });
         self.service.post_encoded("/indexes", body, self.encoder).await
     }
+
+    pub(super) async fn _delete(&self) -> (Value, StatusCode) {
+        let url = format!("/indexes/{}", urlencode(self.uid.as_ref()));
+        self.service.delete(url).await
+    }
+
     pub async fn wait_task(&self, update_id: u64) -> Value {
         // try several times to get status, or panic to not wait forever
         let url = format!("/tasks/{}", update_id);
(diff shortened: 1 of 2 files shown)


meilisearch/tests/index/delete_index.rs
```diff
@@ -5,19 +5,19 @@ use crate::json;
 async fn create_and_delete_index() {
     let server = Server::new().await;
```
irevoire, line 6: Since we're not modifying the state of the server in the test we could use a shared index:
```rust
let index = server.unique_index();
```

meilisearch/tests/index/delete_index.rs
```diff
@@ -26,7 +26,7 @@ async fn create_and_delete_index() {
 async fn error_delete_unexisting_index() {
     let server = Server::new().await;
     let index = server.index("test");
-    let (_, code) = index.delete().await;
+    let (task, code) = index.delete().await;
```
irevoire, line 29: Same as above, except you don’t even need to create the index here.
You can directly use the non-existing shared index from here:
https://github.com/meilisearch/meilisearch/blob/7f20c13f3fc374faa0b812d4e21de541ed95746a/meilisearch/tests/common/mod.rs#L143-L149

irevoire (changes requested):
Hey @DerTimonius,

Nice PR, everything you did is good.
I left a few comments because we can do even more with the shared server/indexes and that’s the most important part.

Thanks for looking into it!

meilisearch/tests/index/delete_index.rs
DerTimonius: I don't know if it's possible to just call `shared_does_not_exists_index()`, as the returned shared index does not have the necessary methods. only the owned indices appear do have the tested methods if I understand correctly

irevoire: Oh sorry, I thought I answered but it looks like I didn’t 😰 

You’re right, it’s not the most obvious one. I think you could add a method `delete_index_fail` like I did with the [`delete_documents_by_filter_fail`](https://github.com/meilisearch/meilisearch/blob/e78da3528781ab03c9b51a12b153a19af8e68f5b/meilisearch/tests/common/index.rs#L262).
The general idea is that you can try to delete a `Shared` index, but if it succeeds, then the test must fail because other tests may be using it, and we want to see the test that caused the issue first, not the other one.

Thanks for your other commit it’s perfect!

DerTimonius: ah, perfect, thanks for the guidance!

meilisearch/tests/common/index.rs
```diff
@@ -314,6 +328,12 @@ impl<State> Index<'_, State> {
         });
         self.service.post_encoded("/indexes", body, self.encoder).await
     }
+
+    pub(super) async fn _delete(&self) -> (Value, StatusCode) {
+        let url = format!("/indexes/{}", urlencode(self.uid.as_ref()));
+        self.service.delete(url).await
+    }
```
irevoire, line 335: awesome

irevoire (approved):
Thanks a lot that’s perfect 👏 

bors merge

curquiza:
bors merge

PR #4969, opened 2024-10-01
Indexer edition 2024 try map

Kerollmops:
This PR brings a new `AppendOnlyVec` struct that we combine with the new `ParallelIteratorExt` trait to improve the data locality of the indexing process. Avoid moving caches between threads for every single document.

### DIFF
--- a/milli/src/update/new/append_only_vec.rs
+++ b/milli/src/update/new/append_only_vec.rs
@@ -99,12 +99,6 @@ fn test_indices() {
     }
 }
 
-impl<T> Default for AppendOnlyVec<T> {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl<T> AppendOnlyVec<T> {
     const EMPTY: UnsafeCell<*mut T> = UnsafeCell::new(ptr::null_mut());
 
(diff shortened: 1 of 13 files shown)


dureuill (approved):
I have reservations about the code of append vec

1. I'm not sure why they're using an UnsafeCell<Ptr> instead of an AtomicPtr, and why it works (seems to me like concurrent read/write access to a ptr that is not atomic is UB)
2. Do we have a hard upper bound on the number of calls to map_with (e.g. the number of threads), or is it occasionally larger due to work stealing and so? if the former we could possibly allocate the whole chunk in front and have each thread select their item from their thread_index rather than by allocating and synchronizing at runtime. It kind of excludes work stealing though...

I don't want to be a blocker on this PR, but maybe we should try running some tests under miri? Although, I'm not sure if we could detect data races 🤔

Kerollmops:
Thank you very much for the review, Louis.

I prefer to keep this "fork" of `AppendOnlyVec` for now and propose these changes to the original crate as a new struct so that they can review the code (as we talked together). So, let's merge and move to something else.

PR #4970, opened 2024-10-01
Create a new export documents meilitool subcommand

Kerollmops:
This subcommand can be useful for extracting documents from an existing database.

### DIFF
--- a/crates/meilitool/src/main.rs
+++ b/crates/meilitool/src/main.rs
@@ -501,7 +501,8 @@ fn export_documents(
             }
 
             let mut stdout = BufWriter::new(std::io::stdout());
-            for (i, ret) in index.all_documents(&rtxn)?.skip(offset.unwrap_or(0)).enumerate() {
+            let all_documents = index.documents_ids(&rtxn)?.into_iter().skip(offset.unwrap_or(0));
+            for (i, ret) in index.iter_documents(&rtxn, all_documents)?.enumerate() {
                 let (id, doc) = ret?;
                 let mut document = obkv_to_json(&all_fields, &fields_ids_map, doc)?;
 
--- a/crates/meilitool/src/main.rs
+++ b/crates/meilitool/src/main.rs
@@ -83,6 +83,10 @@ enum Command {
         /// Do not export vectors with the documents.
         #[arg(long)]
         ignore_vectors: bool,
+
+        /// The number of documents to skip.
+        #[arg(long)]
+        offset: Option<usize>,
     },
 
     /// Attempts to upgrade from one major version to the next without a dump.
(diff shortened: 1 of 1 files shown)


meilitool/src/main.rs
```diff
@@ -750,3 +758,40 @@ fn export_a_dump(
...
+            })?;
+
+            let rtxn = index.read_txn()?;
+            let fields_ids_map = index.fields_ids_map(&rtxn)?;
+            let all_fields: Vec<_> = fields_ids_map.iter().map(|(id, _)| id).collect();
+
+            let mut stdout = BufWriter::new(std::io::stdout());
+            for ret in index.all_documents(&rtxn)? {
+                let (_id, doc) = ret?;
+                let document = obkv_to_json(&all_fields, &fields_ids_map, doc)?;
```
dureuill, line 786: the documents will lack the vectors that are stored in the DB. See logic [in the dump](https://github.com/meilisearch/meilisearch/blob/71b364286bf78bec66c10c22e8f3982ad77dc0d4/index-scheduler/src/batch.rs#L914)

milli/src/index.rs
```diff
@@ -223,7 +223,7 @@ impl Index {
         let documents = env.create_database(&mut wtxn, Some(DOCUMENTS))?;
         wtxn.commit()?;
 
-        Index::set_creation_dates(&env, main, created_at, updated_at)?;
+        // Index::set_creation_dates(&env, main, created_at, updated_at)?;
```
dureuill, line 226: Is it OK to ignore creation dates for indexes going forward?

Kerollmops: Ho Gosh! This is the reason why the CI is failing. Thank you!
However, I disabled this because I was triggering a _deserialize error_ when using this tool on a database, and I think it was related to the way we compile the binaries (from the workspace or crate itself). I'll enable that back and see.

Kerollmops:
Hey @dureuill 👋 I fixed the issue and added support for vectors. Can you review that again? 🙂

dureuill (dismissed):
I can accept the changes as-is, but I'm wondering if we should maybe factor the logic to export the documents between meilitool and the dumps:

this logic changed multiple times (most recently, to inject vectors) and now exists in two different locations. If the next change updates the index-scheduler but forgets about meilitool, then we will get possibly hard to track inconsistencies when using the meilitool export-documents command.

If factoring the logic is too complicated due to them being in different crates not using the same types, maybe a comment on the index-scheduler implementation to also implement the meilitool implementation would be welcome.

What do you think?

crates/meilitool/src/main.rs
```diff
@@ -443,3 +467,105 @@ fn compact_index(db_path: PathBuf, index_name: &str) -> anyhow::Result<()> {
...
+            let fields_ids_map = index.fields_ids_map(&rtxn)?;
+            let all_fields: Vec<_> = fields_ids_map.iter().map(|(id, _)| id).collect();
+            let embedding_configs = index.embedding_configs(&rtxn)?;
+
+            if let Some(offset) = offset {
+                eprintln!("Skipping {offset} documents");
+            }
+
+            let mut stdout = BufWriter::new(std::io::stdout());
+            for (i, ret) in index.all_documents(&rtxn)?.skip(offset.unwrap_or(0)).enumerate() {
```
dureuill, line 504: I fear this way of skipping documents from `all_documents` must be much slower to the one I proposed:

```
           let all_documents = index.documents_ids(&rtxn)?;
            if let Some(offset) = offset {
                eprintln!("Skipping {offset} documents");
            }
            for (document_index, ret) in index.iter_documents(&rtxn, all_documents.into_iter().skip(offset.unwrap_or_default()))?.enumerate() {
```

this avoids spawning each document for real. Care must be taken to add the `document_index` to `offset` though for displaying the correct progress, though

dureuill (approved):
thank you so much!

bors merge

PR #4992, opened 2024-10-09
fix the bad experimental search queue size

### LINKED ISSUE #4991: The search queue size default value of the config files is wrong
**Describe the bug**
The default value of the config file is set to 0 instead of 1000.

Initial discovery was made in https://github.com/meilisearch/meilisearch/issues/4991

Even though the bug is on an experimental feature flag, it actually affects all instances 😬 

**To Reproduce**
1. Starts meilisearch with an empty config file

**Expected behavior**

**Meilisearch version:**
v1.9 to v1.11

irevoire:
# Pull Request

## Related issue
Fixes #4991 

## What does this PR do?
- Set the right default value for the experimental search queue size in the config file


dureuill (approved):
bors merge

PR #5001, opened 2024-10-10
Do not send empty edit document by function

irevoire:
# Pull Request

We realized that we had a huge usage of the feature from user who didn’t enable the feature at all. That shouldn’t be possible.
After a big investigation with @gmourier 
image
We found the issue, it was in the engine

## What does this PR do?
- Do not send the edit by function event to segment if no event was received during this batch


Kerollmops (approved):
bors merge

Issue #5009, opened 2024-10-15
Revamp the analytics


irevoire:
[PRD (internal link)](https://www.notion.so/meilisearch/Analytics-Aggregate-Engine-API-Events-11f4b06b651f805ea60cec7aa7e4362e#1214b06b651f800b83a0febd8941c326)

- [x] Send all the experimental features through the `info` identify event
- [x] For all settings, return the last value OR null if it wasn’t modified
- [x] Aggregate everything
- [x] Store the timestamp of **the first event** and set it correctly
- [x] Store the user-agent of **every requests** and set it correctly
- [x] Add a `requests.total_received` to all the events
- [x] Remove the experimental feature seen

PR #5011, opened 2024-10-17
Revamp analytics

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/5009

## What does this PR do?
- Force every analytics to go through a trait that forces you to handle aggregation correcty
- Put the code to retrieve the `user-agent`, `timestamp` and `requests.total_received` in common between all aggregates, so there is no mistake
- Get rids of all the different channel for each kind of event in favor of an any map
- Ensure that we never [send empty event ever again](https://github.com/meilisearch/meilisearch/pull/5001)
- Merge all the sub-settings route into a global « Settings Updated » event.
- Fix: When using one of the three following feature, we were not sending any analytics IF they were set from the global route
  - /non-separator-tokens
  - /separator-tokens
  - /dictionary


meilisearch/src/routes/indexes/documents.rs
```diff
@@ -100,12 +102,89 @@ pub struct GetDocument {
...
 
+aggregate_methods!(
+    DocumentsGET => "Documents Fetched GET",
+    DocumentsPOST => "Documents Fetched POST",
+);
+
+#[derive(Default, Serialize)]
+pub struct DocumentsFetchAggregator<Method: AggregateMethod> {
+    #[serde(rename = "requests.total_received")]
+    total_received: usize,
```
ManyTheFish, line 113: default of `usize` is `0` so aggregation may always return `0`

meilisearch/src/routes/indexes/documents.rs
```diff
@@ -100,12 +102,89 @@ pub struct GetDocument {
     retrieve_vectors: Param<bool>,
 }
 
+aggregate_methods!(
+    DocumentsGET => "Documents Fetched GET",
+    DocumentsPOST => "Documents Fetched POST",
+);
+
+#[derive(Default, Serialize)]
```
ManyTheFish, line 110: We may prefer avoid the `Default` impl, too error prone.
```suggestion
#[derive(Serialize)]
```

meilisearch/src/routes/indexes/documents.rs
```diff
@@ -129,17 +211,54 @@ pub async fn get_document(
     Ok(HttpResponse::Ok().json(document))
 }
 
+#[derive(Default, Serialize)]
+pub struct DocumentsDeletionAggregator {
+    #[serde(rename = "requests.total_received")]
+    total_received: usize,
```
ManyTheFish, line 217: Is `total_received` necessary here?

meilisearch/src/routes/indexes/documents.rs
```diff
@@ -298,23 +421,72 @@ fn from_char_csv_delimiter(
     }
 }
 
+aggregate_methods!(
+    Replaced => "Documents Added",
+    Updated => "Documents Updated",
+);
+
+#[derive(Default, Serialize)]
```
ManyTheFish, line 429:
```suggestion
#[derive(Serialize)]
```

meilisearch/src/routes/indexes/documents.rs
```diff
@@ -129,17 +211,54 @@ pub async fn get_document(
     Ok(HttpResponse::Ok().json(document))
 }
 
+#[derive(Default, Serialize)]
```
ManyTheFish, line 214:
```suggestion
#[derive(Serialize)]
```

meilisearch/src/routes/indexes/documents.rs
```diff
@@ -599,13 +793,41 @@ pub struct DocumentEditionByFunction {
     pub function: String,
 }
 
+#[derive(Default, Serialize)]
```
ManyTheFish, line 796:
```suggestion
#[derive(Serialize)]
```

ManyTheFish (changes requested):
Some changes before the WE, sorry :/

irevoire:
I think I messed up the settings. When I started using the or instead of the map_or stuff, I should have swapped the order of the self and other field to select the new one instead of the old one.

+ I want to rename the argument of the aggregate function trait to call it "new" instead of "other" so it's clearer which version is new and old

irevoire:
I don’t think we need to integrate the helper function into the trait, we can simply specialize it at compile time and get a ptr to it without putting it in the trait

PR #5019, opened 2024-10-22
Implement facet search extraction

ManyTheFish:
- [ ] use bumpalo in the searchable extractor
- [ ] implement facet search extraction

### DIFF
--- a/milli/src/update/new/facet_search_builder.rs
+++ b/milli/src/update/new/facet_search_builder.rs
@@ -1,0 +1,275 @@
+use std::collections::{BTreeSet, HashMap};
+
+use charabia::{normalizer::NormalizerOption, Language, Normalize, StrDetection, Token};
+use grenad::Sorter;
+use heed::{
+    types::{Bytes, SerdeJson},
+    BytesDecode, BytesEncode, RoTxn,
+};
+
+use crate::{
+    heed_codec::{
+        facet::{FacetGroupKey, FacetGroupKeyCodec},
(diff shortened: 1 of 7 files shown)


milli/src/update/new/fst_merger_builder.rs
```diff
@@ -0,0 +1,155 @@
...
+
+use fst::{Set, SetBuilder, Streamer};
+use memmap2::Mmap;
+use tempfile::tempfile;
+
+use crate::{update::del_add::DelAdd, InternalError, Result};
+
+pub struct FstMergerBuilder<'a> {
+    stream: Option<fst::set::Stream<'a>>,
+    fst_builder: SetBuilder<BufWriter<File>>,
```
ManyTheFish, line 11: We should find a way to store it in RAM if we have the space

milli/src/update/new/facet_search_builder.rs
```diff
@@ -0,0 +1,275 @@
...
+};
+
+use super::{
+    channel::FacetSearchableSender, extract::FacetKind, fst_merger_builder::FstMergerBuilder,
+    KvReaderDelAdd,
+};
+
+pub struct FacetSearchBuilder<'indexer> {
+    registered_facets: HashMap<FieldId, usize>,
+    normalized_facet_string_docids_sorter: Sorter<MergeDeladdBtreesetString>,
```
ManyTheFish, line 31: We should find a way to store it in RAM if we have the space

PR #5025, opened 2024-10-29
test: improve performance of get_documents.rs

### LINKED ISSUE #4840: Improve the meilisearch integration tests

PedroTurik:
# Pull Request

## Related issue
Fixes one item from #4840 

## What does this PR do?
- Applies the changes recommended on the issue for `meilisearch/tests/documents/get_documents.rs`

## PR checklist

### DIFF
--- a/crates/meilisearch/tests/common/index.rs
+++ b/crates/meilisearch/tests/common/index.rs
@@ -9,8 +9,7 @@ use urlencoding::encode as urlencode;
 
 use super::encoder::Encoder;
 use super::service::Service;
-use super::Value;
-use super::{Owned, Shared};
+use super::{Owned, Shared, Value};
 use crate::json;
 
 pub struct Index<'a, State = Owned> {
--- a/crates/meilisearch/tests/common/mod.rs
+++ b/crates/meilisearch/tests/common/mod.rs
@@ -392,20 +392,22 @@ pub static VECTOR_DOCUMENTS: Lazy<Value> = Lazy::new(|| {
 
 pub async fn shared_index_with_test_set() -> &'static Index<'static, Shared> {
     static INDEX: OnceCell<Index<'static, Shared>> = OnceCell::const_new();
-    INDEX.get_or_init(|| async {
-        let server = Server::new_shared();
-        let index = server._index("SHARED_TEST_SET").to_shared();
-        let url = format!("/indexes/{}/documents", urlencoding::encode(index.uid.as_ref()));
-        let (response, code) = index
-            .service
-            .post_str(
-                url,
-                include_str!("../assets/test_set.json"),
-                vec![("content-type", "application/json")],
-            )
-            .await;
-        assert_eq!(code, 202);
-        index.wait_task(response.uid()).await;
-        index
-    }).await
+    INDEX
+        .get_or_init(|| async {
+            let server = Server::new_shared();
+            let index = server._index("SHARED_TEST_SET").to_shared();
+            let url = format!("/indexes/{}/documents", urlencoding::encode(index.uid.as_ref()));
+            let (response, code) = index
+                .service
+                .post_str(
+                    url,
+                    include_str!("../assets/test_set.json"),
+                    vec![("content-type", "application/json")],
+                )
+                .await;
+            assert_eq!(code, 202);
+            index.wait_task(response.uid()).await;
+            index
+        })
+        .await
 }
--- a/crates/meilisearch/tests/common/mod.rs
+++ b/crates/meilisearch/tests/common/mod.rs
@@ -389,3 +389,23 @@ pub static VECTOR_DOCUMENTS: Lazy<Value> = Lazy::new(|| {
       },
     ])
 });
+
+pub async fn shared_index_with_test_set() -> &'static Index<'static, Shared> {
+    static INDEX: OnceCell<Index<'static, Shared>> = OnceCell::const_new();
+    INDEX.get_or_init(|| async {
+        let server = Server::new_shared();
+        let index = server._index("SHARED_TEST_SET").to_shared();
+        let url = format!("/indexes/{}/documents", urlencoding::encode(index.uid.as_ref()));
+        let (response, code) = index
+            .service
+            .post_str(
+                url,
+                include_str!("../assets/test_set.json"),
+                vec![("content-type", "application/json")],
+            )
+            .await;
+        assert_eq!(code, 202);
+        index.wait_task(response.uid()).await;
+        index
+    }).await
+}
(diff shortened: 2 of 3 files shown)


irevoire (changes requested):
Hey, thanks a lot for the PR it’s awesome! This file was really outdated.

> Can someone with more experience in the project provide some guidance?

You're right, the test was just flaky I can't say why. Seems like openAI didn’t answer for a long time.

PedroTurik:
Applied the recommended changes. Thanks for the advice 😁

irevoire (dismissed):
Perfect, thanks a lot @PedroTurik 🎉 

bors merge

irevoire:
bors merge

irevoire (approved):
Thanks

bors merge

irevoire:
bors seems a bit sick sorry

bors cancel

irevoire:
bors merge

PR #5039, opened 2024-11-06
Add 3s timeout to embedding requests made during search

### LINKED ISSUE #5032: Cancel embedding operations for hybrid search after a deadline is met
Currently, any embedding request Meilisearch makes to an embedding server will be retried up to 10 times, with exponential backoff.

The exponential backoff has a time limit of 2 minutes, so the longest theoretical wait time is about 20 minutes.
For indexing, this may be fine: when embedding millions of documents, it might be preferable to sometimes wait multiple minutes rather than cancel the whole operation.

For search, however, this is too long.
- For hybrid search, we want to fallback to full-text search after one second.
- For pure semantic search, we want to error after a couple seconds.

This task is about implementing that change. No API change is expected at this point.

dureuill:
# Pull Request

## Related issue
Fixes #5032 

## What does this PR do?
- Add a 3-second timeout to embedding requests against a remote embedder made in the context of search. The timeout triggers when there are failing requests due to rate-limiting.
- Add a test of that timeout.

### DIFF
--- a/meilisearch/src/search/mod.rs
+++ b/meilisearch/src/search/mod.rs
@@ -796,8 +796,10 @@ fn prepare_search<'t>(
                     let span = tracing::trace_span!(target: "search::vector", "embed_one");
                     let _entered = span.enter();
 
+                    let deadline = std::time::Instant::now() + std::time::Duration::from_secs(10);
+
                     embedder
-                        .embed_one(query.q.clone().unwrap())
+                        .embed_one(query.q.clone().unwrap(), Some(deadline))
                         .map_err(milli::vector::Error::from)
                         .map_err(milli::Error::from)?
                 }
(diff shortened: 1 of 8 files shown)


meilisearch/src/search/mod.rs
```diff
@@ -796,8 +796,10 @@ fn prepare_search<'t>(
                     let span = tracing::trace_span!(target: "search::vector", "embed_one");
                     let _entered = span.enter();
 
+                    let deadline = std::time::Instant::now() + std::time::Duration::from_secs(10);
```
irevoire, line 799:
```suggestion
                    let deadline = std::time::Instant::now() + std::time::Duration::from_secs(3);
```

Shouldn’t it be 3s here as well?

irevoire (changes requested):

meilisearch/src/search/mod.rs
dureuill: This one is for pure semantic search. I thought 10s was better in that case because there's no fallback to full-text search when the embedding fails.

irevoire: you seems ok for me then we can merge 👍

PR #5044, opened 2024-11-06
Adds new metrics to prometheus

### LINKED ISSUE #4998: Add prometheus metrics about the usage of the search queue
To help us debug search performance issues, we should add to the prometheus metrics the:
- Size of the search queue (a single number that shouldn’t change over time)
- The current number of searches being processed
- The number of searches waiting to be processed in the search queue

PedroTurik:
not 100% confident in this solution, especially because i couldn't make the "Search Queue searches waiting" metric give me any value other than 0 with my local testing :laughing:. But i believe it solves the Issue.

# Pull Request

## Related issue
Fixes #4998 

### Adds new metrics to prometheus;
- SearchQueue size, 
- SearchQueue searches running, 
- and Search Queue searches waiting.

## PR checklist

### DIFF
--- a/crates/meilisearch/src/metrics.rs
+++ b/crates/meilisearch/src/metrics.rs
@@ -49,4 +49,18 @@ lazy_static! {
     pub static ref MEILISEARCH_IS_INDEXING: IntGauge =
         register_int_gauge!(opts!("meilisearch_is_indexing", "Meilisearch Is Indexing"))
             .expect("Can't create a metric");
+    pub static ref MEILISEARCH_SEARCH_QUEUE_SIZE: IntGauge = register_int_gauge!(opts!(
+        "meilisearch_search_queue_size",
+        "Meilisearch Search Queue Size"
+    ))
+    .expect("Can't create a metric");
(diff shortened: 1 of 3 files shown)


irevoire:
Hey, yeah I investigated for some time until I decided I would give it a try under heavy load, and it seems like it just work, but it’s hard to catch the right time, I don’t know why 🤔 
image

Since prometheus works on a polling system, I’m wondering if we shouldn’t return something like the max value encountered during the last second. Otherwise, I’m afraid we’ll only see 0.
In my case, to get a 5 I cheated a bit and made all search requests sleep for 1s and then had to run **10** loops spamming search requests at the same time 😭 

But I’m still struggling to understand why it behaves like that I’ll investigate a bit more

irevoire (approved):
bors merge

PedroTurik:
> In my case, to get a 5 I cheated a bit and made all search requests sleep for 1s and then had to run **10** loops spamming search requests at the same time 😭

Exactly :laughing:. I spammed infinite requests as fast as possible with max concurrency in a i7 14700 while checking the metric and it was always zero. I guess that means the product is really damn fast :laughing:!

PR #5051, opened 2024-11-13
Add timeout on read and write operations.

dureuill:
# Pull Request

## Related issue
Addresses #5054 

## What does this PR do?
- Add a timeout for read and write operations in the REST embedder. This might address some issues about tasks that get "stuck" while embedding documents.


irevoire (approved):
Thank you :cow2: 

bors merge

Kerollmops (approved):
Wow! I find 30s of timeout a lot but let's try that way, we can reduce it later on.

Thank you!
Bors merge

PR #5058, opened 2024-11-14
Retry if deserialization of remote response failed

dureuill:
v1.11.2 introduces a new error condition (instead of a possibly infinite wait): deserializing the body of a remote response can cause a timeout and fail immediately.

This PR makes it so that deserialization failures are inside the retry loop, meaning they obey the exponential backoff retry strategy.


irevoire (approved):
Thanks you

bors merge

Issue #4977, opened 2024-10-03
Give visibility about batch details in tasks


curquiza:
Related product team resources: [PRD](https://www.notion.so/meilisearch/Enhance-visibility-on-batched-tasks-1194b06b651f80889e58d2862a4662bc) (_internal only_)

## Motivation

Avoid confusion about understanding how Meilisearch indexing is processing tasks

## Usage

https://www.notion.so/meilisearch/Enhance-visibility-on-batched-tasks-1194b06b651f810b8fe0fab5d72846a8

## Impacted teams

@meilisearch/docs-team
@meilisearch/integration-team

curquiza:
@meilisearch/integration-team @meilisearch/docs-team -> be careful, not everything in the PRD is implemented. 
This issue is still remaining and will be done in pre-release: https://github.com/meilisearch/meilisearch/issues/5068

PR #5060, opened 2024-11-14
Batch route

irevoire:
# Pull Request

See [usage](https://www.notion.so/meilisearch/Enhance-visibility-on-batched-tasks-1194b06b651f810b8fe0fab5d72846a8).

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4977

## What does this PR do?
- For more detailed information, see the PRD.
- Added a `batchUid` to the tasks (that's the cause of all the updates of the dumps):
  - For all enqueued tasks, it's set to `None`
  - For every other tasks it must be set to something
  - :warning: For all the tasks imported in a dump, the `batchUid` will be set to `None` as well.
- Add two new routes:
  - `GET /batches/:uid` - to query a batch by its id
  - `GET /batches` - to retrieve a list of batches. It accepts all the same query parameters that are available on the `GET /tasks` route
- Adds new databases to query the batches directly:
  - When doing a query against the batches, the rule of thumb is that we want to return a batch iif **at least one** task in it matches the provided filter.
  - We don't need a `canceledBy` batch specific database because we can just retrieve the task and if it's a `taskCancelation` retrieve its `batchUid`
- The task cancelation has been updated and simplified a bit:
  - Instead of updating the matching tasks on disk while processing the cancelation task, we instead retrieve the task and let the `tick` function do the work afterward.
  - In the `tick` function, we now have to take care of not missing any tasks
- All the tests applied to the tasks were duplicated and updated to works with the new batches routes
- The deletion of batches doesn't contain any tests because it's already tested in the deletion of tasks (and especially highlighted in the snapshots)

Currently, one part of the PRD is not implemented: it's the progress.


crates/index-scheduler/src/utils.rs
```diff
@@ -25,10 +145,92 @@ impl IndexScheduler {
...
+        }
+        insert_task_datetime(wtxn, self.batch_started_at, batch.started_at, batch.uid)?;
+        insert_task_datetime(wtxn, self.batch_finished_at, batch.finished_at.unwrap(), batch.uid)?;
+
+        Ok(())
+    }
+
+    /// Convert an iterator to a `Vec` of tasks. The tasks MUST exist or a
+    /// `CorruptedTaskQueue` error will be throwed.
+    pub(crate) fn get_existing_tasks_with_processing_batch(
```
Kerollmops, line 216:
```suggestion
    /// Convert an iterator to a `Vec` of tasks and edit the `ProcessingBatch` to add the given tasks.
    ///
    /// The tasks MUST exist, or a `CorruptedTaskQueue` error will be thrown.
    pub(crate) fn get_existing_tasks_for_processing_batch(
```

Kerollmops (changes requested):

Kerollmops (approved):
That is very nice! GG!
bors merge

PR #5070, opened 2024-11-20
Improve the details and stats of the current batch processing

irevoire:
Small improvement we missed over https://github.com/meilisearch/meilisearch/pull/5060

The current batch processing had empty details and stats.

PR #5080, opened 2024-11-21
Fix getting a single batch through the GET route

dureuill:
# Pull Request

## Related issue
Fixes a bug where getting a single batch does not work

Related to #5070 

fix by @Kerollmops


Kerollmops (approved):
bors merge

irevoire:
Wow good catch, that was so dumb and yet since the test trying it was retrieving the batch containing the task 0 it worked I guess :grimacing: 
I'll add a new test on that specifically

Issue #5078, opened 2024-11-21
Indexer edition 2024 1.12 pre-release follow-ups


dureuill:

curquiza:
@Kerollmops @dureuill to clean

PR #5089, opened 2024-11-25
Improve error handling when writing into LMDB

Kerollmops:
This PR exposes two new internal error variants: `StoreDelete` and `StorePut`. So that the error messages are better when we fail at writing into LMDB.

Related to #5078


curquiza (commented):
⚠️ Be careful if you want v1.12, you should not merge on main

dureuill (approved):
LGTM

bors merge

PR #5090, opened 2024-11-25
Use the published crates versions

Kerollmops:
This PR uses the published versions of the obkv, grenad, and roaring crates in milli and Meilisearch.

Related to #5078.


dureuill (approved):
bors merge

PR #5107, opened 2024-12-02
While spamming the batches route we could see a processing batch becoming missing and then finished, this commit ensures the batches goes from processing to finished directly

irevoire:
# Pull Request

## Related issue
Fixes the failed tests from this PR: https://github.com/meilisearch/meilisearch-js/pull/1775
See [this message](https://meilisearch.slack.com/archives/CD7Q2UKGB/p1732784680450749) [private link] for more context

## What does this PR do?
- Ensure we never enter a state where a processing batches (only existing in RAM) becomes « Not found » by removing the processing batches AFTER writing them to disk
- This should also theoretically avoid an issue where a task could go from processing to enqueued and then finished

### DIFF
--- a/crates/index-scheduler/src/lib.rs
+++ b/crates/index-scheduler/src/lib.rs
@@ -1738,11 +1738,8 @@ impl IndexScheduler {
             }
         }
 
-        self.processing_tasks.write().unwrap().stop_processing();
         // We must re-add the canceled task so they're part of the same batch.
-        // processed.processing |= canceled;
         ids |= canceled;
-
         self.write_batch(&mut wtxn, processing_batch, &ids)?;
 
         #[cfg(test)]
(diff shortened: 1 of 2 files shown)


crates/index-scheduler/src/lib.rs
```diff
@@ -1738,20 +1738,21 @@ impl IndexScheduler {
...
 
         wtxn.commit().map_err(Error::HeedTransaction)?;
 
+        // We should stop processing AFTER everything is processed and written to disk otherwise, a batch (which only lives in RAM) may appear in the processing task
+        // and then become « not found » for some time until the commit everything is written and the final commit is made.
+        self.processing_tasks.write().unwrap().stop_processing();
+
         // Once the tasks are committed, we should delete all the update files associated ASAP to avoid leaking files in case of a restart
-        tracing::debug!("Deleting the update files");
+        // tracing::debug!("Deleting the update files");
```
dureuill, line 1755: uncomment or delete?

dureuill (dismissed):
one inline comment otherwise LGTM

crates/index-scheduler/src/lib.rs
irevoire: Oops, it was a typo on my side

irevoire, line 1755:
```suggestion
        tracing::debug!("Deleting the update files");
```

irevoire:
bors merge

Issue #5077, opened 2024-11-21
`tests::test_task_queue_is_full` fails on Mac Mx


dureuill:
**Describe the bug**
Unit test `tests::test_task_queue_is_full`

**To Reproduce**
1. On a Mac M1
2. `cargo test -- test_task_queue_is_full`
3. The test fails

**Expected behavior**
The test passes

**Screenshots**

**Meilisearch version:**
`9a08757a70e8a929b3328bba0032e54b052094c7`

**Additional context**

Tested on Mac M1

irevoire:
Fiexd by #5110

PR #5110, opened 2024-12-03
Increase margin on deletion of task

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/5077

## What does this PR do?
- Increase the margin we keep to enqueue task deletion

The issue was that we had not enough space on the reserved memory to write both the batch and the deletion task we just enqueued.
We could fix it only for this test as it’s not an issue in production where we have 10GiB of margin, but I thought it wasn’t a bad idea either to increase our margin a bit since we’re effectively writing more to lmdb.


dureuill (approved):
bors merge

PR #5119, opened 2024-12-04
Settings opt out error msg

ManyTheFish:
# Pull Request

## Related issue
PRD: https://meilisearch.notion.site/API-usage-Settings-to-opt-out-indexing-features-fff4b06b651f8108ade3f858aeb16b14?pvs=4
## What does this PR do?

Add a new error code and message when the user tries a facet search on an index where the facet search is disabled:
```json
{
  "message": "The facet search is disabled for this index",
  "code": "facet_search_disabled",
  "type": "invalid_request",
  "link": "https://docs.meilisearch.com/errors#invalid_facet_search_disabled"
}
 ```

### DIFF
--- a/crates/meilisearch/tests/search/facet_search.rs
+++ b/crates/meilisearch/tests/search/facet_search.rs
@@ -224,10 +224,10 @@ async fn add_documents_and_deactivate_facet_search() {
     assert_eq!(code, 400, "{}", response);
     snapshot!(response, @r###"
     {
-      "message": "Facet search is disabled for this index",
-      "code": "invalid_search_disabled_facet_search",
+      "message": "The facet search is disabled for this index",
+      "code": "invalid_facet_search_disabled",
       "type": "invalid_request",
-      "link": "https://docs.meilisearch.com/errors#invalid_search_disabled_facet_search"
+      "link": "https://docs.meilisearch.com/errors#invalid_facet_search_disabled"
     }
     "###);
 }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: crates/meilisearch/tests/search/facet_search.rs [hunk 1]
```diff
@@ -221,8 +221,15 @@ async fn add_documents_and_deactivate_facet_search() {
     let (response, code) =
         index.facet_search(json!({"facetName": "genres", "facetQuery": "a"})).await;
 
-    assert_eq!(code, 200, "{}", response);
-    assert_eq!(dbg!(response)["facetHits"].as_array().unwrap().len(), 0);
+    assert_eq!(code, 400, "{}", response);
+    snapshot!(response, @r###"
+    {
+      "message": "The facet search is disabled for this index",
+      "code": "invalid_facet_search_disabled",
```
guimachiavelli, line 228: Minor suggestion that might not fit the actual naming strategy in other parts of the program, but I find `invalid` a bit out of place here.

As in, `invalid_search_vector` makes sense because it indicates the search vector itself is not valid. `invalid_facet_search_disabled` is less clear—what exactly is invalid here?

That said, it could also be read as "Invalid: facet search disabled", I guess, but I'm not sure there's precedent in the existing code.

```suggestion
      "code": "facet_search_disabled",
```

dureuill: If we are seeking consistency, note that we already have the `feature_not_enabled` error code when trying to use an experimental feature without enabling it first.

Here it might make sense to have `facet_search_disabled` though, because the default state of facet search is enabled (so the issue is that it was disabled), while the default state of exp features is disabled (so the issue is that it was not enabled).

In summary, I support `facet_search_disabled` here.

Kerollmops (approved):
bors merge

PR #5122, opened 2024-12-04
Yield the BBQueue writing loop

Kerollmops:
We prefer yielding to let the writing thread do its job instead of spin looping.


ManyTheFish (approved):
bors merge

Issue #5112, opened 2024-12-03
[To be checked] Current batch running sometimes accumulate its details twice instead of one time


irevoire:
I have a meeting and don't have the time to detail everything that happened, but that should be investigated before the end of the week.
It seems like the stats generated in a batch sometimes count the number of tasks contained twice:
I added one document in a single document addition and the stats were showing two documents addition.

curquiza:
Fixed by #5123

Issue #5079, opened 2024-11-21
Duplicated tasks and wrong status in batch details


dureuill:
**Describe the bug**
The batch details reports two tasks being enqueued in the batch details, whereas we have only one task and it is in the processing state

**To Reproduce**
1. Send a `DocumentAdditionOrUpdate` task that will trigger a long indexing operation
2. Check that your task has spawned in the task queue:
```jsonc
// GET http://localhost:7700/tasks/0
{
  "uid": 0,
  "batchUid": 0,
  "indexUid": "mieli",
  "status": "processing",
  "type": "documentAdditionOrUpdate",
  "canceledBy": null,
  "details": {
    "receivedDocuments": 1500000,
    "indexedDocuments": null
  },
  "error": null,
  "duration": null,
  "enqueuedAt": "2024-11-21T16:38:06.145396Z",
  "startedAt": "2024-11-21T16:38:06.149318Z",
  "finishedAt": null
}
```
3. Get the list of batches
```jsonc
// GET http://localhost:7700/batches
{
  "results": [
    {
      "uid": 0,
      "details": {
        "receivedDocuments": 3000000,
        "indexedDocuments": null
      },
      "stats": {
        "totalNbTasks": 2,
        "status": {
          "enqueued": 2
        },
        "types": {
          "documentAdditionOrUpdate": 2
        },
        "indexUids": {
          "mieli": 2
        }
      },
      "duration": null,
      "startedAt": "2024-11-21T16:56:07.325266Z",
      "finishedAt": null
    }
  ],
  "total": 1,
  "limit": 20,
  "from": 0,
  "next": null
}
```
5. The details above indicate a batch of 2 tasks, even though there exists only 1 task
6. The tasks have status enqueued instead of processing.

**Expected behavior**
A batch of a single task with status processing.

**Meilisearch version:**
`9a08757a70e8a929b3328bba0032e54b052094c7`

irevoire:
Fixed by #5123

PR #5123, opened 2024-12-04
Fix batch details

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/5079
Fixes https://github.com/meilisearch/meilisearch/issues/5112

## What does this PR do?
- Make the processing tasks actually processing in the stats of the batch instead of enqueued
- Stop counting one extra task for all non-prioritized batches in the stats
- Add a test

### DIFF
--- a/crates/index-scheduler/src/utils.rs
+++ b/crates/index-scheduler/src/utils.rs
@@ -67,7 +67,7 @@ impl ProcessingBatch {
             task.batch_uid = Some(self.uid);
             // We don't store the statuses in the map since they're all enqueued but we must
             // still store them in the stats since that can be displayed.
-            *self.stats.status.entry(task.status).or_default() += 1;
+            *self.stats.status.entry(Status::Processing).or_default() += 1;
 
             self.kinds.insert(task.kind.as_kind());
             *self.stats.types.entry(task.kind.as_kind()).or_default() += 1;
--- a/crates/index-scheduler/src/batch.rs
+++ b/crates/index-scheduler/src/batch.rs
@@ -496,8 +496,7 @@ impl IndexScheduler {
 
         // 5. We make a batch from the unprioritised tasks. Start by taking the next enqueued task.
         let task_id = if let Some(task_id) = enqueued.min() { task_id } else { return Ok(None) };
-        let mut task = self.get_task(rtxn, task_id)?.ok_or(Error::CorruptedTaskQueue)?;
-        current_batch.processing(Some(&mut task));
+        let task = self.get_task(rtxn, task_id)?.ok_or(Error::CorruptedTaskQueue)?;
 
         // If the task is not associated with any index, verify that it is an index swap and
         // create the batch directly. Otherwise, get the index name associated with the task
--- a/crates/index-scheduler/src/lib.rs
+++ b/crates/index-scheduler/src/lib.rs
@@ -4319,10 +4319,35 @@ mod tests {
         let proc = index_scheduler.processing_tasks.read().unwrap().clone();
 
         let query = Query { statuses: Some(vec![Status::Processing]), ..Default::default() };
-        let (batches, _) = index_scheduler
-            .get_batch_ids_from_authorized_indexes(&rtxn, &proc, &query, &AuthFilter::default())
+        let (mut batches, _) = index_scheduler
+            .get_batches_from_authorized_indexes(query.clone(), &AuthFilter::default())
             .unwrap();
-        snapshot!(snapshot_bitmap(&batches), @"[0,]"); // only the processing batch in the first tick
+        assert_eq!(batches.len(), 1);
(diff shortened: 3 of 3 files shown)


crates/index-scheduler/src/snapshots/lib.rs/swap_indexes/third_empty_swap_processed.snap
```diff
@@ -89,7 +89,7 @@ succeeded [0,1,2,3,4,5,6,]
 ----------------------------------------------------------------------
 ### Batches Kind:
 "indexCreation" [0,1,2,3,]
-"indexSwap" [4,5,6,]
+"indexSwap" [4,5,]
```
dureuill, line 92: I'm not sure I get the change here, could you please explain it as a response to this comment?

crates/index-scheduler/src/utils.rs
```diff
@@ -67,7 +67,7 @@ impl ProcessingBatch {
             task.batch_uid = Some(self.uid);
             // We don't store the statuses in the map since they're all enqueued but we must
             // still store them in the stats since that can be displayed.
-            *self.stats.status.entry(task.status).or_default() += 1;
+            *self.stats.status.entry(Status::Processing).or_default() += 1;
```
dureuill, line 70: that's a way of achieving that 😅

crates/index-scheduler/src/snapshots/lib.rs/fail_in_update_task_after_process_batch_success_for_document_addition/after_batch_succeeded.snap
```diff
@@ -5,7 +5,7 @@ snapshot_kind: text
 ### Autobatching Enabled = true
 ### Processing batch Some(0):
 [0,]
-{uid: 0, details: {"receivedDocuments":2,"indexedDocuments":null}, stats: {"totalNbTasks":2,"status":{"enqueued":2},"types":{"documentAdditionOrUpdate":2},"indexUids":{"doggos":2}}, }
+{uid: 0, details: {"receivedDocuments":1,"indexedDocuments":null}, stats: {"totalNbTasks":1,"status":{"processing":1},"types":{"documentAdditionOrUpdate":1},"indexUids":{"doggos":1}}, }
```
dureuill, line 8: given how the snap is called `after_batch_succeeded.snap`, I would expect the status to be `succeeded`. Are we sure we should be seeing `enqueued`?

dureuill (commented):
Thanks!

A few inline comment + `perform_on_demand_snapshot` is flaky, should we fix it?

crates/index-scheduler/src/snapshots/lib.rs/fail_in_update_task_after_process_batch_success_for_document_addition/after_batch_succeeded.snap
irevoire: What's happening in this test is that we simulate a failure after successfully processing a batch, but **before** committing anything.
So, yes, it's expected that we still see the batch as processing because we've not committed anything yet.

Immediately after this is the next lines:
```rust
        // At this point the next time the scheduler will try to progress it should encounter
        // a critical failure and have to wait for 1s before retrying anything.

        let before_failure = Instant::now();
        handle.advance_till([Start]);
        snapshot!(snapshot_index_scheduler(&index_scheduler), name: "after_failing_to_commit");
```

crates/index-scheduler/src/snapshots/lib.rs/swap_indexes/third_empty_swap_processed.snap
irevoire: Good catch, that was actually a bug.
We handle the empty index swap tasks in another place from the normal index swap. I did not remember that, and I was just missing the task :sweat: 
It's fixed now, thanks :+1:

crates/meilisearch/tests/snapshot/mod.rs
```diff
@@ -129,11 +129,11 @@ async fn perform_on_demand_snapshot() {
...
 
-    server.index("doggo").create(Some("bone")).await;
-    index.wait_task(2).await;
+    let (task, _) = server.index("doggo").create(Some("bone")).await;
+    index.wait_task(task.uid()).await.succeeded();
 
-    server.index("doggo").create(Some("bone")).await;
-    index.wait_task(2).await;
+    let (task, _) = server.index("doggo").create(Some("bone")).await;
+    index.wait_task(task.uid()).await.failed();
```
irevoire, line 136: On my side, this fixed the flakyness of the test.
I hope it was its only cause

dureuill: We'll see in due time :+1: Thank you!

dureuill (approved):
Thank you 🌈 

bors merge

PR #5127, opened 2024-12-05
Log when an entry is too large to fit in the BBQueue

Kerollmops:


crates/milli/src/update/new/channel.rs
```diff
@@ -480,6 +480,8 @@ impl<'b> ExtractorBbqueueSender<'b> {
         let payload_header = EntryHeader::ArroySetVectors(arroy_set_vector);
         let total_length = EntryHeader::total_set_vectors_size(embeddings.len(), dimensions);
         if total_length > capacity {
+            tracing::trace!("We are spilling a large vector that is {total_length} bytes which is larger than the capacity of {capacity} bytes");
```
dureuill, line 483: this can make a lot of traces, is that OK?

Kerollmops:

PR #5130, opened 2024-12-05
Replace the HashMap caches by BTreeMaps

Kerollmops:
This PR is made to benchmark the time we could gain or lose by using BTreeMaps instead of HashMaps. We could gain time by avoiding large sorts when merging entries but lose some when inserting docids when extracting document content.


Kerollmops:

PR #5133, opened 2024-12-07
Finer facet selection

dureuill:
Introduce `DeltaDocument` that can iterate fields to return their `DeltaValue`:

- current value
- updated value
- current **and** updated value

Use this to establish a finer comparison for facet values in facet extraction. When a delta value is `CurrentAndUpdated`, check if it changed, and only register a change if the value actually changed.


dureuill:

dureuill:

ManyTheFish (approved):

dureuill:

dureuill:
Closed as non profitable enough

PR #5138, opened 2024-12-09
Allow xtask bench to proceed without a commit message

dureuill:


Kerollmops (approved):
le plaisir 😍
bors merge

PR #5140, opened 2024-12-09
Fix workload inversion

ManyTheFish:
The used assets were inverted between `workloads/hackernews-modify-facet-numbers.json`
and `workloads/hackernews-modify-facet-strings.json`, now fixed.


dureuill (approved):
bors merge

PR #5141, opened 2024-12-09
Use the right amount of max memory and not impact the settings

### LINKED ISSUE #5132: Change the new indexer memory usage to use 5% of the max indexing memory
As we keep the previous settings indexer and as it uses the max indexing memory in a different way than the new indexer, we cannot define this parameter the same way. We need to keep it like before but internally use only 5% of it for the new indexer.

Related to https://github.com/meilisearch/meilisearch/pull/5125#issue-2719850968 and https://github.com/meilisearch/engine-team/pull/73#discussion_r1871693843

Kerollmops:
Fixes #5132. Related to #5125.

### DIFF
--- a/crates/meilisearch/src/option.rs
+++ b/crates/meilisearch/src/option.rs
@@ -654,9 +654,8 @@ impl Opt {
 
 #[derive(Debug, Default, Clone, Parser, Deserialize)]
 pub struct IndexerOpts {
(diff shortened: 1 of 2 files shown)


ManyTheFish (approved):
bors merge

curquiza:
bors merge

PR #5152, opened 2024-12-11
Make xtasks be able to use the specified binary

Kerollmops:
Makes it possible to specify the binary to run. It is useful to run PGO optimized binaries.

### DIFF
--- a/crates/meilisearch/src/main.rs
+++ b/crates/meilisearch/src/main.rs
@@ -129,6 +129,11 @@ async fn try_main() -> anyhow::Result<()> {
 
     print_launch_resume(&opt, analytics.clone(), config_read_from);
 
+    tokio::spawn(async move {
+        tokio::signal::ctrl_c().await.unwrap();
+        std::process::exit(77);
+    });
+
     run_http(index_scheduler, auth_controller, opt, log_handle, Arc::new(analytics)).await?;
 
     Ok(())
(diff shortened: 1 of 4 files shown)


crates/xtask/src/bench/mod.rs
```diff
@@ -86,6 +86,10 @@ pub struct BenchDeriveArgs {
     /// The maximum time in seconds we allow for fetching the task queue before timing out.
     #[arg(long, default_value_t = 60)]
     tasks_queue_timeout_secs: u64,
+
+    /// The path to the binary to run. By default it compiles the binary with cargo.
```
dureuill, line 90:
```suggestion
    /// The path to the binary to run. 
    /// 
    /// If unspecified, runs `cargo run` after building Meilisearch with `cargo build`
```

crates/meilisearch/src/main.rs
```diff
@@ -129,6 +129,11 @@ async fn try_main() -> anyhow::Result<()> {
 
     print_launch_resume(&opt, analytics.clone(), config_read_from);
 
+    tokio::spawn(async move {
+        tokio::signal::ctrl_c().await.unwrap();
+        std::process::exit(77);
```
dureuill, line 134:
```suggestion
        std::process::exit(130);
```

dureuill (approved):
Thank you, see inline comments

dureuill:
(also we need to run rustfmt)

Kerollmops:
> The PR looks good. I'm just wondering: do we want to do such a significant change as starting to handle the SIGTERM signal so close of v1.12 release?

I think that's it is a really small change compared to all of the new indexer. We can afford this one. We also need it for future development (PGO) anyway.

dureuill (approved):
as I said the code looks good to me 😊

Kerollmops:
bors merge

dureuill:
bors merge

PR #5153, opened 2024-12-11
Return docid in case of errors while rendering the document template

dureuill:
Improves error message:

Before: 

```
ERROR index_scheduler: Batch failed Index `mieli`: user error: missing field in document: liquid: Unknown index
  with:
    variable=doc
    requested index=title
    available indexes=by, id, kids, parent, text, time, type
```

After:

```
ERROR index_scheduler: Batch failed Index `mieli`: user error: missing field in document `11345147`: liquid: Unknown index
  with:
    variable=doc
    requested index=title
    available indexes=by, id, kids, parent, text, time, type
```

### DIFF
--- a/crates/milli/src/prompt/error.rs
+++ b/crates/milli/src/prompt/error.rs
@@ -38,6 +38,16 @@ pub struct RenderPromptError {
     pub fault: FaultSource,
 }
 impl RenderPromptError {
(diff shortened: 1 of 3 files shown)


Kerollmops (approved):
bors merge

PR #5158, opened 2024-12-12
Indexer edition 2024 fix facet fst

ManyTheFish:
# Pull Request
Fix a regression in the new indexer; when several filterable attributes containing strings were set, all the field IDs were shifted, and the last one was overwriting the previous FST.

## What does this PR do?
- Add a test reproducing the bug
- fix the bug


Kerollmops (approved):
bors merge

PR #5159, opened 2024-12-12
Fix the New Indexer Spilling

Kerollmops:
Fix two bugs in the merging of the spilled caches. Thanks to @ManyTheFish and @irevoire 👏


irevoire (approved):
I take the entire responsibility in case of bug because of this PR
bors merge

PR #5166, opened 2024-12-16
fix list indexes

irevoire:
# Pull Request

### Smol benchmark on a meilisearch with 1009 indexes:

**Before** this PR on my computer, it was taking 5.5s to call the `GET /indexes` route on a cold computer where all the indexes were closed.
**After** this PR it takes 0.009s to call the route on the first 20 indexes, and 0.176 for the last 20 indexes (retrieving the first or last indexes on main has no impact on performances).

If my computations are right, that's between 61111.1% and 3125% faster on this test :joy: 

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4694

## What does this PR do?
- Add the primary key to the cache we already have in the index-mapper
- Provide a new route to retrieve the paginated indexes straight from the cache without opening them
- Fix a bug where the cache was not computed when loading a dump and was forcing us to open the indexes to compute their stats on the fly

## Is it breaking?

Since the field I added is an `Option` I think we should consider it as non-breaking and let it update itself automatically on the next operation of this index.
I also tested to run my patch over a DB generated on release-v1.12.0 and it works. The importing a dump also works.

### DIFF
--- a/crates/index-scheduler/src/lib.rs
+++ b/crates/index-scheduler/src/lib.rs
@@ -30,7 +30,7 @@ mod processing;
 mod utils;
 pub mod uuid_codec;
 
-pub type Result<T> = std::result::Result<T, Error>;
+pub type Result<T, E = Error> = std::result::Result<T, E>;
 pub type TaskId = u32;
 
 use std::collections::{BTreeMap, HashMap};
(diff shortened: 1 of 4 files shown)


crates/index-scheduler/src/lib.rs
```diff
@@ -30,7 +30,7 @@ mod processing;
 mod utils;
 pub mod uuid_codec;
 
-pub type Result<T> = std::result::Result<T, Error>;
+pub type Result<T, E = Error> = std::result::Result<T, E>;
```
dureuill, line 33: adding the default error type 😱 🤩

crates/meilisearch/src/routes/indexes/mod.rs
```diff
@@ -104,19 +104,20 @@ pub async fn list_indexes(
...
-            }
-            Ok(Some(
-                IndexView::new(uid.to_string(), index)
-                    .map_err(|e| Error::from_milli(e, Some(uid.to_string())))?,
-            ))
-        })?;
-    // Won't cause to open all indexes because IndexView doesn't keep the `Index` opened.
-    let indexes: Vec<IndexView> = indexes.into_iter().flatten().collect();
-    let ret = paginate.as_pagination().auto_paginate_sized(indexes.into_iter());
+    dbg!("here");
```
dureuill, line 107:
```suggestion
```

crates/meilisearch/src/routes/indexes/mod.rs
```diff
@@ -104,19 +104,20 @@ pub async fn list_indexes(
...
+    let (total, indexes) =
+        index_scheduler.get_paginated_indexes_stats(filters, *paginate.offset, *paginate.limit)?;
+    dbg!("hore");
```
dureuill, line 110:
```suggestion
```

dureuill (commented):
Nice, please remove some `dbg!`

dureuill (approved):
bors merge

PR #5182, opened 2024-12-21
Remove hard coded task ids to prevent flaky tests

### LINKED ISSUE #4840: Improve the meilisearch integration tests

mhmoudr:
# Pull Request

## Related issue
Fixes partial #4840

## What does this PR do?
- Mainly scan the test code for any hard coded task Id and replace it by the returned task Id once the action or task is performed on an index.
- PS: _PR is not split by files as it has one theme applied to all tests which make it easy to review_ 

## PR checklist

### DIFF
--- a/crates/meilisearch/tests/auth/tenant_token.rs
+++ b/crates/meilisearch/tests/auth/tenant_token.rs
@@ -100,11 +100,11 @@ macro_rules! compute_authorized_search {
         let index = server.index("sales");
         let documents = DOCUMENTS.clone();
         let (task1,_status_code) = index.add_documents(documents, None).await;
-        index.wait_task(task1.uid()).await;
+        index.wait_task(task1.uid()).await.succeeded();
         let (task2,_status_code) = index
             .update_settings(json!({"filterableAttributes": ["color"]}))
             .await;
-        index.wait_task(task2.uid()).await;
+        index.wait_task(task2.uid()).await.succeeded();
         drop(index);
 
         for key_content in ACCEPTED_KEYS.iter() {
@@ -147,7 +147,7 @@ macro_rules! compute_forbidden_search {
         let index = server.index("sales");
         let documents = DOCUMENTS.clone();
         let (task,_status_code) = index.add_documents(documents, None).await;
-        index.wait_task(task.uid()).await;
+        index.wait_task(task.uid()).await.succeeded();
         drop(index);
 
         for key_content in $parent_keys.iter() {
--- a/crates/meilisearch/tests/auth/tenant_token.rs
+++ b/crates/meilisearch/tests/auth/tenant_token.rs
@@ -146,7 +146,7 @@ macro_rules! compute_forbidden_search {
         server.use_admin_key("MASTER_KEY").await;
         let index = server.index("sales");
         let documents = DOCUMENTS.clone();
-        let (task,_status_code) = index.add_documents(documents, None).await;
+        let (task, _status_code) = index.add_documents(documents, None).await;
         index.wait_task(task.uid()).await.succeeded();
         drop(index);
 
--- a/crates/meilisearch/tests/auth/tenant_token.rs
+++ b/crates/meilisearch/tests/auth/tenant_token.rs
@@ -99,12 +99,12 @@ macro_rules! compute_authorized_search {
         server.use_admin_key("MASTER_KEY").await;
         let index = server.index("sales");
         let documents = DOCUMENTS.clone();
-        index.add_documents(documents, None).await;
-        index.wait_task(0).await;
-        index
+        let (task1,_status_code) = index.add_documents(documents, None).await;
+        index.wait_task(task1.uid()).await;
+        let (task2,_status_code) = index
             .update_settings(json!({"filterableAttributes": ["color"]}))
             .await;
-        index.wait_task(1).await;
+        index.wait_task(task2.uid()).await;
         drop(index);
 
         for key_content in ACCEPTED_KEYS.iter() {
@@ -146,8 +146,8 @@ macro_rules! compute_forbidden_search {
         server.use_admin_key("MASTER_KEY").await;
         let index = server.index("sales");
         let documents = DOCUMENTS.clone();
-        index.add_documents(documents, None).await;
-        index.wait_task(0).await;
+        let (task,_status_code) = index.add_documents(documents, None).await;
+        index.wait_task(task.uid()).await;
         drop(index);
 
         for key_content in $parent_keys.iter() {
--- a/crates/meilisearch/tests/batches/mod.rs
+++ b/crates/meilisearch/tests/batches/mod.rs
@@ -168,12 +168,9 @@ async fn list_batches_status_filtered() {
 async fn list_batches_type_filtered() {
     let server = Server::new().await;
     let index = server.index("test");
-    let (task, _status_code) = index.create(None).await;
+    let (task, _) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
-    index
-        .add_documents(serde_json::from_str(include_str!("../assets/test_set.json")).unwrap(), None)
-        .await;
-
+    let (task, _) = index.delete().await;
     let (response, code) = index.filtered_batches(&["indexCreation"], &[], &[]).await;
     assert_eq!(code, 200, "{}", response);
     assert_eq!(response["results"].as_array().unwrap().len(), 1);
--- a/crates/meilisearch/tests/batches/mod.rs
+++ b/crates/meilisearch/tests/batches/mod.rs
@@ -143,22 +143,18 @@ async fn list_batches_status_filtered() {
     let index = server.index("test");
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
-    let (task, _status_code) = index
-        .add_documents(serde_json::from_str(include_str!("../assets/test_set.json")).unwrap(), None)
-        .await;
+    let (task, _status_code) = index.create(None).await;
+    index.wait_task(task.uid()).await.failed();
 
     let (response, code) = index.filtered_batches(&[], &["succeeded"], &[]).await;
     assert_eq!(code, 200, "{}", response);
     assert_eq!(response["results"].as_array().unwrap().len(), 1);
 
-    // We can't be sure that the update isn't already processed so we can't test this
-    // let (response, code) = index.filtered_batches(&[], &["processing"]).await;
-    // assert_eq!(code, 200, "{}", response);
-    // assert_eq!(response["results"].as_array().unwrap().len(), 1);
-
-    index.wait_task(task.uid()).await.succeeded();
-
     let (response, code) = index.filtered_batches(&[], &["succeeded"], &[]).await;
+    assert_eq!(code, 200, "{}", response);
+    assert_eq!(response["results"].as_array().unwrap().len(), 1);
+
+    let (response, code) = index.filtered_batches(&[], &["succeeded", "failed"], &[]).await;
     assert_eq!(code, 200, "{}", response);
     assert_eq!(response["results"].as_array().unwrap().len(), 2);
 }
@@ -176,9 +172,13 @@ async fn list_batches_type_filtered() {
     assert_eq!(response["results"].as_array().unwrap().len(), 1);
 
     let (response, code) =
-        index.filtered_batches(&["indexCreation", "documentAdditionOrUpdate"], &[], &[]).await;
+        index.filtered_batches(&["indexCreation", "IndexDeletion"], &[], &[]).await;
     assert_eq!(code, 200, "{}", response);
     assert_eq!(response["results"].as_array().unwrap().len(), 2);
+
+    let (response, code) = index.filtered_batches(&["indexCreation"], &[], &[]).await;
+    assert_eq!(code, 200, "{}", response);
+    assert_eq!(response["results"].as_array().unwrap().len(), 1);
 }
 
 #[actix_rt::test]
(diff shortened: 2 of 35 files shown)


crates/meilisearch/tests/auth/tenant_token.rs
```diff
@@ -99,12 +99,12 @@ macro_rules! compute_authorized_search {
         server.use_admin_key("MASTER_KEY").await;
         let index = server.index("sales");
         let documents = DOCUMENTS.clone();
-        index.add_documents(documents, None).await;
-        index.wait_task(0).await;
-        index
+        let (task1,_status_code) = index.add_documents(documents, None).await;
+        index.wait_task(task1.uid()).await;
```
irevoire, line 103:
```suggestion
        index.wait_task(task1.uid()).await.succeeded();
```

irevoire (changes requested):
Hey, it seems good to me.
After calling the `.wait_task` method, could also call the `.succeeded()` or `.failed()` method just so we're sure we got the expected status?
Otherwise, that doesn't entirely prevent the flaky tests.

Also, just wondering, did you use a tool like coccinelle or something else to generate this PR?

mhmoudr:
Hey Thanks for review ... 
I have added the `.succeeded()` when possible, as in some calls the success check was performed manually after. 

PS1: tbh, I am not sure if this succeeded call will help stabilising the test as if the task have failed it is expected that the test will fail too, the earlier change was approaching the assumption about the task Ids which was hard coded/numbered everywhere in the test - happy to be corrected otherwise?

PS2: As for the change is was mostly find all and manual change as regx replace did play well for me in this case, many needed manual check.

curquiza:
Hello @mhmoudr
can you fix the git conflicts?
We made a use release yesterday, so we brought a lot of new code on `main`

mhmoudr:
Code is rebased to  latest origin/main.

mhmoudr:
**Don't merge** the PR I just noticed a rebase problem (some code is unintentionally removed) ... I will fix later

mhmoudr:
The rebase issue has been fixed, feel free to merge after review.

crates/meilisearch/tests/batches/mod.rs
```diff
@@ -167,17 +168,18 @@ async fn list_batches_status_filtered() {
...
     let server = Server::new().await;
     let index = server.index("test");
-    let (task, _) = index.create(None).await;
-    index.wait_task(task.uid()).await.succeeded();
-    let (task, _) = index.delete().await;
+    let (task,_status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
+    index
+        .add_documents(serde_json::from_str(include_str!("../assets/test_set.json")).unwrap(), None)
+        .await;
```
irevoire, line 175: Hey, this test is wrong now.
I think it comes down to a bad rebase, it's one of the test that I recently updated because it was flaky (and it still is after your PR :sweat_smile:).
Can you put back my version here

crates/meilisearch/tests/batches/mod.rs
```diff
@@ -1021,11 +1023,12 @@ async fn test_summarized_batch_cancelation() {
...
-    index.wait_task(0).await;
-    server.cancel_tasks("uids=0").await;
-    index.wait_task(1).await;
-    let (batch, _) = index.get_batch(1).await;
+    let (task,_status_code) = index.create(None).await;
+    index.wait_task(task.uid()).await.succeeded();
+    let (task,_status_code) = server.cancel_tasks("uids=0").await;
+    index.wait_task(task.uid()).await.succeeded();
+    //TODO: create a get_batch function interface that accepts u64, and remove the following cast.
+    let (batch, _) = index.get_batch(task.uid().to_u32().unwrap()).await;
```
irevoire, line 1031: Hey, you're right that it would be better to retrieve the actual batch used in the task.
But we should definitely not use the task uid as a batch uid here.
Either we keep the previous version which was using a `1` or if you have the time you could add a method very similar to the `.uid()` but named `.batch_uid()` that returns the batch uid.
Feel free to also change the signature of the `get_batch` function so it can accept whatever you want, it's not really important. 
In the Meilisearch code it cannot accept anything else than a `u32` though.

irevoire (changes requested):
Hey @mhmoudr it's awesome, I reviewed everything and I believe we're almost good to merge.
I left two small comments.
The CI is also broken because of a small warning, that should be easy to fix.
And from what I see you'll also need to run `cargo +nightly fmt` because a lot of formatting is wrong.

Then we should be good to merge :ok_hand: 
Thanks again for your interest in the project!

crates/meilisearch/tests/batches/mod.rs
mhmoudr: sure ... working on it

crates/meilisearch/tests/batches/mod.rs
mhmoudr: I'v tried to get batch ID from the JSON payload but I was not able to find it in the returned string 


```
{
  "taskUid": 1,
  "indexUid": null,
  "status": "enqueued",
  "type": "taskCancelation",
  "enqueuedAt": "[date]"
}
```

my implementation was 
```
pub fn batch_uid(&self) -> u64 {
        if let Some(batch_uid) = self["batch_uid"].as_u64() {
            batch_uid
        } else if let Some(batch_uid) = self["batchUid"].as_u64() {
            batch_uid
        } else {
            panic!("Didn't find any task batch id in: {self}");
        }
    }
```

For now I will flip it back to hard coded values of 1 as it was, unless you have another suggestions.

mhmoudr:
addressed all the mentioned issues, please review.

crates/meilisearch/tests/batches/mod.rs
mhmoudr: if interested in checking the error you can try pull changes from this commit (on a different branch): https://github.com/meilisearch/meilisearch/commit/0d380e005d2ef6c176afa7b23e47ce69082030f5

irevoire (changes requested):
Hey, thanks for the update!
I still found a few strange things but we're almost good :thinking:

crates/meilisearch/tests/batches/mod.rs
irevoire: Hey, yeah, it's because the `batchUid` is only present when retrieving a task from the task queue and not when enqueuing a new task.
In your commit, the call to the task queue is made on the `.wait_task` so it should be updated to something like;
```rust
let task = server.wait_task(task.uid());
task.succeeded();
let batch = task.batch_uid(); // now that works
```

But no problem with keeping the hard coded values :+1:

crates/meilisearch/tests/documents/add_documents.rs
```diff
@@ -1,4 +1,5 @@
 use actix_web::test;
+use actix_web::web::resource;
```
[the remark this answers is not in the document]
mhmoudr: removed ..

crates/meilisearch/tests/batches/mod.rs
```diff
@@ -170,14 +171,12 @@ async fn list_batches_type_filtered() {
     let (task, _) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
     let (task, _) = index.delete().await;
-    index.wait_task(task.uid()).await.succeeded();
-
```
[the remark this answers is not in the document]
mhmoudr: Added back .. 
the previous rebase seems to have messed with may files :(

ManyTheFish:
Hey @mhmoudr,
Let us know when you think the PR is ready to be reviewed again!

See you!

mhmoudr:
Hi @ManyTheFish , 
It is ready for review ..  as I am trying to minimise the recent changes just to address any feedback on the PR

ManyTheFish:
Hey @mhmoudr,
The test CI doesn't pass. Can you fix them before we can review your PR again, please? 😄 

`cargo test --locked --release --all` is the failing command ran by the CI.

I converted the PR to a draft; you can put it back as RFR when ready. This will let us know when we can review it.

Thank you!

irevoire (approved):
Hey @mhmoudr, 

We really need this PR merged, so I fixed the remaining issues myself.
Thanks again for your contribution :fire: 

bors merge

PR #5205, opened 2025-01-06
Incremental facets on v1.12

dureuill:
# Pull Request

## Related issue

Fixes https://github.com/meilisearch/meilisearch/issues/5213

## What does this PR do?
- Add `new_incremental`module that computes incremental facet indexing for the new indexer
- Change heuristics for incremental vs bulk facet choice: under 1000 operations is now always incremental, over 100_000 operations is now always bulk)
- Add sanity checks in debug

## Future improvements

- Use multi ops from Roaring to decrease the number of allocations
- Consider removing or adding multiple levels at once instead of max once per update
- Consider using information about the tree structure + the operations that were done (e.g. only addition) to avoid recomputing the group from all children 
- Consider making the algorithm parallel and looking into the roaring values to know which actually changed

### DIFF
--- a/crates/milli/src/update/facet/new_incremental.rs
+++ b/crates/milli/src/update/facet/new_incremental.rs
@@ -474,21 +474,4 @@ impl FacetsUpdateIncrementalInner {
 #[derive(Debug)]
 pub struct FacetFieldIdChange {
     pub facet_value: Box<[u8]>,
-    pub operation: FacetFieldIdOperation,
-}
-
-#[derive(Debug, Clone, Copy)]
-pub enum FacetFieldIdOperation {
-    /// The docids have been modified for an existing facet value
-    ///
-    /// The modification must be propagated to upper levels, without changing the structure of the tree
-    InPlace,
-    /// A new value has been inserted
-    ///
-    /// The modification must be propagated to upper levels, splitting nodes and adding new levels as necessary.
-    Insert,
-    /// An existing value has been deleted
-    ///
-    /// The modification must be propagated to upper levels, merging nodes and removing levels as necessary.
-    Remove,
 }
--- a/crates/milli/src/update/facet/new_incremental.rs
+++ b/crates/milli/src/update/facet/new_incremental.rs
@@ -144,12 +144,26 @@ impl FacetsUpdateIncrementalInner {
                             .prefix_iter_mut(wtxn, &parent_level_left_bound)?;
                         match parent_it.next() {
                             // 1. left of the current left bound, or
-                            Some(Ok((first_key, _first_value))) => 'change_left_bound: {
+                            Some(Ok((first_key, _first_value))) => {
                                 // make sure we don't spill on the neighboring fid (level also included defensively)
                                 if first_key.field_id != self.field_id
                                     || first_key.level != parent_level
                                 {
-                                    break 'change_left_bound;
+                                    // max level reached, exit
+                                    drop(parent_it);
+                                    self.compute_parent_group(
+                                        wtxn,
+                                        child_level,
+                                        child.facet_value,
+                                    )?;
+                                    for child in child_it.by_ref() {
+                                        self.compute_parent_group(
+                                            wtxn,
+                                            child_level,
+                                            child.facet_value,
+                                        )?;
+                                    }
+                                    return Ok(());
                                 }
                                 // remove old left bound
                                 unsafe { parent_it.del_current()? };
@@ -170,8 +184,6 @@ impl FacetsUpdateIncrementalInner {
                                         child.facet_value,
                                     )?;
                                 }
-
-                                break 'current_level;
                             }
                             Some(Err(err)) => return Err(err.into()),
                             // 2. max level reached, exit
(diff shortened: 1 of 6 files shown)


crates/milli/src/update/facet/new_incremental.rs
```diff
@@ -0,0 +1,477 @@
...
+                .map(|change| change.facet_value.into()),
+        )?;
+
+        self.inner.add_or_delete_level(wtxn)
+    }
+}
+
+impl FacetsUpdateIncrementalInner {
+    /// WARNING: `touched_children` must be sorted in **reverse** lexicographic order.
+    fn find_touched_parents(
```
ManyTheFish, line 85: I understand what you mean by `touched` but I am not sure about the name, what about `impacted`/`modified`

crates/milli/src/update/facet/new_incremental.rs
```diff
@@ -0,0 +1,477 @@
...
+        let mut last_parent: Option<Vec<u8>> = None;
+
+        for child in &mut touched_children {
+            if !valid_facet_value(&child) {
+                continue;
+            }
+
+            if let Some(last_parent) = &last_parent {
+                if child.as_slice() >= last_parent.as_slice() {
+                    self.compute_parent_group(wtxn, child_level, child)?;
```
ManyTheFish, line 105: I am not sure why we need to call `compute_parent_group`, is it in the case the current child is a totally new one and should be registered in the parent?

crates/milli/src/update/facet/new_incremental.rs
```diff
@@ -0,0 +1,477 @@
...
+        }
+        if !touched_parents.is_empty() {
+            self.find_touched_parents(
+                wtxn,
+                parent_level,
+                touched_parents
+                    // no need to `rev` here because the parents were already visited in reverse order
+                    .into_iter(),
+            )?;
+        }
```
ManyTheFish, line 191: I understand the recursive here, I am completely ok with it. I just want your thoughts about doing an iterative implementation using a BTreeSet<(level, Vec<u8>)>?

Maybe poping and pushing in a BTreeSet is not the best idea, and the recursive calls can't be that high. 🤔

crates/milli/src/update/index_documents/helpers/mod.rs
```diff
@@ -16,6 +16,10 @@ pub fn valid_lmdb_key(key: impl AsRef<[u8]>) -> bool {
     key.as_ref().len() <= MAX_WORD_LENGTH * 2 && !key.as_ref().is_empty()
 }
 
+pub fn valid_facet_value(facet_value: impl AsRef<[u8]>) -> bool {
+    facet_value.as_ref().len() <= (MAX_WORD_LENGTH * 2) - 3 && !facet_value.as_ref().is_empty()
+}
```
ManyTheFish, line 21: I know it has been copy-pasted, but there is a [MAX_LMDB_KEY_LENGTH](https://github.com/meilisearch/meilisearch/blob/f00a285a6d4d35d0184c14493120617934da0c90/crates/milli/src/lib.rs#L116-L117), 
LEGACY....... 😅
```suggestion
pub fn valid_facet_value(facet_value: impl AsRef<[u8]>) -> bool {
    facet_value.as_ref().len() <= MAX_LMDB_KEY_LENGTH - 3 && !facet_value.as_ref().is_empty()
}
```

ManyTheFish (changes requested):

crates/milli/src/update/facet/new_incremental.rs
dureuill: Since we are applying a list of changes, I suggest `changed` instead, would that do?

crates/milli/src/update/facet/new_incremental.rs
dureuill: `compute_parent_group` considers its argument (here, the `child`) as a group, and recomputes it from its own children.

When the level argument to `compute_parent_group` is `0`, then the call is a no-op.

Otherwise, we need to compute the group before discarding the `child`, otherwise some groups will not be refreshed.

I hope this clarifies this part of the algorithm, available for a call if more information is needed

crates/milli/src/update/facet/new_incremental.rs
dureuill: the recursive calls are max 255, since level is a u8 and we exit early if it is u8::MAX. Also to reach 255 level we need a number of nodes whose `log4` equals 255, which requires approximatively `1_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000` nodes. Actual values we reach is level 12 which requires 15 millions of nodes already. `u32::MAX` (max number of documents) fits in 16 levels.

However, as the call is terminal, it is true that it could be rewritten as a for loop, no BTree necessary.

crates/milli/src/update/index_documents/helpers/mod.rs
dureuill: done

ManyTheFish:
bors retry

Kerollmops:
bors retry

Kerollmops:
bors merge

curquiza:
bors merge

PR #5210, opened 2025-01-06
Improve test performance of get_index.rs 

DerTimonius:
# Pull Request

## Related issue
related to #4840

## What does this PR do?
This PR aims to improve the performance of the tests in `get_index.rs`.

There is a small issue though: 
the `list_multiple_indexes` test works great when ran alone, but when running with other tests it fails with a `corrupted task queue` error. I guess this has something to do with using a shared server, but I was not really able to pinpoint the issue.

Also, the `no_index_return_empty_list` does not work a shared server (as there now will always be at least one index on the server) and I was not really sure if rebuilding the whole suite for `get_and_paginate_indexes` should be viable? While waiting for feedback on the issue mentioned above, I'll try to change the `get_and_paginate_indexes` test so that it can use the shared server

## PR checklist


irevoire:
Hey @DerTimonius thank for your interest, and sorry for the delay :pray: 

> the list_multiple_indexes test works great when ran alone, but when running with other tests it fails with a corrupted task queue error. I guess this has something to do with using a shared server, but I was not really able to pinpoint the issue.

This was a bug on main that I fixed yesterday: https://github.com/meilisearch/meilisearch/pull/5239

> Also, the no_index_return_empty_list does not work a shared server (as there now will always be at least one index on the server)

Yes I think you're right, we should not use a shared server in these tests :+1:

crates/meilisearch/tests/index/get_index.rs
```diff
@@ -26,13 +26,13 @@ async fn create_and_get_index() {
 
 #[actix_rt::test]
 async fn error_get_unexisting_index() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
```
irevoire, line 30: I believe you can directly use the special [`shared_does_not_exists_index`](https://github.com/meilisearch/meilisearch/blob/8f006eeaf3c001a0fe72a43772bfb871e588debf/crates/meilisearch/tests/common/mod.rs#L163C14-L169) here 

```suggestion
    let index = shared_does_not_exists_index();
```

crates/meilisearch/tests/index/get_index.rs
```diff
@@ -26,13 +26,13 @@ async fn create_and_get_index() {
...
 
     let (response, code) = index.get().await;
 
     let expected_response = json!({
-        "message": "Index `test` not found.",
+        "message": format!("Index `{}` not found.", index.uid),
```
irevoire, line 35: If we can do the previous comment, then we can hardcode the index name here:
```suggestion
        "message": "Index `DOES_NOT_EXISTS` not found.",
```

crates/meilisearch/tests/index/get_index.rs
```diff
@@ -53,19 +53,25 @@ async fn no_index_return_empty_list() {
...
-    server.index("test1").create(Some("key")).await;
+    let server = Server::new_shared();
+    let index1 = server.unique_index();
+    let index2 = server.unique_index();
 
-    server.index("test").wait_task(1).await;
+    let (task1, _) = index1.create(None).await;
+    let (_task2, _) = index2.create(Some("key")).await;
+
+    index1.wait_task(task1.uid()).await.succeeded();
```
irevoire, line 63: The `wait_task(1)` was referring to the second task :sweat_smile: 
That's also what was causing your task queue corruption I believe (+ the fact we had a bug on main)

```suggestion
    index1.wait_task(task2.uid()).await.succeeded();
```

crates/meilisearch/tests/index/get_index.rs
```diff
@@ -53,19 +53,25 @@ async fn no_index_return_empty_list() {
...
     assert!(response["results"].is_array());
     let arr = response["results"].as_array().unwrap();
-    assert_eq!(arr.len(), 2);
-    assert!(arr.iter().any(|entry| entry["uid"] == "test" && entry["primaryKey"] == Value::Null));
-    assert!(arr.iter().any(|entry| entry["uid"] == "test1" && entry["primaryKey"] == "key"));
+    assert!(arr.len() >= 2);
+    assert!(arr
+        .iter()
+        .any(|entry| entry["uid"] == index1.uid && entry["primaryKey"] == Value::Null));
+    assert!(arr.iter().any(|entry| entry["uid"] == index2.uid && entry["primaryKey"] == "key"));
```
irevoire, line 74: I think you're right, for this test we should not use a shared index otherwise it's practically impossible to do these asserts

DerTimonius:
@irevoire thanks! I will rebase this and check if this fixes the issues

irevoire (changes requested):
Hey @DerTimonius,

Thanks a lot for your interest and sorry for the huge delay :pray: 

> the list_multiple_indexes test works great when ran alone, but when running with other tests it fails with a corrupted task queue error. I guess this has something to do with using a shared server, but I was not really able to pinpoint the issue.

That was a bug in the engine on the `main` branch. I found and fixed it only yesterday: https://github.com/meilisearch/meilisearch/pull/5239

> Also, the no_index_return_empty_list does not work a shared server (as there now will always be at least one index on the server) and I was not really sure if rebuilding the whole suite for get_and_paginate_indexes should be viable? While waiting for feedback on the issue mentioned above, I'll try to change the get_and_paginate_indexes test so that it can use the shared server

Huum tbh after reviewing the test I think you should stop using a shared server.
Since the very purpose of the test is to check something on the whole index list we don't have the choice but to have the control over the whole index list. Thus without a shared server :pensive:

crates/meilisearch/tests/index/get_index.rs
DerTimonius: oh..did not realize that it was supposed to be the second task 😅

crates/index-scheduler/src/scheduler/test.rs
```diff
@@ -899,16 +899,16 @@ fn create_and_list_index() {
 
     index_scheduler.index("kefir").unwrap();
     let list = index_scheduler.get_paginated_indexes_stats(&AuthFilter::default(), 0, 20).unwrap();
-    snapshot!(json_string!(list, { "[1][0][1].created_at" => "[date]", "[1][0][1].updated_at" => "[date]" }), @r#"
+    snapshot!(json_string!(list, { "[1][0][1].created_at" => "[date]", "[1][0][1].updated_at" => "[date]" }), @r###"
```
irevoire, line 902: Hey, looks like this test was badly written by me sorry :grimacing: 
Here's the fix

```suggestion
    snapshot!(json_string!(list, { "[1][0][1].created_at" => "[date]", "[1][0][1].updated_at" => "[date]", "[1][0][1].database_size" => "[bytes]", "[1][0][1].used_database_size" => "[bytes]" }), @r###"
```

irevoire (changes requested):

crates/meilisearch/tests/index/get_index.rs
DerTimonius: I somehow missed this comment 🙈 
reverted back to the old version as I could not come up with a feasible way to get the asserts running

irevoire (dismissed):
Hey @DerTimonius, that's perfect, thanks.

We really need to merge https://github.com/meilisearch/meilisearch/pull/5264 first, and I'm afraid you'll have a conflict.
I'll try to merge your PR on monday

DerTimonius:
@irevoire let me know once you merged it, I'll try to solve the conflicts 👍

irevoire:
@DerTimonius Awesome thanks, it just landed, and as expected, there is a small conflict sorry :pray:

DerTimonius:
@irevoire resolved the merge conflicts and the tests are still looking good 👍

irevoire (approved):
Awesome, thanks a lot @DerTimonius 

bors merge

irevoire:
bors cancel

irevoire (approved):
Oops, I just had to remove a few unused files

bors merge

PR #5224, opened 2025-01-09
Fix facet distribution

### LINKED ISSUE #5222: Facet distribution is incorrect when using PUT documents
**Describe the bug**
When updating documents incrementally with the PUT method, without specifying all the fields of the documents, then the fields that are not repeated are mistakenly removed from the field distribution.

**To Reproduce**
1. POST a document with `{"id": 0, "some_field": "aaa", "other_field": "aaa" }`
2. Index stats reports: `{ "fieldDistribution": {"id": 1, "some_field": 1, "other_field": 1}`
3. PUT an update to that document `{"id": 0, "some_field": "bbb"}`. `other_field` is not repeated so keeps its `aaa` value.
4. Index stats now reports `{ "fieldDistribution": {"id": 1, "some_field": 1}`

**Expected behavior**
Field distribution should not have changed

**Meilisearch version:**
v1.12.0..v1.12.2

dureuill:
# Pull Request

## Related issue
Fixes #5222 

## What does this PR do?
- Fixes the computation of facet distribution when incrementally updating documents
- Add a unit test that fails in v1.12.2 and succeeds after the fix
- Add support to `meilitool` for offline upgrade to v1.12.3 by rebuilding the facet distribution.

### DIFF
--- a/crates/meilitool/Cargo.toml
+++ b/crates/meilitool/Cargo.toml
@@ -14,11 +14,12 @@ arroy_v04_to_v05 = { package = "arroy", git = "https://github.com/meilisearch/ar
 clap = { version = "4.5.9", features = ["derive"] }
 dump = { path = "../dump" }
 file-store = { path = "../file-store" }
-indexmap = {version = "2.7.0", features = ["serde"]}
+indexmap = { version = "2.7.0", features = ["serde"] }
 meilisearch-auth = { path = "../meilisearch-auth" }
 meilisearch-types = { path = "../meilisearch-types" }
+milli = { path = "../milli" }
 serde = { version = "1.0.209", features = ["derive"] }
-serde_json = {version = "1.0.133", features = ["preserve_order"]}
+serde_json = { version = "1.0.133", features = ["preserve_order"] }
 tempfile = "3.14.0"
 time = { version = "0.3.36", features = ["formatting", "parsing", "alloc"] }
 uuid = { version = "1.10.0", features = ["v4"], default-features = false }
--- a/crates/meilitool/Cargo.toml
+++ b/crates/meilitool/Cargo.toml
@@ -17,7 +17,6 @@ file-store = { path = "../file-store" }
 indexmap = { version = "2.7.0", features = ["serde"] }
 meilisearch-auth = { path = "../meilisearch-auth" }
 meilisearch-types = { path = "../meilisearch-types" }
-milli = { path = "../milli" }
 serde = { version = "1.0.209", features = ["derive"] }
 serde_json = { version = "1.0.133", features = ["preserve_order"] }
 tempfile = "3.14.0"
--- a/crates/meilitool/src/upgrade/v1_12.rs
+++ b/crates/meilitool/src/upgrade/v1_12.rs
@@ -23,7 +23,12 @@ use uuid::Uuid;
 use crate::try_opening_database;
 use crate::uuid_codec::UuidCodec;
 
-pub fn v1_11_to_v1_12(db_path: &Path) -> anyhow::Result<()> {
+pub fn v1_11_to_v1_12(
+    db_path: &Path,
+    _origin_major: &str,
+    _origin_minor: &str,
+    _origin_patch: &str,
+) -> anyhow::Result<()> {
     println!("Upgrading from v1.11.0 to v1.12.0");
 
     convert_update_files(db_path)?;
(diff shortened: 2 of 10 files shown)


crates/meilitool/src/upgrade/v1_12.rs
```diff
@@ -1,24 +1,58 @@
...
 
+pub fn v1_12_to_v1_12_3(
+    db_path: &Path,
+    origin_major: &str,
+    origin_minor: &str,
+    origin_patch: &str,
+) -> anyhow::Result<()> {
+    println!("Upgrading from v1.12.{{0, 1, 2}} to v1.12.3");
+
+    if origin_minor == "12" {
```
irevoire, line 47: If this function is called, then we're guaranteed that the previous version was the v1.12.{0,1,2}.
We don't need this check, and we don't need to retrieve the previous version, I think 🤔

crates/meilitool/src/upgrade/v1_12.rs
```diff
@@ -77,3 +111,184 @@ fn convert_update_files(db_path: &Path) -> anyhow::Result<()> {
...
+                })?;
+
+            let mut index_txn = index.write_txn()?;
+
+            milli::update::new::reindex::field_distribution(&index, &mut index_txn, &progress)
+                .context("while rebuilding field distribution")?;
+
+            let stats = IndexStats::new(&index, &index_txn)
+                .with_context(|| format!("computing stats for index `{uid}`"))?;
+            store_stats_of(stats_db, uuid, &mut sched_wtxn, &uid, &stats)?;
```
irevoire, line 188: Very nice you thought about that 👌

crates/milli/src/update/new/reindex.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+#[derive(Default)]
+pub struct FieldDistributionIndexProgress;
+
+impl NamedStep for FieldDistributionIndexProgress {
+    fn name(&self) -> &'static str {
+        "documents"
+    }
+}
```
irevoire, line 47: I don't think we use it anywhere?

irevoire (changes requested):

crates/meilitool/src/upgrade/v1_12.rs
dureuill: There are two cases here:

1. This function is called in a chain of calls coming from e.g. v1.10 to v1.12.3:
    1. v1.10 to v1.11
    2. v1.11 to v1.12
    3. v1.12 to v1.12.3
2. This function is directly called from v1.12.x:
    1. v1.12 to v1.12.3

We only want to rebuild facet distribution in case (2), as it wasn't corrupted in case (1).
The functions were updated to accept the `origin_major`, `origin_minor`, `origin_patch`, which are the version of the DB **before any upgrade operation started**.

Thus, it should allow us to distinguish between case (1) and case (2)

crates/meilitool/src/upgrade/v1_12.rs
dureuill: (ahaha, I actually did not. I saw it in testing)

crates/milli/src/update/new/reindex.rs
dureuill: true that, removing

curquiza:
⚠️ @dureuill, I think you know it, but just in case because it can be dangerous -> the current merge is in v1.12.2 and not v1.12.3

dureuill:
thanks curqui, no I wasn't aware

crates/meilitool/src/upgrade/v1_12.rs
irevoire: ah yes got it! 👍

crates/meilitool/Cargo.toml
```diff
@@ -14,11 +14,12 @@ arroy_v04_to_v05 = { package = "arroy", git = "https://github.com/meilisearch/ar
 clap = { version = "4.5.9", features = ["derive"] }
 dump = { path = "../dump" }
 file-store = { path = "../file-store" }
-indexmap = {version = "2.7.0", features = ["serde"]}
+indexmap = { version = "2.7.0", features = ["serde"] }
 meilisearch-auth = { path = "../meilisearch-auth" }
 meilisearch-types = { path = "../meilisearch-types" }
+milli = { path = "../milli" }
```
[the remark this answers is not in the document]
dureuill: done

crates/meilitool/src/upgrade/v1_12.rs
```diff
@@ -1,24 +1,58 @@
...
+use std::sync::atomic::AtomicBool;
 
 use anyhow::Context;
 use file_store::FileStore;
 use indexmap::IndexMap;
 use meilisearch_types::milli::documents::DocumentsBatchReader;
+use milli::heed::types::{SerdeJson, Str};
+use milli::heed::{Database, EnvOpenOptions, RoTxn, RwTxn};
+use milli::progress::Step;
+use milli::{FieldDistribution, Index};
```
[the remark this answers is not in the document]
dureuill: done

crates/milli/src/update/new/reindex.rs
dureuill: done

Kerollmops (approved):
bors merge

irevoire (approved):
perfecto :ok_hand:

Kerollmops:
bors merge

dureuill:
I don't really get this 🤔 

@Kerollmops think you can merge directly?

Issue #5228, opened 2025-01-13
Panicking after trying to delete unknown entries while indexing documents


Kerollmops:
Meilisearch is panicking while trying to delete an unknown entry from an index database. This has been reported internally, and more information can be found [in this Slack thread](https://meilisearch.slack.com/archives/C05DH9MM7JP/p1736513253729999). From the customer:

> We initially saw the same error (different key) in meilisearch 1.12.1, and then upgraded to 1.12.2 and saw the issue there. It looks like incremental facet indexing was introduced in 1.12.2 - so it sounds like that change can be ruled out.

```
panicked at crates/milli/src/update/new/indexer/mod.rs:586:29: internal error: entered unreachable code: We tried to delete an unknown key: [0, 1, 2, 54, 20, 3, 104, 97, 108, 97, 32, 101, 108, 45, 103, 104, 97, 122, 97, 108, 121]
```

The following is a better error message coming from v1.12.3. It explicitly tells us that the issue comes from deleting an unknown key from the `field-id-docid-facet-strings` database.

```
ERROR meilisearch: info=panicked at crates/milli/src/update/new/indexer/mod.rs:587:29: internal error: entered unreachable code: We tried to delete an unknown key from field-id-docid-facet-strings: "\0\x01\x026\x14\x03hala el-ghazaly"
```

**To Reproduce**
1. Send a lot of documents into an index.
2. Wait for the panic to happen.

**Expected behavior**
Meilisearch shouldn't panic but, more importantly, shouldn't delete unknown entries from the index.

**Meilisearch version:**
v1.12.2

**Additional context**
Meilisearch was running on a Linux.

dureuill:
From the key, it was in the `field_id_docid_facet_strings` DB.

- 2 bytes of fid
- 4 bytes of docid
- string

curquiza:
Closed by #5258

PR #5229, opened 2025-01-13
Improve the panic message when deleting an unknown entry

Kerollmops:
This pull request is related to #5228 and improves the error message so that we can better debug the issue. It displays the database in which the deletion tentative happened and displays the key as an escaped string instead of an array of bytes.


dureuill (approved):
bors merge

Kerollmops:
bors merge

PR #5230, opened 2025-01-13
Use BitPacker8x to measure performances

Kerollmops:
This PR is an experiment using [a `BitPacker8x` instead of a `BitPacker4x`](https://docs.rs/bitpacking/latest/bitpacking/index.html) in [`Bbbul`](https://docs.rs/bumparaw-collections/latest/bumparaw_collections/bbbul/struct.Bbbul.html). Keep in mind that the benchmarks will show the performance of the benchmark machine, which is an intel machine (massive CPU: Intel(R) Xeon(R) E-2246G CPU @ 3.60GHz).

### DIFF
--- a/crates/milli/src/update/new/extract/cache.rs
+++ b/crates/milli/src/update/new/extract/cache.rs
@@ -69,7 +69,7 @@ use std::io::BufReader;
 use std::{io, iter, mem};
 
 use bumpalo::Bump;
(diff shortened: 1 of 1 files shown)


Kerollmops:
Ok, it doesn't change much. It also probably means that the Bbbul type is shadowed by something else (grenad merges, Roaring bitmaps unions...).

PR #5234, opened 2025-01-13
Parse ollama URL to adapt configuration depending on the endpoint

### LINKED ISSUE #5002: ensure forward compatibility with ollama
As outlined in #4983, the ollama endpoint that Meilisearch has been using (`/api/embeddings`) is deprecated.

A new `/api/embed` endpoint was introduced in v0.3.4.

Switching to the new endpoint has advantages, as the new endpoint supports batching, and the backward compatibility policy towards the deprecated endpoint is unclear (it is prudent to avoid using deprecated endpoints).

However, it breaks compatibility with `ollama < v0.3.4`

This issue is about determining a compatibility policy for Meilisearch with ollama, and making the necessary code changes to implement it.

The current proposal is given below as an RFC:

1. the `url` parameter now supports either the `/api/embed` and `/api/embeddings` endpoints.
2. internally, Meilisearch will use the API that matches the selected endpoint
3. it is technically a breaking change, because passing an unsupported endpoint will now trigger an error on Meilisearch's end. I don't think this breaking change will have any impact in practice, because using a different endpoint was already an error on ollama's end.

---

Alternatives:

1. Change url meaning:
    1. Change the meaning of `MEILI_OLLAMA_URL` and of the `url` parameter to require it points to the **root url** of the ollama server. This is a breaking change, because it is currently required to point to the direct endpoint that must be called
    4. when spawning the underlying REST embedder, try the newer endpoint first. If it succeeds, then use that. If it fails, then try the older endpoint.
   5. This may require again that the ollama server is up at meilisearch configuration time

dureuill:
# Pull Request

## Related issue
Fixes #5002 

## What does this PR do?
- Parses `url` parameter of `ollama` to recognize supported endpoint and adapt the REST configuration to the recognized endpoint
- Throws a new error if no endpoint is recognized
- Add a test for the various recognized endpoints

Thanks to @Guikingone for the original report and PR

### DIFF
--- a/crates/milli/src/vector/error.rs
+++ b/crates/milli/src/vector/error.rs
@@ -67,7 +67,7 @@ pub enum EmbedErrorKind {
     #[error("could not authenticate against {embedding} server{server_reply}{hint}", embedding=match *.1 {
         ConfigurationSource::User => "embedding",
         ConfigurationSource::OpenAi => "OpenAI",
-        ConfigurationSource::Ollama => "ollama"
+        ConfigurationSource::Ollama => "Ollama"
     },
     server_reply=option_info(.0.as_deref(), "server replied with "),
     hint=match *.1 {
@@ -306,6 +306,10 @@ impl NewEmbedderError {
             fault: FaultSource::User,
         }
     }
+
+    pub(crate) fn ollama_unsupported_url(url: String) -> NewEmbedderError {
+        Self { kind: NewEmbedderErrorKind::OllamaUnsupportedUrl(url), fault: FaultSource::User }
+    }
 }
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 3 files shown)


crates/milli/src/vector/error.rs
```diff
@@ -67,7 +67,7 @@ pub enum EmbedErrorKind {
     #[error("could not authenticate against {embedding} server{server_reply}{hint}", embedding=match *.1 {
         ConfigurationSource::User => "embedding",
         ConfigurationSource::OpenAi => "OpenAI",
-        ConfigurationSource::Ollama => "ollama"
+        ConfigurationSource::Ollama => "Ollama"
```
Kerollmops, line 70: Do we need to inform the gui about this change?

Kerollmops (approved):
Nice changes! Thank you!
bors merge

crates/milli/src/vector/error.rs
dureuill: hmmm, I'm not sure, it looks pretty minor?

PR #5237, opened 2025-01-14
Bring back v1.12.3 changes into main

dureuill:
This brings back the (already reviewed) changes of v1.12.3 into main:

1. fix the field distribution issue
2. improve the error message when trying to delete a non-existing key


irevoire (approved):
Thank you! :dog2: 

bors merge

Issue #5238, opened 2025-01-15
Internal error: Corrupted task queue when retrieving index


irevoire:
**Describe the bug**
When an index is created, if we try to retrieve it immediately while it's being created, we can trigger an internal error.
The issue was discovered in the meilisearch-ruby SDK CI.
It was introduced a few days/weeks ago on main and has never been released. 

The bug was revealed by: https://github.com/meilisearch/meilisearch/pull/5166
But the previous behavior has always been a bit buggy, it just wouldn't throw an internal error and instead return dummy values for the dates.

**To Reproduce**
1. See the test I wrote in the first commit

**Expected behavior**
It doesn't crash.

**Screenshots**
![Image](https://github.com/user-attachments/assets/e30630f4-e92c-44f7-a963-1713479c2288)

**Meilisearch version:**
None, only present on main

PR #5239, opened 2025-01-15
Fix corrupted task queue errors on index creation

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/5238

## What does this PR do?
- Add a test that reproduces the issue and ensure we never introduce the bug again
- Fix the bug by storing the stats of the index upon creation instead of waiting for the update index task to do it


dureuill (approved):
bors merge

dureuill:
bors retry

PR #5242, opened 2025-01-16
Fix infinite loop

dureuill:
- Fix possible infinite loop by releasing `writer_receiver` as soon as writing to DB panics
- Demote panic to error log


Kerollmops (approved):
Merci Louis 👍
bors merge

dureuill:
bors retry

PR #5246, opened 2025-01-16
Fix dump import

dureuill:
- Fix: handle the change of format of the update files
  - Correctly handle update files as JSON stream rather than obkv when exporting a dump with enqueued tasks
  - Correctly recreate update files as JSON stream rather than obkv when importing a dump
  - As the dump format itself didn't change, all dumps are still compatible
- Temporary workaround for https://github.com/meilisearch/meilisearch/issues/5247: set the batch uid of tasks to `null` at dump export time.
- Changes to meilitool
  - Export dump with update files in new format if DB >= v1.12
  - offline upgrade now supports upgrading from [1.9.0-1.12.5] to [1.10.0-1.12.5].
  - offline upgrade supports no-op upgrades and has better error messages

### DIFF
--- a/crates/index-scheduler/src/batch.rs
+++ b/crates/index-scheduler/src/batch.rs
@@ -819,6 +819,13 @@ impl IndexScheduler {
                         t.started_at = Some(started_at);
                         t.finished_at = Some(finished_at);
                     }
+
+                    // Patch the task to remove the batch uid, because as of v1.12.5 batches are not persisted.
+                    // This prevent from referencing *future* batches not actually associated with the task.
+                    //
+                    // See <https://github.com/meilisearch/meilisearch/issues/5247> for details.
+                    t.batch_uid = None;
+
                     let mut dump_content_file = dump_tasks.push_task(&t.into())?;
 
                     // 2.1. Dump the `content_file` associated with the task if there is one and the task is not finished yet.
--- a/crates/index-scheduler/src/batch.rs
+++ b/crates/index-scheduler/src/batch.rs
@@ -829,21 +829,18 @@ impl IndexScheduler {
                         if status == Status::Enqueued {
                             let content_file = self.file_store.get_update(content_file)?;
 
-                            let reader = DocumentsBatchReader::from_reader(content_file)
-                                .map_err(|e| Error::from_milli(e.into(), None))?;
-
-                            let (mut cursor, documents_batch_index) =
-                                reader.into_cursor_and_fields_index();
-
-                            while let Some(doc) = cursor
-                                .next_document()
-                                .map_err(|e| Error::from_milli(e.into(), None))?
+                            for document in
+                                serde_json::de::Deserializer::from_reader(content_file).into_iter()
                             {
-                                dump_content_file.push_document(
-                                    &obkv_to_object(doc, &documents_batch_index)
-                                        .map_err(|e| Error::from_milli(e, None))?,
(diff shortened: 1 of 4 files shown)


crates/meilitool/src/main.rs
```diff
@@ -254,18 +252,39 @@ fn export_a_dump(
...
-                    while let Some(doc) = cursor.next_document().with_context(|| {
-                        format!("While iterating on content file {:?}", content_file_uuid)
-                    })? {
-                        dump_content_file
-                            .push_document(&obkv_to_object(doc, &documents_batch_index)?)?;
+                    if (
+                        detected_version.0.as_str(),
+                        detected_version.1.as_str(),
+                        detected_version.2.as_str(),
+                    ) < ("1", "12", "0")
```
irevoire, line 259: Nice, I didn't know we could do that

crates/meilitool/src/upgrade/mod.rs
```diff
@@ -54,17 +87,27 @@ impl OfflineUpgrade {
             ("1", "10", _) => 0,
             ("1", "11", _) => 1,
             ("1", "12", x) if x == "0" || x == "1" || x == "2" => 2,
-            ("1", "12", "3") => 3,
+            ("1", "12", x) if x == "3" || x == "4" || x == "5" => 3,
```
irevoire, line 90:
```suggestion
            ("1", "12", 3 | 4 | 5) => 3,
```

I you want

irevoire (dismissed):
Good to me!

crates/meilitool/src/main.rs
dureuill: this is using the blanked Ord impl of `(T,U,V)` for `T, U, V: Ord`. https://doc.rust-lang.org/std/cmp/trait.Ord.html#impl-Ord-for-(T,)

It is doing lexicographic comparisons, which is what we want.

crates/meilitool/src/upgrade/mod.rs
dureuill: oh yes that is much better

dureuill: done

crates/meilitool/src/main.rs
irevoire: :fire:

irevoire (approved):
Re-approving, but feel free to wait for kero review because he know more than me on how the new-indexer works :sweat_smile:

crates/index-scheduler/src/batch.rs
```diff
@@ -819,6 +819,13 @@ impl IndexScheduler {
                         t.started_at = Some(started_at);
                         t.finished_at = Some(finished_at);
                     }
+
+                    // Patch the task to remove the batch uid, because as of v1.12.5 batches are not persisted.
+                    // This prevent from referencing *future* batches not actually associated with the task.
+                    //
+                    // See <https://github.com/meilisearch/meilisearch/issues/5247> for details.
+                    t.batch_uid = None;
```
Kerollmops, line 827: Nice one!

crates/meilitool/src/main.rs
Kerollmops: That will work only before we release releases like 1.100.0 as _100 < 12_, for example.

Kerollmops (approved):
Looks good to me! Thank you!
bors merge

PR #5254, opened 2025-01-20
WIP

ManyTheFish:
# Pull Request

## Related issue
Fixes #<issue_number>

## What does this PR do?
- ...

## PR checklist

### DIFF
--- a/crates/milli/src/attribute_patterns.rs
+++ b/crates/milli/src/attribute_patterns.rs
@@ -1,0 +1,58 @@
+use deserr::Deserr;
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Deserr, ToSchema)]
+#[repr(transparent)]
+#[serde(transparent)]
+pub struct AttributePatterns {
+    #[schema(value_type = Vec<String>)]
+    pub patterns: Vec<String>,
+}
+
+impl From<Vec<String>> for AttributePatterns {
+    fn from(patterns: Vec<String>) -> Self {
+        Self { patterns }
+    }
+}
+
+impl AttributePatterns {
+    pub fn match_str(&self, str: &str) -> bool {
+        self.patterns.iter().any(|pattern| match_pattern(pattern, str))
+    }
+}
+
+fn match_pattern(pattern: &str, str: &str) -> bool {
+    if pattern == "*" {
+        true
+    } else if pattern.starts_with('*') && pattern.ends_with('*') {
+        str.contains(&pattern[1..pattern.len() - 1])
+    } else if let Some(pattern) = pattern.strip_prefix('*') {
+        str.ends_with(pattern)
+    } else if let Some(pattern) = pattern.strip_suffix('*') {
+        str.starts_with(pattern)
+    } else {
+        pattern == str
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_match_pattern() {
+        assert!(match_pattern("*", "test"));
+        assert!(match_pattern("test*", "test"));
+        assert!(match_pattern("test*", "testa"));
+        assert!(match_pattern("*test", "test"));
+        assert!(match_pattern("*test", "atest"));
+        assert!(match_pattern("*test*", "test"));
+        assert!(match_pattern("*test*", "atesta"));
+        assert!(match_pattern("*test*", "atest"));
+        assert!(match_pattern("*test*", "testa"));
+        assert!(!match_pattern("test*test", "test"));
+        assert!(!match_pattern("*test", "testa"));
+        assert!(!match_pattern("test*", "atest"));
+    }
+}
--- a/crates/milli/src/fields_ids_map/metadata.rs
+++ b/crates/milli/src/fields_ids_map/metadata.rs
@@ -5,14 +5,14 @@ use charabia::Language;
 use heed::RoTxn;
 
 use super::FieldsIdsMap;
-use crate::{FieldId, Index, LocalizedAttributesRule, Result};
+use crate::{FieldId, FilterableAttributesSettings, Index, LocalizedAttributesRule, Result};
 
 #[derive(Debug, Clone, Copy)]
 pub struct Metadata {
     pub searchable: bool,
-    pub filterable: bool,
     pub sortable: bool,
     localized_attributes_rule_id: Option<NonZeroU16>,
+    filterable_attributes_rule_id: Option<NonZeroU16>,
 }
 
 #[derive(Debug, Clone)]
@@ -111,7 +111,7 @@ impl Metadata {
 #[derive(Debug, Clone)]
 pub struct MetadataBuilder {
     searchable_attributes: Vec<String>,
-    filterable_attributes: HashSet<String>,
+    filterable_attributes: Vec<FilterableAttributesSettings>,
     sortable_attributes: HashSet<String>,
     localized_attributes: Option<Vec<LocalizedAttributesRule>>,
 }
@@ -134,7 +134,7 @@ impl MetadataBuilder {
 
     pub fn new(
         searchable_attributes: Vec<String>,
-        filterable_attributes: HashSet<String>,
+        filterable_attributes: Vec<FilterableAttributesSettings>,
         sortable_attributes: HashSet<String>,
         localized_attributes: Option<Vec<LocalizedAttributesRule>>,
     ) -> Self {
@@ -152,8 +152,6 @@ impl MetadataBuilder {
             .iter()
             .any(|attribute| attribute == "*" || attribute == field);
 
-        let filterable = self.filterable_attributes.contains(field);
-
         let sortable = self.sortable_attributes.contains(field);
 
         let localized_attributes_rule_id = self
@@ -164,7 +162,24 @@ impl MetadataBuilder {
             // saturating_add(1): make `id` `NonZero`
             .map(|id| NonZeroU16::new(id.saturating_add(1).try_into().unwrap()).unwrap());
 
-        Metadata { searchable, filterable, sortable, localized_attributes_rule_id }
+        let filterable_attributes_rule_id = self
+            .filterable_attributes
+            .iter()
+            .position(|attribute| match attribute {
+                FilterableAttributesSettings::Field(field_name) => field_name == field,
+                FilterableAttributesSettings::Pattern(patterns) => {
+                    patterns.patterns.match_str(field)
+                }
+            })
+            // saturating_add(1): make `id` `NonZero`
+            .map(|id| NonZeroU16::new(id.saturating_add(1).try_into().unwrap()).unwrap());
+
+        Metadata {
+            searchable,
+            sortable,
+            localized_attributes_rule_id,
+            filterable_attributes_rule_id,
+        }
     }
 
     pub fn searchable_attributes(&self) -> &[String] {
@@ -175,7 +190,7 @@ impl MetadataBuilder {
         &self.sortable_attributes
     }
 
-    pub fn filterable_attributes(&self) -> &HashSet<String> {
+    pub fn filterable_attributes(&self) -> &[FilterableAttributesSettings] {
         &self.filterable_attributes
     }
 
--- a/crates/milli/src/index.rs
+++ b/crates/milli/src/index.rs
@@ -10,10 +10,11 @@ use roaring::RoaringBitmap;
 use rstar::RTree;
 use serde::{Deserialize, Serialize};
 
-use crate::constants::RESERVED_VECTORS_FIELD_NAME;
+use crate::constants::{RESERVED_GEO_FIELD_NAME, RESERVED_VECTORS_FIELD_NAME};
 use crate::documents::PrimaryKey;
 use crate::error::{InternalError, UserError};
 use crate::fields_ids_map::FieldsIdsMap;
+use crate::filterable_fields::matching_field_ids;
 use crate::heed_codec::facet::{
     FacetGroupKeyCodec, FacetGroupValueCodec, FieldDocIdFacetF64Codec, FieldDocIdFacetStringCodec,
     FieldIdCodec, OrderedF64Codec,
@@ -25,8 +26,9 @@ use crate::vector::{ArroyWrapper, Embedding, EmbeddingConfig};
 use crate::{
     default_criteria, CboRoaringBitmapCodec, Criterion, DocumentId, ExternalDocumentsIds,
     FacetDistribution, FieldDistribution, FieldId, FieldIdMapMissingEntry, FieldIdWordCountCodec,
-    FieldidsWeightsMap, GeoPoint, LocalizedAttributesRule, ObkvCodec, Result, RoaringBitmapCodec,
-    RoaringBitmapLenCodec, Search, U8StrStrCodec, Weight, BEU16, BEU32, BEU64,
+    FieldidsWeightsMap, FilterableAttributesSettings, GeoPoint, LocalizedAttributesRule, ObkvCodec,
+    Result, RoaringBitmapCodec, RoaringBitmapLenCodec, Search, U8StrStrCodec, Weight, BEU16, BEU32,
+    BEU64,
 };
 
 pub const DEFAULT_MIN_WORD_LEN_ONE_TYPO: u8 = 5;
@@ -787,7 +789,7 @@ impl Index {
     pub(crate) fn put_filterable_fields(
         &self,
         wtxn: &mut RwTxn<'_>,
-        fields: &HashSet<String>,
+        fields: &Vec<FilterableAttributesSettings>,
     ) -> heed::Result<()> {
         self.main.remap_types::<Str, SerdeJson<_>>().put(
             wtxn,
@@ -802,7 +804,10 @@ impl Index {
     }
 
     /// Returns the filterable fields names.
-    pub fn filterable_fields(&self, rtxn: &RoTxn<'_>) -> heed::Result<HashSet<String>> {
+    pub fn filterable_fields(
+        &self,
+        rtxn: &RoTxn<'_>,
+    ) -> heed::Result<Vec<FilterableAttributesSettings>> {
         Ok(self
             .main
             .remap_types::<Str, SerdeJson<_>>()
@@ -815,14 +820,9 @@ impl Index {
         let fields = self.filterable_fields(rtxn)?;
         let fields_ids_map = self.fields_ids_map(rtxn)?;
 
-        let mut fields_ids = HashSet::new();
-        for name in fields {
-            if let Some(field_id) = fields_ids_map.id(&name) {
-                fields_ids.insert(field_id);
-            }
-        }
+        let matching_field_ids = matching_field_ids(&fields, &fields_ids_map);
 
-        Ok(fields_ids)
+        Ok(matching_field_ids)
     }
 
     /* sortable fields */
@@ -876,6 +876,13 @@ impl Index {
         )
     }
 
+    /// Returns true if the geo feature is activated.
+    pub fn is_geo_activated(&self, rtxn: &RoTxn<'_>) -> Result<bool> {
+        let geo_filter = self.filterable_fields(rtxn)?.iter().any(|field| field.has_geo());
+        let geo_sortable = self.sortable_fields(rtxn)?.contains(RESERVED_GEO_FIELD_NAME);
+        Ok(geo_filter || geo_sortable)
+    }
+
     /// Returns the faceted fields names.
     pub fn faceted_fields(&self, rtxn: &RoTxn<'_>) -> heed::Result<HashSet<String>> {
         Ok(self
@@ -906,7 +913,8 @@ impl Index {
     ///
     /// The user faceted fields are the union of all the filterable, sortable, distinct, and Asc/Desc fields.
     pub fn user_defined_faceted_fields(&self, rtxn: &RoTxn<'_>) -> Result<HashSet<String>> {
-        let filterable_fields = self.filterable_fields(rtxn)?;
+        let fields_ids_map = self.fields_ids_map(rtxn)?;
+        let filterable_fields = self.filterable_fields_ids(rtxn)?;
         let sortable_fields = self.sortable_fields(rtxn)?;
         let distinct_field = self.distinct_field(rtxn)?;
         let asc_desc_fields =
@@ -915,7 +923,14 @@ impl Index {
                 _otherwise => None,
             });
 
-        let mut faceted_fields = filterable_fields;
+        let mut faceted_fields: HashSet<_> = filterable_fields
+            .into_iter()
+            .filter_map(|field_id| {
+                let field_name = fields_ids_map.name(field_id);
+                debug_assert!(field_name.is_some(), "field name not found for {field_id}");
+                field_name.map(|field| field.to_string())
+            })
+            .collect();
         faceted_fields.extend(sortable_fields);
         faceted_fields.extend(asc_desc_fields);
         if let Some(field) = distinct_field {
@@ -923,21 +938,6 @@ impl Index {
         }
 
         Ok(faceted_fields)
-    }
-
-    /// Identical to `user_defined_faceted_fields`, but returns ids instead.
-    pub fn user_defined_faceted_fields_ids(&self, rtxn: &RoTxn<'_>) -> Result<HashSet<FieldId>> {
-        let fields = self.user_defined_faceted_fields(rtxn)?;
-        let fields_ids_map = self.fields_ids_map(rtxn)?;
-
-        let mut fields_ids = HashSet::new();
-        for name in fields {
-            if let Some(field_id) = fields_ids_map.id(&name) {
-                fields_ids.insert(field_id);
-            }
-        }
-
-        Ok(fields_ids)
     }
 
     /* faceted documents ids */
--- a/crates/meilisearch-types/src/settings.rs
+++ b/crates/meilisearch-types/src/settings.rs
@@ -11,7 +11,9 @@ use fst::IntoStreamer;
 use milli::index::{IndexEmbeddingConfig, PrefixSearch};
 use milli::proximity::ProximityPrecision;
 use milli::update::Setting;
-use milli::{Criterion, CriterionError, Index, DEFAULT_VALUES_PER_FACET};
+use milli::{
+    Criterion, CriterionError, FilterableAttributesSettings, Index, DEFAULT_VALUES_PER_FACET,
+};
 use serde::{Deserialize, Serialize, Serializer};
 use utoipa::ToSchema;
 
@@ -202,8 +204,8 @@ pub struct Settings<T> {
     /// Attributes to use for faceting and filtering. See [Filtering and Faceted Search](https://www.meilisearch.com/docs/learn/filtering_and_sorting/search_with_facet_filters).
     #[serde(default, skip_serializing_if = "Setting::is_not_set")]
     #[deserr(default, error = DeserrJsonError<InvalidSettingsFilterableAttributes>)]
-    #[schema(value_type = Option<Vec<String>>, example = json!(["release_date", "genre"]))]
-    pub filterable_attributes: Setting<BTreeSet<String>>,
+    #[schema(value_type = Option<Vec<FilterableAttributesSettings>>, example = json!(["release_date", "genre"]))]
+    pub filterable_attributes: Setting<Vec<FilterableAttributesSettings>>,
     /// Attributes to use when sorting search results.
     #[serde(default, skip_serializing_if = "Setting::is_not_set")]
     #[deserr(default, error = DeserrJsonError<InvalidSettingsSortableAttributes>)]
--- a/crates/milli/src/filterable_fields.rs
+++ b/crates/milli/src/filterable_fields.rs
@@ -1,0 +1,75 @@
+use deserr::{DeserializeError, Deserr, ValuePointerRef};
+use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
+use utoipa::ToSchema;
+
+use crate::{
+    constants::RESERVED_GEO_FIELD_NAME, is_faceted_by, AttributePatterns, FieldId, FieldsIdsMap,
+};
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Debug, ToSchema)]
+#[serde(untagged)]
+pub enum FilterableAttributesSettings {
+    Field(String),
+    Pattern(FilterableAttributesPatterns),
+}
+
+impl FilterableAttributesSettings {
+    pub fn match_str(&self, field: &str) -> bool {
+        match self {
+            FilterableAttributesSettings::Field(field_name) => is_faceted_by(field, field_name),
+            FilterableAttributesSettings::Pattern(patterns) => patterns.patterns.match_str(field),
+        }
+    }
+
+    pub fn has_geo(&self) -> bool {
+        /// TODO: This is a temporary solution to check if the geo field is activated.
+        matches!(self, FilterableAttributesSettings::Field(field_name) if field_name == RESERVED_GEO_FIELD_NAME)
+    }
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Debug, Deserr, ToSchema)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+#[deserr(rename_all = camelCase, deny_unknown_fields)]
+pub struct FilterableAttributesPatterns {
+    pub patterns: AttributePatterns,
+    pub features: Option<FilterableAttributesFeatures>,
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Debug, Deserr, ToSchema)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+#[deserr(rename_all = camelCase, deny_unknown_fields)]
+pub struct FilterableAttributesFeatures {
+    facet_search: Option<String>,
+    filter: Option<String>,
+}
+
+impl<E: DeserializeError> Deserr<E> for FilterableAttributesSettings {
+    fn deserialize_from_value<V: deserr::IntoValue>(
+        value: deserr::Value<V>,
+        location: ValuePointerRef,
+    ) -> Result<Self, E> {
+        if value.kind() == deserr::ValueKind::Map {
+            Ok(Self::Pattern(FilterableAttributesPatterns::deserialize_from_value(
+                value, location,
+            )?))
+        } else {
+            Ok(Self::Field(String::deserialize_from_value(value, location)?))
+        }
+    }
+}
+
+pub fn matching_field_ids(
+    filterable_attributes: &[FilterableAttributesSettings],
+    fields_ids_map: &FieldsIdsMap,
+) -> HashSet<FieldId> {
+    let mut result = HashSet::new();
+    for (field_id, field_name) in fields_ids_map.iter() {
+        for filterable_attribute in filterable_attributes {
+            if filterable_attribute.match_str(field_name) {
+                result.insert(field_id);
+            }
+        }
+    }
+    result
+}
--- a/crates/milli/src/lib.rs
+++ b/crates/milli/src/lib.rs
@@ -9,11 +9,13 @@ pub static ALLOC: mimalloc::MiMalloc = mimalloc::MiMalloc;
 pub mod documents;
 
 mod asc_desc;
+mod attribute_patterns;
 mod criterion;
 mod error;
 mod external_documents_ids;
 pub mod facet;
 mod fields_ids_map;
+mod filterable_fields;
 pub mod heed_codec;
 pub mod index;
 mod localized_attributes_rules;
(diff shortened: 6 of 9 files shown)


crates/milli/src/update/index_documents/extract/extract_facet_string_docids.rs
```diff
@@ -92,15 +87,25 @@ fn extract_facet_string_docids_document_update<R: io::Read + io::Seek>(
...
         let document_id = u32::from_be_bytes(document_id_bytes);
 
         let normalized_value = str::from_utf8(normalized_value_bytes)?;
 
         // Facet search normalization
-        if facet_search {
-            let locales = localized_field_ids.locales(field_id);
+        let features =
+            metadata.filterable_attributes_features(&settings.filterable_attributes_rules);
+        if features.is_facet_searchable() {
```
ManyTheFish, line 107: is global facet search feature activated?

crates/milli/src/index.rs
```diff
@@ -876,11 +876,11 @@ impl Index {
...
     /* filterable fields */
 
-    /// Writes the filterable fields names in the database.
-    pub(crate) fn put_filterable_fields(
+    /// Writes the filterable attributes rules in the database.
+    pub(crate) fn put_filterable_attributes_rules(
         &self,
         wtxn: &mut RwTxn<'_>,
-        fields: &HashSet<String>,
+        #[allow(clippy::ptr_arg)] fields: &Vec<FilterableAttributesRule>,
```
ManyTheFish, line 871: &&[FilterableAttributesRule]

crates/milli/src/attribute_patterns.rs
```diff
@@ -0,0 +1,128 @@
...
+            return PatternMatch::Parent;
+        }
+    }
+    // If the field does not match the distinct field and is not a parent of a nested field that matches the distinct field, return NoMatch
+    PatternMatch::NoMatch
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PatternMatch {
+    /// The field is a parent of the of a nested field that matches the pattern
```
dureuill, line 101:
```suggestion
    /// The field is a parent of a nested field that matches the pattern
    /// For example, the field is `toto`, and the pattern is `toto.titi`
```

crates/milli/src/attribute_patterns.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// The field does not match the pattern
+    NoMatch,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_match_pattern() {
```
dureuill, line 114: would be great to have some tests for `PatternMatch::Parent`

`toto.*`, `toto` => Parent

and so on

crates/milli/src/filterable_attributes_rules.rs
```diff
@@ -0,0 +1,204 @@
...
+            FilterableAttributesRule::Pattern(patterns) => patterns.features(),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Debug, Deserr, ToSchema)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+#[deserr(rename_all = camelCase, deny_unknown_fields)]
+pub struct FilterableAttributesPatterns {
+    pub patterns: AttributePatterns,
```
dureuill, line 43: the usage page states that this should be `attributePatterns`, here it will be `patterns` unless I'm mistaken

crates/milli/src/filterable_attributes_rules.rs
```diff
@@ -0,0 +1,204 @@
...
+    #[serde(default)]
+    #[deserr(default)]
+    pub features: FilterableAttributesFeatures,
```
dureuill, line 46: the usage pages sometimes states that this should be `filter`. Am I misunderstanding something?

crates/milli/src/filterable_attributes_rules.rs
```diff
@@ -0,0 +1,204 @@
...
+impl FilterableAttributesRule {
+    pub fn match_str(&self, field: &str) -> PatternMatch {
+        match self {
+            FilterableAttributesRule::Field(pattern) => match_field_legacy(pattern, field),
+            FilterableAttributesRule::Pattern(patterns) => patterns.match_str(field),
+        }
+    }
+
+    pub fn has_geo(&self) -> bool {
+        matches!(self, FilterableAttributesRule::Field(field_name) if field_name == RESERVED_GEO_FIELD_NAME)
```
dureuill, line 28: This means that only `"_geo"` as its string from is allowed, right?

In particular, the PRD has an example with:
```jsonc
{
    "filterableAttributes": [
	    // Good way
	    {
	      "attributePatterns": ["_geo"]
	    }
    ]
}
``` 

But reading the code it appears that only the following would work:

```jsonc
{
    "filterableAttributes": [ "_geo" ]
}
```

crates/milli/src/filterable_attributes_rules.rs
```diff
@@ -0,0 +1,204 @@
...
+            Ok(Self::Pattern(FilterableAttributesPatterns::deserialize_from_value(
+                value, location,
+            )?))
+        } else {
+            Ok(Self::Field(String::deserialize_from_value(value, location)?))
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Debug, Deserr, ToSchema)]
```
dureuill, line 131: this could be made `Copy`

crates/milli/src/filterable_attributes_rules.rs
```diff
@@ -0,0 +1,204 @@
...
+    pub fn match_str(&self, field: &str) -> PatternMatch {
+        self.patterns.match_str(field)
+    }
+
+    pub fn features(&self) -> FilterableAttributesFeatures {
+        self.features.clone()
+    }
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Debug, Deserr, ToSchema)]
```
dureuill, line 59: This could be made `Copy`

crates/milli/src/filterable_attributes_rules.rs
```diff
@@ -0,0 +1,204 @@
...
+    pub features: FilterableAttributesFeatures,
+}
+
+impl FilterableAttributesPatterns {
+    pub fn match_str(&self, field: &str) -> PatternMatch {
+        self.patterns.match_str(field)
+    }
+
+    pub fn features(&self) -> FilterableAttributesFeatures {
+        self.features.clone()
```
dureuill, line 55: the `clone` here could be removed with `FilterableAttributesFeatures` `Copy`

crates/milli/src/filterable_attributes_rules.rs
```diff
@@ -202,3 +202,107 @@ impl Default for FilterFeatures {
...
+            PatternMatch::Match => return PatternMatch::Match,
+            PatternMatch::Parent => selection = PatternMatch::Parent,
+            PatternMatch::NoMatch => (),
+        }
+    }
+
+    selection
+}
+
+fn match_pattern_by_features(
```
dureuill, line 339: I'm a bit surprised at the logic in this function, because I thought that the first rule with a matching pattern would "win".

Here it looks to me like we are trying to pattern match only if the filter returns true.

Taking again the two rules: `"toto" => "filter": false` and `"t*t*" => filter: true`, then to match against field `"toto"`, the first rule will be ignored because the filter doesn't match, and the second rule will be considered.

It looks to me like the proper semantic here is to consider the first rule that matches the pattern, and so to return `NoMatch`.

Now, the issue here is the logic with `Parent` vs direct `Match`. What I suggest here is that the caller should decide if `PatternMatch::Parent` is considered a match. An easy way to do this is to provide two functions: `match_pattern_by_features` and `match_pattern_or_parent_by_features`.

Is there a use case for `Parent` on the search side, I thought it was more for indexing?

crates/milli/src/search/facet/filter.rs
```diff
@@ -409,7 +441,7 @@ impl<'a> Filter<'a> {
         &self,
         rtxn: &heed::RoTxn<'_>,
         index: &Index,
-        filterable_fields: &HashSet<String>,
+        filterable_fields: &[FilterableAttributesRule],
```
dureuill, line 444: rename to `filterable_rules` or `filterable_attributes_rules` or `rules`?

crates/milli/src/search/facet/filter.rs
```diff
@@ -434,36 +466,56 @@ impl<'a> Filter<'a> {
...
+                        let field_ids_map = index.fields_ids_map(rtxn)?;
+                        if let Some(fid) = field_ids_map.id(fid.value()) {
+                            Self::evaluate_operator(rtxn, index, fid, universe, op, &features)
+                        } else {
+                            Ok(RoaringBitmap::new())
+                        }
+                    }
+                    _ => {
+                        let field_ids_map = index.fields_ids_map(rtxn)?;
+                        Err(fid.as_external_error(FilterError::AttributeNotFilterable {
```
dureuill, line 510: I thought this case was checked in `evaluate` already.

Do we need to keep the pre-check in `evaluate` then?

crates/milli/src/search/facet/filter.rs
```diff
@@ -434,36 +466,56 @@ impl<'a> Filter<'a> {
...
             FilterCondition::Condition { fid, op } => {
-                if crate::is_faceted(fid.value(), filterable_fields) {
-                    let field_ids_map = index.fields_ids_map(rtxn)?;
-                    if let Some(fid) = field_ids_map.id(fid.value()) {
-                        Self::evaluate_operator(rtxn, index, fid, universe, op)
-                    } else {
-                        Ok(RoaringBitmap::new())
+                match matching_features(fid.value(), filterable_fields) {
+                    Some(features) if features.is_filterable() => {
+                        let field_ids_map = index.fields_ids_map(rtxn)?;
```
dureuill, line 501: I don't think this is new from this PR, but I also don't think we should be deserializing a whole new `field_ids_map` on each condition of the filter...

I think it should be spawned in `evaluate` and passed to all subfunctions instead.

crates/milli/src/search/facet/filter.rs
```diff
@@ -434,36 +466,56 @@ impl<'a> Filter<'a> {
...
                 }
             }
             FilterCondition::Condition { fid, op } => {
-                if crate::is_faceted(fid.value(), filterable_fields) {
-                    let field_ids_map = index.fields_ids_map(rtxn)?;
-                    if let Some(fid) = field_ids_map.id(fid.value()) {
-                        Self::evaluate_operator(rtxn, index, fid, universe, op)
-                    } else {
-                        Ok(RoaringBitmap::new())
+                match matching_features(fid.value(), filterable_fields) {
```
dureuill, line 499: I find it a bit sad that we'll be looking up matching rules each time a fid appears in a condition, but I guess in the common case where a fid appears only once in a filter, it is faster than memorizing the features in a map 🤔

crates/milli/src/search/facet/filter.rs
```diff
@@ -1164,7 +1271,11 @@ mod tests {
...
             .update_settings(|settings| {
                 settings.set_primary_key("id".to_owned());
-                settings.set_filterable_fields(hashset! { S("id"), S("one"), S("two") });
+                settings.set_filterable_fields(vec![
+                    FilterableAttributesRule::Field("id".to_string()),
+                    FilterableAttributesRule::Field("one".to_string()),
+                    FilterableAttributesRule::Field("two".to_string()),
+                ]);
             })
             .unwrap();
```
dureuill, line 1281: Would be nice to have some added tests here, testing "negative rules" (e.g. a first rule that indicates that "toto" is not filterable, then a second rule that would also match "toto" and make it filterable. The expected result is that "toto" is not filterable, as the first matching rule wins) and case with "parent" (e.g. a first rule on "toto.titi", a second one on just "toto". Field "toto" is expected to match the second rule, and "toto.titi" the first one)

crates/milli/src/search/facet/search.rs
```diff
@@ -73,10 +76,16 @@ impl<'a> SearchForFacetValues<'a> {
         let index = self.search_query.index;
         let rtxn = self.search_query.rtxn;
 
-        let filterable_fields = index.filterable_fields(rtxn)?;
-        if !filterable_fields.contains(&self.facet) {
+        let filterable_attributes_rules = index.filterable_attributes_rules(rtxn)?;
+        if !is_field_facet_searchable(&self.facet, &filterable_attributes_rules) {
```
dureuill, line 80: check that facet search is enabled at all here?

dureuill (commented):
Comments on commit `Refactor search and facet-search`

crates/milli/src/filterable_attributes_rules.rs
ManyTheFish: You are right and it must be consistent with the localized attributes:
https://www.meilisearch.com/docs/reference/api/settings#localized-attributes

crates/milli/src/filterable_attributes_rules.rs
ManyTheFish: the `filter` is contained in the `features` with the `facet_search`

crates/milli/src/search/facet/facet_distribution.rs
```diff
@@ -287,37 +288,23 @@ impl<'a> FacetDistribution<'a> {
...
-                    }
-                    .into());
-                } else {
-                    facets.iter().map(|(name, _)| name).cloned().collect()
-                }
-            }
-            None => filterable_fields,
-        };
+        let fields_ids_map = self.index.fields_ids_map(self.rtxn)?;
+        let fields_ids_map = FieldIdMapWithMetadata::new(
```
dureuill, line 298: spawning a field id map with metadata might be a perf regression here, to check with benchmarks

depending on the results it might be better to do this more manually by matching the filterable attributes rules only of the members of `facets`

this is only while we don't store the `FieldIdMapWithMetadata` in the index data

crates/milli/src/search/facet/facet_distribution.rs
```diff
@@ -385,6 +357,61 @@ impl<'a> FacetDistribution<'a> {
...
+    /// Select a field if it is faceted and in the facets.
+    fn select_field(
+        &self,
+        name: &str,
+        metadata: &Metadata,
+        filterable_attributes_rules: &[FilterableAttributesRule],
+    ) -> bool {
+        match &self.facets {
+            Some(facets) => {
+                // The list of facets provided by the user is a legacy pattern ("dog.age" must be selected with "dog").
```
dureuill, line 370: can this create an inconsistency, if the user uses the new rules syntax, and has set only `dog` as filterable? On the indexing side, `dog.age` might not be filterable, but here we will return it as filterable.

crates/milli/src/search/facet/facet_distribution.rs
```diff
@@ -385,6 +357,61 @@ impl<'a> FacetDistribution<'a> {
...
+        &self,
+        fields_ids_map: &FieldIdMapWithMetadata,
+        filterable_attributes_rules: &[FilterableAttributesRule],
+    ) -> Result<()> {
+        let mut invalid_facets = BTreeSet::new();
+        if let Some(facets) = &self.facets {
+            for (field, _) in facets {
+                let is_valid_faceted_field =
+                    fields_ids_map.id_with_metadata(field).map_or(false, |(_, metadata)| {
+                        metadata.is_faceted(filterable_attributes_rules)
```
dureuill, line 388: I seem to remember that facet distribution would only work with filterable attributes, **not** distinct, sortable, etc. 

The error message of `fn format_invalid_filter_distribution` explicitly mentions `filterable` attributes.

Can we confirm this behavior on a v1.13 and a v1.12?

dureuill (commented):
Comments for `Refactor the FieldIdMapWithMetadata`

crates/milli/src/search/new/ranking_rule_graph/fid/mod.rs
```diff
@@ -88,10 +88,10 @@ impl RankingRuleGraphTrait for FidGraph {
         }
 
         // always lookup the max_fid if we don't already and add an artificial condition for max scoring
-        let max_weight: Option<u16> = weights_map.max_weight();
+        let max_weight_fid = weights_map.max_weight_fid();
 
-        if let Some(max_weight) = max_weight {
-            if !all_fields.contains(&max_weight) {
+        if let Some((max_fid, max_weight)) = max_weight_fid {
+            if !all_fields.contains(&max_fid) {
```
dureuill, line 94: Thank you for fixing this!

Note that we probably want the condition to check if any fid with the max_weight exists in the edges, rather than just checking for `max_fid`, so that we only add the condition when **no** other fid has the `max_weight`, but that is already fixing a bug.

crates/milli/src/filterable_attributes_rules.rs
```diff
@@ -202,3 +202,107 @@ impl Default for FilterFeatures {
...
+) -> BTreeSet<&'fim str> {
+    let mut result = BTreeSet::new();
+    for (_, field_name) in fields_ids_map.iter() {
+        for filterable_attribute in filterable_attributes {
+            if filterable_attribute.match_str(field_name) == PatternMatch::Match {
+                let features = filterable_attribute.features();
+                if filter(&features) {
+                    result.insert(field_name);
+                }
+            }
```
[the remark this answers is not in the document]
ManyTheFish: you are totally right

crates/milli/src/filterable_attributes_rules.rs
```diff
@@ -202,3 +202,107 @@ impl Default for FilterFeatures {
...
+        .map_or(false, |features| features.is_filterable())
+}
+
+pub fn is_field_facet_searchable(
+    field_name: &str,
+    filterable_attributes: &[FilterableAttributesRule],
+) -> bool {
+    matching_features(field_name, filterable_attributes)
+        .map_or(false, |features| features.is_facet_searchable())
+}
```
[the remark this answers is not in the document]
ManyTheFish: > if these helpers are not frequently used together, that is OK, and matching the rules should generally be fast as there shouldn't be that many rules, but this is something to keep in mind.

These two helpers are used in different places, one in the filter parser, the other in the facet search.

crates/milli/src/filterable_attributes_rules.rs
ManyTheFish: My mistake in the commits, this is only used during indexing.

crates/milli/src/search/facet/filter.rs
ManyTheFish: > Do we need to keep the pre-check in evaluate then?
We could remove it, however, I feel the check is an early return error in case the last filter is wrong avoiding to evaluate the other filter for nothing, on the other side, we could replace each match case by `unreachable!` knowing that we already checked if the field is filterable but I am not sure if we want to add "possible" panics in the search 🤔 

I don't have a strong opinion here, I'll follow the call

crates/milli/src/search/facet/search.rs
ManyTheFish: I am not sure to understand, do you speak about the following check?
https://github.com/meilisearch/meilisearch/blob/683a2ac685210402848466a6a0c9e8b406a4df52/crates/meilisearch/src/search/mod.rs#L1529-L1534

crates/milli/src/filterable_attributes_rules.rs
dureuill: Alright about the commit, makes it clearer, however my point about the risk of picking a rule after the first matching rule remains.

crates/milli/src/search/facet/filter.rs
dureuill: > I feel the check is an early return error in case the last filter is wrong avoiding to evaluate the other filter for nothing,

I don't feel that optimizing for the error case is right, especially when the evaluation of a filter is supposed to be a relatively quick operation (so that search is quick).

I'd rather we remove the pre-check then, if we are sure this won't create an observable behavior difference.

crates/milli/src/search/facet/search.rs
dureuill: ah, my bad, I didn't see this.

crates/milli/src/search/facet/filter.rs
ManyTheFish: I've tried to fix it by relying more on the `FieldIdMapWithMetadata` https://github.com/meilisearch/meilisearch/pull/5254/commits/b88aa9cc76b6af7b3b5f94072b6b63e427fc7b02

crates/milli/src/filterable_attributes_rules.rs
ManyTheFish: Are you sure?
I think the following commit fixes the problem:
https://github.com/meilisearch/meilisearch/pull/5254/commits/5fa4b5c50abdcaf7675ef34bbfeb6acdfb29903b

dureuill: I'm sorry I'm still not sure this fixes the problem (also, having the fix in a test commit is definitely unexpected):

consider the rules:

1. "pattern": "toto.dog", "filter": false,
2. "pattern": "toto.dog", "filter": true

Consider the field "toto".
In the current implementation it will try rule #1, which will fail the filter, and then try rule #2, which will succeed and return `Parent`. Yet, the answer here should be `NoMatch`

crates/milli/src/search/facet/facet_distribution.rs
ManyTheFish: you're right https://github.com/meilisearch/meilisearch-support/issues/702#issuecomment-2703403394

crates/milli/src/update/index_documents/transform.rs
```diff
@@ -734,12 +739,22 @@ impl<'a, 'i> Transform<'a, 'i> {
...
-                obkv_writer.insert(id, val)?;
+            } else {
+                let facet_operation = necessary_faceted_field(id);
+                let searchable_operation = settings_diff.reindex_searchable_id(id);
+                let operation = facet_operation
+                    // TODO: replace `zip.map` with `zip_with` once stable
+                    .zip(searchable_operation)
+                    .map(|(op1, op2)| op1.merge(op2))
+                    .or(facet_operation)
+                    .or(searchable_operation);
```
dureuill, line 753: style, but I found this a bit hard to read.

Would have preferred an explicit match here:

```rust
let operation = match (facet_operation, searchable_operation) {
    (Some(facet_operation), Some(searchable_operation)) => Some(facet_operation.merge(searchable_operation)),
    (Some(operation), None) | (None, Some(operation)) => Some(operation),
    (None, None) => None
};
```

crates/milli/src/search/new/ranking_rule_graph/fid/mod.rs
ManyTheFish: true, I'll try to enhance that

dureuill (commented):
Comments on `Refactor Settings Indexing process`

crates/milli/src/update/new/extract/searchable/tokenize_document.rs
```diff
@@ -265,6 +276,10 @@ mod test {
                 2,
                 16,
             ]: "catto",
+            [
+                3,
+                0,
+            ]: "unsearchable",
```
dureuill, line 282: why did this test change?

dureuill (commented):
Comment for `Refactor Document indexing process (searchables)`

crates/milli/src/index.rs
```diff
@@ -2189,7 +2190,7 @@ pub(crate) mod tests {
         let rtxn = index.read_txn().unwrap();
 
         let real = index.searchable_fields(&rtxn).unwrap();
-        assert_eq!(real, &["doggo", "name"]);
+        assert!(real.is_empty());
```
dureuill, line 2203: why did the result change?

dureuill (commented):
Comment on `Update tests using filterable attributes rules`

dureuill (commented):
No comment on `Index document in filterable attributes tests`

crates/milli/src/fields_ids_map/metadata.rs
```diff
@@ -106,76 +121,215 @@ impl Metadata {
...
-        let filterable = self.filterable_attributes.contains(field);
+        let filterable_attributes_rule_id = self
+            .filterable_attributes
+            .iter()
+            .position(|attribute| attribute.match_str(field) == PatternMatch::Match)
+            // saturating_add(1): make `id` `NonZero`
+            .map(|id| NonZeroU16::new(id.saturating_add(1).try_into().unwrap()).unwrap());
 
-        let sortable = self.sortable_attributes.contains(field);
+        if match_field_legacy(RESERVED_GEO_FIELD_NAME, field) == PatternMatch::Match {
```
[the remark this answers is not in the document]
dureuill: already handled by the `match_field_legacy`

crates/milli/src/fields_ids_map/metadata.rs
```diff
@@ -106,76 +121,215 @@ impl Metadata {
...
+                filterable_attributes_rule_id: None,
+            };
+        }
+
+        // A field is sortable if it is faceted by a sortable attribute
+        let sortable = self
+            .sortable_attributes
             .iter()
-            .any(|attribute| attribute == "*" || attribute == field);
+            .any(|pattern| match_field_legacy(pattern, field) == PatternMatch::Match);
```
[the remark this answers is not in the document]
dureuill: `sortable` from the metadata wasn't used in previous versions, so my point is moot. Resolving...

crates/milli/src/filterable_attributes_rules.rs
ManyTheFish: let's modify the doc to fit the current behavior, enhance the doc

dureuill (commented):
No comment on `Update snapshot using the new filterableAttributes type`

crates/milli/src/index.rs
ManyTheFish: There is no document in the database

crates/meilisearch/tests/search/facet_search.rs
```diff
@@ -436,3 +494,124 @@ async fn deactivate_facet_search_add_documents_and_reset_facet_search() {
...
+        |response, code| {
+            snapshot!(code, @"400 Bad Request");
+            snapshot!(response["message"], @r###""Attribute `invalid` is not facet-searchable. Available facet-searchable attributes are: `genres`. To make it facet-searchable add it to the `filterableAttributes` index settings.""###);
+        },
+    )
+    .await;
+
+    test_settings_documents_indexing_swapping_and_facet_search(
+      &DOCUMENTS,
+      &json!({"filterableAttributes": [{"patterns": ["genres"]}]}),
```
dureuill: This error is a bit confusing because `genres` **is** already in the list of patterns of the `filterableAttributes` index settings. I think you should specify that `features.facetSearch` should be set to `true` as well.

The error message could become:

```md
Attribute `genres` is not facet-searchable.
  - note: this index does not have a filterable attribute pattern that is facet searchable.
  - hint: add `{ "patterns": ["genres"], "features": { "facetSearch":true } }` to `filterableAttributes`
```

crates/meilisearch/tests/search/filters.rs
```diff
@@ -0,0 +1,625 @@
...
+    index
+        .search(
+            json!({
+                "filter": "doggos.age > 2"
+            }),
+            |response, code| {
+                snapshot!(code, @"400 Bad Request");
+                snapshot!(json_string!(response), @r###"
+                {
+                  "message": "Index `test`: Filter operator `>` is not allowed for the attribute `doggos.age`, allowed operators: OR, AND, NOT, =, !=, IN, IS EMPTY, IS NULL, EXISTS.",
```
dureuill: btw, it is a details, but I would to have:

```md
Index `test`: filter operator `>` is not allowed for the attribute `doggos.age`
  - note: allowed operators for `doggos.age`: OR, AND, NOT, =, !=, IN, IS EMPTY, IS NULL, EXISTS.
```

dureuill (commented):
Comment on `Add and enhance tests`

crates/milli/src/attribute_patterns.rs
```diff
@@ -41,22 +42,35 @@ impl AttributePatterns {
...
+/// # Arguments
+///
+/// * `pattern` - The pattern to match against.
+/// * `str` - The string to match against the pattern.
 fn match_pattern(pattern: &str, str: &str) -> PatternMatch {
+    // If the pattern is a wildcard, return Match
     if pattern == "*" {
         return PatternMatch::Match;
     } else if pattern.starts_with('*') && pattern.ends_with('*') {
+        // If the starts and ends with a wildcard, return Match if the string contains the pattern without the wildcards
```
dureuill, line 58:
```suggestion
        // If the pattern starts and ends with a wildcard, return Match if the string contains the pattern without the wildcards
```

dureuill (commented):
Comment on `Add some documentation in modules`

crates/milli/src/update/new/extract/searchable/tokenize_document.rs
ManyTheFish: The check to know if a field is searchable is made a bit latter in the code, so this test no more have the searchable field filter

crates/meilisearch/tests/search/multi/mod.rs
```diff
@@ -3604,22 +3604,28 @@ async fn federation_non_faceted_for_an_index() {
 
     let index = server.index("fruits");
 
+    let documents = FRUITS_DOCUMENTS.clone();
+    let (value, _) = index.add_documents(documents, None).await;
+    index.wait_task(value.uid()).await.succeeded();
```
dureuill, line 3609: this change shouldn't be necessary anymore, now that we display the filterable *patterns* in error messages and that we compute the max weight regardless of which fields actually exist in the dataset.

Should we consider undoing the changes where this PR is adding documents to indexes?

crates/milli/src/search/facet/filter.rs
```diff
@@ -75,14 +77,14 @@ impl<'a> From<BadGeoError> for FilterError<'a> {
...
         match self {
-            Self::AttributeNotFilterable { attribute, filterable_fields } => {
+            Self::AttributeNotFilterable { attribute, filterable_patterns } => {
                 write!(f, "Attribute `{attribute}` is not filterable.")?;
-                if filterable_fields.is_empty() {
+                if filterable_patterns.is_empty() {
                     write!(f, " This index does not have configured filterable attributes.")
                 } else {
-                    write!(f, " Available filterable attributes are: ")?;
+                    write!(f, " Available filterable attributes patterns are: ")?;
```
dureuill, line 85:
```suggestion
                    write!(f, " Available filterable attribute patterns are: ")?;
```

dureuill:
Please fix clippy errors

crates/milli/src/fields_ids_map/metadata.rs
```diff
@@ -106,76 +121,247 @@ impl Metadata {
...
-        searchable_attributes: Vec<String>,
-        filterable_attributes: HashSet<String>,
+        searchable_attributes: Option<Vec<String>>,
+        filterable_attributes: Vec<FilterableAttributesRule>,
+        sortable_attributes: HashSet<String>,
+        localized_attributes: Option<Vec<LocalizedAttributesRule>>,
+        distinct_attribute: Option<String>,
+        asc_desc_attributes: HashSet<String>,
+    ) -> Self {
+        Self::_new(
```
dureuill, line 249: I don't understand the need for `_new` here. Can you inline it in `new` and replace its callers by callers to `new`?

crates/milli/src/fields_ids_map/metadata.rs
```diff
@@ -106,76 +121,247 @@ impl Metadata {
...
-            .any(|attribute| attribute == "*" || attribute == field);
+            .any(|pattern| match_field_legacy(pattern, field) == PatternMatch::Match);
 
-        let filterable = self.filterable_attributes.contains(field);
+        let filterable_attributes_rule_id = self
+            .filterable_attributes
+            .iter()
+            .position(|attribute| attribute.match_str(field) == PatternMatch::Match)
+            // saturating_add(1): make `id` `NonZero`
+            .map(|id| NonZeroU16::new(id.saturating_add(1).try_into().unwrap()).unwrap());
```
dureuill, line 308: Do we enforce anywhere in the settings that the max number of rules is `NonZeroU16::MAX - 1`? If not, should we?

crates/milli/src/search/facet/facet_distribution.rs
```diff
@@ -385,6 +349,62 @@ impl<'a> FacetDistribution<'a> {
 
         Ok(distribution)
     }
+
+    /// Select a field if it is faceted and in the facets.
```
dureuill, line 353:
```suggestion
    /// Select a field if it is filterable and in the facets.
```

crates/milli/src/search/facet/facet_distribution.rs
```diff
@@ -385,6 +349,62 @@ impl<'a> FacetDistribution<'a> {
...
+        match &self.facets {
+            Some(facets) => {
+                // The list of facets provided by the user is a legacy pattern ("dog.age" must be selected with "dog").
+                facets.keys().any(|key| match_field_legacy(key, name) == PatternMatch::Match)
+            }
+            None => true,
+        }
+    }
+
+    /// Check if the fields in the facets are valid faceted fields.
```
dureuill, line 375:
```suggestion
    /// Check if the fields in the facets are valid filterable fields.
```

crates/milli/src/fields_ids_map/metadata.rs
ManyTheFish: I don't want new to be accessible from anything else that test

crates/milli/src/fields_ids_map/metadata.rs
ManyTheFish: No but it's the same for the localized attribute rules 🤔

crates/milli/src/fields_ids_map/metadata.rs
dureuill: Ah. I see. I don't think it would cause much harm to remove the `cfg[test]` here, but if you want to keep it, maybe you could highlight this with a comment on top of `_new`:

```rust
    // Note: `pub fn new` is #[cfg(test)], this function is private
    fn _new(
```

crates/milli/src/update/new/document_change.rs
```diff
@@ -173,7 +173,7 @@ impl<'doc> Update<'doc> {
     /// Otherwise `false`.
     pub fn has_changed_for_fields<'t, Mapper: FieldIdMapper>(
         &self,
-        fields: Option<&[&str]>,
+        selector: &mut impl FnMut(&str) -> PatternMatch,
```
dureuill, line 176: worth updating the documentation of this function?

```rust
    /// Returns whether the updated version of the document is different from the current version for the subset of fields selected by `selector`.
    ///
    /// `true` if at least one top-level-field that is exactly a selected field or a parent of a selected field changed.
    /// Otherwise `false`.
    /// 
    /// - Note: `_geo` and `_vectors` are not taken into account by this function.
```

crates/milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
```diff
@@ -76,9 +76,9 @@ pub fn extract_fid_docid_facet_values<R: io::Read + io::Seek>(
     let mut strings_key_buffer = Vec::new();
 
     let old_faceted_fids: BTreeSet<_> =
-        settings_diff.old.faceted_fields_ids.iter().copied().collect();
+        settings_diff.list_faceted_fields_from_fid_map(DelAdd::Deletion);
     let new_faceted_fids: BTreeSet<_> =
-        settings_diff.new.faceted_fields_ids.iter().copied().collect();
+        settings_diff.list_faceted_fields_from_fid_map(DelAdd::Addition);
 
     if !settings_diff.settings_update_only || settings_diff.reindex_facets() {
```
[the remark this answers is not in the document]
ManyTheFish: > I thought list_faceted_fields_from_fid_map could not be used for document additions (such as dump import)

Yes, because the field id map may not contain all the fields.
However, in the old indexer, we fill the field id map in Transform when flattening the documents ([flatten_from_fields_ids_map](https://github.com/meilisearch/meilisearch/blob/d0dda78f3dae9c560cd4a068310fecde29ba6efc/crates/milli/src/update/index_documents/transform.rs#L372)), that's why it's "acceptable" to do it.
That said, we may want to migrate the dump import to the new indexer (v1.15?) to remove this uncertainty.

> I thought old_faceted_fids and new_faceted_fids could only differ in a settings diff (settings_update_only)

No, if you add a document with a new nested field, this would extend the list

> Yet here we guard it the use of old_faceted_fids and new_faceted_fids behind a ! settings_diff.settings_update_only

This condition is badly expressed, here is how I interpret it:
`if (there are any new documents) OR (a setting impacting the facet databases changed)`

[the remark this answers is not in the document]
dureuill: Alright, it looks fine then :-)

crates/milli/src/fields_ids_map/metadata.rs
dureuill: Created as https://github.com/meilisearch/meilisearch/issues/5409, resolving for now

crates/meilisearch/tests/search/facet_search.rs
dureuill: Referenced and detailed as https://github.com/meilisearch/meilisearch/issues/5411.

Resolving for now

dureuill (approved):
A major feature, well designed, tested and documented.

Thank you for your tremendous work 🚀

ManyTheFish:
bors merge

Issue #5162, opened 2024-12-14
Dumpless upgrade


curquiza:
## Motivation

Ease Meilisearch way to upgrade between to minor versions

## Usage

Usage here: https://meilisearch.notion.site/Dumpless-upgrade-fff4b06b651f81f1acafe24d4687b3f7?pvs=74

### Reminders when modifying the Setting API

- [ ] Update the CONTRIBUTING.md
- [ ] Update the issue template
- [ ] Update the progress part

#### Special cases when adding a setting for an experimental feature

- [ ] ⚠️ API stability: The setting does not appear on the main settings route when the feature has never been enabled (e.g. mark it `Unset` when returned from the index in this situation. See [an example](https://github.com/meilisearch/meilisearch/blob/7a89abd2a025606a42f8b219e539117eb2eb029f/meilisearch-types/src/settings.rs#L608))
- [ ] The setting cannot be set when the feature is disabled, either by the main settings route or the subroute (see [`validate_settings` function](https://github.com/meilisearch/meilisearch/blob/7a89abd2a025606a42f8b219e539117eb2eb029f/meilisearch/src/routes/indexes/settings.rs#L811))
- [ ] If possible, the setting is reset when the feature is disabled (hard if it requires reindexing)

## Impacted teams

@meilisearch/docs-team @meilisearch/cloud-team

francoism90:
This would be a great addition. :)

On MySQL you have an option to auto upgrade (also on minor upgrades), but on PostgreSQL you haven't. 

It's no excuse not to make snapshots of your databases, but sometimes it isn't that important (like you can sync it manually later, e.g. Laravel Scout) and creating dumps of static data, isn't something you want to do on a daily basis.

PR #5264, opened 2025-01-21
Dumpless upgrade

irevoire:
# Pull Request
Usage: https://meilisearch.notion.site/Dumpless-upgrade-fff4b06b651f81f1acafe24d4687b3f7?pvs=74

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/5162

## What does this PR do?
- Implement the dumpless upgrade with multiple hooks:
  - In meilisearch directly before the task queue has been opened
  - In the index-scheduler while processing the task
  - In milli while upgrading the indexes
- There is no hook at search/query time to handle the old version of a database. That's left to the next person upgrading a database
- A new special type of task (`upgradeDatabase`) that can be retried has been introduced
- A new experimental cli flag has been introduced
- The version has been upgraded to the v1.13.0 in this PR otherwise it was a lot of useless work to test the dumpless upgrade
- Multiple tests have been introduced

## PR checklist
- [ ] Update the issue template we use for features, mentioning what we should do in case of a database upgrade
- [x] The experimental feature discussion should be opened and updated in the PR
- [x] Update the PRD
    - [x] Add the new error codes => No new error code since it's an internal error when it fails
    - [x] Add the task details
    - [x] Add the telemetry

## Notes

The new tests introduced are not _that_ slow
image


crates/meilisearch/src/lib.rs
```diff
@@ -334,13 +339,35 @@ fn open_or_create_database_unchecked(
     }
 }
 
+/// Ensures Meilisearch version is compatible with the database, returns an error versions mismatch.
```
dureuill, line 342:
```suggestion
/// Ensures Meilisearch version is compatible with the database, returns an error in case of version mismatch.
```

crates/index-scheduler/src/upgrade/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    let start = match from {
+        (1, 12, _) => 0,
+        (major, minor, patch) => {
+            if major > current_major
+                || (major == current_major && minor > current_minor)
+                || (major == current_major && minor == current_minor && patch > current_patch)
+            {
+                bail!(
+                "Database version {major}.{minor}.{patch} is higher than the binary version {current_major}.{current_minor}.{current_patch}. Downgrade is not supported",
```
dureuill, line 28:
```suggestion
                "Database version {major}.{minor}.{patch} is higher than the Meilisearch version {current_major}.{current_minor}.{current_patch}. Downgrade is not supported",
```

crates/index-scheduler/src/upgrade/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+                );
+            } else if major < 1 || (major == current_major && minor < 12) {
+                bail!(
+                "Database version {major}.{minor}.{patch} is too old for the experimental dumpless upgrade feature. Please generate a dump using the v{major}.{minor}.{patch} and imports it in the v{current_major}.{current_minor}.{current_patch}",
```
dureuill, line 32:
```suggestion
                "Database version {major}.{minor}.{patch} is too old for the experimental dumpless upgrade feature. Please generate a dump using the v{major}.{minor}.{patch} and import it in the v{current_major}.{current_minor}.{current_patch}",
```

crates/index-scheduler/src/upgrade/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    info!("Upgrading the task queue");
+    for (upgrade, upgrade_name) in upgrade_functions[start..].iter() {
+        info!("{upgrade_name}");
+        (upgrade)(tasks_path)?;
+    }
+
+    let env = unsafe {
+        heed::EnvOpenOptions::new()
+            .max_dbs(19)
```
dureuill, line 48: can we fetch this number from a constant?

crates/index-scheduler/src/upgrade/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+            // Since that's the only database memory-mapped currently we don't need to check the budget yet
+            .map_size(100 * 1024 * 1024)
```
dureuill, line 50: can we fetch the map size from a constant?

crates/meilisearch-types/src/tasks.rs
```diff
@@ -360,6 +369,7 @@ impl From<&KindWithContent> for Option<Details> {
             }),
             KindWithContent::DumpCreation { .. } => Some(Details::Dump { dump_uid: None }),
             KindWithContent::SnapshotCreation => None,
+            KindWithContent::UpgradeDatabase { .. } => None,
```
dureuill, line 372:
```suggestion
            KindWithContent::UpgradeDatabase { from: .. } => Some(Details::UpgradeDatabase { upgrade_from: Some()}),
```

crates/index-scheduler/src/scheduler/mod.rs
```diff
@@ -221,7 +223,7 @@ impl IndexScheduler {
                     self.queue
                         .tasks
                         .update_task(&mut wtxn, &task)
-                        .map_err(|e| Error::TaskDatabaseUpdate(Box::new(e)))?;
+                        .map_err(|e| Error::TaskDatabaseUpgrade(Box::new(e)))?;
```
dureuill, line 226: consider removing the original `TaskDatabaseUpdate` and all its uses here.

crates/index-scheduler/src/scheduler/process_batch.rs
```diff
@@ -314,6 +315,26 @@ impl IndexScheduler {
...
                 Ok(vec![task])
             }
+            Batch::UpgradeDatabase { mut tasks } => {
+                let ret = catch_unwind(AssertUnwindSafe(|| self.process_upgrade(progress)));
+                match ret {
+                    Ok(Ok(())) => (),
+                    Ok(Err(e)) => return Err(Error::TaskDatabaseUpgrade(Box::new(e))),
+                    Err(_e) => {
+                        return Err(Error::TaskDatabaseUpgrade(Box::new(
+                            Error::ProcessBatchPanicked,
```
dureuill, line 325: todo in a new issue: have `ProcessBatchPanicked` take the payload, and rebuild a message from it (look in `crates/xtask/src/bench` for way to downcast a panic payload)

crates/index-scheduler/src/scheduler/process_upgrade/mod.rs
```diff
@@ -0,0 +1,40 @@
+use meilisearch_types::milli;
+use meilisearch_types::milli::progress::{Progress, VariableNameStep};
+
+use crate::processing::UpgradeDatabaseProgress;
+use crate::{Error, IndexScheduler, Result};
+
+impl IndexScheduler {
+    pub(super) fn process_upgrade(&self, progress: Progress) -> Result<()> {
+        progress.update_progress(UpgradeDatabaseProgress::EnsuringCorrectnessOfTheSwap);
```
dureuill, line 9: this might require a change

crates/index-scheduler/src/scheduler/process_upgrade/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+                .map_err(|e| Error::from_milli(e, Some(uid.to_string())))?;
+            if regen_stats {
+                let stats = crate::index_mapper::IndexStats::new(&index, &wtxn)
+                    .map_err(|e| Error::from_milli(e, Some(uid.to_string())))?;
+                // Release wtxn as soon as possible because it stops us from registering tasks
+                let mut index_schd_wtxn = self.env.write_txn()?;
+                self.index_mapper.store_stats_of(&mut index_schd_wtxn, uid, &stats)?;
+                index_schd_wtxn.commit()?;
+            }
+            wtxn.commit()?;
```
dureuill, line 35: consider committing the index before regenerating the stats

crates/milli/src/update/upgrade/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+use crate::{Index, InternalError, Result};
+
+/// Return true if the cached stats of the index must be regenerated
+pub fn upgrade(wtxn: &mut RwTxn, index: &Index, progress: Progress) -> Result<bool> {
+    let from = index.get_version(wtxn)?;
+    let upgrade_functions = [
+        (
+            v1_12_to_v1_12_3 as fn(&mut RwTxn, &Index, Progress) -> Result<bool>,
+            "Upgrading from v1.12.(0/1/2) to v1.12.3",
+        ),
```
dureuill, line 16: consider introducing a Trait here:

```rust
trait Upgrade {
    fn upgrade(&mut RwTxn, &Index, Progress) -> Result<bool>;
    fn message() -> &'static str;
    fn target_version() -> (usize, usize, usize);
}
```

crates/milli/src/update/upgrade/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+    enum UpgradeVersion {}
+    let upgrade_path = &upgrade_functions[start..];
+
+    let mut regenerate_stats = false;
+    for (i, (upgrade_function, upgrade_msg)) in upgrade_path.iter().enumerate() {
+        progress.update_progress(VariableNameStep::<UpgradeVersion>::new(
+            upgrade_msg.to_string(),
+            i as u32,
+            upgrade_path.len() as u32,
+        ));
```
dureuill, line 43: call the trait and then call `index.put_version(trait_object.target_version())`

crates/milli/src/index.rs
```diff
@@ -253,7 +252,22 @@ impl Index {
             vector_arroy,
             embedder_category_id,
             documents,
-        })
+        };
+        if this.get_version(&wtxn)?.is_none() {
+            this.put_version(
```
dureuill, line 257: this is only true for indexes that are indeed created at this point.

older indexes should be assigned `v1.12.0` abruptly or be recovered from Meilisearch.

crates/index-scheduler/src/queue/batches.rs
```diff
@@ -172,30 +178,91 @@ impl BatchQueue {
...
+        // Update the enqueued_at, we cannot retrieve the previous enqueued at from the previous batch and
+        // must instead go through the db looking for it. We cannot look at the task contained in this batch either
+        // because they may have been removed.
+        // What we know though is that the task date from before the enqueued_at and only two timestamp have been written
+        // to the DB per batches.
+        if let Some(ref old_batch) = old_batch {
+            let started_at = old_batch.started_at.unix_timestamp_nanos();
+            let mut exit = false;
+            let mut iterator = self.enqueued_at.rev_iter_mut(wtxn)?;
+            while let Some(entry) = iterator.next() {
```
dureuill, line 234: The early exit won't work for batches containing a single task. Please handle this case separately

crates/index-scheduler/src/test_utils.rs
```diff
@@ -349,4 +390,18 @@ impl IndexSchedulerHandle {
...
+    // Wait for one failed batch.
+    #[track_caller]
+    pub(crate) fn scheduler_is_down(&mut self) {
+        loop {
+            match self
+            .test_breakpoint_rcv
+            .recv_timeout(std::time::Duration::from_secs(1)) {
+                Ok((_, true)) => continue,
+                Ok((b, false)) => panic!("The scheduler was supposed to be down but successfully moved to the next breakpoint: {b:?}"),
+                Err(RecvTimeoutError::Timeout | RecvTimeoutError::Disconnected) => break,
```
dureuill, line 403: does that work with just disconnected instead of the timeout?

crates/index-scheduler/src/scheduler/test_failure.rs
```diff
@@ -249,3 +250,68 @@ fn panic_in_process_batch_for_index_creation() {
...
+    let kind = index_creation_task("girafo", "leaves");
+    let _task = index_scheduler.register(kind, None, false).unwrap();
+    // The scheduler is up and running
+    handle.advance_one_successful_batch();
+    handle.advance_one_successful_batch();
+    handle.advance_one_failed_batch(); // doggo already exists
+    handle.advance_one_successful_batch();
+    snapshot!(snapshot_index_scheduler(&index_scheduler), name: "after_processing_everything");
+
+    // When deleting the single upgrade task it should remove the three associated batches
```
dureuill, line 303:
```suggestion
    // When deleting the single upgrade task it should remove the associated batche
```

crates/index-scheduler/src/lib.rs
```diff
@@ -366,6 +369,7 @@ impl IndexScheduler {
                     match ret {
                         Ok(Ok(TickOutcome::TickAgain(_))) => (),
                         Ok(Ok(TickOutcome::WaitForSignal)) => run.scheduler.wake_up.wait(),
+                        Ok(Ok(TickOutcome::StopProcessingForever)) => break,
```
dureuill, line 372: can we check that `run` is never called again?

crates/milli/src/update/upgrade/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+        ),
+    ];
+
+    let start = match from {
+        // If there was no version it means we're coming from the v1.12
+        None | Some((1, 12, 0..=2)) => 0,
+        Some((1, 12, 3..)) => 1,
+        // We must handle the current version in the match because in case of a failure some index may have been upgraded but not other.
+        Some((1, 13, _)) => return Ok(false),
+        Some((major, minor, patch)) => {
```
dureuill, line 29: would be good to find a way to test that a newly released test version is handled here

crates/meilisearch/tests/upgrade/v1_12/v1_12_0.rs
```diff
@@ -0,0 +1,268 @@
...
+    "#);
+
+    // Everything worked
+    let (keys, _) = server.list_api_keys("").await;
+    snapshot!(json_string!(keys, { ".results[].updatedAt" => "[date]" }), name: "list_all_keys_after_removing_kefir");
+}
+
+/// We must ensure the index-scheduler database is still working:
+/// 1. We can query the indexes and their metadata
+/// 2. The upgrade task has been spawned and has been processed (wait for it to finish or it'll be flaky)
```
dureuill, line 102: a way for the test not to be flaky and have unprocessed task would be to test the case of a failing upgrade.

Properties to check:

1. After the upgrade task fails, no task is ever processed
2. The VERSION file has been rewritten
3. After the upgrade task fails, it is still possible to enqueue new tasks
4. Restarting the server after the task failed restart the task and attempts it again

- Testing (1), (2), (3) can be done by doctoring a data.ms to contain e.g. an index with the wrong version that will cause the upgrade to fail.
- Testing (4) can be done by doctoring a data.ms to contain a failed upgrade task (for example, as in the previous bullet point), then doctoring again such that the task will pass (for example, by removing the wrong version in the faulty index)

crates/meilisearch-types/src/tasks.rs
```diff
@@ -607,6 +622,9 @@ pub enum Details {
     IndexSwap {
         swaps: Vec<IndexSwap>,
     },
+    UpgradeDatabase {
+        from: (usize, usize, usize),
```
dureuill, line 626: Would be great to also have `to: (usize, usize, usize)` here, so that we can track from which to which versions a DB has been upgraded

irevoire: But we're always upgrading to the latest version, right?
If we add this information, we should update the details of the task, every time we restart the task I guess? :thinking:

dureuill: If I look at an upgrade task I did 2 weeks ago I might be happy to know if it was targeting v1.13.2 or v1.13.9

crates/meilisearch-types/src/tasks.rs
irevoire: There was other issues in the end, it's fixed and tested :ok_hand:

crates/index-scheduler/src/test_utils.rs
irevoire: I checked, and no, since the index-scheduler receiver channel has not been dropped it can still answer with `Timeout` :pensive:

crates/meilisearch-types/src/tasks.rs
irevoire: So from what I understand, you would prefer the `to` to be created once when it's enqueued and then never change again?

crates/index-scheduler/src/scheduler/process_upgrade/mod.rs
irevoire: Actually I don't think we have to write anything here, I got rid of this type entirely

crates/meilisearch/src/lib.rs
```diff
@@ -334,13 +339,35 @@ fn open_or_create_database_unchecked(
...
+}
+
 /// Ensure you're in a valid state and open the IndexScheduler + AuthController for you.
 fn open_or_create_database(
     opt: &Opt,
     empty_db: bool,
 ) -> anyhow::Result<(IndexScheduler, AuthController)> {
     if !empty_db {
-        check_version_file(&opt.db_path)?;
+        check_version_and_update_task_queue(&opt.db_path, opt.experimental_dumpless_upgrade)?;
```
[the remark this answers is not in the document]
irevoire: Now we write the version file immediately and exit this function.
Later on, when the index-scheduler is opened, it checks its version and updates itself in the `new` function.
To make that works, I also had to store the version in the index-scheduler

crates/index-scheduler/src/lib.rs
irevoire: I checked manually and that seems to be the case.
If you were thinking about something automatic, I think we should open an issue and do it in another PR since the index-scheduler was already not supposed to work with two run functions executing at the same time.

crates/index-scheduler/src/scheduler/mod.rs
irevoire: I went for opt 2

crates/index-scheduler/src/scheduler/process_batch.rs
irevoire: I opened the issue for the v1.13

crates/index-scheduler/src/queue/tasks.rs
```diff
@@ -9,12 +9,17 @@ use time::OffsetDateTime;
...
 use super::{Query, Queue};
 use crate::processing::ProcessingTasks;
-use crate::utils::{self, insert_task_datetime, keep_ids_within_datetimes, map_bound};
+use crate::utils::{
+    self, insert_task_datetime, keep_ids_within_datetimes, map_bound, remove_task_datetime,
+};
 use crate::{Error, Result, TaskId, BEI128};
 
+/// The number of database used by the task queue
+const NUMBER_OF_DATABASES: u32 = 8;
```
dureuill, line 18: For the future: wondering if we couldn't get the number of database from an enum rather than manually counting the `const`s

crates/index-scheduler/src/versioning.rs
```diff
@@ -0,0 +1,68 @@
...
+        let patch = versioning::VERSION_PATCH.parse().unwrap();
+        self.set_version(wtxn, (major, minor, patch))
+    }
+
+    /// Create an index scheduler and start its run loop.
+    pub(crate) fn new(env: &Env, db_version: (u32, u32, u32)) -> Result<Self> {
+        let mut wtxn = env.write_txn()?;
+        let version = env.create_database(&mut wtxn, Some(db_name::VERSION))?;
+        let this = Self { version };
+        let from = this.get_version(&wtxn)?.unwrap_or(db_version);
```
dureuill, line 48: Don't we need to write the index-scheduler version immediately if it is `None`. This is because at this point we already rewrote the version file?

Ideally we would:

1. read the index-scheduler version, defaulting to db version
2. write the db version as the index-scheduler version if it was None, committing the txn
3. Only then, write the VERSION file
4. We probably want to register the the task and change the index-scheduler version to latest in the same wtxn as well

crates/index-scheduler/src/upgrade/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+    to: (u32, u32, u32),
+) -> anyhow::Result<()> {
+    let current_major = to.0;
+    let current_minor = to.1;
+    let current_patch = to.2;
+
+    let upgrade_functions: &[&dyn UpgradeIndexScheduler] = &[&V1_12_ToCurrent {}];
+
+    let start = match from {
+        (1, 12, _) => 0,
```
dureuill, line 30: We will need to explicitly add v1.13.0 to this match when releasing v1.13.1

dureuill (dismissed):
Thank you for your phenomenal work :tada:

Let's usher into a new era of dumpless upgrade :heart:

bors merge

dureuill (approved):
bors merge

Issue #5256, opened 2025-01-20
Fix the flaky test


irevoire:
The tests are still flaky on main because of the batches, that's not expected.
See: https://github.com/meilisearch/meilisearch/actions/workflows/flaky-tests.yml

irevoire:
Should be fixed by #5272, but let's wait a bit before closing the issue since these flaky tests are hard to find

irevoire:
The new remote search work introduced a new flaky test ; https://github.com/meilisearch/meilisearch/actions/runs/13244914553/job/36968929414?pr=5326

PR #5272, opened 2025-01-22
Fix Batches Deletion and flaky tests

### LINKED ISSUE #5263: Removing a batch leaks some internal dates in the database
**Describe the bug**
When we remove a task, that can trigger the deletion of a batch.
When we remove this batch we remove a lot of metadata associated to the batch, but we don't remove the associated dates from the `enqueued_at`, `started_at` and `finished_at` databases.

**To Reproduce**
We can see it typically in this snapshot test: https://github.com/meilisearch/meilisearch/blob/main/crates/index-scheduler/src/scheduler/snapshots/test.rs/task_deletion_deleteable/task_deletion_processed.snap

We deleted a task, the associated batch has been removed and only the batch with the uid `1` remains.
That's successfully reflected in the mapping, statuses and kinds, but in all the time we can still see mention of the batch with the ID `0`.

**NOTE**:
This doesn't cause a panic while retrieving batches!
So the issue is really minor for now but should still be fixed because otherwise we may end up filling a queue with useless data.

**Expected behavior**
The batches are successfully removed.

**Meilisearch version:**
v1.12.0
Since the introduction of the batches.

Kerollmops:
- This issue fixes #5263 by removing the batches from the date and time databases.
- It also introduces a new `enqueued_at` field in the batch object to quickly retrieve them in the `batches.enqueued_at` database
- Finally, it probably fixes all the flaky tests of the batches: https://github.com/meilisearch/meilisearch/issues/5256


crates/index-scheduler/src/scheduler/snapshots/test.rs/task_deletion_deleteable/task_deletion_processed.snap
```diff
@@ -54,6 +54,7 @@ succeeded [1,]
 ### Batches Index Tasks:
 ----------------------------------------------------------------------
 ### Batches Enqueued At:
+[timestamp] [0,]
```
Kerollmops: Is it normal to add 0 back in here?

crates/index-scheduler/src/queue/batches_test.rs
```diff
@@ -102,30 +102,33 @@ fn query_batches_simple() {
...
+          "indexCreation": 1
+        },
+        "indexUids": {
+          "catto": 1
         }
-        "#);
+      },
+      "startedAt": "1970-01-01T00:00:00Z",
+      "finishedAt": null,
+      "enqueuedAt": null
```
Kerollmops, line 129: We now have an enqueued at? I suppose it's an object with oldest and earliest? Can you add a test to check that, please?

crates/index-scheduler/src/queue/batches.rs
```diff
@@ -239,33 +239,21 @@ impl BatchQueue {
...
-                        }
-                    }
-                }
+                // If we don't have the enqueued at in the batch it means the database comes from the v1.12
+                // and we still need to find the date by scrolling the database
+                remove_n_tasks_datetime_earlier_than(
+                    wtxn,
+                    self.enqueued_at,
+                    old_batch.started_at,
+                    if old_batch.stats.total_nb_tasks >= 2 { 2 } else { 1 },
```
Kerollmops, line 248: Can we use a clamp here?

```rust
old_batch.stats.total_nb_tasks.clamp(1, 2),
```

crates/index-scheduler/src/queue/batches_test.rs
irevoire: Yes, there was already a test, but I improved it here;
https://github.com/meilisearch/meilisearch/pull/5272/commits/c98c3ad95a9db6fbe28744d1cfaf80d19b361f99

Now we 100% ensure we always have an enqueued_at and that it matches the tasks contained.

crates/index-scheduler/src/queue/batches.rs
irevoire: Done, for consistency I also replaced your if when removing batches with my clamp :grin: 
https://github.com/meilisearch/meilisearch/pull/5272/commits/9f21abd4f300ed4aee2988f8ea09ec799611b878

Kerollmops (commented):
Thank you very much! I let you Approve and bors merge 👌

irevoire (approved):
I fixed another flaky test

bors merge

irevoire:
bors merge

Issue #5273, opened 2025-01-22
Have `ProcessBatchPanicked` take the payload, and rebuild a message from it


irevoire:
Context:
              todo in a new issue: have `ProcessBatchPanicked` take the payload, and rebuild a message from it (look in `crates/xtask/src/bench` for way to downcast a panic payload)

_Originally posted by @dureuill in https://github.com/meilisearch/meilisearch/pull/5264#discussion_r1924075346_

dureuill:
Fixed by #5300

PR #5300, opened 2025-01-29
Improve unexpected panic message

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/5273

## What does this PR do?
- When an unexpected panic happens in the index-scheduler we catch it and rebuild an error message from the join_error
- Same when the upgrade index-scheduler fails


irevoire:
bors merge

PR #5307, opened 2025-02-03
Parallelize bulk facets & word prefix fid/position docids

### LINKED ISSUE #5290: Index facets by reading the facet tree in parallel
I recently measured the time the indexer spent on each part of the indexation. We did that with 240 million small documents. There are five filterable attributes; all the attributes are searchable, but they are very short. Note that we were indexing in batches of ten million documents, so the bulk indexing of facets was probably always used (taking more time than the incremental one). You can see [the Slack conversation there](https://meilisearch.slack.com/archives/C07NRD7JTQW/p1737718244430489).

```
processing tasks > indexing > post-processing facets , 21725s
processing tasks > indexing > extracting facets > document , 20020s
processing tasks > indexing > extracting words > document , 6548s
processing tasks > indexing > post-processing words , 5224s
processing tasks > computing document changes > preparing update file > payload , 1060s
processing tasks > indexing > writing to database , 976s
processing tasks > indexing > finalizing , 951s
processing tasks > indexing > extracting documents > document , 100s
processing tasks > computing document changes > preparing update file , 94s
processing tasks , 55s
processing tasks > indexing , 20s
processing tasks > indexing > extracting words , 2s
```

The longest part is the post-processing of facets. With these PRs on heed and LMDB, [it is possible to read an uncommitted transaction from multiple threads](https://github.com/meilisearch/heed/pull/307) in parallel. We still need to figure out how to build an algorithm to compute the facets using this advantage. Remember that the heed PR is unsafe and only here to experiment. If there is one such algorithm, we need to port an upstream patch to LMDB that is sustainable and maintainable.

Kerollmops:
This fixes #5290, and it will probably be followed by another PR to work on the incremental part of it. A prototype is available, called `prototype-speedup-post-processing-2`.

I replicated a dataset of about 30M documents to see how it behaves, and we can clearly see that the timings are divided by even more than the number of CPUs (4 in this case) and are closer to being divided by 6. Which is absolutely great 🎉 Both list of batches are linked below.

As a reviewer, you must know that you have to review the algorithms in this PR but also [the PR that modifies LMDB](https://github.com/meilisearch/lmdb/pull/3) to implement the feature to be able to read uncommitted changes from multiple threads (creating nested transactions).

[batches-original.json](https://github.com/user-attachments/files/22594032/gethookd.json)
[batches-replicated.json](https://github.com/user-attachments/files/22594038/gethookd-replicated.json)

### To Do
- [x] Use heed@0.22.1-nested-rtxns.
  - [x] In hannoy.
  - [x] In arroy.
  - [x] In Cellulite.
- [x] Introduce two experimental CLI arguments to disable the post-processing of prefixes and facets.
- [x] (will be done) [Disable it by default on the Cloud](https://linear.app/meilisearch/issue/SCA-236/disable-it-by-default-on-the-cloud).
- [ ] ~Create a follow-up PR for the incremental part~.
- [x] Create [a follow-up PR](https://linear.app/meilisearch/issue/SCA-237/fix-ghost-doc-ids-in-prefixes-with-single-threaded-prefix-post) to fix the bug in the non-multithreaded version.
- [x] Review the facet bulk implementation.
- [x] Test on big customers who seek fast word prefix docids.
- [x] Use env var `MEILI_TEST_FULL_SNAPS` to see the real snap content.

### DIFF
--- a/crates/milli/src/update/new/words_prefix_docids.rs
+++ b/crates/milli/src/update/new/words_prefix_docids.rs
@@ -4,7 +4,7 @@ use std::io::{BufReader, BufWriter, Read, Seek, Write};
 use std::iter;
 
 use hashbrown::HashMap;
-use heed::types::Bytes;
+use heed::types::{Bytes, DecodeIgnore};
 use heed::{BytesDecode, Database, Error, RoTxn, RwTxn};
 use rayon::iter::{IndexedParallelIterator as _, IntoParallelIterator, ParallelIterator as _};
 use roaring::MultiOps;
@@ -225,37 +225,33 @@ impl<'i> WordPrefixIntegerDocids<'i> {
                         continue;
                     }
 
-                    let mut bitmaps_bytes_at_positions = HashMap::new();
-                    let mut prev_pos = None;
+                    let mut bitmap_bytes_at_positions = HashMap::new();
                     for result in self
                         .database
-                        .remap_data_type::<Bytes>()
                         .prefix_iter(&rtxn, prefix.as_bytes())?
+                        .remap_types::<StrBEU16Codec, Bytes>()
                     {
-                        let (key, bitmap_bytes) = result?;
-                        let (_word, pos) =
-                            StrBEU16Codec::bytes_decode(key).map_err(Error::Decoding)?;
-
-                        // Track positions with no corresponding bitmap bytes,
-                        // these means that the prefix no longer exists in the database
-                        // and must, therefore, be removed from the index.
-                        if let Some(prev_pos) = prev_pos {
-                            // A range with the left bigger than the right bound results in an empty range.
-                            for pos in prev_pos..pos {
-                                // They are represented by an empty set of bitmaps.
-                                bitmaps_bytes_at_positions.entry(pos).or_insert_with(Vec::new);
-                            }
-                        }
-
-                        bitmaps_bytes_at_positions
+                        let ((_word, pos), bitmap_bytes) = result?;
+                        bitmap_bytes_at_positions
                             .entry(pos)
                             .or_insert_with(Vec::new)
                             .push(bitmap_bytes);
-
-                        prev_pos = Some(pos);
                     }
 
-                    for (pos, bitmaps_bytes) in bitmaps_bytes_at_positions {
+                    // We track positions with no corresponding bitmap bytes,
+                    // these means that the prefix no longer exists in the database
+                    // and must, therefore, be removed from the index.
+                    for result in self
+                        .prefix_database
+                        .prefix_iter(&rtxn, prefix.as_bytes())?
+                        .remap_types::<StrBEU16Codec, DecodeIgnore>()
+                    {
+                        let ((_word, pos), ()) = result?;
+                        // They are represented by an empty set of bitmaps.
+                        bitmap_bytes_at_positions.entry(pos).or_insert_with(Vec::new);
+                    }
+
+                    for (pos, bitmaps_bytes) in bitmap_bytes_at_positions {
                         if bitmaps_bytes.is_empty() {
                             indexes.push(PrefixIntegerEntry {
                                 prefix,


crates/milli/src/update/new/indexer/post_processing/facet_bulk.rs
```diff
@@ -0,0 +1,164 @@
...
+                );
+                db.remap_data_type::<Bytes>().put(wtxn, &key, facet_group_value)?;
+            }
+        }
+
+        base_level += 1;
+
+        // If the next level will have the minimum required groups, continue.
+        (level_size / FACET_GROUP_SIZE as usize) >= FACET_MIN_LEVEL_SIZE as usize
+    } {}
```
ManyTheFish, line 65: I think it's the same 🤔 
I moved up the `level = base_level.checked_add(1).unwrap();` and assigned `base_level = level` after the loops.
Let me know if I am wrong:
```suggestion
    while {
        let mut level_size = 0;
        let level = base_level.checked_add(1).unwrap();
        for reader in compute_level(index, wtxn, db, field_id, base_level)? {
            let mut cursor = reader.into_cursor()?;
            while let Some((left_bound, facet_group_value)) = cursor.move_on_next()? {
                level_size += 1;
                let key = FacetGroupKey { field_id, level, left_bound };
                debug_assert!(
                    db.get(wtxn, &key).transpose().is_none(),
                    "entry must not be there and must have already been deleted: {key:?}"
                );
                db.remap_data_type::<Bytes>().put(wtxn, &key, facet_group_value)?;
            }
        }

        base_level = level;

        // If the next level will have the minimum required groups, continue.
        (level_size / FACET_GROUP_SIZE as usize) >= FACET_MIN_LEVEL_SIZE as usize
    } {}
```

crates/milli/src/update/new/words_prefix_docids.rs
```diff
@@ -177,7 +185,144 @@ impl WordPrefixIntegerDocids {
...
+                        )?;
+                    }
+                    None => {
+                        self.prefix_database.delete(wtxn, &key_buffer)?;
+                    }
+                }
+            }
+        }
+
+        Ok(())
```
ManyTheFish, line 312: The source database is serialized as `word, 0, position -> docids`, which means it will be ordered as below:
```
he, 0, 0 -> [docids]
he, 0, 1 -> [docids]
he, 0, 2 -> [docids]
hell, 0, 0 -> [docids]
hell, 0, 1 -> [docids]
hell, 0, 2 -> [docids]
hello, 0, 0 -> [docids]
hello, 0, 1 -> [docids]
hello, 0, 2 -> [docids]
heritage, 0, 0 -> [docids]
heritage, 0, 1 -> [docids]
heritage, 0, 2 -> [docids]
```

Meaning that for the prefix `h`, it would create more than one `PrefixIntegerEntry` per position:

```rust
PrefixIntegerEntry { prefix: "h", pos: 0, .. } // he, 0, 0
PrefixIntegerEntry { prefix: "h", pos: 1, .. } // he, 0, 1 : because we changed position
PrefixIntegerEntry { prefix: "h", pos: 2, .. } // he, 0, 2 : because we changed position
PrefixIntegerEntry { prefix: "h", pos: 0, .. } // hell, 0, 0 : we come back to position 0 and we push the same PrefixIntegerEntry as the first one but with different docids.
...
```

then when writing into the database, we iterate over the indexes `for PrefixIntegerEntry { prefix, pos, serialized_length } in index` and override the value.

I think the persisted value for each prefixes will only contain the  `docids` of the last word starting with the prefix for each position.
In our example, we would only have the `heritage` docids for the `h` prefix.

Am I right, or am I missing something? 🤔

crates/milli/src/update/new/words_prefix_docids.rs
```diff
@@ -262,7 +407,7 @@ impl WordPrefixIntegerDocids {
     }
 }
 
-/// Represents a prefix and the lenght the bitmap takes on disk.
+/// Represents a prefix and the length the bitmap takes on disk.
```
ManyTheFish, line 397: @ManyTheFish is raeli goud in Anglish, ok?

crates/milli/src/update/new/words_prefix_docids.rs
Kerollmops: I may be wrong but the databases we are iterating on are `word_prefix_fid_docids` and `word_position_docids` which are of type `Database<StrBEU16Codec, CboRoaringBitmapCodec>` which means that the key is not composed of the zero you are showing in the middle, right?

Kerollmops: Ok, we discussed and the bug is indeed here. What I plan to do to fix it is to collect all the bitmaps under their positions for a given prefix in an hashmap.

crates/milli/src/update/new/words_prefix_docids.rs
```diff
@@ -177,7 +185,135 @@ impl WordPrefixIntegerDocids {
...
+                        // Track positions with no corresponding bitmap bytes,
+                        // these means that the prefix no longer exists in the database
+                        // and must, therefore, be removed from the index.
+                        if let Some(prev_pos) = prev_pos {
+                            // A range with the left bigger than the right bound results in an empty range.
+                            for pos in prev_pos..pos {
+                                // They are represented by an empty set of bitmaps.
+                                bitmaps_bytes_at_positions.entry(pos).or_insert_with(Vec::new);
+                            }
+                        }
```
Kerollmops: What do you think about this @ManyTheFish. Previously, in the non-multi-threaded algorithm, we never deleted prefixes because we never created entries for positions that we did not encounter while looping over the words. Now we create an empty list of bitmaps in the HashMap and therefore enter the special case to delete (probably non-existing) prefix entries.

However, now that I think a bit more about this, I see two issues:
- Deleting non-existing entries in LMDB in the wtxn loop will do more useless IOs.
- I do not generate empty entries (tombstones) for positions that are higher than the highest I encounter. This means that we will never delete entries for a prefix at high positions that are more likely to disappear (i.e., a small number of documents with a lot of prefix far in a field).

What I propose is to use a BTreeMap or a Vec with all possible positions for a prefix (I think we have a maximum somewhere), perform the entire prefix loop, and fill the different _prefix at position_ lists of bitmaps without interruption. Once we are done, we go through the prefixes and verify that the ones that must be deleted actually exist before deleting them (writing the deletion in the operation file). The advantage is that we perform all these checks in parallel, rather than in a single transaction loop.

crates/milli/src/update/new/words_prefix_docids.rs
```diff
@@ -177,7 +185,131 @@ impl WordPrefixIntegerDocids {
...
+                    // and must, therefore, be removed from the index.
+                    for result in self
+                        .prefix_database
+                        .prefix_iter(&rtxn, prefix.as_bytes())?
+                        .remap_types::<StrBEU16Codec, DecodeIgnore>()
+                    {
+                        let ((_word, pos), ()) = result?;
+                        // They are represented by an empty set of bitmaps.
+                        bitmap_bytes_at_positions.entry(pos).or_insert_with(Vec::new);
+                    }
```
Kerollmops, line 252: @ManyTheFish, I fixed the bug where prefixes were never deleted by iterating on the current prefixes and listing all the positions that were not being modified after this update. I do this in parallel and avoid doing too many get operations by directly iterating over the existing ones instead.

I haven't yet fixed the non-multi-threaded algorithm. I am waiting for your validation before spending too much time on a fix, but I would like to fix it the same way (iterating over current prefixes).

crates/milli/src/update/new/words_prefix_docids.rs
ManyTheFish: In the previous implementation, we were calling delete prefixes with a list of prefixes that are no longer into the prefix FST:
https://github.com/meilisearch/meilisearch/blob/b98e2cef8165d2f7fd4fd9c1607fa7d37589617a/crates/milli/src/update/new/words_prefix_docids.rs#L43

However, you are right, because if a prefix doesn't disappear from the FST, BUT disappears completely from all the "title" of every document, we could have residual empty bitmaps, but I can't say if there will be a lot of them 🤔 


To me, the bug comes from the way we evaluate prefixes, we are relying on the global FST that is not aware of the word distribution in the fields.

crates/milli/src/update/new/words_prefix_docids.rs
ManyTheFish: yes you are right

crates/milli/src/update/new/words_prefix_docids.rs
ManyTheFish: > "we could have residual empty bitmaps"

I am wrong here, the bitmaps are filled with wrong IDs, and this may make the search crash in the case the document ID doesn't exist anymore.

So it's more concerning that I thought

crates/milli/src/update/new/indexer/post_processing/mod.rs
```diff
@@ -239,9 +244,14 @@ fn compute_facet_level_database(
...
-                tracing::debug!(%fid, "bulk string facet processing");
-                FacetsUpdateBulk::new_not_updating_level_0(index, vec![fid], FacetType::String)
-                    .execute(wtxn)?
+                if grenad_parameters.experimental_no_edition_2024_for_facet_post_processing {
+                    tracing::debug!(%fid, "bulk string facet processing");
+                    FacetsUpdateBulk::new_not_updating_level_0(index, vec![fid], FacetType::String)
+                        .execute(wtxn)?
+                } else {
+                    tracing::debug!(%fid, "bulk string facet processing in parallel");
+                    generate_facet_levels(index, wtxn, fid, FacetType::String)?
```
Kerollmops, line 253: Hey @ManyTheFish 👋 Don't you think I forgot to call this very function on the Bulk Numbers part too? (30 lines below) 🙊

PR #5309, opened 2025-02-04
Meilitool Export Documents v1.12.8

Kerollmops:
This PR super seeds #4970 and is based on v1.12.8. We are creating a new prototype from that to help a customer to extract the documents. This will eventually land on v1.13.

# To Do
- [ ] Introduce the offset param.


dureuill:
do you think you can rebase this again on v1.13.0, add the `--offset` option and the progress on stderr? Then we can open an issue saying we should eventually migrate to the docs API and land in a coming RC.

Kerollmops:
@dureuill, I'm closing this in favor of #4970 that is already in v1.13 now.

Issue #5311, opened 2025-02-04
Stop retrieving the OSS metrics every hours


irevoire:
The metrics help us know if the features we develop are used and how.

Today, we have so many users that we no longer need to gather that many metrics.
In the previous version, we already removed many events from the list of tracked actions; this time, we also want to reduce how often we retrieve these metrics.

curquiza:
fixed by #5312

PR #5312, opened 2025-02-04
Send the OSS analytics once per day instead of once per hour

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/5311

## What does this PR do?
- If the instance is OSS => we send the analytics once every day
- If the instance is on the meilisearch cloud => we send the analytics every hour


ManyTheFish (approved):
LGTM,
bors merge

PR #5322, opened 2025-02-06
Make sure arroy is using the rayon thread-pool

### LINKED ISSUE #5249: High CPU usage during indexing, even with max_indexing_threads and default settings
Our meilisearch instance is running on a 128Gib RAM 16 core server. 

There is a single index with 5-6 attributes per document and 768 dimension embeddings supplied by `userProvided` embedder.

New documents are sent in 15.000 document batches to be added to the index. During indexing, the server CPU resources seem to be pushed to 100% on all cores. Setting `max_indexing_threads` does not seem to change anything (tried 6, 12 and leaving for default).

Image

We are at 7.5 million documents and 122.5Gb size on this index. Is it pushing too hard for this server specs? Am I missing something on the resource use characteristics?

**Meilisearch version:**
1.12.3

Kerollmops:
This PR fixes #5249 by ensuring arroy uses the rayon thread pool.

### DIFF
--- a/crates/milli/src/update/new/indexer/mod.rs
+++ b/crates/milli/src/update/new/indexer/mod.rs
@@ -191,13 +191,16 @@ where
 
         indexing_context.progress.update_progress(IndexingStep::WritingEmbeddingsToDatabase);
 
-        build_vectors(
-            index,
-            wtxn,
-            index_embeddings,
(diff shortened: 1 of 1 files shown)


dureuill (approved):
bors merge

PR #5326, opened 2025-02-10
Expose a route to get the file content associated with a task

Kerollmops:
This PR exposes a new `/tasks/{taskUid}/file` route, exposing the update file associated with a task.

## To Do
- [ ] (optional) Change the route to `/tasks/{taskUid}/documents` @dureuill.
- [ ] Update the PRD
- [ ] (optional) Make this route experimental and enable it via the experimental route.

---

Hey @meilisearch/docs-team 👋

This PR introduces a new route that is under `/tasks/{taskUid}/file`, which will return the content of the update file (documents_ in NDJSON format and 404 if either the task doesn't exist (*task_not_found*) or if this kind of task does not expose any updated file, i.e., Settings, index deletion (*task_file_not_found*). The latter error code is a new one.

### DIFF
--- a/crates/index-scheduler/src/error.rs
+++ b/crates/index-scheduler/src/error.rs
@@ -109,6 +109,8 @@ pub enum Error {
     InvalidIndexUid { index_uid: String },
     #[error("Task `{0}` not found.")]
     TaskNotFound(TaskId),
+    #[error("Task `{0}` does not provide any content file.")]
+    TaskFileNotFound(TaskId),
     #[error("Batch `{0}` not found.")]
     BatchNotFound(BatchId),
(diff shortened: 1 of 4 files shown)


crates/index-scheduler/src/error.rs
```diff
@@ -109,6 +109,8 @@ pub enum Error {
     InvalidIndexUid { index_uid: String },
     #[error("Task `{0}` not found.")]
     TaskNotFound(TaskId),
+    #[error("Task `{0}` does not provide any content file.")]
```
irevoire, line 112:
```suggestion
    #[error("Task `{0}` does not contain any documents. Only `documentAdditionOrUpdate` tasks with the statuses `enqueued` or `processing` contain documents")]
```

Hey, what do you think of this more specific message to reduce questions when someone use it wrong?

irevoire (dismissed):
Hey, overall the code seems good to me, I proposed a slight change on the error message to guide users (or ourselves) more.
Feel free to merge with or without the message.

Also there is no test at all, nor for the happy path or the error path, I guess that’s ok as long as the feature stays experimental but we should really remember to write one if we want to stabilize it ahah

irevoire (approved):

PR #5340, opened 2025-02-12
Randomly checking arroy keys before a last iteration

Kerollmops:
This PR is based on #5336 and multithread the hair drying command to make it go X times fast from 22 minutes to XX minutes for a 305GiB database.


Kerollmops:
This PR is stalled, and we could reimplement it if needed.

PR #5371, opened 2025-02-26
Composite embedders

### LINKED ISSUE #5343: Composite embedders
Related product team resources: [PRD](https://www.notion.so/meilisearch/AI-search-Performance-Optimization-Independent-models-at-search-and-indexing-stages-14a4b06b651f808b98a7e5fce25c844e) (_internal only_)
Related product discussion:

## Motivation

Search speed and latency reduction when using AI and embedders.

## Usage

https://meilisearch.notion.site/Composite-embedder-usage-14a4b06b651f81859dc3df21e8cd02a0?pvs=4

- [ ] Update the openAPI file with utoipa:
  - [ ] If a new module has been introduced, create a new structure deriving [the OpenAPI proc-macro](https://docs.rs/utoipa/latest/utoipa/derive.OpenApi.html) and nest it in the main [openAPI structure](https://github.com/meilisearch/meilisearch/blob/f2185438eed60fa32d25b15480c5ee064f6fba4a/crates/meilisearch/src/routes/mod.rs#L64-L78).
  - [ ] If a new route has been introduced, add the [path decorator](https://docs.rs/utoipa/latest/utoipa/attr.path.html) to it and add the route at the top of the file in its openAPI structure.
  - [ ] If a structure which is deserialized or serialized in the API has been introduced or modified, it must derive the [`schema`](https://docs.rs/utoipa/latest/utoipa/macro.schema.html) or the [`IntoParams`](https://docs.rs/utoipa/latest/utoipa/derive.IntoParams.html) proc-macro.
  - [ ] Once everything is done, start Meilisearch with the swagger flag: `cargo run --features swagger`, open `http://localhost:7700/scalar` on your browser, and ensure everything works as expected.

## Impacted teams

@meilisearch/docs-team, @meilisearch/integration-team, @meilisearch/cloud-team

dureuill:
# Pull Request

## Related issue
Fixes #5343 

## What does this PR do?
- Implement [public usage](https://www.notion.so/meilisearch/Composite-embedder-usage-14a4b06b651f81859dc3df21e8cd02a0)
- Refactor the way we check if a parameter is mandatory/allowed/disallowed for a given source
- Take the "nesting context" into account for computer if a parameter is mandatory/allowed/disallowed
- Add tests checking all parameters with all sources, and made sure the results didn't change compared with v1.13

## Dumpless Upgrade

- This adds a new value for an existing parameter => compatible without change
- This adds new optional parameters => compatible without change


crates/milli/src/vector/settings.rs
```diff
@@ -333,6 +739,223 @@ impl SettingsDiff {
...
+        new_api_key: Setting<String>,
+        new_dimensions: Setting<usize>,
+        new_document_template: Setting<String>,
+        new_document_template_max_bytes: Setting<usize>,
+        new_url: Setting<String>,
+        new_request: Setting<serde_json::Value>,
+        new_response: Setting<serde_json::Value>,
+        new_headers: Setting<BTreeMap<String, String>>,
+    ) {
+        // **Warning**: do not use short-circuiting || here, we want all these operations applied
```
irevoire, line 882: I was a bit surprised to see that the `diff` function was actually mostly modifying its parameter, would be nice to find a better name, but I don’t have one 😅

crates/milli/src/vector/settings.rs
```diff
@@ -333,6 +739,223 @@ impl SettingsDiff {
...
+        }
+        if dimensions.apply(new_dimensions) {
+            match *source {
+                // regenerate on dimensions change in OpenAI since truncation is supported
+                Setting::Set(EmbedderSource::OpenAi) | Setting::Reset => {
+                    ReindexAction::push_action(reindex_action, ReindexAction::FullReindex);
+                }
+                // for all other embedders, the parameter is a hint that should not be able to change the result
+                // and so won't cause a reindex by itself.
+                _ => {}
```
irevoire, line 920: And from what I saw it’s caught earlier in the process and would return an error right?

crates/milli/src/vector/settings.rs
```diff
@@ -411,148 +1044,374 @@ fn apply_default_for_source(
...
+            }),
+            (FieldStatus::Mandatory, _) => Err(UserError::MissingFieldForSource {
+                field: field.name(),
+                source_: source,
+                embedder_name: embedder_name.to_owned(),
+            }),
+        }
+    }
+
+    pub(crate) fn field_status(
```
irevoire, line 1314: That’s a banger function

irevoire (changes requested):
Maybe it’s not the right best place or time to ask this, but shouldn’t we let people user provide their embeddings at indexing time but let meilisearch compute them at search time?
Since Meilisearch is not supposed to be your main database I feel like it makes sense to generate the embeddings for your documents in your main database, but let meilisearch handle the search requests 🤔 
Also, I may be wrong but I feel like « a lot » of our clients come with their own embeddings but are still using common models that Meilisearch could run for them.

crates/milli/src/vector/settings.rs
dureuill: `apply_and_diff`?

crates/milli/src/vector/settings.rs
dureuill: not necessarily, it could be accompanied by changes of models that actually changes the dimensions.

This code supports the use case of e.g. ollama or rest, where dimensions is optional and is used as a hint to spare us a test embedding.

- Adding or removing the hint should not change the actual dimensions, so no reindexing
- Adding a hint with a wrong dimensions will be caught and result in an error
- Changing the dimensions is possible if the model is changed at the same time, but then the model change will cause reindexing

crates/meilisearch/tests/settings/vectors.rs
```diff
@@ -0,0 +1,269 @@
...
+    value[param] = valid_parameter(source, param).0;
+    let (response, code) = index
+        .update_settings(crate::json!({
+            "embedders": {
+                "test": value
+            }
+        }))
+        .await;
+    snapshot!(code, name: stringify!($source-$param-sending_code));
+    snapshot!(json_string!(response, {".enqueuedAt" => "[enqueuedAt]", ".taskUid" => "[taskUid]"}), name: stringify!($source-$param-sending_result));
```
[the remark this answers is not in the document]
dureuill: fixed! thanks for concat!

crates/milli/src/vector/composite.rs
```diff
@@ -0,0 +1,280 @@
...
+            SubEmbedderOptions::UserProvided(options) => {
+                Self::UserProvided(manual::Embedder::new(options))
+            }
+            SubEmbedderOptions::Rest(options) => {
+                Self::Rest(rest::Embedder::new(options, rest::ConfigurationSource::User)?)
+            }
+        })
+    }
+
+    pub fn embed(
```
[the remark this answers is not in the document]
dureuill: After discussing, we removed instead the `Embedder::embed` in the parent module

crates/milli/src/vector/hf.rs
```diff
@@ -255,34 +255,8 @@ impl Embedder {
...
-        let embeddings = self
-            .model
-            .forward(&token_ids, &token_type_ids, None)
-            .map_err(EmbedError::model_forward)?;
-
-        let embeddings = Self::pooling(embeddings, self.pooling)?;
-
-        let embeddings: Vec<Embedding> = embeddings.to_vec2().map_err(EmbedError::tensor_shape)?;
-        Ok(embeddings)
+    pub fn embed(&self, texts: Vec<String>) -> std::result::Result<Vec<Embedding>, EmbedError> {
```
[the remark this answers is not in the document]
dureuill: Same as above

crates/milli/src/vector/composite.rs
```diff
@@ -0,0 +1,280 @@
...
+            header: arroy::distances::Cosine::new_header(&right),
+            vector: right,
+        };
+
+        let distance = arroy::distances::Cosine::built_distance(&left, &right);
+
+        if distance > super::MAX_COMPOSITE_DISTANCE {
+            return Err(NewEmbedderError::composite_embedding_value_mismatch(distance, hint));
+        }
+    }
```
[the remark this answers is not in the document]
dureuill: there might be, we will make this feature experimental in a subsequent PR, also removing the error is not a breaking change

dureuill (commented):
N/A

irevoire (dismissed):
Thank you

bors merge

irevoire (approved):
bors merge

PR #5395, opened 2025-03-06
Update process for dumpless upgrade

irevoire:
# Pull Request

## What does this PR do?
- Add a list of checks to do on the sprint issue
- Enforce the usage of the `db change` or `no db change` label on PRs

### DIFF
--- a/.github/ISSUE_TEMPLATE/sprint_issue.md
+++ b/.github/ISSUE_TEMPLATE/sprint_issue.md
@@ -22,6 +22,18 @@ Related product discussion:
 
 <!---If necessary, create a list with technical/product steps-->
 
+### Are you modifying a database?
+- [ ] If not, add the `no db change` label to your PR, and you're good to go.
+- [ ] If yes, add the `db change` label to your PR and you're good to go.
+  - [ ] /!\ Ensure all the read operations still work!
+    - If the change happened in milli, you may need to check the version of the database before doing any read operation
+    - If the change happened in the index-scheduler, make sure the new code can immediately read the old database
+    - If the change happened in the meilisearch-auth database, reach out to the team; we don't know yet how to handle these changes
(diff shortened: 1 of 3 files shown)


dureuill:
It works well! 

I'm still wondering if putting the instructions in the sprint issue is the best location. I know I don't often read the instructions that are already there (settings API changes), I wonder if we could somehow have it more in our face.

For example, setting the `db change` label could have a bot send us a comment with the instructions on the PR.

Also, it is unclear to me what label I should put if I change the db in a **forward-compatible way**. Should I use `no db change` or `db change`. If the latter, maybe we should add that case to the instructions?

irevoire:
> Also, it is unclear to me what label I should put if I change the db in a forward-compatible way. Should I use no db change or db change. If the latter, maybe we should add that case to the instructions?

I would say it’s the latter because we might want to double-check this PR in the future, but it’s true that the naming is not really good 🤔 
Feel free to propose something else if you have an idea, I don’t currently. Maybe I should just add a third label instead?

> For example, setting the db change label could have a bot send us a comment with the instructions on the PR.

With a third label that would be nice, we would receive exactly the instruction we need to read.

> I'm also wondering if we should maybe remove the label when a new commit arrives, as a means of reevaluating the situation. Or is that too drastic 🤔

I feel like it’s going to bother us a lot, I would be in favor of trying without it at first and then seeing if we make mistakes

dureuill (dismissed):
Approving the initial version, feel free to extend with my suggestion

dureuill (approved):
I love it, thank you for listening to my feedback :heart:

dureuill:
bors merge

dureuill:
bors merge

dureuill:
sounds like bors is waiting forever on `Check db change labels (Running)`

irevoire:
bors merge

PR #5406, opened 2025-03-11
Bump heed to v0.22 and arroy to v0.6

Kerollmops:
This PR upgrades heed to v0.22 and arroy (using the same heed version) to v0.6. It also updates the dumpless upgrade path from arroy 0.5 to 0.6. I discovered that the auth env was opened a second time when generating a snapshot. I had to store it in the index scheduler to avoid this mistake and open it only once. Finally, it makes the seed used for arroy random instead of always using `42`.

## To do
- [ ] ~Make sure the dumpless upgrade with arroy works.~ Testing before `rc.0`.
- [x] Make sure snapshots and dumps work.

### DIFF
--- a/crates/meilisearch/tests/features/mod.rs
+++ b/crates/meilisearch/tests/features/mod.rs
@@ -144,14 +144,6 @@ async fn experimental_feature_metrics() {
     let (response, code) = server.get_metrics().await;
     meili_snap::snapshot!(code, @"200 OK");
     meili_snap::snapshot!(response, @"null");
-
-    // startup without flag respects persisted metrics value
-    let disable_metrics =
-        Opt { experimental_enable_metrics: false, ..default_settings(dir.path()) };
-    let server_no_flag = Server::new_with_options(disable_metrics).await.unwrap();
-    let (response, code) = server_no_flag.get_metrics().await;
-    meili_snap::snapshot!(code, @"200 OK");
-    meili_snap::snapshot!(response, @"null");
 }
 
 #[actix_rt::test]
--- a/crates/meilisearch/tests/upgrade/mod.rs
+++ b/crates/meilisearch/tests/upgrade/mod.rs
@@ -64,9 +64,6 @@ async fn version_requires_downgrade() {
 #[actix_rt::test]
 async fn upgrade_to_the_current_version() {
     let temp = tempfile::tempdir().unwrap();
-    let server = Server::new_with_options(default_settings(temp.path())).await.unwrap();
-    drop(server);
-
     let server = Server::new_with_options(Opt {
         experimental_dumpless_upgrade: true,
         ..default_settings(temp.path())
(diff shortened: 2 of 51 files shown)


Kerollmops:
This is a benchmark that uses the new heed compiled with the `NDEBUG` C define which disabled plenty of asserts and should gives us performances gains 👇

Kerollmops:
This one uses the new heed compiled *without* the `NDEBUG` C define 👇

crates/meilisearch/tests/features/mod.rs
```diff
@@ -138,14 +138,6 @@ async fn experimental_feature_metrics() {
...
     meili_snap::snapshot!(code, @"200 OK");
     meili_snap::snapshot!(response, @"null");
-
-    // startup without flag respects persisted metrics value
-    let disable_metrics =
-        Opt { experimental_enable_metrics: false, ..default_settings(dir.path()) };
-    let server_no_flag = Server::new_with_options(disable_metrics).await.unwrap();
-    let (response, code) = server_no_flag.get_metrics().await;
-    meili_snap::snapshot!(code, @"200 OK");
-    meili_snap::snapshot!(response, @"null");
```
Kerollmops, line 148: Opening multiple Servers never worked, as the Envs were kept open anyway...

crates/meilisearch/tests/upgrade/mod.rs
```diff
@@ -64,8 +64,8 @@ async fn version_requires_downgrade() {
 #[actix_rt::test]
 async fn upgrade_to_the_current_version() {
     let temp = tempfile::tempdir().unwrap();
-    let server = Server::new_with_options(default_settings(temp.path())).await.unwrap();
-    drop(server);
+    // let server = Server::new_with_options(default_settings(temp.path())).await.unwrap();
+    // drop(server);
```
Kerollmops, line 68: Same here: Opening multiple Servers never worked, as the Envs were kept open anyway...
What do you think @irevoire of removing these lines?

crates/milli/src/update/upgrade/v1_14.rs
```diff
@@ -0,0 +1,41 @@
...
+
+        let rtxn = index.read_txn()?;
+        arroy::upgrade::from_0_5_to_0_6::<Cosine>(
+            &rtxn,
+            index.vector_arroy.remap_data_type(),
+            wtxn,
+            index.vector_arroy.remap_data_type(),
+        )?;
+
+        Ok(true)
```
irevoire, line 35: We don’t need to regenerate the stats

```suggestion
        Ok(false)
```

irevoire (approved):
Perfecto, thank you 👏 

bors merge

Kerollmops:
bors merge

Kerollmops:
bors merge

PR #5418, opened 2025-03-13
Cache embeddings in search

dureuill:
# Pull Request

## Related issue
TBD

## What does this PR do?
- Adds a cache for embeddings produced in search
- The cache is disabled by default, and can be enabled following the instructions [here](https://github.com/orgs/meilisearch/discussions/818).
- Had to accommodate the `timeout` test for openai that uses a mock that simulates a timeout on subsequent responses: since the test was reusing the same query, the cache would kick-in and no request would be made to the mock, meaning no timeout any longer and so a failing test 😅 
- `Embedder::embed_search` now accepts a reference instead of an owned `String`.

## Manual testing

- I created 4 indexes on a fresh DB with the same settings (one embedder from openai)
- I sent 1/4 of movies.json to each index
- I sent a federated search request against all 4 indexes, with the same query for each index, using the embedder of each index.

Results:

- The first call took 400ms to 1s. Before this change, it took in the 3s range.
- Any repeated call with the same query took in the range of 25ms.
- Looking at the details at trace log level, I can see that the first index that needs the embedding is taking most of the 400ms in `embed_one`. The other indexes report that the query text is found in the cache and they each take a few µs.


Cargo.lock
```diff
@@ -3517,6 +3517,15 @@ version = "0.4.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "90ed8c1e510134f979dbc4f070f87d4313098b704861a105fe34231c70a3901c"
 
+[[package]]
+name = "lru"
```
Kerollmops, line 3521: Fair!

crates/index-scheduler/src/scheduler/test_embedders.rs
```diff
@@ -104,10 +104,9 @@ fn import_vectors() {
 
         let configs = index_scheduler.embedders("doggos".to_string(), configs).unwrap();
         let (hf_embedder, _, _) = configs.get(&simple_hf_name).unwrap();
-        let beagle_embed =
-            hf_embedder.embed_search(S("Intel the beagle best doggo"), None).unwrap();
-        let lab_embed = hf_embedder.embed_search(S("Max the lab best doggo"), None).unwrap();
-        let patou_embed = hf_embedder.embed_search(S("kefir the patou best doggo"), None).unwrap();
+        let beagle_embed = hf_embedder.embed_search("Intel the beagle best doggo", None).unwrap();
```
Kerollmops, line 107: Fair!

crates/milli/src/vector/mod.rs
```diff
@@ -542,6 +543,46 @@ pub enum Embedder {
...
+    data: Option<Mutex<lru::LruCache<String, Embedding>>>,
+}
+
+impl EmbeddingCache {
+    const MAX_TEXT_LEN: usize = 2000;
+
+    pub fn new(cap: u16) -> Self {
+        let data = (|| Some(Mutex::new(lru::LruCache::new(NonZeroUsize::new(cap as usize)?))))();
+        Self { data }
+    }
```
Kerollmops, line 557: Why is it possible that it's None and why all this complexity around creating the Mutexed lru?

Kerollmops: Wow! Is it just to return None if the `NonZeroUsize` returns `None`?
```suggestion
    data: Option<Mutex<lru::LruCache<String, Embedding>>>,
}

impl EmbeddingCache {
    const MAX_TEXT_LEN: usize = 2000;

    pub fn new(cap: u16) -> Self {
        Self { data: NonZeroUsize::new(cap as usize).map(|cap| Mutex::new(lru::LruCache::new(cap))) }
    }
```

Kerollmops: Or even this would be more clear than all those parentheses 😂
```suggestion
    data: Option<Mutex<lru::LruCache<String, Embedding>>>,
}

impl EmbeddingCache {
    const MAX_TEXT_LEN: usize = 2000;

    pub fn new(cap: u16) -> Self 
        Self { data: NonZeroUsize::new(cap as usize).map(lru::LruCache::new).map(Mutex::new) }
    }
```

crates/milli/src/vector/mod.rs
```diff
@@ -642,19 +683,36 @@ impl Embedder {
     #[tracing::instrument(level = "debug", skip_all, target = "search")]
     pub fn embed_search(
         &self,
-        text: String,
+        text: &str,
```
Kerollmops, line 686: Better as, indeed, there is a change the `&str` is in the cache and doesn't need to be copied.

Kerollmops (changes requested):
Looks good to me, but some parentheses issues.

Kerollmops (dismissed):
Nice addition 😍 

bors merge

Kerollmops:
bors cancel

(need rebase)

Kerollmops (approved):
bors merge

dureuill:
bors merge

PR #5421, opened 2025-03-14
Accept total batch size in human size

Kerollmops:
This PR fixes the new `experimental-limit-batched-tasks-total-size` to accept human-defined sizes in bytes.


irevoire (approved):
Perfecto, that’ll be way better!

bors merge

PR #5422, opened 2025-03-16
Add more progress levels to measure merging

Kerollmops:
I found out that Meilisearch was not correctly reporting the long indexing times in the progress and that a lot of time was spent on extracting words with all documents already extracted. The reason was that there was no step to report merging the cache and sending the entries to write to the writer thread. This PR adds these entries to the progress.


crates/milli/src/update/new/steps.rs
```diff
@@ -13,6 +13,9 @@ pub enum IndexingStep {
     ExtractingWords,
     ExtractingWordProximity,
     ExtractingEmbeddings,
+    MergingFacetCaches,
+    MergingWordCaches,
+    MergingWordProximity,
     WritingGeoPoints,
     WaitingForDatabaseWrites,
     WaitingForExtractors,
```
irevoire, line 21: Huuum, not really required in this PR but if we were using the `make_enum_progress!` macro it would handle all the derive, repr and implementation of the `IndexingStep` trait

irevoire (approved):
Hey, that seems good to me 👍 

I don’t merge myself just in case you want to rewrite the part I mentioned below but it’s not mandatory

irevoire (approved):
Perfecto presto


bors merge

Kerollmops:
bors merge

PR #5446, opened 2025-03-20
Fix _matchesPosition length calculate

### LINKED ISSUE #5429: _matchesPosition length calc with char length instead of byte
**Describe the bug**

https://github.com/meilisearch/meilisearch/blob/2500e3c06776c332b8566f9e5c8e1b34c4fff0ff/crates/milli/src/search/new/matches/mod.rs#L232-L233
**Expected behavior**

As [the document](https://www.meilisearch.com/docs/reference/api/search#show-matches-position) said that:

> start and length are measured in bytes and not the number of characters. For example, ü represents two bytes but one character.

I think either the code should be fixed or the documentation should be modified to avoid confusion.

**Meilisearch version:**

Latest

shaokeyibb:
# Pull Request

## Related issue
Fixes #5429 

## What does this PR do?
- It fixed #5429, which incorrectly using char length instead of byte length as result of `_matchesPosition`

## PR checklist

### DIFF
--- a/crates/milli/src/search/new/matches/mod.rs
+++ b/crates/milli/src/search/new/matches/mod.rs
@@ -8,6 +8,7 @@ use std::cmp::{max, min};
 
 use charabia::{Language, SeparatorKind, Token, Tokenizer};
 use either::Either;
+use itertools::Itertools;
 pub use matching_words::MatchingWords;
 use matching_words::{MatchType, PartialMatch};
 use r#match::{Match, MatchPosition};
@@ -229,12 +230,7 @@ impl<'t, 'tokenizer> Matcher<'t, 'tokenizer, '_, '_> {
                 .iter()
                 .map(|m| MatchBounds {
                     start: tokens[m.get_first_token_pos()].byte_start,
-                    length: (m.get_first_token_pos()..m.get_last_token_pos() + 1)
-                        .map(|i| tokens[i].clone())
-                        .flat_map(|token| token.char_map.clone().unwrap_or(vec![(1, 1); token.char_end - token.char_start] /* Some token doesn't have a char map, here we treat them as single byte chars. */))
-                        .map(|(original, _)| original as usize)
-                        .take(m.char_count)
-                        .sum(),
+                    length: self.calc_byte_length(&tokens, m),
                     indices: if array_indices.is_empty() {
                         None
                     } else {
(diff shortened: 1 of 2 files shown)


crates/milli/src/search/new/matches/mod.rs
```diff
@@ -229,8 +229,12 @@ impl<'t, 'tokenizer> Matcher<'t, 'tokenizer, '_, '_> {
...
                 .map(|m| MatchBounds {
                     start: tokens[m.get_first_token_pos()].byte_start,
-                    // TODO: Why is this in chars, while start is in bytes?
-                    length: m.char_count,
+                    length: (m.get_first_token_pos()..m.get_last_token_pos() + 1)
+                        .map(|i| tokens[i].clone())
+                        .flat_map(|token| token.char_map.clone().unwrap_or(vec![(1, 1); token.char_end - token.char_start] /* Some token doesn't have a char map, here we treat them as single byte chars. */))
+                        .map(|(original, _)| original as usize)
+                        .take(m.char_count)
+                        .sum(),
```
ManyTheFish, line 237: We may avoid all these `clone()` and take the byte lengths of the characters when the char map is missing instead of counting 1 for each character.

```suggestion
                    length: (m.get_first_token_pos()..=m.get_last_token_pos())
                        .scan(m.char_count, |i, remaining_chars| {
                            let mut bytes = 0;
                            match &tokens[i].char_map {
                                Some(char_map) => {
                                  // iterate over the char map, retrieving the original bytes len
                                  for (original_bytes_in_char, _) in char_map {
                                    bytes += original_bytes_in_char;
                                    *remaining_chars -= 1;
                                  
                                    if remaining_chars == 0 {
                                      break;
                                    }
                                  }
                                }
                                None => {
                                  // iterate over the chars, retrieving the bytes len
                                  for c in tokens[i].lemma().chars() {
                                    bytes += c.len_utf8();
                                    *remaining_chars -= 1;
                                  
                                    if remaining_chars == 0 {
                                      break;
                                    }
                                  }
                                }
                              }
                            
                              bytes
                          })
                        .sum(),
```

But this should be put in a dedicated function to make the code cleaner.

shaokeyibb:
Hi @ManyTheFish, here's the new implemention of byte length calculate without clone according to your code

ManyTheFish (approved):

PR #5464, opened 2025-03-26
Prefer camelCase for internal database sizes db name

Kerollmops:
Convert the internal database names in the stats into camelCase.


irevoire (approved):
👍

PR #5471, opened 2025-03-27
Support EC private key

### LINKED ISSUE #5427: Meilisearch does not accept some encodings of EC keys
**Describe the bug**
Meilisearch does not accept EC server keys with a "-----BEGIN EC PRIVATE KEY-----" header. It does accept the same key when converted to the more generic "-----BEGIN PRIVATE KEY-----" header.

Confusingly, the error message mentions RSA instead:
```
ERROR meilisearch: info=panicked at crates/meilisearch/src/option.rs:888:9: assertion failed: !rsa_keys.is_empty()
```

**To Reproduce**
```bash
# create an EC cert pair
openssl req -x509 -newkey ec -pkeyopt ec_paramgen_curve:prime256v1 -days 90 -nodes -keyout server.key -out server.crt -subj "/CN=localhost"

# convert it to the more EC specific format
openssl ec -in ./server.key -check -out server-ec.key

# this errors:
meilisearch --ssl-cert-path=./server.crt  --ssl-key-path=./server-ec.key
# ERROR meilisearch: info=panicked at crates/meilisearch/src/option.rs:888:9: assertion failed: !rsa_keys.is_empty()

# whereas this works:
meilisearch --ssl-cert-path=./server.crt  --ssl-key-path=./server.key
```

**Expected behavior**
Both formats of the EC key should work. If that's not possible, the error message should be more clear and definitely not reference RSA.

**Meilisearch version:**
1.13.3

**Additional context**
I ran into this because our cert generation tool is in Ruby and Ruby's openssl gem outputs the "-----BEGIN EC PRIVATE KEY-----" format by default. 

We successfully use this output format regularly with many other programs, but I'm not sure if any of them are in Rust, so it could be something with Rust's TLS implementation or it may be specific to Meilisearch.

HDT3213:
# Pull Request

## Related issue
Fixes #5427 

## What does this PR do?
- Add support for EC private key

## PR checklist

### DIFF
--- a/crates/meilisearch/src/option.rs
+++ b/crates/meilisearch/src/option.rs
@@ -16,7 +16,7 @@ use meilisearch_types::milli::update::IndexerConfig;
 use meilisearch_types::milli::ThreadPoolNoAbortBuilder;
 use rustls::server::{ServerSessionMemoryCache, WebPkiClientVerifier};
 use rustls::RootCertStore;
-use rustls_pemfile::{certs, rsa_private_keys};
+use rustls_pemfile::{certs, ec_private_keys, rsa_private_keys};
 use serde::{Deserialize, Serialize};
 use sysinfo::{MemoryRefreshKind, RefreshKind, System};
 use url::Url;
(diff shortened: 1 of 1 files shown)


irevoire (changes requested):
Hey @HDT3213, thanks for your contribution!

We’re almost good to go I just left a small comment

HDT3213:
Hi @irevoire, thanks  for your review.  I have modified the code as you suggested

irevoire (approved):
Perfect, thank you 🌻

PR #5480, opened 2025-03-31
Bump Rust version to 1.85.1

Kerollmops:
This PR bumps Rust from 1.81 to 1.85.1. Most of the changes are around stabilization and can help #5479.


dureuill (approved):
sad about all the anonymized lifetimes, but such is *progress*

PR #5487, opened 2025-04-02
fix ranking rules after _geo do not work

### LINKED ISSUE #5405: ranking rules after `_geo` do not work as "tiebreaker"
**Describe the bug**
In a search that involves `geo`, documents might be returned in the wrong order when:

- Multiple documents have the exact same distance to the searched point
- There is another ranking rule to "break the tie" after the geo sort.

**To Reproduce**
1. Add two or more documents with the same `_geo` fields, and some other way to differentiate between them (such as a `boost` field)
2. Setup your ranking rules to take `sort`
3. Make a search with `sort = [geoPoint(...,...):asc, boost:desc]` and the detailed ranking scores
4. Check the ranking scores: documents might be out of order

**Expected behavior**
Documents should respect the ranking rules after `geoPoint` in case of a tie

**Meilisearch version:**
at least v1.10.3 and at least up to v1.13.3

**Additional context**
As the order of documents after geo sort is arbitrary, it might sometimes be the right order. Invert the second sort rule to observe the bug in this case.

HDT3213:
# Pull Request

## Related issue
Fixes #5405

## What does this PR do?

`GeoSort::next_bucket` now returns all points with the same distance to the searched point at a time, so that ranking rules after _geo can do their work 

I built a test dataset that was larger than GeoSort’s cache size and used it to verify that my code could refill the cache after it was exhausted.

Since the  ranking rules after geo may select any doc in the last bucket, `next_bucket` must always return a complete bucket to ensure the correct result, even if this makes the total amount exceed `limit+offset`.

Fortunately, `bucket_sort` will stop iterating `next_bucket` when the total returned exceeds `limit+offset`, so we don't need to worry about "unbounded" work

## PR checklist

### DIFF
--- a/crates/milli/src/search/new/geo_sort.rs
+++ b/crates/milli/src/search/new/geo_sort.rs
@@ -289,7 +289,7 @@ impl<'ctx, Q: RankingRuleQueryTrait> RankingRule<'ctx, Q> for GeoSort<Q> {
                 if geo_candidates.contains(id) {
                     let distance = distance_between_two_points(&self.point, &point);
                     if let Some((point0, bucket_distance)) = current_distance.as_ref() {
-                        if (bucket_distance - &distance).abs() > self.distance_error_margin {
+                        if (bucket_distance - distance).abs() > self.distance_error_margin {
                             // different distance, point belongs to next bucket
                             put_back(&mut self.cached_sorted_docids, (id, point));
                             return Ok(Some(RankingRuleOutput {
--- a/crates/milli/src/search/new/geo_sort.rs
+++ b/crates/milli/src/search/new/geo_sort.rs
@@ -1,6 +1,4 @@
 use std::collections::VecDeque;
-use std::iter::FromIterator;
-
 use heed::types::{Bytes, Unit};
 use heed::{RoPrefix, RoTxn};
 use roaring::RoaringBitmap;
@@ -245,7 +243,7 @@ impl<'ctx, Q: RankingRuleQueryTrait> RankingRule<'ctx, Q> for GeoSort<Q> {
     ) -> Result<Option<RankingRuleOutput<Q>>> {
         let query = self.query.as_ref().unwrap().clone();
 
-        let geo_candidates = &self.geo_candidates & universe;
+        let mut geo_candidates = &self.geo_candidates & universe;
 
         if geo_candidates.is_empty() {
             return Ok(Some(RankingRuleOutput {
@@ -267,24 +265,79 @@ impl<'ctx, Q: RankingRuleQueryTrait> RankingRule<'ctx, Q> for GeoSort<Q> {
                 cache.pop_back()
             }
         };
-        while let Some((id, point)) = next(&mut self.cached_sorted_docids) {
-            if geo_candidates.contains(id) {
-                return Ok(Some(RankingRuleOutput {
-                    query,
-                    candidates: RoaringBitmap::from_iter([id]),
-                    score: ScoreDetails::GeoSort(score_details::GeoSort {
-                        target_point: self.point,
-                        ascending: self.ascending,
-                        value: Some(point),
-                    }),
-                }));
+        let put_back = |cache: &mut VecDeque<_>, x: _| {
+            if ascending {
+                cache.push_front(x)
+            } else {
+                cache.push_back(x)
+            }
+        };
+
+        let mut current_bucket = RoaringBitmap::new();
+        // current_distance stores the first point and distance in current bucket
+        // The farthest distance between two points on earth is about 2e7 meters, u32 is big enough to hold any distance.
+        let mut current_distance: Option<([f64; 2], u32)> = None;
+        loop {
+            // The loop will only exit when we have found all points with equal distance or have exhausted the candidates.
+            if let Some((id, point)) = next(&mut self.cached_sorted_docids) {
+                if geo_candidates.contains(id) {
+                    let distance = distance_between_two_points(&self.point, &point).round() as u32;
+                    if let Some((point0, bucket_distance)) = current_distance.as_ref() {
+                        if bucket_distance != &distance {
+                            // different distance, point belongs to next bucket
+                            put_back(&mut self.cached_sorted_docids, (id, point));
+                            return Ok(Some(RankingRuleOutput {
+                                query,
+                                candidates: current_bucket,
+                                score: ScoreDetails::GeoSort(score_details::GeoSort {
+                                    target_point: self.point,
+                                    ascending: self.ascending,
+                                    value: Some(point0.to_owned()),
+                                }),
+                            }));
+                        } else {
+                            // same distance, point belongs to current bucket
+                            current_bucket.push(id);
+                            // remove from cadidates to prevent it from being added to the cache again
+                            geo_candidates.remove(id); 
+                        }
+                    } else {
+                        // first doc in current bucket
+                        current_distance = Some((point, distance));
+                        current_bucket.push(id);
+                        geo_candidates.remove(id); 
+                    }
+                }
+            } else {
+                // cache exhausted, we need to refill it
+                self.fill_buffer(ctx, &geo_candidates)?;
+
+                if self.cached_sorted_docids.is_empty() {
+                    // candidates exhausted, exit
+                    if let Some((point0, _)) = current_distance.as_ref() {
+                        return Ok(Some(RankingRuleOutput {
+                            query,
+                            candidates: current_bucket,
+                            score: ScoreDetails::GeoSort(score_details::GeoSort {
+                                target_point: self.point,
+                                ascending: self.ascending,
+                                value: Some(point0.to_owned()),
+                            }),
+                        }));
+                    } else {
+                        return Ok(Some(RankingRuleOutput {
+                            query,
+                            candidates: universe.clone(),
+                            score: ScoreDetails::GeoSort(score_details::GeoSort {
+                                target_point: self.point,
+                                ascending: self.ascending,
+                                value: None,
+                            }),
+                        }));
+                    }
+                }
             }
         }
-
-        // if we got out of this loop it means we've exhausted our cache.
-        // we need to refill it and run the function again.
-        self.fill_buffer(ctx, &geo_candidates)?;
-        self.next_bucket(ctx, logger, universe)
     }
 
     #[tracing::instrument(level = "trace", skip_all, target = "search::geo_sort")]
--- a/crates/milli/src/search/new/geo_sort.rs
+++ b/crates/milli/src/search/new/geo_sort.rs
@@ -40,7 +40,6 @@ fn facet_number_values<'a>(
 }
 
 #[derive(Debug, Clone, Copy)]
-
 pub struct Parameter {
     // Define the strategy used by the geo sort
     pub strategy: Strategy,
@@ -107,7 +106,7 @@ pub struct GeoSort<Q: RankingRuleQueryTrait> {
 
 impl<Q: RankingRuleQueryTrait> GeoSort<Q> {
     pub fn new(
-        parameter: &Parameter,
+        parameter: Parameter,
         geo_faceted_docids: RoaringBitmap,
         point: [f64; 2],
         ascending: bool,
@@ -115,15 +114,15 @@ impl<Q: RankingRuleQueryTrait> GeoSort<Q> {
         let Parameter { strategy, max_bucket_size, distance_error_margin } = parameter;
         Ok(Self {
             query: None,
-            strategy: *strategy,
+            strategy,
             ascending,
             point,
             geo_candidates: geo_faceted_docids,
             field_ids: None,
             rtree: None,
             cached_sorted_docids: VecDeque::new(),
-            max_bucket_size: *max_bucket_size,
-            distance_error_margin: *distance_error_margin,
+            max_bucket_size,
+            distance_error_margin,
         })
     }
 
--- a/crates/milli/src/search/new/geo_sort.rs
+++ b/crates/milli/src/search/new/geo_sort.rs
@@ -1,8 +1,8 @@
-use std::collections::VecDeque;
 use heed::types::{Bytes, Unit};
 use heed::{RoPrefix, RoTxn};
 use roaring::RoaringBitmap;
 use rstar::RTree;
+use std::collections::VecDeque;
 
 use super::facet_string_values;
 use super::ranking_rules::{RankingRule, RankingRuleOutput, RankingRuleQueryTrait};
@@ -238,7 +238,7 @@ impl<'ctx, Q: RankingRuleQueryTrait> RankingRule<'ctx, Q> for GeoSort<Q> {
     fn next_bucket(
         &mut self,
         ctx: &mut SearchContext<'ctx>,
-        logger: &mut dyn SearchLogger<Q>,
+        _logger: &mut dyn SearchLogger<Q>,
         universe: &RoaringBitmap,
     ) -> Result<Option<RankingRuleOutput<Q>>> {
         let query = self.query.as_ref().unwrap().clone();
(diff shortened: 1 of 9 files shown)


crates/milli/src/search/new/geo_sort.rs
```diff
@@ -267,24 +265,79 @@ impl<'ctx, Q: RankingRuleQueryTrait> RankingRule<'ctx, Q> for GeoSort<Q> {
...
+                        if bucket_distance != &distance {
+                            // different distance, point belongs to next bucket
+                            put_back(&mut self.cached_sorted_docids, (id, point));
+                            return Ok(Some(RankingRuleOutput {
+                                query,
+                                candidates: current_bucket,
+                                score: ScoreDetails::GeoSort(score_details::GeoSort {
+                                    target_point: self.point,
+                                    ascending: self.ascending,
+                                    value: Some(point0.to_owned()),
```
irevoire, line 295: Outputting the coordinates of one random points in the bucket is pretty bad and could be misleading.
We should stop returning this value entirely and only output the distance of this bucket from the required point.

But that would be a breaking change and cannot be done in this PR. I’ll open an issue for v2.

crates/milli/src/search/new/geo_sort.rs
```diff
@@ -267,24 +265,79 @@ impl<'ctx, Q: RankingRuleQueryTrait> RankingRule<'ctx, Q> for GeoSort<Q> {
...
+                    } else {
+                        return Ok(Some(RankingRuleOutput {
+                            query,
+                            candidates: universe.clone(),
+                            score: ScoreDetails::GeoSort(score_details::GeoSort {
+                                target_point: self.point,
+                                ascending: self.ascending,
+                                value: None,
+                            }),
+                        }));
```
irevoire, line 336: Hey, I don’t understand why we’re returning nothing here?
Shouldn’t we continue with the loop and check if we have other candidates matching our bucket?

It would be nice if you could add a test where two candidates must be grouped together but are not part of the same bucket, as this is now a bit tricky to handle.

irevoire (changes requested):
Hey, thanks for your contribution!

Could you add a test with two points pointing to the same spot in `crates/milli/src/search/new/tests/geo_sort.rs`.
By looking at the other test you’ll also see that you can select the scoring strategy to use the rtree or not so you should run your test twice with and without the rtree!

crates/milli/src/search/new/geo_sort.rs
HDT3213: If self.cached_sorted_docids is still empty after refill, it means we have exhausted the entire candidate set. At this time, if current_bucket is not empty, I should return current_bucket. But if current_bucket is empty, I can only return an empty result set.

Also, when you say "grouped together", do you mean they appear in the cache at the same time?

crates/milli/src/search/new/geo_sort.rs
```diff
@@ -267,24 +265,79 @@ impl<'ctx, Q: RankingRuleQueryTrait> RankingRule<'ctx, Q> for GeoSort<Q> {
...
+        };
 
-        // if we got out of this loop it means we've exhausted our cache.
-        // we need to refill it and run the function again.
-        self.fill_buffer(ctx, &geo_candidates)?;
-        self.next_bucket(ctx, logger, universe)
+        let mut current_bucket = RoaringBitmap::new();
+        // current_distance stores the first point and distance in current bucket
+        // The farthest distance between two points on earth is about 2e7 meters, u32 is big enough to hold any distance.
+        let mut current_distance: Option<([f64; 2], u32)> = None;
```
HDT3213, line 279: In fact, distance can be represented by float. I originally wanted to use distance as the key of a map and then use u32. Now using u32 is a mistake.

crates/milli/src/search/new/geo_sort.rs
HDT3213: OK, I'll leave this question alone for now.

By the way, as this [comment](https://github.com/meilisearch/meilisearch/issues/5405#issuecomment-2774762727) says, small differences in distance should be treated as equal. Perhaps we should allow users to customize this difference

crates/milli/src/search/new/geo_sort.rs
irevoire: Yes why not.
But when comparing floats do not forget that it’s recommended to run `left - right < f64::EPSILON` instead of a normal `==`.

crates/milli/src/search/new/geo_sort.rs
irevoire: We should think about it, but it’s more of a product question.

I think your PR is great as meilisearch returns the `_geoDistance` in meter it makes sense that when two documents share the same `_geoDistance` they’re bucketed together 👍

crates/milli/src/search/new/geo_sort.rs
```diff
@@ -267,24 +265,79 @@ impl<'ctx, Q: RankingRuleQueryTrait> RankingRule<'ctx, Q> for GeoSort<Q> {
...
+                                    ascending: self.ascending,
+                                    value: Some(point0.to_owned()),
+                                }),
+                            }));
+                        } else {
+                            // same distance, point belongs to current bucket
+                            current_bucket.push(id);
+                            // remove from cadidates to prevent it from being added to the cache again
+                            geo_candidates.remove(id);
+                        }
```
irevoire, line 303: While chatting with the team, we thought about this else and the way it could loop through the whole dataset if all documents share the same distance.
As a security hatch, we think it would be great if you could exit the loop anyway once we reach a certain number of documents in the same bucket.

1000 seems like a good value, but it would be awesome if you could make the thresholds configurable so we can easily test it out

HDT3213: I am worried that this will cause another problem. The ranking rules after geo need to get all the documents to sort correctly.If we only give 1000 documents, it may lead to wrong results.

For example, let's assume that all documents have the same distance, and the user sorts ["geoPoint...", "score:desc"]. Due to the security hatch, the document with the highest score may not be able to enter the bucket. in this case, the first item in the sorted result is not the document with the highest score.

irevoire: Yes, the idea is that;
- Generally speaking there is no reason to have more than a few documents exactly at the same distance, that’s really an edge case
- If a malevolent actor notices want to DOS an instance, they could push a ton of document exactly in the same spot and then every search requests would be irresponsive 

So, we think it’s better to get results slightly wrong in a case that should never happen that risk a DOS 😅

crates/milli/src/search/new/geo_sort.rs
irevoire: Aaaah yes sorry my bad, I thought the else was applied to the other branch my bad 🤦 
So you’re actually doing what I was expecting 👌

HDT3213:
I have added test cases as requested, and added max_bucket_size and distance_error_margin to GeoSort to limit the number of points in a bucket and the maximum distance (they default to 1000 and 1 meter). 

And I found that I had mistakenly used the RoaringBitmap::push instead of the RoaringBitmap::insert. This error was not caught by the previous tests, which could be dangerous.

HDT3213:
To be honest, I don't know what my code has to do with the proxy😮‍💨

```
failures:

---- search::multi::proxy::error_remote_does_not_answer stdout ----
{
  "remotes": {
    "ms0": {
      "url": "http://127.0.0.1:46263/"
    },
    "ms1": {
      "url": "http://127.0.0.1:44643/"
    },
    "ms2": {
      "url": "https://thiswebsitedoesnotexist.example/"
    }
  }
}
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: error_remote_does_not_answer
Source: crates/meilisearch/tests/search/multi/proxy.rs:1327
────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────
   35    35 │   "offset": 0,
   36    36 │   "estimatedTotalHits": 3,
   37    37 │   "remoteErrors": {
   38    38 │     "ms2": {
   39       │-      "message": "error sending request",
   40       │-      "code": "remote_could_not_send_request",
         39 │+      "message": "remote host did not answer before the deadline",
         40 │+      "code": "remote_timeout",
   41    41 │       "type": "system",
   42       │-      "link": "https://docs.meilisearch.com/errors#remote_could_not_send_request"
         42 │+      "link": "https://docs.meilisearch.com/errors#remote_timeout"
   43    43 │     }
   44    44 │   }
   45    45 │ }
────────────┴───────────────────────────────────────────────────────────────────
Stopped on the first failure. Run `cargo insta test` to run all snapshots.

thread 'search::multi::proxy::error_remote_does_not_answer' panicked at /github/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/insta-1.39.0/src/runtime.rs:562:9:
snapshot assertion for 'error_remote_does_not_answer' failed in line 1327
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: insta::runtime::assert_snapshot
   3: integration::search::multi::proxy::error_remote_does_not_answer::{{closure}}
   4: tokio::task::local::LocalSet::run_until::{{closure}}
   5: core::ops::function::FnOnce::call_once
```

HDT3213:
Hi @irevoire , I have completed the changes but I am getting some unrelated test failures. Actually I am also getting some random test failures in the main branch. So how should we handle these test failures?

crates/milli/src/search/new/tests/geo_sort.rs
```diff
@@ -95,6 +95,47 @@ fn test_geo_sort() {
...
+
+    s.sort_criteria(vec![
+        AscDesc::Desc(Member::Geo([0., 0.])),
+        AscDesc::Desc(Member::Field("score".to_string())),
+    ]);
+    let (ids, scores) = execute_iterative_and_rtree_returns_the_same(&rtxn, &index, &mut s);
+    insta::assert_snapshot!(format!("{ids:?}"), @"[12, 13, 14, 15, 6, 7, 8, 9, 10, 11, 1, 4, 3, 2, 5]");
+    insta::assert_snapshot!(format!("{scores:#?}"));
+}
+
```
irevoire, line 138: These tests are awesome.
I was just expecting two more test with a very small bucket of like 2 elements.
Let me know if it makes sense and if you can add them?

Then I think we’ll be good to merge 👌

HDT3213: In fact, execute_iterative_and_rtree_returns_the_same covers both the cases where the bucket size is greater than the cache size (need to refill the cache) and less than the cache size (no refill is required). I wonder which case a very small bucket is for?

irevoire: Huum from what I see we have a test where the cache is small and must be refilled before we reach our thresholds. That’s good.
But I don’t see any test doing the opposite, basically, we have a very low threshold and must stop early even though the bucket is not empty

HDT3213: Hi, @irevoire I hope you had a great holiday.

You wanted to test the case where max_bucket_size is reached, right? I think I misunderstood you earlier, I will add the corresponding tests later.

irevoire: Yes exactly!
Thank you 🙏 
Since there is two caches at the same time it’s not easy to explain ahah 😅

crates/milli/src/search/new/geo_sort.rs
```diff
@@ -73,6 +71,26 @@ impl Strategy {
...
+pub fn set_default_max_bucket_size(n: u64) {
+    let mut size = DEFAULT_MAX_BUCKET_SIZE.lock().unwrap();
+    *size = n;
+}
+
+#[cfg(test)]
+fn default_max_bucket_size() -> u64 {
+    let max_size = *(DEFAULT_MAX_BUCKET_SIZE.lock().unwrap());
+    max_size
+}
```
irevoire, line 92: Hey, sorry but since multiple tests are run in parallel sharing the same variable this will be super flaky.
One test may change the default max bucket size for the other 😭 

Maybe the easiest thing is to just shove the value into the `Strategy` like we did for the cache size.
Either you just add it to every variant, or you could create a structure `Parameter` that contains both the max_bucket_size and the cache_size + a new unit enum `Strategy` that doesn’t hold any data.

When calling this value from Meilisearch we always set it to `default` iirc, so you can do whatever you prefer it’s only local to this file.

HDT3213: 😂 I think I would prefer to add a Parameter to hold all the relevant parameters

crates/milli/src/search/new/geo_sort.rs
```diff
@@ -41,6 +39,22 @@ fn facet_number_values<'a>(
     Ok(iter)
 }
 
+#[derive(Debug, Clone, Copy)]
+
+pub struct Parameter {
```
irevoire, line 44:
```suggestion
#[derive(Debug, Clone, Copy)]
pub struct Parameter {
```

crates/milli/src/search/new/geo_sort.rs
```diff
@@ -84,24 +98,32 @@ pub struct GeoSort<Q: RankingRuleQueryTrait> {
...
+    // Limit the number of docs in a single bucket to avoid unexpectedly large overhead
+    max_bucket_size: u64,
+    // Considering the errors of GPS and geographical calculations, distances less than distance_error_margin will be treated as equal
+    distance_error_margin: f64,
 }
 
 impl<Q: RankingRuleQueryTrait> GeoSort<Q> {
     pub fn new(
-        strategy: Strategy,
+        parameter: &Parameter,
```
irevoire, line 110:
```suggestion
        parameter: Parameter,
```

When a structure derive `Copy` it’s generally best to take it by copy instead of reference

crates/milli/src/search/new/geo_sort.rs
```diff
@@ -84,24 +98,32 @@ pub struct GeoSort<Q: RankingRuleQueryTrait> {
...
         geo_faceted_docids: RoaringBitmap,
         point: [f64; 2],
         ascending: bool,
     ) -> Result<Self> {
+        let Parameter { strategy, max_bucket_size, distance_error_margin } = parameter;
         Ok(Self {
             query: None,
-            strategy,
+            strategy: *strategy,
```
irevoire, line 118:
```suggestion
            strategy,
```

crates/milli/src/search/new/geo_sort.rs
```diff
@@ -84,24 +98,32 @@ pub struct GeoSort<Q: RankingRuleQueryTrait> {
...
             ascending,
             point,
             geo_candidates: geo_faceted_docids,
             field_ids: None,
             rtree: None,
             cached_sorted_docids: VecDeque::new(),
+            max_bucket_size: *max_bucket_size,
+            distance_error_margin: *distance_error_margin,
```
irevoire, line 126:
```suggestion
            max_bucket_size,
            distance_error_margin,
```

crates/milli/src/search/new/mod.rs
```diff
@@ -299,7 +300,7 @@ fn get_ranking_rules_for_placeholder_search<'ctx>(
                     &mut ranking_rules,
                     &mut sorted_fields,
                     &mut geo_sorted,
-                    geo_strategy,
+                    &geo_param,
```
irevoire, line 303:
```suggestion
                    geo_param,
```

crates/milli/src/search/new/mod.rs
```diff
@@ -375,7 +376,7 @@ fn get_ranking_rules_for_vector<'ctx>(
...
```
irevoire, line 379:
```suggestion
                    geo_param,
```

crates/milli/src/search/new/mod.rs
```diff
@@ -477,7 +478,7 @@ fn get_ranking_rules_for_query_graph_search<'ctx>(
...
```
irevoire, line 481:
```suggestion
                    geo_param,
```

irevoire (changes requested):
Awesome, one little comment, a type implementing `Copy` is generally not shared by ref but by copy directly

irevoire (approved):

PR #5488, opened 2025-04-02
add "batcher stopped because" field to batch objects

dureuill:
Fixes https://github.com/meilisearch/meilisearch/issues/5511

[Public Usage](https://www.notion.so/meilisearch/v1-15-batch-updates-1d64b06b651f80edbe09dcdfed63161d?pvs=4#1d64b06b651f80488a5bdc2992ae63d6)

---

- [x] Detail the change to the DB format and why they are forward compatible
  - Adds a new `stopReason` field in the batch objects in the DB
  - When deserializing a batch object, if that field is missing, as is the case for batch objects created before this PR, then the field will be set to `unspecified`
    - I checked that the batch reason was `unspecified`


irevoire (approved):
Thank you 👓

PR #5498, opened 2025-04-08
Stop compacting the snapshot

dureuill:
# Pull Request

## Related issue
Fixes TBD

## What does this PR do?
- No longer enable the compaction option when performing a snapshot


irevoire (approved):
Good to me, I thought we were also going to stop compressing the snapshots but if it was not intended you can merge 👍

dureuill:
compaction was the most urgent matter. compression I'll do if I manage to smuggle the parameters

PR #5512, opened 2025-04-16
Task queue metrics

### LINKED ISSUE #5333: Expose the size of the status of the task queue in the metrics
We sometimes run into issues in production with the task queue that we can’t really explain, and we end up asking the user to change their workflow or remove all their tasks entirely which is not good.

All of this should already be handled and is tested, but lately the test has been behaving very strangely and [need an investigation](https://github.com/meilisearch/meilisearch/issues/5301).
However, in any case, we think it would help us understand these issues and help our clients if we had a metric showing us how full the task queue is.

We should add all of the following metrics:
- The maximum capacity of the task queue
- The maximum size allowed until the task queue stops you from registering tasks
- The current size used in the task queue

DanasFi:
# Pull Request

## Related issue
Fixes #5333

## What does this PR do?
-  Add the maximum possible size of the task queue, the size allowed until the task queue stop registering tasks, and the current size used in the task queue, to the metrics endpoint.

## PR checklist


crates/index-scheduler/src/lib.rs
```diff
@@ -425,6 +427,17 @@ impl IndexScheduler {
         Ok(self.env.non_free_pages_size()?)
     }
 
+    /// Return the maximum possible database size
+    pub fn max_size(&self) -> Result<u64> {
+        Ok(self.env.info().map_size as u64)
+    }
+
+    /// Return the max size of task allowed until the task queue stop receiving.
+    pub fn remaining_size_until_task_queu_stop(&self) -> Result<u64> {
```
irevoire, line 436:
```suggestion
    pub fn remaining_size_until_task_queue_stop(&self) -> Result<u64> {
```

crates/index-scheduler/src/lib.rs
```diff
@@ -425,6 +427,17 @@ impl IndexScheduler {
...
+        Ok((self.env.info().map_size as u64 * TASK_SCHEDULER_SIZE_THRESHOLD_PERCENT_INT / 100)
+            .saturating_sub(self.used_size()?))
```
irevoire, line 438: The used_size should is in bytes and should be made as a percentage I believe before subtracting it?

irevoire (changes requested):
Hey @DanasFi, thanks for your contribution, I’m not sure the formulae is right though 🤔 

Also, do you think you could update the Grafana dashboard to include these three new values?

crates/index-scheduler/src/lib.rs
DanasFi: Unless i'm not seeing something,  it's 40% of the maximum size (bytes) - the current used size (bytes), which should be the bytes of tasks remaining until the queue stop.
Does the metric need too be the percentage until the task queue stop?

irevoire: ah no my bad you’re right!
Thanks

crates/meilisearch/db.snapshot
irevoire, line 1: Hey, sorry I don’t think this file was supposed to go in the rep 😅 
I’ll edit the git history to get rids of it and merge afterward!

PR #5519, opened 2025-04-16
Fix ruleset workflow

Kerollmops:
Make the action that adds a rule to the ruleset work better.

### DIFF
--- a/.github/workflows/milestone-workflow.yml
+++ b/.github/workflows/milestone-workflow.yml
@@ -150,6 +150,7 @@ jobs:
 
   update-ruleset:
     runs-on: ubuntu-latest
+    if: github.event.action == 'created'
     steps:
       - uses: actions/checkout@v3
       - name: Install jq
--- a/.github/workflows/milestone-workflow.yml
+++ b/.github/workflows/milestone-workflow.yml
@@ -163,15 +163,18 @@ jobs:
           RULESET_ID: 4253297
           BRANCH_NAME: ${{ github.event.inputs.branch_name }}
         run: |
+          echo "RULESET_ID: ${{ env.RULESET_ID }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/milestone-workflow.yml
```diff
@@ -162,15 +163,18 @@ jobs:
...
+          echo "RULESET_ID: ${{ env.RULESET_ID }}"
+          echo "BRANCH_NAME: ${{ env.BRANCH_NAME }}"
+
           # Get current ruleset conditions
-          CONDITIONS=$(gh api repos/meilisearch/meilisearch/rulesets/$RULESET_ID --jq '{ conditions: .conditions }')
+          CONDITIONS=$(gh api repos/meilisearch/meilisearch/rulesets/${{ env.RULESET_ID }} --jq '{ conditions: .conditions }')
 
           # Update the conditions by appending the milestone version
-          UPDATED_CONDITIONS=$(echo $CONDITIONS | jq '.conditions.ref_name.include += ["refs/heads/release-'$MILESTONE_VERSION'"]')
+          UPDATED_CONDITIONS=$(echo $CONDITIONS | jq '.conditions.ref_name.include += ["refs/heads/release-'${{ env.MILESTONE_VERSION }}'"]')
```
curquiza, line 173: Where do we defined `MILESTONE_VERSION`? My internet is really bad, I cannot make more research, so I approve in case I missed a stupid thing

curquiza (approved):
I don't want to block so I validate in case my question is not relevant

.github/workflows/milestone-workflow.yml
Kerollmops: At the top of the file 😇

PR #5523, opened 2025-04-22
Allow rollbacking updates

dureuill:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/5529

## What does this PR do?

This PR adds the ability to rollback the migration to a newer version.
To use this new ability, one must cancel a **processing** or **failed** (or enqueued) `DatabaseUpgrade` task.

Doing so will also cancel **all** enqueued, processing or failed `DatabaseUpgrade` tasks.

The rollback takes place during the processing of the `taskCancelation` task itself.

The rollback executes the following algorithm:

1. For each index, attempt to rollback the index
    1. Make sure the index is closed so that there aren't any remaining reference to the index
    2. Reopen the index normally, check if its version is the target version (index was never upgraded in the first place). If it is, move to the next index.
    3. If the version is different from the target version, attempt to open the index in a special mode where its data is the data from before the last write to the index.
    4. If the version of the index in the special mode is the target version, then write something to the index, and commit, discarding the last write to the index.
    5. Otherwise, the rollback cannot be performed. Restart Meilisearch to attempt the upgrade again.
2. After all the indexes have been rollbacked, the index scheduler itself is downgraded to the target version: **note that currently only the version of the index scheduler is rewritten**, in case of non-backward compatible changes in the task queue one should incorporate **a proper downgrade procedure**
3. Lastly, the `VERSION` file is rewritten to the target version.

As the index rollback uses the `PREV_SNAPSHOT` feature from LMDB, it has good properties:
- rollback should not require free disk space
- rollback should be extremely fast even if the upgrade was long
- rollback has not a lot of "points of failure"

### Implementation

- Add an `Index::rollback` function that attempts to rollback an index to a previous version using the described rollback algorithm
- Add an `IndexMapper::rollback_index` function that puts the index in safe conditions to be rollbacked, regarding the cache of indexes.
- Modify the task batching algorithm (`tick` function):
    1. Check `taskCancelation` tasks before `databaseUpgrade` tasks. When starting a Meilisearch that has an upgrade task and a task cancelation, this allows executing the canceling task before the upgrade task.
    2. Check that the version of the index scheduler is the expected version, refuse to batch any task that isn't a task cancelation or a database upgrade if it isn't the case. This is because a successful downgrade via a `taskCancelation` should disable the execution of all other tasks (*should deletion tasks be allowed as well?*)
- Add cancelling support to `upgrade`, so that calling the cancel route can abort a `databaseUpgrade` task  that is currently processing. For now the support is rather superficial (one check between each index), it could be improved to add cancel points while an index is upgrading.
-  Modify the processing of `taskCancelation` tasks to call the rollback routine when at least one database upgrade task is canceled.
- Move test module out of the giant `index.rs`


crates/index-scheduler/src/scheduler/create_batch.rs
```diff
@@ -459,17 +473,21 @@ impl IndexScheduler {
...
+        let version = self.version.get_version(rtxn)?;
+        let package_version = (
+            meilisearch_types::versioning::VERSION_MAJOR,
+            meilisearch_types::versioning::VERSION_MINOR,
+            meilisearch_types::versioning::VERSION_PATCH,
+        );
+        if version != Some(package_version) {
+            return Err(Error::UnrecoverableError(Box::new(
+                Error::IndexSchedulerVersionMismatch {
+                    index_scheduler_version: version.unwrap_or((1, 12, 0)),
```
Kerollmops, line 487: Is it expected to see a raw 1.12.0 version here?

Kerollmops: Ho! It's the default version if we don't have one 🤔

crates/index-scheduler/src/scheduler/process_batch.rs
```diff
@@ -144,11 +145,22 @@ impl IndexScheduler {
                     self.index_mapper.index(&rtxn, &index_uid)?
                 };
 
+                let mut index_wtxn = index.write_txn()?;
+
+                let index_version = index.get_version(&index_wtxn)?.unwrap_or((1, 12, 0));
+                let package_version = (VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH);
+                if index_version != (VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH) {
```
Kerollmops, line 152:
```suggestion
                let package_version = (VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH);
                if index_version != package_version {
```

crates/index-scheduler/src/scheduler/process_upgrade/mod.rs
```diff
@@ -1,5 +1,5 @@
-use meilisearch_types::milli;
 use meilisearch_types::milli::progress::{Progress, VariableNameStep};
+use meilisearch_types::milli::{self};
```
Kerollmops, line 2:
```suggestion
use meilisearch_types::milli;
```

Kerollmops (changes requested):
It looks impressively easy to implement (apart from some little quirks here and there). I still have one question.

> To use this new ability, one must cancel a **processing** or **failed** (or enqueued) `DatabaseUpgrade` task.

I don't get how someone could cancel a failed task. Is it because Meilisearch enqueues one upgrade task by index, and if one fails, it is a dedicated task in the task queue that can be rolled back? I am not sure if the task queue has been modified to accept such user requests to cancel failed (already processed) tasks.

dureuill:
Thank you for the review @Kerollmops 

> I don't get how someone could cancel a failed task. 

This is because of a small targeted modification to the code that handles cancelation. This only applies to `databaseUpgrade` tasks.

> Is it because Meilisearch enqueues one upgrade task by index

No, Meilisearch does not do this. Meilisearch enqueues a single upgrade task to upgrade all indexes.

> and if one fails, it is a dedicated task in the task queue that can be rolled back?

No, if the upgrade of a single index fails, the entire upgrade task fails.

 > I am not sure if the task queue has been modified to accept such user requests to cancel failed (already processed) tasks.

Alright, let's see how this happens, step by step.

1. When the user requests a cancelation in the `/tasks/cancel` route, a filter is applied to select applicable tasks. https://github.com/meilisearch/meilisearch/blob/rollback-updates/crates/meilisearch/src/routes/tasks.rs#L384
2. That filter is the same for task cancelation and task deletion, so it doesn't prevent matching failed tasks (except if the user specified a `statuses` filter that doesn't include `failed`)
3. The matching task uids are stored in the task and ultimately passed to `cancel_matched_tasks` https://github.com/meilisearch/meilisearch/blob/rollback-updates/crates/index-scheduler/src/scheduler/process_batch.rs#L661
4. In this function, the matching tasks used to be filtered by the enqueued tasks to remove any finished tasks (processing tasks are canceled using a different mechanism), but this PR changes the logic to check if any `DatabaseUpgrade` task matches the filter https://github.com/meilisearch/meilisearch/blob/rollback-updates/crates/index-scheduler/src/scheduler/process_batch.rs#L677
5. When it is the case, the matched tasks are specifically modified (https://github.com/meilisearch/meilisearch/blob/rollback-updates/crates/index-scheduler/src/scheduler/process_batch.rs#L680):
    - Add all the upgrade tasks that are either failed or enqueued (processing were already accounted for)

Kerollmops (approved):
Amazing performances 👍

PR #5525, opened 2025-04-22
Display and sum the time spent in arroy

Kerollmops:
This Pull Request introduces a new line in the logs to display the time spent searching the vector store. This will allow us to determine whether some performance issues are related to querying the vector store or awaiting network requests.


dureuill (approved):
:+1:

PR #5527, opened 2025-04-24
Use all CPUs during an import dump 

### LINKED ISSUE #5514: Improve dump import by using all CPUs
When importing a dump, the Meilisearch HTTP server is not ready and listening to HTTP requests. This is because Meilisearch is importing the data, indexes do not exist yet, and the dump import is not pushed in the task queue (it was discussed but not done).

By default, Meilisearch uses 2/3 of all the available CPUs to index content to ensure search requests can be served without being blocked by indexing operations. However, as explained above, we cannot serve search requests when importing a dump. That's an opportunity to use all the CPUs in this situation.

We can easily change that by using all the CPUs [in the following part of the codebase](https://github.com/meilisearch/meilisearch/blob/8cecc6989a01db45b153e9403bdc53f57acb234c/crates/meilisearch/src/lib.rs#L555-L565). To increase the number of CPUs to the full indexing threads provided from the command line; you'll have to tweak the `indexer_config` to set more CPUs.

nnethercott:
# Pull Request

## Related issue
Fixes #5514 

## What does this PR do?
- Derives Clone for `IndexerConfig` so a copy of the settings with higher memory can be used when indexing an import dump
- Exposes `total_memory_bytes` publicly through `MaxMemory` as an associated function
- Ensures import_dump tests still passing 

## PR checklist


nnethercott:
nit: are we sure we want `system.total_memory()` and not `system.available_memory()` ?

Kerollmops (changes requested):
Hey @nnethercott 👋

The PR title explains that the dump import must use all the CPUs. However, the PR only modifies the `IndexerConfig` to make it `Clone`able and changes the amount of memory used during dump import.

Did you use an AI/LLM to work on this PR?
Would you please remove all commits and start from scratch by creating another, temporary thread pool that uses all threads, before importing a dump?

Have a nice weekend ✈️

nnethercott:
> Hey @ook 👋
> 
> The PR title explains that the dump import must use all the CPUs. However, the PR only modifies the `IndexerConfig` to make it `Clone`able and changes the amount of memory used during dump import.
> 
> Did you use an AI/LLM to work on this PR? Would you please remove all commits and start from scratch by creating another, temporary thread pool that uses all threads, before importing a dump?
> 
> Have a nice weekend ✈️

Hey thanks for the feedback ! No llms were used haha, I guess I just interpreted the issue differently from how it was intended. 

In the original issue #5514 you mention 2/3 of the CPUs are used by default, but looking at the [code](https://github.com/meilisearch/meilisearch/blob/3f683c42382169a36b7c8c854fa1d9a9b8b18b78/crates/meilisearch/src/option.rs#L844) you're using 1/2 of the CPUs  and [2/3 of the available RAM](https://github.com/meilisearch/meilisearch/blob/3f683c42382169a36b7c8c854fa1d9a9b8b18b78/crates/meilisearch/src/option.rs#L790). I think my brain saw the "2/3" and fixated on that. This feedback makes it clearer though :)

As for the `Clone`ing business in my PR, your indexer_config is actually a deref of an `Arc<IndexerConfig>` by the time its used in `import_dump` with a strong count of 2, so you can't really modify it to allocate the CPUs without either making it Clone or by adding interior mutability. At the time I chose the route with the least refactoring (make a _copy_ of the indexer_config with more resources to configure your threadpools for indexing a dump)

Anyways I'll take a look today at all this again today !

ook:
(still don’t understand why I got ping, but I’m happy to see the project progressing :) )

Kerollmops:
My mistake @ook 😂 I was on mobile and may have clicked too fast on the person to ping, sorry 😅

crates/milli/src/update/indexer_config.rs
```diff
@@ -11,7 +13,7 @@ pub struct IndexerConfig {
     pub max_memory: Option<usize>,
     pub chunk_compression_type: CompressionType,
     pub chunk_compression_level: Option<u32>,
-    pub thread_pool: Option<ThreadPoolNoAbort>,
+    pub thread_pool: RwLock<Option<ThreadPoolNoAbort>>,
```
Kerollmops, line 16: It seems that you complicated the problem a bit. I would prefer that we keep this type unchanged but propose a new method to create an `IndexerConfig::clone_ignore_thread_pool` and then assign the `thread_pool` by hand. This method would clone the `IndexerConfig` but always keep the `thread_pool` at `None`.

Kerollmops (changes requested):
Thank you for the changes, but it seems that we use the indexer config only once in the dump import process, and fewer changes would be preferable. I proposed another way to tackle the problem below 👇

crates/milli/src/update/indexer_config.rs
nnethercott: haha yeah this is a much better idea than mucking around with interior mutability mdr, I'll refactor that real quick (third time's the charm) :-)

crates/milli/src/update/indexer_config.rs
```diff
@@ -25,6 +25,20 @@ impl IndexerConfig {
...
+            max_nb_chunks: other.max_nb_chunks,
+            documents_chunk_size: other.documents_chunk_size,
+            max_memory: other.max_memory,
+            chunk_compression_type: other.chunk_compression_type,
+            chunk_compression_level: other.chunk_compression_level,
+            max_positions_per_attributes: other.max_positions_per_attributes,
+            skip_index_budget: other.skip_index_budget,
+            thread_pool: None,
+        }
+    }
```
Kerollmops, line 41:
```suggestion
    pub fn clone_no_threadpool(&self) -> Self {
        Self {
            log_every_n: self.log_every_n,
            max_nb_chunks: self.max_nb_chunks,
            documents_chunk_size: self.documents_chunk_size,
            max_memory: self.max_memory,
            chunk_compression_type: self.chunk_compression_type,
            chunk_compression_level: self.chunk_compression_level,
            max_positions_per_attributes: self.max_positions_per_attributes,
            skip_index_budget: self.skip_index_budget,
            thread_pool: None,
        }
    }
```

crates/meilisearch/src/lib.rs
```diff
@@ -503,7 +504,19 @@ fn import_dump(
     let network = dump_reader.network()?.cloned().unwrap_or_default();
     index_scheduler.put_network(network)?;
 
-    let indexer_config = index_scheduler.indexer_config();
+    let mut indexer_config = IndexerConfig::clone_no_threadpool(index_scheduler.indexer_config());
```
Kerollmops, line 507:
```suggestion
    let mut indexer_config = index_scheduler.indexer_config().clone_no_threadpool();
```

crates/meilisearch/src/lib.rs
```diff
@@ -503,7 +504,19 @@ fn import_dump(
...
+
+    // 3.1 Use all cpus to index the import dump
+    indexer_config.thread_pool = {
+        let all_cpus = num_cpus::get();
+
+        let temp_pool = ThreadPoolNoAbortBuilder::new()
+            .thread_name(|index| format!("indexing-thread:{index}"))
+            .num_threads(all_cpus)
+            .build()?;
```
Kerollmops, line 516: Sorry about the misunderstanding, but when I was talking about using all the CPUs, I meant all the CPUs that were defined through the settings. This means that you'll need to use `Opt` to get the `max-index-threads` value for the `ThreadPoolNoAbortBuilder::num_threads` or don't call this method if `max-index-threads` is not set 🙏

Kerollmops (changes requested):
Hey 👋 Your changes are looking very close to what I actually expect. Just a couple of changes more.

crates/meilisearch/src/lib.rs
nnethercott: If that's the case I think this is already the default behaviour in your code. 

If `max_indexing_threads` isn't defined in the settings then it gets the default `MaxThreads::default()` = num_cpus/2 [(here](https://github.com/meilisearch/meilisearch/blob/8cecc6989a01db45b153e9403bdc53f57acb234c/crates/meilisearch/src/option.rs#L726) and [here)](https://github.com/meilisearch/meilisearch/blob/032c67662d68b2ee4d561b3c8a054f8eabc6c554/crates/meilisearch/src/option.rs#L842); otherwise it reads the value from the env - hence its always defined.  When the `indexer_config` is constructed from the `Opt` a thread_pool with `max_indexing_threads` is then built and assigned to it [(here](https://github.com/meilisearch/meilisearch/blob/main/crates/meilisearch/src/lib.rs#L226) and [here)](https://github.com/meilisearch/meilisearch/blob/8cecc6989a01db45b153e9403bdc53f57acb234c/crates/meilisearch/src/option.rs#L762).

So creating a temporary thread_pool using the `Opt`s `max_indexing_threads` is a no-op since it'll just use the same number of CPUs as before => you're already using `max_indexing_threads` to index the dump.

ManyTheFish: Hey @nnethercott,
Kero is on vacation this week, so I will temporarily take the review.

The goal is to change the default behavior of max_indexing_threads when importing a dump; however, if the max_indexing_threads config is set, we don't want to override it.

So when max_indexing_threads is not set:
- The default number of threads for an indexing process should be the same as before (1/2 of the available)
- The default number of threads for a dump import should be changed to take all available threads

Your changes should only be triggered if max_indexing_threads is not set.

Your current implementation overrides both default behavior and set behavior.

curquiza:
@nnethercott thank you for your PR, let us know when @Kerollmops or @ManyTheFish can review again

crates/meilisearch/src/lib.rs
nnethercott: I'll take a look again tonight.  I think the vibe is gonna be to make `MaxThreads` like `MaxMemory` and have an optional inner type. This way if a user provides `max_indexing_threads` we can build the thread pool, otherwise we defer the pool construction later inside the `index_dump` call which will use all available threads.

ManyTheFish (changes requested):
Ok, I understand what you are doing, you build a thread pool with the default value, but you don't use it in the dump import. What about storing the thread pool in an option and adding a getter returning an `Option<&ThreadPool>` later?
Then, when retrieving the threadpool, use a match case on the getter, this would avoid having the `clone_no_threadpool` function.

nnethercott:
> Ok, I understand what you are doing, you build a thread pool with the default value, but you don't use it in the dump import. What about storing the thread pool in an option and adding a getter returning an `Option<&ThreadPool>` later? Then, when retrieving the threadpool, use a match case on the getter, this would avoid having the `clone_no_threadpool` function.

My understanding of your suggestion is that you want to do something like this:

```rust
impl TryFrom<&IndexerOpts> for IndexerConfig {
   ... 
    fn try_from(other: &IndexerOpts) -> Result<Self, Self::Error> {
        
        // defaults to None if `max_indexing_threads` unspecified
        let thread_pool = other.max_indexing_threads.and_then(|max_threads| {
            ThreadPoolNoAbortBuilder::new()
                .thread_name(|index| format!("indexing-thread:{index}"))
                .num_threads(max_threads)
                .build()
                .ok()
        });

        Ok(Self {
            thread_pool: thread_pool,
            ...
        })

    }
}
```
then later in the dump

```rust 
    let indexer_config = index_scheduler.indexer_config();

    let thread_pool = match indexer_config.thread_pool.as_ref() {
        Some(t) => t,
        None => ThreadPoolNoAbortBuilder::new()
            .thread_name(|index| format!("indexing-thread:{index}"))
            .num_threads(num_cpus::get())
            .build(),
    };
    // use new thread pool in the dump 
```
(Note the above^ won't compile cause the thread_pool None arm returns ThreadPoolNoAbort not a &ThreadPoolNoAbort but its easier to read this way :p)

If this is the case, I'm going to claim that you have no way of passing this new threadpool to an `IndexerConfig` without either cloning, interior mutability, or unsafe. 

## a rant
How do we assign the new threadpool to the indexer_config to [build the `IndexDocuments` instance](https://github.com/meilisearch/meilisearch/blob/71ab11f1fe911edacf0a4990917fed61958c3677/crates/meilisearch/src/lib.rs#L555) ?

In the import_dump function the indexer_config comes from [deref coercion on an `Arc<IndexerConfig>`](https://github.com/meilisearch/meilisearch/blob/main/crates/index-scheduler/src/index_mapper/mod.rs#L573) => it's not mut and in fact its strong count is already 2 by the time we use it (it is cloned building the index mapper [here](https://github.com/meilisearch/meilisearch/blob/71ab11f1fe911edacf0a4990917fed61958c3677/crates/index-scheduler/src/index_mapper/mod.rs#L181)). So we can't use something like Arc::get_mut safely to assign it the new threadpool.  

I [previously suggested](https://github.com/meilisearch/meilisearch/pull/5527/commits/648b2876f65a5f11d12f3e69781c710630bdd46d) adding interior mutability to the threadpool using an RwLock to make the thread_pool like `RwLock<Option<ThreadPoolNoAbort>>` but was advised against it. I'm also assuming we don't want to use an unsafe block and mess around with raw pointers to swap in the new thread pool. That really only leaves cloning, hence the `clone_no_threadpool` method. 

Another solution is to decouple the thread_pool entirely from the indexer config, but that involves a lot more refactoring and requires you to remember to create a threadpool manually based on the config before indexing.

Defaulting the thread_pool of the IndexerConfig to None also has the same drawbacks - namely how do you set it to use 1/2 the cpu threads later on without interior mutability or cloning?

Curious to hear your thoughts on all this

ManyTheFish:
@nnethercott, right...

To be honest, I am not quite happy to have the structure built twice for the same process, which could lead to misinterpretation of the real behavior in the future. 🤔
Moreover, as you said, the structure might be cloned in the index mapper, making the default behavior even more obscure. 😬 
However, having an RwLock does not look good either. 😞 

Sorry, I am rereading the whole code, and I think we should get rid of the Option for the thread_pool in the IndexerConfig struct. The [default impl of IndexerConfig](https://github.com/meilisearch/meilisearch/blob/71ab11f1fe911edacf0a4990917fed61958c3677/crates/milli/src/update/indexer_config.rs#L30-L44), mainly used in the tests, [should be instantiated like this](https://github.com/meilisearch/meilisearch/blob/71ab11f1fe911edacf0a4990917fed61958c3677/crates/milli/src/update/index_documents/mod.rs#L235-L246).

This should simplify a lot of code. Then, about your changes on the dumps, we will keep them as they are; it will be a special case, but it must be the only one. 🤔 

Again, I'm sorry for making up my mind in the review flow.

To summarize:
1) We keep your current changes
2) Could you change the [IndexerConfig struct](https://github.com/meilisearch/meilisearch/blob/71ab11f1fe911edacf0a4990917fed61958c3677/crates/milli/src/update/indexer_config.rs#L6-L17) by changing the `thread_pool: Option<ThreadPoolNoAbort>` into `thread_pool: ThreadPoolNoAbort` and simplify the related code.
3) [The default impl of IndexerConfig](https://github.com/meilisearch/meilisearch/blob/71ab11f1fe911edacf0a4990917fed61958c3677/crates/milli/src/update/indexer_config.rs#L30-L44) must name the ThreadPoolNoAbort's threads to `.thread_name(|i| format!("indexing-thread-{i}"))` and limit the number of threads to 1 if it is a test `#[cfg(test)]`, moreover, to avoid the thread pool being overridden in a dump import, we have to set max_threads to something else to None depending on the case.

nnethercott:
> @nnethercott, right...
> 
> To be honest, I am not quite happy to have the structure built twice for the same process, which could lead to misinterpretation of the real behavior in the future. 🤔 Moreover, as you said, the structure might be cloned in the index mapper, making the default behavior even more obscure. 😬 However, having an RwLock does not look good either. 😞
> 
> Sorry, I am rereading the whole code, and I think we should get rid of the Option for the thread_pool in the IndexerConfig struct. The [default impl of IndexerConfig](https://github.com/meilisearch/meilisearch/blob/71ab11f1fe911edacf0a4990917fed61958c3677/crates/milli/src/update/indexer_config.rs#L30-L44), mainly used in the tests, [should be instantiated like this](https://github.com/meilisearch/meilisearch/blob/71ab11f1fe911edacf0a4990917fed61958c3677/crates/milli/src/update/index_documents/mod.rs#L235-L246).
> 
> This should simplify a lot of code. Then, about your changes on the dumps, we will keep them as they are; it will be a special case, but it must be the only one. 🤔
> 
> Again, I'm sorry for making up my mind in the review flow.
> 
> To summarize:
> 
> 1. We keep your current changes
> 2. Could you change the [IndexerConfig struct](https://github.com/meilisearch/meilisearch/blob/71ab11f1fe911edacf0a4990917fed61958c3677/crates/milli/src/update/indexer_config.rs#L6-L17) by changing the `thread_pool: Option<ThreadPoolNoAbort>` into `thread_pool: ThreadPoolNoAbort` and simplify the related code.
> 3. [The default impl of IndexerConfig](https://github.com/meilisearch/meilisearch/blob/71ab11f1fe911edacf0a4990917fed61958c3677/crates/milli/src/update/indexer_config.rs#L30-L44) must name the ThreadPoolNoAbort's threads to `.thread_name(|i| format!("indexing-thread-{i}"))` and limit the number of threads to 1 if it is a test `#[cfg(test)]`, moreover, to avoid the thread pool being overridden in a dump import, we have to set max_threads to something else to None depending on the case.

Thanks for the clear feedback & direction 🙂 I think removing the Option around the thread_pool is a step in the right direction for sure. Also lets us delete a lot of repeated checks in the tests which is nice. 

Only drawback I can think of with the code now is that the `Default` impl for the `IndexerConfig` executes a fallible operation (namely rayon ThreadPoolBuilder::build()) which could fail at runtime. Since it was optional before we just used `thread_pool: None`

ManyTheFish:
> Only drawback I can think of with the code now is that the Default impl for the IndexerConfig executes a fallible operation (namely rayon ThreadPoolBuilder::build()) which could fail at runtime. Since it was optional before we just used thread_pool: None

I would be okay with panicking here, but I don't think Meilisearch can recover from this. Another solution would be to implement a custom default method returning an error instead of the official Default implementation.

ManyTheFish (changes requested):
Small change before the merge!

Thank you for your contribution!

crates/meilisearch/src/option.rs
```diff
@@ -828,31 +834,31 @@ fn total_memory_bytes() -> Option<u64> {
...
+        usize::from_str(s).map(Some).map(MaxThreads)
     }
 }
 
 impl fmt::Display for MaxThreads {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.0)
+        match self.0 {
+            Some(threads) => write!(f, "{}", threads),
+            None => Ok(()),
```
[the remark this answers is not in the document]
nnethercott: sounds good I'll add that now !

crates/meilisearch/src/lib.rs
```diff
@@ -500,7 +501,22 @@ fn import_dump(
...
     index_scheduler.put_network(network)?;
 
-    let indexer_config = index_scheduler.indexer_config();
+    // 3.1 Use all cpus to process dump if `max_indexing_threads` not configured
+    let backup_config;
+    let base_config = index_scheduler.indexer_config();
+
+    let indexer_config = if base_config.max_threads.is_none() {
+        let thread_pool = ThreadPoolNoAbortBuilder::new()
+            .thread_name(|index| format!("indexing-thread:{index}"))
```
martin-g, line 510: This changes the thread name from `indexing-thread-N` to `indexing-thread:N`. Is this intentional ?

crates/meilisearch/src/option.rs
```diff
@@ -746,26 +746,32 @@ impl IndexerOpts {
...
+        }
     }
 }
 
 impl TryFrom<&IndexerOpts> for IndexerConfig {
     type Error = anyhow::Error;
 
     fn try_from(other: &IndexerOpts) -> Result<Self, Self::Error> {
+        // use 1/2 cpu threads if no value specified
+        let max_indexing_threads = other.max_indexing_threads.unwrap_or(num_cpus::get() / 2);
```
martin-g, line 763:
```suggestion
        let max_indexing_threads = other.max_indexing_threads.unwrap_or_else(|| num_cpus::get() / 2);
```
Do not make additional computation unless really necessary.

crates/milli/src/update/indexer_config.rs
```diff
@@ -29,14 +30,32 @@ impl IndexerConfig {
 
 impl Default for IndexerConfig {
     fn default() -> Self {
+        #[allow(unused_mut)]
```
martin-g, line 33: Personal option: I think it would be cleaner to declare a helper method with two impls - one for `cfg(test)` and another for non-test that return a tuple with immutable and pre-configured values for the pool and max_threads.

crates/meilisearch/src/lib.rs
nnethercott: I went with the [same convention the TryFrom](https://github.com/meilisearch/meilisearch/blob/main/crates/meilisearch/src/option.rs#L761), but since you pointed it out the threads are indeed named "indexing-thread-N" in `process_index_operations.rs` like [here](https://github.com/meilisearch/meilisearch/blob/677e8b122caf47b2da6c18c88002aa4a8ffb98bf/crates/index-scheduler/src/scheduler/process_index_operation.rs#L122) -- Maybe the names should be consolidated ?

ManyTheFish: Let's keep the `TryFrom` one, it's good to see this PR will unifythe namings

crates/milli/src/update/indexer_config.rs
ManyTheFish: Good idea, this could be used in the other constructor by passing the max_threads as a paameter as well:
https://github.com/meilisearch/meilisearch/blob/e96c1d4b0fa044da9606c5247e921e30cfe58abb/crates/meilisearch/src/option.rs#L761-L780

nnethercott: Factorized the test and non-test behaviour out in [this commit](https://github.com/meilisearch/meilisearch/pull/5527/commits/865f24cfefbdfd2d0c6f6be2006266977f9d4cee) but didn't go as far as abstracting it away in the TryFrom. I think the new builder method `ThreadPoolNoAbortBuilder::new_for_indexing()` cleans up a lot of stuff

martin-g (approved):

crates/meilisearch/src/lib.rs
```diff
@@ -500,7 +501,20 @@ fn import_dump(
...
     index_scheduler.put_network(network)?;
 
-    let indexer_config = index_scheduler.indexer_config();
+    // 3.1 Use all cpus to process dump if `max_indexing_threads` not configured
+    let backup_config;
+    let base_config = index_scheduler.indexer_config();
+
+    let indexer_config = if base_config.max_threads.is_none() {
+        let thread_pool =
+            ThreadPoolNoAbortBuilder::new_for_indexing().num_threads(num_cpus::get()).build()?;
```
ManyTheFish, line 510: Only use one thread in test suits me better:
```suggestion
        let (thread_pool, _) = default_thread_pool_and_threads()?;
```

crates/meilisearch/src/option.rs
```diff
@@ -828,31 +830,31 @@ fn total_memory_bytes() -> Option<u64> {
...
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        usize::from_str(s).map(Self)
-    }
-}
-
-impl Default for MaxThreads {
-    fn default() -> Self {
-        MaxThreads(num_cpus::get() / 2)
+    fn from_str(s: &str) -> Result<MaxThreads, Self::Err> {
+        if s.is_empty() || s == "unlimited" {
```
ManyTheFish, line 840: My bad on this, I thought it was for debugging, but it's used for the environment variable, revert 😬  
```suggestion
        if s.is_empty() {
```

crates/meilisearch/src/option.rs
```diff
@@ -828,31 +830,31 @@ fn total_memory_bytes() -> Option<u64> {
...
+        usize::from_str(s).map(Some).map(MaxThreads)
     }
 }
 
 impl fmt::Display for MaxThreads {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.0)
+        match self.0 {
+            Some(threads) => write!(f, "{}", threads),
+            None => write!(f, "unlimited"),
```
ManyTheFish, line 851: My bad on this, I thought it was for debugging, but it's used for the environment variable, revert 😬
```suggestion
            None => (),
```

ManyTheFish (changes requested):
Final review before merging,

thank you for your work, it will definitelly improve the codebase!

crates/meilisearch/src/option.rs
nnethercott: no worries haha did the same thing last night

ManyTheFish (approved):
LGTM! 

@Kerollmops, do you want to run a last review as the issue creator? 😄 

context:
- the code has been factorized for every thread pool users
- an exception has been made for import dump, if max_threads is None, then we use all the threads instead of half of them

PR #5545, opened 2025-05-07
[PROTOTYPE] Use unstable arroy version containing incremental indexing

ManyTheFish:


Kerollmops:
Closing as we are no longer using arroy and we are switching everyone to hannoy.

PR #5552, opened 2025-05-12
Add v1.15 in index-scheduler upgrade

ManyTheFish:
The dumpless upgrade for v1.15 was not fully implemented. Implement it for:
- index scheduler


dureuill (approved):
perfect thank you

PR #5556, opened 2025-05-13
Chat route

Kerollmops:
This PR exposes the new `POST /chats/$workspace/chat/completions` route compatible with the underlying LLM and a `PUT+GET /chats/$workspace/settings`.  If the LLM setup from the `/chats/$workspace/chat` is OpenAI, the non-streaming and streaming APIs will return the content of OpenAI. If the LLM is Mistral, the content will be the Mistral responses.

- Create a new _Default Chat API Key_ which [supports tenant tokens](https://www.meilisearch.com/docs/learn/security/tenant_token_reference#example-payload).
- The tool calling is capable of handling any index that is visible to the key.
- The chat completions API supports workspaces so users can define different prompts or LLM providers on the same instance.
- The chat completions feature must be enabled through the experimental-features route.

You can read more about this chat completions feature on [the official usage page](https://meilisearch.notion.site/Feature-Meilisearch-chat-route-usage-1f84b06b651f80eab264f8821e9f3a00?pvs=4). You'll find better API descriptions, concerns, details about key management, and even code examples. The PRD can be found [on this page (internal)](https://www.notion.so/meilisearch/Chat-Completions-API-1f84b06b651f80faaf6cc61cbf862d4a).

There is [a prototype available on Docker](https://github.com/meilisearch/meilisearch/actions/runs/15564372590) with the following tag/command:
```
docker pull getmeili/meilisearch:prototype-chat-route-16
```

---

### This PR makes forward-compatible changes to the DB format

- [x] Detail the change to the DB format and why they are forward compatible: 
    - this PR adds new settings to the index that are treated as default if missing
    - this PR adds a new database to the index-scheduler. This database is created on opening the index-scheduler environment if missing

---

## To Do

- [x] Deactivate the non-streamed route.
  - [ ] ~~The non-streamed version is buggy when calling tools (+ panic on error)~~
- [x] Redact API Key from PATCH chats settings
- [x] Make sure the Actions::ALL is doing everything. [It is](https://github.com/meilisearch/meilisearch/blob/a7f5d3bb7a711df25d29630c6623a6f12a4bfba6/crates/meilisearch-auth/src/store.rs#L91).
- [x] Rename `baseApi` into `baseUrl`
- [x] Remove `preQuery`
- [x] Add a `GET /chats/{chatWorkspace}` route (akin to get index)
- [ ] Maybe rename the `prompts` parameter?
- [ ] Add analytics


crates/meilisearch/src/routes/chat.rs
```diff
@@ -0,0 +1,466 @@
...
+
+    setup_search_tool(&mut chat_completion, &chat_settings.prompts);
+
+    let (tx, rx) = tokio::sync::mpsc::channel(10);
+    let _join_handle = Handle::current().spawn(async move {
+        let client = Client::with_config(config.clone());
+        let mut global_tool_calls = HashMap::<u32, Call>::new();
+        let mut finish_reason = None;
+
+        'main: while finish_reason.map_or(true, |fr| fr == FinishReason::ToolCalls) {
```
Kerollmops, line 273: I am not sure we should run a new conversation when the original stream stopped without any finish reason. We should stop.
```suggestion
        'main: while finish_reason.map_or(false, |fr| fr == FinishReason::ToolCalls) {
```

crates/meilisearch-types/src/features.rs
```diff
@@ -2,6 +2,13 @@ use std::collections::BTreeMap;
...
+pub const DEFAULT_CHAT_SEARCH_DESCRIPTION_PROMPT: &str =
+    "Search the database for relevant JSON documents using an optional query.";
+pub const DEFAULT_CHAT_SEARCH_Q_PARAM_PROMPT: &str = "The search query string used to find relevant documents in the index. This should contain keywords or phrases that best represent what the user is looking for. More specific queries will yield more precise results.";
+pub const DEFAULT_CHAT_SEARCH_INDEX_UID_PARAM_PROMPT: &str = "The name of the index to search within. An index is a collection of documents organized for search. Selecting the right index ensures the most relevant results for the user query. You can access to two indexes: movies, steam. The movies index contains movies with overviews. The steam index contains steam games from the Steam platform with their prices";
```
martin-g:
```suggestion
pub const DEFAULT_CHAT_SEARCH_INDEX_UID_PARAM_PROMPT: &str = "The name of the index to search within. An index is a collection of documents organized for search. Selecting the right index ensures the most relevant results for the user query. You have access to two indexes: movies, steam. The movies index contains movies with overviews. The steam index contains steam games from the Steam platform with their prices";
```

crates/meilisearch-types/src/features.rs
```diff
@@ -37,3 +45,204 @@ pub struct Network {
...
+            VLlm => System,
+        }
+    }
+
+    /// Returns true if the model is an old OpenAI model.
+    ///
+    /// Old OpenAI models use the system role while new ones use the developer role.
+    fn old_openai_model(model: &str) -> bool {
+        ["gpt-3.5", "gpt-4", "gpt-4.1", "gpt-4.5", "gpt-4o", "chatgpt-4o"].iter().any(|old| {
+            model.starts_with(old) && model.chars().nth(old.len()).is_none_or(|last| last == '-')
```
Kerollmops:
```suggestion
            model.starts_with(old) && model.chars().nth(old.chars().count()).is_none_or(|last| last == '-')
```

crates/meilisearch-types/src/features.rs
```diff
@@ -37,3 +45,204 @@ pub struct Network {
...
+        })
+    }
+
+    pub fn base_url(&self) -> Option<&'static str> {
```
Kerollmops, line 152: Check correct error when None

crates/meilisearch-types/src/keys.rs
```diff
@@ -308,6 +323,25 @@ pub enum Action {
     #[serde(rename = "network.update")]
     #[deserr(rename = "network.update")]
     NetworkUpdate,
+    // TODO should we rename it chatCompletions.get ?
+    #[serde(rename = "chatCompletion")]
+    #[deserr(rename = "chatCompletion")]
+    ChatCompletions,
```
Kerollmops:
```suggestion
    #[serde(rename = "chatCompletions")]
    #[deserr(rename = "chatCompletions")]
    ChatCompletions,
```

crates/meilisearch-types/src/keys.rs
```diff
@@ -308,6 +323,25 @@ pub enum Action {
     #[serde(rename = "network.update")]
     #[deserr(rename = "network.update")]
     NetworkUpdate,
+    // TODO should we rename it chatCompletions.get ?
```
Kerollmops:
```suggestion
```

crates/meilisearch-types/src/keys.rs
```diff
@@ -308,6 +323,25 @@ pub enum Action {
...
+    ChatsSettingsAll,
+    #[serde(rename = "chatsSettings.get")]
+    #[deserr(rename = "chatsSettings.get")]
+    ChatsSettingsGet,
+    #[serde(rename = "chatsSettings.update")]
+    #[deserr(rename = "chatsSettings.update")]
+    ChatsSettingsUpdate,
+    #[serde(rename = "chatsSettings.delete")]
+    #[deserr(rename = "chatsSettings.delete")]
+    ChatsSettingsDelete,
```
Kerollmops:
```suggestion
    #[serde(rename = "chats.delete")]
    #[deserr(rename = "chats.delete")]
    ChatsDelete,
```

crates/meilisearch-types/src/keys.rs
```diff
@@ -333,6 +367,12 @@ impl Action {
             SETTINGS_ALL => Some(Self::SettingsAll),
             SETTINGS_GET => Some(Self::SettingsGet),
             SETTINGS_UPDATE => Some(Self::SettingsUpdate),
+            CHAT_COMPLETIONS => Some(Self::ChatCompletions),
+            CHATS_GET => Some(Self::ChatsGet),
+            CHATS_SETTINGS_ALL => Some(Self::ChatsSettingsAll),
+            CHATS_SETTINGS_GET => Some(Self::ChatsSettingsGet),
+            CHATS_SETTINGS_UPDATE => Some(Self::ChatsSettingsUpdate),
+            CHATS_SETTINGS_DELETE => Some(Self::ChatsSettingsDelete),
```
Kerollmops:
```suggestion
            CHATS_DELETE => Some(Self::ChatsDelete),
```

crates/meilisearch/src/routes/chats/config.rs
```diff
@@ -0,0 +1,87 @@
...
+                if let Some(org_id) = chat_settings.org_id.as_ref() {
+                    config = config.with_org_id(org_id);
+                }
+                if let Some(project_id) = chat_settings.project_id.as_ref() {
+                    config = config.with_project_id(project_id);
+                }
+                if let Some(api_key) = chat_settings.api_key.as_ref() {
+                    config = config.with_api_key(api_key);
+                }
+                if let Some(base_api) = chat_settings.base_api.as_ref() {
```
Kerollmops: We must return an error if vLLM (and one more) is not giving its base url

crates/milli/src/prompt/mod.rs
```diff
@@ -64,7 +64,7 @@ fn default_template() -> liquid::Template {
     new_template(default_template_text()).unwrap()
 }
 
-fn default_template_text() -> &'static str {
+pub fn default_template_text() -> &'static str {
```
Kerollmops, line 67:
```suggestion
fn default_template_text() -> &'static str {
```

crates/milli/src/update/settings.rs
```diff
@@ -123,6 +124,15 @@ impl<T> Setting<T> {
         *self = new;
         true
     }
+
+    #[track_caller]
+    pub fn unwrap(self) -> T {
```
Kerollmops: Do I need this again?

crates/meilisearch/src/routes/chats/settings.rs
```diff
@@ -0,0 +1,264 @@
...
+use crate::extractors::authentication::policies::ActionPolicy;
+use crate::extractors::authentication::GuardedData;
+use crate::extractors::sequential_extractor::SeqHandler;
+
+pub fn configure(cfg: &mut web::ServiceConfig) {
+    cfg.service(
+        web::resource("")
+            .route(web::get().to(SeqHandler(get_settings)))
+            .route(web::patch().to(SeqHandler(patch_settings)))
+            .route(web::delete().to(SeqHandler(delete_settings))),
```
Kerollmops: Resets the settings to the default ones and add a delete method at the workspace level

crates/meilisearch/src/routes/chats/settings.rs
```diff
@@ -0,0 +1,264 @@
...
+        },
+        api_key: match new.api_key {
+            Setting::Set(new_api_key) => Some(new_api_key),
+            Setting::Reset => None,
+            Setting::NotSet => old_settings.api_key,
+        },
+        prompts,
+    };
+
+    // TODO send analytics
```
Kerollmops, line 136:
```suggestion
    // TODO send analytics
```

crates/meilisearch/src/routes/chats/settings.rs
```diff
@@ -0,0 +1,264 @@
...
+    //     &req,
+    // );
+
+    index_scheduler.put_chat_settings(&mut wtxn, &workspace_uid, &settings)?;
+    wtxn.commit()?;
+
+    Ok(HttpResponse::Ok().json(settings))
+}
+
+async fn delete_settings(
```
Kerollmops: Rename reset settings and do not delete. Delete at the workspace level instead.

crates/meilisearch/src/routes/chats/settings.rs
```diff
@@ -0,0 +1,264 @@
...
+    index_scheduler: GuardedData<
+        ActionPolicy<{ actions::CHATS_SETTINGS_DELETE }>,
```
Kerollmops:
```suggestion
        ActionPolicy<{ actions::CHATS_SETTINGS_UPDATE }>,
```

crates/meilisearch/src/routes/chats/settings.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    // TODO do a spawn_blocking here
+    let mut wtxn = index_scheduler.write_txn()?;
+    if index_scheduler.delete_chat_settings(&mut wtxn, &workspace_uid)? {
+        wtxn.commit()?;
+        Ok(HttpResponse::NoContent().finish())
+    } else {
+        Err(ResponseError::from_msg(
+            format!("Chat `{workspace_uid}` not found"),
+            Code::ChatWorkspaceNotFound,
```
Kerollmops:
```suggestion
            Code::ChatWorkspaceNotFound,
```

Discuss with gui, do we want to leak this term?

crates/meilisearch/src/routes/chats/settings.rs
```diff
@@ -0,0 +1,264 @@
...
+        ))
+    }
+}
+
+#[derive(Debug, Clone, Deserialize, Deserr, ToSchema)]
+#[deserr(error = DeserrJsonError, rename_all = camelCase, deny_unknown_fields)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+#[schema(rename_all = "camelCase")]
+pub struct GlobalChatSettings {
```
Kerollmops:
```suggestion
pub struct ChatWorkspaceSettings {
```

crates/meilisearch/src/routes/chats/settings.rs
```diff
@@ -0,0 +1,264 @@
...
+    #[schema(value_type = Option<String>, example = json!("2024-02-01"))]
+    pub api_version: Setting<String>,
+    #[serde(default)]
+    #[deserr(default, error = DeserrJsonError<InvalidChatCompletionDeploymentId>)]
+    #[schema(value_type = Option<String>, example = json!("1234abcd..."))]
+    pub deployment_id: Setting<String>,
+    #[serde(default)]
+    #[deserr(default, error = DeserrJsonError<InvalidChatCompletionBaseApi>)]
+    #[schema(value_type = Option<String>, example = json!("https://api.mistral.ai/v1"))]
+    pub base_api: Setting<String>,
```
Kerollmops:
```suggestion
    pub base_url: Setting<String>,
```

crates/meilisearch/src/routes/chats/settings.rs
```diff
@@ -0,0 +1,264 @@
...
+    #[schema(inline, value_type = Option<ChatPrompts>)]
+    pub prompts: Setting<ChatPrompts>,
+}
+
+#[derive(Default, Debug, Clone, Copy, Serialize, Deserialize, Deserr, ToSchema)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+#[deserr(error = DeserrJsonError, rename_all = camelCase, deny_unknown_fields)]
+pub enum ChatCompletionSource {
+    #[default]
+    OpenAi,
```
Kerollmops, line 219:
```suggestion
    OpenAi,
    ...
```

crates/meilisearch/src/routes/chats/settings.rs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+impl From<ChatCompletionSource> for DbChatCompletionSource {
+    fn from(source: ChatCompletionSource) -> Self {
+        match source {
+            ChatCompletionSource::OpenAi => DbChatCompletionSource::OpenAi,
```
Kerollmops:
```suggestion
            ChatCompletionSource::OpenAi => DbChatCompletionSource::OpenAi,
            ...
```

crates/meilisearch/src/routes/chats/utils.rs
```diff
@@ -0,0 +1,243 @@
...
+
+    /// Ask the front-end user to append this tool *call* to the conversation
+    pub async fn append_tool_call_conversation_message(
+        &self,
+        resp: CreateChatCompletionStreamResponse,
+        call_id: String,
+        function_name: String,
+        function_arguments: String,
+    ) -> Result<(), SendError<Event>> {
+        #[allow(deprecated)]
```
Kerollmops:
```suggestion
        #[allow(deprecated)] // function_call
```

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,719 @@
...
+                                            function_name: name.as_ref().unwrap().clone(),
+                                            arguments: arguments.as_ref().unwrap().clone(),
+                                        }
+                                    } else {
+                                        Call::External
+                                    }
+                                });
+
+                            if global_tool_calls.get(index).is_some_and(Call::is_external) {
+                                todo!("Support forwarding external tool calls");
```
Kerollmops: Is it ouf ? no!

crates/index-scheduler/src/lib.rs
```diff
@@ -269,6 +279,7 @@ impl IndexScheduler {
         let features = features::FeatureData::new(&env, &mut wtxn, options.instance_features)?;
         let queue = Queue::new(&env, &mut wtxn, &options)?;
         let index_mapper = IndexMapper::new(&env, &mut wtxn, &options, budget)?;
+        let chat_settings = env.create_database(&mut wtxn, Some("chat-settings"))?;
```
dureuill: nitpick: should that be a constant instead of a literal?

crates/index-scheduler/src/lib.rs
```diff
@@ -292,12 +303,21 @@ impl IndexScheduler {
...
         Ok(this)
     }
 
+    pub fn write_txn(&self) -> Result<RwTxn> {
+        self.env.write_txn().map_err(|e| e.into())
+    }
+
+    pub fn read_txn(&self) -> Result<RoTxn<WithoutTls>> {
+        self.env.read_txn().map_err(|e| e.into())
+    }
```
dureuill: nitpick: I think I'd rather we keep these functions private and use the txns internally in the methods of the index-scheduler.

(note that read_txn was already public, but unless I'm mistaken it wasn't actually necessary)

crates/index-scheduler/src/lib.rs
```diff
@@ -536,6 +552,25 @@ impl IndexScheduler {
         ret.map(|ret| (total, ret))
     }
 
+    /// Returns the total number of chat workspaces available ~~for the specified filter~~.
+    /// And a `Vec` of the workspace_uids
+    pub fn paginated_chat_workspace_uids(
+        &self,
+        _filters: &meilisearch_auth::AuthFilter,
```
dureuill: do we have to keep the `filters` argument if it isn't used?

or is that actually a `TODO`?

either way isn't clear to me.

crates/meilisearch/Cargo.toml
```diff
@@ -48,6 +49,7 @@ is-terminal = "0.4.13"
 itertools = "0.14.0"
 jsonwebtoken = "9.3.0"
 lazy_static = "1.5.0"
+liquid = "0.26.9"
```
dureuill: appears unused, I suggest removal


```suggestion
```

crates/milli/src/index.rs
```diff
@@ -1917,13 +1943,97 @@ pub struct IndexEmbeddingConfig {
     pub user_provided: RoaringBitmap,
 }
 
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct ChatConfig {
```
dureuill, line 1947: note for later: this is not especially on the ChatConfig (IndexEmbeddingConfig) is the same, but index.rs is starting to become really crowded. We should consider splitting it into multiple modules.

crates/milli/src/index.rs
```diff
@@ -1917,13 +1943,97 @@ pub struct IndexEmbeddingConfig {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub matching_strategy: Option<MatchingStrategy>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub attributes_to_search_on: Option<Vec<String>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub ranking_score_threshold: Option<RankingScoreThreshold>,
+}
+
+#[derive(Debug, Clone, Copy, Default, Deserialize, Serialize, PartialEq, Deserr, ToSchema)]
+#[deserr(try_from(f64) = TryFrom::try_from -> InvalidSearchRankingScoreThreshold)]
```
dureuill: is the error code correct here? Should be `InvalidSettingsRankingScoreThreshold` or not?

crates/milli/src/index.rs
```diff
@@ -1917,13 +1943,97 @@ pub struct IndexEmbeddingConfig {
...
+#[serde(rename_all = "camelCase")]
+pub enum MatchingStrategy {
+    /// Remove query words from last to first
+    #[default]
+    Last,
+    /// All query words are mandatory
+    All,
+    /// Remove query words from the most frequent to the least
+    Frequency,
+}
```
dureuill, line 2037: Nitpick: that duplication is unfortunate :/ Maybe we could comment on it at the very least?

crates/meilisearch-types/src/features.rs
```diff
@@ -2,6 +2,12 @@ use std::collections::BTreeMap;
...
+pub const DEFAULT_CHAT_SEARCH_DESCRIPTION_PROMPT: &str =
+    "Search the database for relevant JSON documents using an optional query.";
+pub const DEFAULT_CHAT_SEARCH_Q_PARAM_PROMPT: &str = "The search query string used to find relevant documents in the index. This should contain keywords or phrases that best represent what the user is looking for. More specific queries will yield more precise results.";
+pub const DEFAULT_CHAT_SEARCH_INDEX_UID_PARAM_PROMPT: &str = "The name of the index to search within. An index is a collection of documents organized for search. Selecting the right index ensures the most relevant results for the user query. You have access to two indexes: movies, steam. The movies index contains movies with overviews. The steam index contains steam games from the Steam platform with their prices";
```
dureuill: should we hardcode "you have access to two indexes: movies, steam"? Seems wrong to me.

crates/meilisearch-types/src/deserr/mod.rs
```diff
@@ -117,6 +121,16 @@ impl<C: Default + ErrorCode> DeserializeError for DeserrQueryParamError<C> {
     }
 }
 
+impl Deserr<DeserrError<DeserrJson, deserr_codes::InvalidSettingsIndexChat>> for ChatSettings {
```
dureuill, line 124: I don't understand why the manual implementation was necessary, nor why it had to be done in this file rather than at the definition of `ChatSettings`. Maybe you could help me with a comment?

crates/meilisearch/src/search/mod.rs
```diff
@@ -56,7 +57,7 @@ pub const DEFAULT_HIGHLIGHT_PRE_TAG: fn() -> String = || "<em>".to_string();
 pub const DEFAULT_HIGHLIGHT_POST_TAG: fn() -> String = || "</em>".to_string();
 pub const DEFAULT_SEMANTIC_RATIO: fn() -> SemanticRatio = || SemanticRatio(0.5);
 
-#[derive(Clone, Default, PartialEq, Deserr, ToSchema)]
+#[derive(Clone, PartialEq, Deserr, ToSchema)]
```
dureuill: why are we switching to a manual implementation of `Default` here?

crates/meilisearch/src/search/mod.rs
```diff
@@ -119,9 +120,73 @@ pub struct SearchQuery {
...
+                    .unwrap_or_else(DEFAULT_SEMANTIC_RATIO),
+                embedder,
+            }),
+            limit: limit.unwrap_or_else(DEFAULT_SEARCH_LIMIT),
+            sort,
+            distinct,
+            matching_strategy: matching_strategy.map(MatchingStrategy::from).unwrap_or_default(),
+            attributes_to_search_on,
+            ranking_score_threshold: ranking_score_threshold.map(RankingScoreThreshold::from),
+            ..Default::default()
```
dureuill: I would strongly suggest we avoid the `..Default::default()` here.

Adding a parameter to `SearchQuery` *should* break this code, so that we get the opportunity to review if we also want to add it to `SearchParameters`.

crates/meilisearch/src/search/mod.rs
```diff
@@ -136,6 +201,14 @@ impl std::convert::TryFrom<f64> for RankingScoreThreshold {
     }
 }
 
+impl From<index::RankingScoreThreshold> for RankingScoreThreshold {
+    fn from(threshold: index::RankingScoreThreshold) -> Self {
+        let threshold = threshold.as_f64();
+        assert!((0.0..=1.0).contains(&threshold));
```
dureuill, line 192: floats are a bit feeble, I hope a roundtrip won't make a value stored in the index slightly bigger than 1, otherwise 💥

crates/milli/src/update/chat.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub limit: Setting<usize>,
+
+    // #[serde(default, skip_serializing_if = "Setting::is_not_set")]
+    // #[deserr(default)]
+    // pub attributes_to_retrieve: Option<BTreeSet<String>>,
+
+    // #[serde(default, skip_serializing_if = "Setting::is_not_set")]
+    // #[deserr(default)]
+    // pub filter: Option<Value>,
+    //
```
dureuill: to remove?

crates/milli/src/update/settings.rs
dureuill: seems to be unused. Reading the code it appears `unwrap()` is the same as `set().unwrap()`

crates/milli/src/update/settings.rs
```diff
@@ -1238,6 +1249,109 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
...
+                };
+
+                let prompt = PromptData {
+                    template: match new_document_template {
+                        Setting::Set(new) => new.clone(),
+                        Setting::Reset => default_template_text().to_string(),
+                        Setting::NotSet => prompt.template.clone(),
+                    },
+                    max_bytes: match new_document_template_max_bytes {
+                        Setting::Set(m) => NonZeroUsize::new(*m),
```
dureuill: This will turn 0 into `None`, I don't think this is intended behavior.

Compare with the behavior for embedders <https://github.com/meilisearch/meilisearch/blob/29d82ade5646f7fec5f19b7618ea2beaec39d273/crates/milli/src/update/settings.rs#L1820>

crates/milli/src/update/settings.rs
```diff
@@ -1238,6 +1249,109 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
...
+                            attributes_to_search_on,
+                            ranking_score_threshold,
+                        } = sp;
+
+                        SearchParameters {
+                            hybrid: match hybrid {
+                                Setting::Set(hybrid) => Some(crate::index::HybridQuery {
+                                    semantic_ratio: *hybrid.semantic_ratio,
+                                    embedder: hybrid.embedder.clone(),
+                                }),
```
dureuill, line 1302: should we allow incremental update of `hybrid`? like, just sending the semantic_ratio, without updating the embedder

crates/meilisearch/src/routes/indexes/settings.rs
```diff
@@ -557,6 +569,10 @@ pub async fn update_all(
     debug!(parameters = ?new_settings, "Update all settings");
     let new_settings = validate_settings(new_settings, &index_scheduler)?;
 
+    if !new_settings.chat.is_not_set() {
+        index_scheduler.features().check_chat_completions("setting `chat` in the index route")?;
```
dureuill:
```suggestion
        index_scheduler.features().check_chat_completions("passing `chat` in the settings")?;
```

crates/meilisearch/src/routes/indexes/settings.rs
```diff
@@ -741,5 +762,9 @@ fn validate_settings(
         }
     }
 
+    if let Setting::Set(_chat) = &settings.chat {
+        features.check_chat_completions("setting `chat` in the index settings")?;
```
dureuill, line 762: I don't understand why both checks are needed

crates/meilisearch/src/routes/chats/utils.rs
```diff
@@ -0,0 +1,243 @@
...
+            index: 0,
+            id: Some(uuid::Uuid::new_v4().to_string()),
+            r#type: Some(ChatCompletionToolType::Function),
+            function: Some(FunctionCallStream {
+                name: Some(MEILI_SEARCH_PROGRESS_NAME.to_string()),
+                arguments: Some(call_text),
+            }),
+        };
+
+        resp.choices[0] = ChatChoiceStream {
```
dureuill, line 126: this will panic if choices is empty. Could we comment on why this happening would necessarily be a bug?

crates/meilisearch/src/routes/chats/utils.rs
```diff
@@ -0,0 +1,243 @@
...
+            index: 0,
+            id: Some(uuid::Uuid::new_v4().to_string()),
+            r#type: Some(ChatCompletionToolType::Function),
+            function: Some(FunctionCallStream {
+                name: Some(MEILI_SEARCH_SOURCES_NAME.to_string()),
+                arguments: Some(call_text),
+            }),
+        };
+
+        resp.choices[0] = ChatChoiceStream {
```
dureuill, line 171: same remark as above

crates/meilisearch/src/routes/chats/utils.rs
```diff
@@ -0,0 +1,243 @@
...
+        call_id: &str,
+        documents: &[Document],
+    ) -> Result<(), SendError<Event>> {
+        #[derive(Debug, Clone, Serialize)]
+        /// Provides sources of the search.
+        struct MeiliSearchSources<'a> {
+            /// The call ID to track the original search associated to those sources.
+            call_id: &'a str,
+            /// The documents associated with the search (call_id).
+            /// Only the displayed attributes of the documents are returned.
```
dureuill, line 155: it looks like this property must be verified by the caller of this function?

crates/meilisearch/src/routes/chats/utils.rs
```diff
@@ -0,0 +1,243 @@
...
+    ) -> Result<(), SendError<Event>> {
+        self.send_json(resp).await
+    }
+
+    pub async fn send_error(&self, error: &StreamErrorEvent) -> Result<(), SendError<Event>> {
+        self.send_json(error).await
+    }
+
+    pub async fn stop(self) -> Result<(), SendError<Event>> {
+        self.0.send(Event::Data(sse::Data::new("[DONE]"))).await
```
dureuill: should we use a constant here instead of a literal? is this literal imbued with a specific meaning to clients? We could maybe comment on it

crates/meilisearch/src/routes/chats/utils.rs
```diff
@@ -0,0 +1,243 @@
...
+    index: &Index,
+    doc_alloc: &'doc Bump,
+    internal_docids: Vec<DocumentId>,
+) -> Result<Vec<&'doc str>, ResponseError> {
+    let ChatConfig { prompt: PromptData { template, max_bytes }, .. } = index.chat_config(rtxn)?;
+
+    let prompt = Prompt::new(template, max_bytes).unwrap();
+    let fid_map = index.fields_ids_map(rtxn)?;
+    let metadata_builder = MetadataBuilder::from_index(index, rtxn)?;
+    let fid_map_with_meta = FieldIdMapWithMetadata::new(fid_map.clone(), metadata_builder);
```
dureuill, line 225: Performance concern: this is a potentially costly operation that we've avoided to use on the search side up to now ...

At the very least, as it might be difficult to rearchitect the code not to use it, I suggest making extra sure this is used only once per search

crates/meilisearch/src/routes/chats/utils.rs
```diff
@@ -0,0 +1,243 @@
...
+    let external_ids: Vec<String> = index
+        .external_id_of(rtxn, internal_docids.iter().copied())?
+        .into_iter()
+        .collect::<Result<_, _>>()?;
+
+    let mut renders = Vec::new();
+    for (docid, external_docid) in internal_docids.into_iter().zip(external_ids) {
+        let document = match DocumentFromDb::new(docid, rtxn, index, &fid_map)? {
+            Some(doc) => doc,
+            None => continue,
```
dureuill: should this be unreachable?

dureuill (commented):
I'm not done with the routes directory, but I reviewed most of the rest

crates/index-scheduler/src/lib.rs
Kerollmops: The reason was because I wanted to add filters on the API Keys but I'll do that later. Let's remove this param ✅

crates/milli/src/index.rs
Kerollmops: Hum, yeah, you are right!

crates/meilisearch/src/routes/chats/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+
+pub mod chat_completions;
+mod config;
+mod errors;
+pub mod settings;
+mod utils;
+
+/// The function name to report search progress.
+const MEILI_SEARCH_PROGRESS_NAME: &str = "_meiliSearchProgress";
+/// The function name to append a conversation message in the user conversation.
```
dureuill, line 32: I feel like the comment here should elaborate on *why* this is needed

crates/meilisearch-types/src/features.rs
Kerollmops: Haha yeah, wrong manipulation here 😂

crates/meilisearch/src/search/mod.rs
Kerollmops: The reason is because the default implementation just put zero values instead of the default of serde, i.e. the default limit is zero in Rust but 20 by using serde. So I unified everything with this impl.

crates/meilisearch/src/routes/chats/settings.rs
```diff
@@ -0,0 +1,262 @@
...
+        ActionPolicy<{ actions::CHATS_SETTINGS_GET }>,
+        Data<IndexScheduler>,
+    >,
+    chats_param: web::Path<ChatsParam>,
+) -> Result<HttpResponse, ResponseError> {
+    index_scheduler.features().check_chat_completions("using the /chats/settings route")?;
+
+    let ChatsParam { workspace_uid } = chats_param.into_inner();
+
+    // TODO do a spawn_blocking here ???
```
dureuill: I don't think that is necessary: this operation is not expected to take a long time in general, unlike, say, a search.

crates/meilisearch/src/routes/chats/settings.rs
```diff
@@ -0,0 +1,262 @@
...
+        ActionPolicy<{ actions::CHATS_SETTINGS_UPDATE }>,
+        Data<IndexScheduler>,
+    >,
+    chats_param: web::Path<ChatsParam>,
+    web::Json(new): web::Json<ChatWorkspaceSettings>,
+) -> Result<HttpResponse, ResponseError> {
+    index_scheduler.features().check_chat_completions("using the /chats/settings route")?;
+    let ChatsParam { workspace_uid } = chats_param.into_inner();
+
+    // TODO do a spawn_blocking here
```
dureuill: same as above I think we can remove the comment here

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,713 @@
...
+    };
+    chat_completion.messages.insert(0, system_message);
+
+    Ok(FunctionSupport { report_progress, report_sources, append_to_conversation })
+}
+
+/// Process search request and return formatted results
+async fn process_search_request(
+    index_scheduler: &GuardedData<
+        ActionPolicy<{ actions::CHAT_COMPLETIONS }>,
```
dureuill, line 230: should we request `actions::SEARCH` here?

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,713 @@
...
+    search_queue: web::Data<SearchQueue>,
+    web::Json(chat_completion): web::Json<CreateChatCompletionRequest>,
+) -> impl Responder {
+    let ChatsParam { workspace_uid } = chats_param.into_inner();
+
+    assert_eq!(
+        chat_completion.n.unwrap_or(1),
+        1,
+        "Meilisearch /chat only support one completion at a time (n = 1, n = null)"
+    );
```
dureuill: Should we make this an error code or do we know that chat_completion.n will never be different from one in the input?

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,713 @@
...
+                req,
+                chat_completion,
+            )
+            .await,
+        )
+    }
+}
+
+#[derive(Default, Debug, Clone, Copy)]
+pub struct FunctionSupport {
```
dureuill, line 95: I wonder if it shouldn't be made part of the stream wrapper in the utils file wdyt

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,713 @@
...
+    index_scheduler: &Data<IndexScheduler>,
+    filters: &meilisearch_auth::AuthFilter,
+    chat_completion: &mut CreateChatCompletionRequest,
+    prompts: &DbChatCompletionPrompts,
+    system_role: SystemRole,
+) -> Result<FunctionSupport, ResponseError> {
+    let tools = chat_completion.tools.get_or_insert_default();
+    if tools.iter().any(|t| t.function.name == MEILI_SEARCH_IN_INDEX_FUNCTION_NAME) {
+        panic!("{MEILI_SEARCH_IN_INDEX_FUNCTION_NAME} function already set");
+    }
```
dureuill: should we return a 400 here?

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,713 @@
...
+    index_scheduler.try_for_each_index::<_, ()>(|name, index| {
+        // Make sure to skip unauthorized indexes
+        if !filters.is_index_authorized(name) {
+            return Ok(());
+        }
+
+        let rtxn = index.read_txn()?;
+        let chat_config = index.chat_config(&rtxn)?;
+        let index_description = chat_config.description;
+        let _ = writeln!(&mut function_description, "\n\n - {name}: {index_description}\n");
```
dureuill, line 169: that's a lot of \n, did you mean to use `writeln` or `write`?

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,713 @@
...
+                append_to_conversation = true;
+                false
+            }
+            _ => true, // keep other tools
+        }
+    });
+
+    let mut index_uids = Vec::new();
+    let mut function_description = prompts.search_description.clone();
+    index_scheduler.try_for_each_index::<_, ()>(|name, index| {
```
dureuill, line 160: this does not scale well to many indexes, as it will open all the indexes which can take multiple seconds (on macos) and will trash the cache (on all oses).

this might be hard to fix for now, but maybe we should configure a worskpace to only consider a subset of all indexes?

crates/meilisearch/src/search/mod.rs
Kerollmops: Done, thank you @dureuill.

crates/meilisearch/src/search/mod.rs
Kerollmops: Ok, I removed the default impl to fix the comment below 👇

crates/milli/src/update/chat.rs
Kerollmops: Yup, for a future implementation 👌

crates/milli/src/update/settings.rs
Kerollmops: Adding that to the followup tracking issue.

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,713 @@
...
+                    "type": "object",
+                    "properties": {
+                        "index_uid": {
+                            "type": "string",
+                            "enum": index_uids,
+                            "description": prompts.search_index_uid_param,
+                        },
+                        "q": {
+                            // Unfortunately, Mistral does not support an array of types, here.
+                            // "type": ["string", "null"],
```
dureuill, line 191: could we branch on the source type then?

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,713 @@
...
+        Data<IndexScheduler>,
+    >,
+    auth_ctrl: web::Data<AuthController>,
+    search_queue: &web::Data<SearchQueue>,
+    auth_token: &str,
+    index_uid: String,
+    q: Option<String>,
+) -> Result<(Index, Vec<Document>, String), ResponseError> {
+    // TBD
+    // let mut aggregate = SearchAggregator::<SearchPOST>::from_query(&query);
```
dureuill: TODO analytics?

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,713 @@
...
+    let mut documents = Vec::new();
+    if let Ok((ref rtxn, ref search_result)) = output {
+        // aggregate.succeed(search_result);
+        if search_result.degraded {
+            MEILISEARCH_DEGRADED_SEARCH_REQUESTS.inc();
+        }
+
+        let fields_ids_map = index.fields_ids_map(rtxn)?;
+        let displayed_fields = index.displayed_fields_ids(rtxn)?;
+        for &document_id in &search_result.documents_ids {
```
dureuill, line 288: for later: should we also use the ranking score as a mesure of confidence shared with the assistant?

crates/meilisearch/src/routes/chats/utils.rs
Kerollmops: This will not panic as we test that choices n == 1 above

crates/meilisearch/src/routes/chats/utils.rs
Kerollmops: This will not panic as we test that choices n == 1 above

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,721 @@
...
+    search_queue: &web::Data<SearchQueue>,
+    auth_token: &str,
+    client: &Client<C>,
+    chat_completion: &mut CreateChatCompletionRequest,
+    tx: &SseEventSender,
+    global_tool_calls: &mut HashMap<u32, Call>,
+    function_support: FunctionSupport,
+) -> Result<ControlFlow<Option<FinishReason>, ()>, SendError<Event>> {
+    let mut finish_reason = None;
+    // safety: The unwrap can only happen if the stream is not correctly configured.
```
dureuill:
```suggestion
    // unwrap: can only happens if `stream` was set to `false`
```

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,721 @@
...
+                match tool_calls {
+                    Some(tool_calls) => {
+                        for chunk in tool_calls {
+                            let ChatCompletionMessageToolCallChunk {
+                                index,
+                                id,
+                                r#type: _,
+                                function,
+                            } = chunk;
+                            let FunctionCallStream { name, arguments } = function.as_ref().unwrap();
```
dureuill, line 529: unwrap?

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,721 @@
...
+
+                            global_tool_calls
+                                .entry(*index)
+                                .and_modify(|call| {
+                                    if call.is_internal() {
+                                        call.append(arguments.as_ref().unwrap())
```
dureuill, line 535: unwrap?

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,721 @@
...
+                                            function_name: name.as_ref().unwrap().clone(),
+                                            arguments: arguments.as_ref().unwrap().clone(),
+                                        }
+                                    } else {
+                                        Call::External
+                                    }
+                                });
+
+                            if global_tool_calls.get(index).is_some_and(Call::is_external) {
+                                todo!("Support forwarding external tool calls");
```
dureuill: should be changed to an `unimplemented!` if cannot be ready for this PR

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,721 @@
...
+                                    .tool_calls(meili_calls.clone())
+                                    .build()
+                                    .unwrap()
+                                    .into(),
+                            );
+
+                            assert!(
+                                other_calls.is_empty(),
+                                "We do not support external tool forwarding for now"
+                            );
```
dureuill: user error?

dureuill: note that panics will only be reported in the logs, which are not visible by cloud users

crates/meilisearch/src/routes/chats/utils.rs
Kerollmops: I mean, yeah sure. We use the `format_documents` method only once by search request, soooo. Yup 👌

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,721 @@
...
+    External,
+}
+
+impl Call {
+    fn is_internal(&self) -> bool {
+        matches!(self, Call::Internal { .. })
+    }
+
+    fn is_external(&self) -> bool {
+        matches!(self, Call::External { .. })
```
dureuill: I had no idea you could refer to a unit variant with the braces

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,721 @@
...
+        };
+
+        let tool = ChatCompletionRequestMessage::Tool(ChatCompletionRequestToolMessage {
+            tool_call_id: call.id.clone(),
+            content: ChatCompletionRequestToolMessageContent::Text(answer),
+        });
+
+        if append_to_conversation {
+            tx.append_conversation_message(resp.clone(), &tool).await?;
+        }
```
dureuill, line 688: This part could be made easier to understand with comments outlining the *why*

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -0,0 +1,721 @@
...
+impl Call {
+    fn is_internal(&self) -> bool {
+        matches!(self, Call::Internal { .. })
+    }
+
+    fn is_external(&self) -> bool {
+        matches!(self, Call::External { .. })
+    }
+
+    fn append(&mut self, more: &str) {
```
dureuill, line 719:
```suggestion
    /// # Panics
    ///
    /// - if called on external calls
    fn append(&mut self, more: &str) {
```

crates/meilisearch/src/routes/chats/chat_completions.rs
Kerollmops: I returned an unimplement with a status code not implemented

crates/meilisearch/src/routes/chats/chat_completions.rs
Kerollmops: We cannot do that right now. Let's do that later.

crates/meilisearch/src/routes/chats/chat_completions.rs
Kerollmops: Yup, good idea. It's easy.

crates/meilisearch/src/routes/chats/chat_completions.rs
Kerollmops: I'll do the analytics later

crates/meilisearch/src/routes/chats/chat_completions.rs
Kerollmops: I moved the function verification above and it is now synchronous

crates/meilisearch/src/routes/chats/chat_completions.rs
Kerollmops: We could. We should use [the `byot` feature in the future](https://github.com/64bit/async-openai/pull/368#issuecomment-2952916801).

dureuill (approved):
thank you for this huge feature

dureuill (approved):

crates/meilisearch/src/routes/chats/chat_completions.rs
Kerollmops: We could also let the index_scheduler check the API Key indexes list to ensure it doesn't try to open any that are not allowed to be opened.

Issue #5547, opened 2025-05-09
`MeiliSearchApiError: task 123 panicked with message "assertion failed: self.stable_store.len() < u16::MAX as usize"`


underflow00:
**Describe the bug**
Search operation throws error
```
MeiliSearchApiError: task 123 panicked with message "assertion failed: self.stable_store.len() < u16::MAX as usize"
```

**To Reproduce**

Run Node.js script (as per https://www.npmjs.com/package/meilisearch):
```
import MeiliSearch from "meilisearch";

const NUM_DOCS = 65_535;
const manyDocs = [];
for (let i = 0; i < NUM_DOCS; i++) {
    manyDocs.push({
        'id': `doc${i}`,
        'title': `title${i}`,
    });
}

(async () => {
    const client = new MeiliSearch({
        host: 'http://localhost:7700',
        apiKey: 'masterKey',
    })
    let task;

    await client.deleteIndexIfExists('movies')
    const index = client.index('movies')

    task = await index.addDocuments(manyDocs)
    await client.tasks.waitForTask(task.taskUid);

    await index.search('title')
})()
```

**Expected behavior**
No error

**Screenshots**

**Meilisearch version:**

Node.js package: "0.50.0"

Server:
```
Config file path:       "none"
Database path:          "./data.ms"
Server listening on:    "http://localhost:7700"
Environment:            "development"
Commit SHA:             "94b43001dbfaf9fd8db3ef446e7f5fc67e4b0f8d"
Commit date:            "2025-04-03T15:46:46Z"
Package version:        "1.14.0"
```

**Additional context**

OS: Windows x64

dureuill:
Hello,

I'm sorry you encountered an internal error with Meilisearch. Thank you for the reproducer, I will try it on my end and report back.

dureuill:
Found the issue:

In [`find_one_typo_derivation`](https://github.com/meilisearch/meilisearch/blob/v1.14.0/crates/milli/src/search/new/query_term/compute_derivations.rs#L96), we are unconditionally adding the derived word to the interner, without correctly bounding the number of steps we will take. We only exit the loop [after finding `MAX_ONE_TYPO_COUNT` single typoes](https://github.com/meilisearch/meilisearch/blob/v1.14.0/crates/milli/src/search/new/query_term/compute_derivations.rs#L285), which is not guaranteed to happen: here, we have only 10 one-typo results, so we will iterate over the whole FST that contains `usize::MAX` terms.

[...]

PR #5551, opened 2025-05-12
Only intern in case of single-typo when looking for single typoes

dureuill:
# Pull Request

## Related issue
Fixes #5547 

## What does this PR do?
- When looking for single typoes, do not intern strings before knowing for sure that they correspond to a single-typo.
- Without this check, it was possible to build an index and a search query such that an arbitrary number of strings would be interned, causing a crash.
- Add test failing on v1.14.0, but not after this PR.

### Test output before this PR

```
thread 'tokio-runtime-worker' panicked at crates/milli/src/search/new/interner.rs:60:13:
assertion failed: self.stable_store.len() < u16::MAX as usize
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'search::bug_5547' panicked at crates/meilisearch/tests/search/mod.rs:131:5:
assertion `left == right` failed
  left: 500
 right: 200
```

PR #5564, opened 2025-05-14
Port to v1.15: Only intern in case of single-typo when looking for single typoes

dureuill:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/5547

Cf #5551


irevoire (approved):
👌

PR #5565, opened 2025-05-14
Fix 0 batched task

### LINKED ISSUE #5562: No tasks should be processed when the max number of batched tasks is set to 0
**Describe the bug**
No tasks should be processed when the max number of batched tasks is set to 0.
Currently, all the tasks that are not related to a single index are still processed as usual.

irevoire:
# Pull Request

## Related issue
Fixes #5562 

## What does this PR do?
- fix the bug
- add a test


ManyTheFish (approved):
LGTM

crates/index-scheduler/src/lib.rs
```diff
@@ -381,6 +381,11 @@ impl IndexScheduler {
     /// This function will execute in a different thread and must be called
     /// only once per index scheduler.
     fn run(&self) {
+        // If the number of batched tasks is 0, we don't need to run the scheduler at all.
```
dureuill, line 384: I'm worried that other parts of the code might assume some of the initialization here was done, so maybe we should not put the change so high up in the logic?

Alternatively we could put that check inside of `create_batch` or even `tick` (returning TickOutcome::NeverCallAgain`)

dureuill (commented):
Clippy needs fixing

crates/index-scheduler/src/lib.rs
irevoire: That's what I wanted to do initially, but then I realized it would have caused a bug because if the task queue it too full, the very first lines of the tick function are going to batch a new deletion task.

> so maybe we should not put the change so high up in the logic?

So, after seeing this issue, I thought the opposite, basically. 
I could send the `NeverCallAgain`, but it exits the loop as well, and if the only additional thing we do is to exit the tick function, I don't think it's really worth the additional complexity 🤔

dureuill: alright then, I trust your analysis :+1:

dureuill:
new test needs fixing

dureuill (approved):
thank you

PR #4528, opened 2024-03-22
allow to listen on unix domain sockets

kristofvandam:
# Pull Request

## What does this PR do?
- adds the ability to set a Unix domein socket as host_addr
- show the protocol used on startup

## reason
It's easier to create multiple instances and create (nginx for example) vhosts for them

## PR checklist
Please check if your PR fulfills the following requirements:
- [ x ] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [ x ] Have you read the contributing guidelines?
- [ x ] Have you made sure that the title is accurate and descriptive of the changes?

Thank you so much for contributing to Meilisearch!

PR #5571, opened 2025-05-15
feat: Add support for Unix Domain Socket connection

martin-g:
This PR is built on top of https://github.com/meilisearch/meilisearch/pull/4528 
It updates it to current `main` branch and adds a test case (!!! that starts meilisearch http server !!!).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -80,7 +80,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.101",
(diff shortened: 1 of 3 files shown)


irevoire:
Hey @martin-g, since pretty much no one is asking for it and there is not much interest in this PR for us either, I think we should just close it for now.
Maybe we'll revisit one day if people ask for it, but it's not planned anytime soon, sorry 😖

mantyr:
It's a shame. I'd like to use this in a Mac OS X APK build, and I really don't want to have to open network connections instead of a socket for it.

EkoNesLeg:
I'd very much like a Unix socket option. I run a number services in their own network spaces and Unix sockets make it easier and more direct to transverse them. Not to mention the potential performance benefits this PR would bring.

PR #5578, opened 2025-05-19
perf: Faster index::get_index IT tests

martin-g:
# Pull Request

## Related issue
#4840

## What does this PR do?
- Use a shared server for all tests in get_index.rs

### DIFF
--- a/crates/meilisearch/tests/index/get_index.rs
+++ b/crates/meilisearch/tests/index/get_index.rs
@@ -43,7 +43,7 @@ async fn error_get_unexisting_index() {
 
 #[actix_rt::test]
 async fn no_index_return_empty_list() {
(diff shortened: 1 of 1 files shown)


irevoire (changes requested):
Hey thanks for the contribution again!

This file is pretty hard because it touches a lot of stuff that are global to the instance. We can't really apply the changes I want everywhere 😓 

I left a review explaining everything the best I can, let me know if something is not clear.
And keep in mind that when you run `cargo test`, cargo starts one threads per integration test and run them all at the exact same time.
That's why in the comment I'm saying that we can't check a property on the whole list of indexes, another test may be modifying it at the same time 😩

irevoire (approved):
Awesome, thank you!

PR #5584, opened 2025-05-20
tests: Faster index::search::mod IT tests

martin-g:
# Pull Request

## Related issue
#4840

## What does this PR do?
* Use shared index where possible.
* Call .succeeded/.failed when waiting for a task.
* Use newer format_args syntax
* Do not use fully qualified name for meili_snap:: functions. The functions are already imported in scope

### DIFF
--- a/crates/meilisearch/tests/search/mod.rs
+++ b/crates/meilisearch/tests/search/mod.rs
@@ -110,8 +110,8 @@ async fn simple_search() {
 /// See <https://github.com/meilisearch/meilisearch/issues/5547>
 #[actix_rt::test]
 async fn bug_5547() {
-    let server = Server::new().await;
-    let index = server.index("big_fst");
+    let server = Server::new_shared();
+    let index = server.unique_index();
     let (response, _code) = index.create(None).await;
     index.wait_task(response.uid()).await.succeeded();
 
@@ -130,8 +130,8 @@ async fn bug_5547() {
 #[actix_rt::test]
 async fn search_with_stop_word() {
     // related to https://github.com/meilisearch/meilisearch/issues/4984
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
 
     let (_, code) = index
         .update_settings(json!({"stopWords": ["the", "The", "a", "an", "to", "in", "of"]}))
@@ -195,8 +195,8 @@ async fn search_with_stop_word() {
 #[actix_rt::test]
 async fn search_with_typo_settings() {
     // related to https://github.com/meilisearch/meilisearch/issues/5240
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
 
     let (_, code) = index
         .update_settings(json!({"typoTolerance": { "disableOnAttributes": ["title", "id"]}}))
@@ -229,8 +229,8 @@ async fn search_with_typo_settings() {
 #[actix_rt::test]
 async fn phrase_search_with_stop_word() {
     // related to https://github.com/meilisearch/meilisearch/issues/3521
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
 
     let (_, code) = index.update_settings(json!({"stopWords": ["the", "of"]})).await;
     snapshot!(code, @"202 Accepted");
@@ -312,8 +312,8 @@ async fn non_negative_search() {
 
 #[actix_rt::test]
 async fn negative_special_cases_search() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
 
     let documents = DOCUMENTS.clone();
     let (task, _status_code) = index.add_documents(documents, None).await;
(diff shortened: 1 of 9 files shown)


crates/meilisearch/tests/search/mod.rs
```diff
@@ -688,7 +683,7 @@ async fn search_is_hard_limited() {
                 "limit": 1500,
             }),
             |response, code| {
```
irevoire, line 685: same

crates/meilisearch/tests/search/mod.rs
```diff
@@ -757,7 +752,7 @@ async fn faceting_max_values_per_facet() {
                 "facets": ["number"]
             }),
             |response, code| {
```
irevoire, line 754: Same

crates/meilisearch/tests/search/mod.rs
```diff
@@ -1075,8 +1052,8 @@ async fn test_degraded_score_details() {
                 "showRankingScoreDetails": true,
```
irevoire, line 1052: Same here

crates/meilisearch/tests/search/mod.rs
```diff
@@ -1162,8 +1139,8 @@ async fn camelcased_words() {
```
irevoire, line 1139: same here

crates/meilisearch/tests/search/mod.rs
```diff
@@ -1377,17 +1354,17 @@ async fn simple_search_with_strange_synonyms() {
```
irevoire, line 1354: Same

crates/meilisearch/tests/search/mod.rs
```diff
@@ -1462,8 +1439,8 @@ async fn change_attributes_settings() {
                 "attributesToRetrieve": ["id", "doggos"]
```
irevoire, line 1439: same here

crates/meilisearch/tests/search/mod.rs
```diff
@@ -1955,7 +1932,7 @@ async fn change_facet_casing() {
         }))
```
irevoire, line 1932: same here

crates/meilisearch/tests/search/mod.rs
```diff
@@ -39,55 +40,49 @@ async fn test_settings_documents_indexing_swapping_and_search(
     let index = server.index("test");
 
     let (task, code) = index.add_documents(documents.clone(), None).await;
```
irevoire, line 42: This function is called a tons of time. We can't use a shared index but we can use a shared `Server` for sure

irevoire (changes requested):
Hey, that was a strange file, half of the shared index/server were already done but not the other half 😓

crates/meilisearch/tests/search/mod.rs
```diff
@@ -141,7 +136,7 @@ async fn search_with_stop_word() {
     let (_, code) = index
```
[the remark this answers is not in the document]
martin-g: Right!
I didn't do it because of https://github.com/meilisearch/meilisearch/issues/4840#issuecomment-2891233068
But the settings should not be modified for the shared **index**, not the shared _server_!

crates/meilisearch/tests/search/mod.rs
```diff
@@ -604,7 +599,7 @@ async fn displayed_attributes() {
```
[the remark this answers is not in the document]
martin-g: I decided to use a unique Index for all tests that modify the settings.

crates/meilisearch/tests/search/mod.rs
martin-g: This leads to an assertion error at https://github.com/meilisearch/meilisearch/blob/8047cfe4389dcdbf78589818a80be4a5cdc95498/crates/meilisearch/tests/search/errors.rs#L732

The index name is a UUID.

irevoire: Aaaah, that's awful 😭 

Do you have an idea of how we could fix that?
I'm thinking that maybe we could make something work with that: https://insta.rs/docs/redactions/#dynamic-redactions to find and replace the uuids in the error fields.
Ideally, we should do that in the [`Display` implementation of our `Value`](https://github.com/meilisearch/MeiliSearch/blob/main/crates/meilisearch/tests/common/mod.rs#L118-L135), so it works everywhere by default, but that doesn't seem obvious.

irevoire: Hey, I agree that would be super nice, but no, we clearly cannot do that because users do have indexes named after UUID and they want to use them 😬

We should really try to shove that logic only in the tests and not in the actual code.
So, ideally, with insta redactions. 

Worst worst case scenario would be to keep your function but put a big cfg(test) on top of the if. But I'm not a fan, as it makes the actual source code harder to understand 😩

crates/meilisearch/tests/search/mod.rs
```diff
@@ -135,13 +121,13 @@ async fn bug_5547() {
...
+    let index = server.unique_index();
 
     let (_, code) = index
         .update_settings(json!({"stopWords": ["the", "The", "a", "an", "to", "in", "of"]}))
         .await;
-    meili_snap::snapshot!(code, @"202 Accepted");
+    snapshot!(code, @"202 Accepted");
 
     let documents = DOCUMENTS.clone();
     index.add_documents(documents, None).await;
```
irevoire, line 133: This test is flaky and blocking the CI because right below this comment we're still waiting for the task `1` on a shared instance.
Instead, we should retrieve the actual id of the task 👍

irevoire (changes requested):
Hey I think we have one last little bug, and then we'll be good to merge this huge PR.
Thanks a lot for your work, that was probably one of the hardest file 🦸

irevoire (approved):
Hey, awesome, thanks for your work 💪 
Merging now

PR #5588, opened 2025-05-22
Rename batch creation complete

dureuill:
- Rename `batchCreationComplete` to `batchStrategy`
- Adjust messages to fit the new name (*and* the messages are more descriptive :tada:)
- Update snapshots

### No db changes

- as only the displayed batch object is modified, the database format doesn't change at all in this PR
- the values of the stop reason change in this PR compared with v1.15.0-rc.{0-2}, but this is compatible in both directions


irevoire (approved):
Perfecto, thank you! 🌵

PR #5592, opened 2025-05-26
Decouple geo facet extraction from rest of document

### LINKED ISSUE #5440: Separate Geo facet extraction from the standard one
During the facet extraction, there are 3 cases:
- when [the document is added](https://github.com/meilisearch/meilisearch/blob/7df2bdfb153b68e50ae65a7d7d1d12554a8aabaf/crates/milli/src/update/new/extract/faceted/extract_facets.rs#L192-L215), we extract the data from the new document
- when [the document is deleted](https://github.com/meilisearch/meilisearch/blob/7df2bdfb153b68e50ae65a7d7d1d12554a8aabaf/crates/milli/src/update/new/extract/faceted/extract_facets.rs#L94-L118), we extract the data from the deleted document
- when [the document is modified](https://github.com/meilisearch/meilisearch/blob/7df2bdfb153b68e50ae65a7d7d1d12554a8aabaf/crates/milli/src/update/new/extract/faceted/extract_facets.rs#L119-L191), we extract the data from both versions of the document starting by the deleted version

The extraction process called in these 3 cases will always:
1) [extract the standard facets by iterating over the fields of the document](https://github.com/meilisearch/meilisearch/blob/7df2bdfb153b68e50ae65a7d7d1d12554a8aabaf/crates/milli/src/update/new/extract/faceted/facet_document.rs#L51-L103)
2) [extract the geo data from the document](https://github.com/meilisearch/meilisearch/blob/7df2bdfb153b68e50ae65a7d7d1d12554a8aabaf/crates/milli/src/update/new/extract/faceted/facet_document.rs#L105-L117)

However, when a document is modified, the user may have modified only a part of the document, such as the geo fields.
There are already several checks[ made when a document is modified](https://github.com/meilisearch/meilisearch/blob/7df2bdfb153b68e50ae65a7d7d1d12554a8aabaf/crates/milli/src/update/new/extract/faceted/extract_facets.rs#L120-L138), but they are not used independently to process only the geo extraction or the facet extraction. It would be better to isolate the [geo data extraction](https://github.com/meilisearch/meilisearch/blob/7df2bdfb153b68e50ae65a7d7d1d12554a8aabaf/crates/milli/src/update/new/extract/faceted/facet_document.rs#L105-L117) in a function and call it only if it's needed.
The geo extraction should always occur for a deletion or an addition if the geo feature is activated.

nnethercott:
# Pull Request

## Related issue
Fixes #5440 

## What does this PR do?
- introduces new function `extract_geo_document` for _geo facet extraction & removes that logic from `extract_document_facets` 
- adds a new `macro_rules!` macro to generate `add` and `del` facet functions, limiting code duplication 

Note: Since each variant of the macro uses a &mut ref, we can't just define top level helpers to use throughout the match block. Ideally a factory function would be used in place of this macro but since closures have anonymous types we run into issues defining the signature of the factory, unless we wanna toss in `Box<dyn ...>`. I think the macro is a nice trade-off here

## PR checklist

### DIFF
--- a/crates/milli/src/update/new/extract/faceted/extract_facets.rs
+++ b/crates/milli/src/update/new/extract/faceted/extract_facets.rs
@@ -153,7 +153,7 @@ impl FacetedDocidsExtractor {
                 }
             }
             DocumentChange::Update(inner) => {
-                let has_changed = inner.has_changed_for_fields(
+                let has_changed_for_facets = inner.has_changed_for_fields(
                     &mut |field_name| {
                         match_faceted_field(
                             field_name,
@@ -171,7 +171,7 @@ impl FacetedDocidsExtractor {
                     inner.has_changed_for_geo_fields(rtxn, index, context.db_fields_ids_map)?;
 
                 // 1. Maybe update doc
-                if has_changed {
+                if has_changed_for_facets {
                     extract_document_facets(
                         inner.current(rtxn, index, context.db_fields_ids_map)?,
                         new_fields_ids_map.deref_mut(),
--- a/crates/milli/src/update/new/extract/faceted/extract_facets.rs
+++ b/crates/milli/src/update/new/extract/faceted/extract_facets.rs
@@ -90,8 +90,48 @@ impl FacetedDocidsExtractor {
         let mut cached_sorter = context.data.borrow_mut_or_yield();
         let mut del_add_facet_value = DelAddFacetValue::new(&context.doc_alloc);
         let docid = document_change.docid();
+
+        macro_rules! facet_fn_factory {
+            (del) => {
+                |fid: FieldId, meta: Metadata, depth: perm_json_p::Depth, value: &Value| {
+                    Self::facet_fn_with_options(
+                        &context.doc_alloc,
+                        cached_sorter.deref_mut(),
+                        BalancedCaches::insert_del_u32,
+                        &mut del_add_facet_value,
+                        DelAddFacetValue::insert_del,
+                        docid,
+                        fid,
+                        meta,
+                        filterable_attributes,
+                        depth,
+                        value,
+                    )
+                }
+            };
+            (add) => {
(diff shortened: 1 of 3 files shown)


crates/milli/src/update/new/extract/faceted/extract_facets.rs
```diff
@@ -133,91 +169,72 @@ impl FacetedDocidsExtractor {
...
+                        filterable_attributes,
+                        sortable_fields,
+                        asc_desc_fields,
+                        distinct_field,
+                        &mut facet_fn!(add),
+                    )?;
                 }
 
+                // 2. Maybe update geo
+                if is_geo_enabled && has_changed_for_geo_fields {
```
ManyTheFish, line 197:
```suggestion
                if is_geo_enabled && inner.has_changed_for_geo_fields(rtxn, index, context.db_fields_ids_map)? {
```

crates/milli/src/update/new/extract/faceted/extract_facets.rs
```diff
@@ -90,17 +90,11 @@ impl FacetedDocidsExtractor {
...
-                new_fields_ids_map.deref_mut(),
-                filterable_attributes,
-                sortable_fields,
-                asc_desc_fields,
-                distinct_field,
-                is_geo_enabled,
-                &mut |fid, meta, depth, value| {
+
+        // Macro expanding to an insertion/deletion facet fn
+        macro_rules! facet_fn {
```
ManyTheFish, line 95: Can you explain why the macro is necessary here, please?

```suggestion
        // Macro expanding to an insertion/deletion facet fn,
        // using a macro avoid to borrow the parameters as mutable in both closures at the same time by postponing their creation 
        macro_rules! facet_fn {
```

ManyTheFish (commented):
hey @nnethercott,

Nice PR, I wrote some change requests ,

thank you for contributing to Meilisearch!

PR #5593, opened 2025-05-27
Remove TemplateChecker

dureuill:
- Remove `template_checker` module and type as the check are both missing some wrong variable uses and, more importantly, reporting legit variable use as wrong (anytime a document field is used in a filter that expects an array)
- Mark as ignored the tests that relied on the checker: they won't pass without the checker, but the hope is that we can enable them again if https://github.com/cobalt-org/liquid-rust/issues/588 is implemented
- Similarly allow the now unused error code and error factory function


ManyTheFish (approved):
LGTM

PR #5600, opened 2025-05-29
tests: Faster search::facet_search IT tests

martin-g:
# Pull Request

## Related issue
#4840

## What does this PR do?
- Use shared server + unique indices where possible. 
- Assert .succeeded() for the waited tasks.
- Drop usage of dbg!() in the assertions. It caused noise in the logs

### DIFF
--- a/crates/meilisearch/tests/search/facet_search.rs
+++ b/crates/meilisearch/tests/search/facet_search.rs
@@ -50,13 +50,11 @@ async fn test_settings_documents_indexing_swapping_and_facet_search(
 
     let (task, code) = index.add_documents(documents.clone(), None).await;
     assert_eq!(code, 202, "{}", task);
-    let response = index.wait_task(task.uid()).await.succeeded();
-    assert!(response.is_success(), "{response:?}");
+    index.wait_task(task.uid()).await.succeeded();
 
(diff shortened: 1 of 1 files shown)


crates/meilisearch/tests/search/facet_search.rs
```diff
@@ -50,13 +50,13 @@ async fn test_settings_documents_indexing_swapping_and_facet_search(
 
     let (task, code) = index.add_documents(documents.clone(), None).await;
     assert_eq!(code, 202, "{}", task);
-    let response = index.wait_task(task.uid()).await;
-    assert!(response.is_success(), "{:?}", response);
+    let response = index.wait_task(task.uid()).await.succeeded();
+    assert!(response.is_success(), "{response:?}");
```
ManyTheFish: both lines are checking the same thing, see the [succeeded method implementation](https://github.com/meilisearch/meilisearch/blob/d416b3b3905ebe684e0aabc6a8c5343ad4aa5df7/crates/meilisearch/tests/common/mod.rs#L53-L58).

```suggestion
    let response = index.wait_task(task.uid()).await.succeeded();
```

Same everywhere in the file

ManyTheFish (changes requested):

PR #5603, opened 2025-05-29
tests: Faster search::multi IT tests

martin-g:
# Pull Request

## Related issue
#4840

## What does this PR do?
- Use shared server + unique indices where possible

### DIFF
--- a/crates/meilisearch/tests/search/multi/mod.rs
+++ b/crates/meilisearch/tests/search/multi/mod.rs
@@ -855,10 +855,6 @@ async fn search_multiple_indexes_dont_exist() {
         ]}))
         .await;
     snapshot!(code, @"400 Bad Request");
-
-    let (list_response, _code) = server.list_indexes(Some(0), Some(1_000_000)).await;
-    dbg!(list_response);
-
     snapshot!(json_string!(response), @r###"
     {
       "message": "Inside `.queries[0]`: Index `test` not found.",


irevoire:
>  One thing that bothers me a bit is that now the snapshots use [uuid] for all index names and it is not very clear which index brings the hit, e.g. see federation_inconsistent_merge_order() test case.

Hey yeah, you're right, that's really hard to read 😩 

In the beginning, you are using a lot of shared indices, and since these ones have a fixed name, I find the test pretty good to read.
From what I see, almost all the tests rely on like five different datasets (and some are already available as shared index). Do you think it would make sense to make all of these datasets into a unique, named and shared index for all test?

And if a test must update the settings or something, then we don't use the shared server?
It's not ideal, but I feel like it would fix the issue for 99% of the tests while keeping the test very easy to read (maybe even more than before since we won't have to repeat all the initialization work)

martin-g:
@irevoire https://github.com/meilisearch/meilisearch/pull/5603/commits/1824fbd1b513c1b00f68f45cebdcd2c6f4252fbe introduces `Index::unique_index_with_prefix(&str)`
Using it produces index names like `movies-[uuid]` in the response JSON. This way the index is both unique and readable.

irevoire:
Oh yes that's easy and smart, I love it 🔥

martin-g:
@irevoire I think this big boy is ready for review!
You may make yourself some🍿 ! 😄

martin-g:
One of the tests fails due to racing:
```
---- search::multi::search_multiple_indexes_dont_exist stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: search_multiple_indexes_dont_exist
Source: crates/meilisearch/tests/search/multi/mod.rs:794
────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────
    0     0 │ {
    1       │-  "message": "Inside `.queries[0]`: Index `test` not found.",
          1 │+  "message": "Inside `.queries[1]`: Index `nested` not found.",
    2     2 │   "code": "index_not_found",
    3     3 │   "type": "invalid_request",
    4     4 │   "link": "https://docs.meilisearch.com/errors#index_not_found"
    5     5 │ }
```
Both `test` and `nested` indices do not exist in the shared server, but sometimes the error is about the first index (`test`) and sometimes for the second one (`nested`)...

**Update**: It looks like the `test` index is actually existing in the shared server. Using unique index names consistently returns errors with the first index name in the details.

crates/meilisearch/tests/search/multi/mod.rs
```diff
@@ -414,29 +382,22 @@ async fn federation_search_illegal_index_uid() {
...
 
-    let documents = DOCUMENTS.clone();
-    let (task, _status_code) = index.add_documents(documents, None).await;
-    index.wait_task(task.uid()).await.succeeded();
-
-    let index = server.index("nested");
-    let documents = NESTED_DOCUMENTS.clone();
-    let (add_task, _status_code) = index.add_documents(documents, None).await;
-    index.wait_task(add_task.uid()).await.succeeded();
+    let nested_index = shared_index_with_nested_documents().await;
```
irevoire: That's what I like to see 😁

crates/meilisearch/tests/search/multi/mod.rs
```diff
@@ -848,18 +783,21 @@ async fn federation_one_index_doesnt_exist() {
 
 #[actix_rt::test]
 async fn search_multiple_indexes_dont_exist() {
-    let server = Server::new().await;
+    let server = Server::new_shared();
+
+    let index_1 = format!("index_1-{}", Uuid::new_v4());
+    let index_2 = format!("index_2-{}", Uuid::new_v4());
```
irevoire: Why are we not using the `unique_with_prefix` as you suggested?
It seems "safer" to use

crates/meilisearch/tests/search/multi/mod.rs
```diff
@@ -956,29 +880,20 @@ async fn federation_one_query_error() {
...
+    let server = Server::new_shared();
+    let index = shared_index_with_documents().await;
+    let nested_index = shared_index_with_nested_documents().await;
 
     let (response, code) = server
         .multi_search(json!({"federation": {}, "queries": [
-        {"indexUid" : "test", "q": "glass"},
-        {"indexUid": "nested", "q": "pésti", "sort": ["doggos:desc"]},
+        {"indexUid" : index.uid, "q": "glass"},
+        {"indexUid": nested_index.uid, "q": "pésti", "sort": ["mother:desc"]},
```
irevoire, line 950: good catch

crates/meilisearch/tests/search/multi/mod.rs
```diff
@@ -1707,9 +1573,8 @@ async fn federation_sort_same_indexes_different_criterion_same_direction() {
 
 #[actix_rt::test]
 async fn federation_sort_same_indexes_different_criterion_opposite_direction() {
-    let server = Server::new().await;
-
-    let index = server.index("nested");
+    let server = Server::new_shared();
+    let index = server.unique_index_with_prefix("nested");
```
irevoire, line 1637: noice

crates/meilisearch/tests/search/multi/mod.rs
```diff
@@ -2058,15 +1922,15 @@ async fn federation_sort_different_ranking_rules() {
...
+    let batman_index = server.unique_index_with_prefix("batman");
 
     let documents = SCORE_DOCUMENTS.clone();
-    let (value, _) = index.add_documents(documents, None).await;
-    index.wait_task(value.uid()).await.succeeded();
+    let (value, _) = batman_index.add_documents(documents, None).await;
+    batman_index.wait_task(value.uid()).await.succeeded();
 
-    let (value, _) = index
+    let (value, _) = batman_index
```
irevoire: From what I see here we're using the default ranking rules and could use the shared index

https://www.meilisearch.com/docs/learn/relevancy/ranking_rules#list-of-built-in-ranking-rules

crates/meilisearch/tests/search/multi/mod.rs
```diff
@@ -2241,15 +2105,15 @@ async fn federation_sort_different_ranking_rules() {
...
+    let movies_index = server.unique_index_with_prefix("movies");
 
     let documents = DOCUMENTS.clone();
-    let (value, _) = index.add_documents(documents, None).await;
-    index.wait_task(value.uid()).await.succeeded();
+    let (value, _) = movies_index.add_documents(documents, None).await;
+    movies_index.wait_task(value.uid()).await.succeeded();
 
-    let (value, _) = index
+    let (value, _) = movies_index
```
irevoire: Same question

crates/meilisearch/tests/search/multi/mod.rs
```diff
@@ -2262,15 +2126,15 @@ async fn federation_sort_different_indexes_same_criterion_opposite_direction() {
...
+    let batman_index = server.unique_index_with_prefix("batman");
 
     let documents = SCORE_DOCUMENTS.clone();
-    let (value, _) = index.add_documents(documents, None).await;
-    index.wait_task(value.uid()).await.succeeded();
+    let (value, _) = batman_index.add_documents(documents, None).await;
+    batman_index.wait_task(value.uid()).await.succeeded();
 
-    let (value, _) = index
+    let (value, _) = batman_index
```
irevoire: Same question

irevoire (changes requested):
Hey!

I saw this chunk of code literally dozens of times I think:
```rust

    let (value, _) = movies_index.add_documents(documents, None).await;
    movies_index.wait_task(value.uid()).await.succeeded();

    let (value, _) = movies_index
        .update_settings(json!({
          "sortableAttributes": ["title"],
          "filterableAttributes": ["title", "color"],
          "rankingRules": [
            "sort",
            "words",
            "typo",
            "proximity",
            "attribute",
            "exactness"
          ]
        }))
        .await;
    movies_index.wait_task(value.uid()).await.succeeded();

    let batman_index = server.unique_index_with_prefix("batman");

    let documents = SCORE_DOCUMENTS.clone();
    let (value, _) = batman_index.add_documents(documents, None).await;
    batman_index.wait_task(value.uid()).await.succeeded();

    let (value, _) = batman_index
        .update_settings(json!({
          "sortableAttributes": ["title"],
          "filterableAttributes": ["title"],
          "rankingRules": [
            "sort",
            "words",
            "typo",
            "proximity",
            "attribute",
            "exactness"
          ]
        }))
        .await;
    batman_index.wait_task(value.uid()).await.succeeded();
```

I think it would make sense to create two new shared index for these two datasets 🤔 

Since playing with the ranking rules seems very specific to the multisearch and federation I think they could be local to this file though (at the top would be the best)

Otherwise it looks pretty good! I'll take a look at your issue right now

irevoire:
> Update: It looks like the test index is actually existing in the shared server. Using unique index names consistently returns errors with the first index name in the details.

Ah I see you found the bug while I was reviewing your PR.
But oooof that's a big bug we're never supposed to be able to create a non-unique index with the shared index and that could come from any test 🤔 
Maybe a quick way to find the issue would be to update the index creation call to crash when it's called with an index name that contains `test`?

crates/meilisearch/tests/search/multi/mod.rs
martin-g: Initially I used `server.unique_index_with_prefix("index_1")` but then I simplified it to plain strings because `server.unique_index()` looks like it is creating the index. Using a plain string makes it more obvious that this is just a name.
I don't mind changing it back to `server.unique_index_with_prefix()` if you think it would be better!

crates/meilisearch/tests/search/multi/mod.rs
martin-g: Here the order of the ranking rules is not the same as the default ones.
Here:
```
 "rankingRules": [
            "sort",
            "words",
            "typo",
            "proximity",
            "attribute",
            "exactness"
          ]
```

Defaults:
```
[
  "words",
  "typo",
  "proximity",
  "attribute",
  "sort",
  "exactness"
]
```

crates/meilisearch/tests/search/multi/mod.rs
martin-g: Here the order of the ranking rules is not the same as the default ones.
Here:
```
 "rankingRules": [
            "sort",
            "words",
            "typo",
            "proximity",
            "attribute",
            "exactness"
          ]
```

Defaults:
```
[
  "words",
  "typo",
  "proximity",
  "attribute",
  "sort",
  "exactness"
]
```

crates/meilisearch/tests/search/multi/mod.rs
martin-g: Done!

martin-g:
> But oooof that's a big bug we're never supposed to be able to create a non-unique index with the shared index and that could come from any test

I just tried to list all indexes in the shared server and confirm my assumption above with:
```
let (response, _code) = server.list_indexes(Some(0), Some(1_000_000)).await;
    dbg!(response);
```

but it returns:
```
       START             meilisearch::integration search::multi::search_multiple_indexes_dont_exist

running 1 test
[crates/meilisearch/tests/search/multi/mod.rs:792:5] response = Value(
    Object {
        "results": Array [],
        "offset": Number(0),
        "limit": Number(1000000),
        "total": Number(0),
    },
)
```

The shared server says "I have no indexes"!

And this is expected since every IT test setups its own Actix-Web App!
So my explanation [above](https://github.com/meilisearch/meilisearch/pull/5603#issuecomment-2961407945) is not correct!
It should be a racing issue in the backend ...

martin-g:
I was using `cargo nextest` and this caused the problem with the empty list of indexes!
Using `cargo test` shows a list with many entries but all of the items are either unique or shared. All looks good!

martin-g:
AFAIS the queries are processed in the order from the request body JSON and there is no racing - https://github.com/meilisearch/meilisearch/blob/aefebdeb8baf85fcba538b04b010fc5d676836a4/crates/meilisearch/src/routes/multi_search.rs#L214-L240
Mistery!

martin-g:
@irevoire Here is the `"test"` in a shared server - https://github.com/meilisearch/meilisearch/blob/c3368e6859bef4b39621b7d4bf81f8fa215d6e34/crates/meilisearch/tests/index/create_index.rs#L50
Noice! :-)

irevoire:
>  Noice! :-)

Oof good catch!

That's why I feels like even if we're not creating the indexes we should always just use the method that makes them unique ahah

PR #5607, opened 2025-05-30
tests: Faster settings::get_settings IT tests

martin-g:
# Pull Request

## Related issue
#4840

## What does this PR do?

### DIFF
--- a/crates/meilisearch/tests/settings/get_settings.rs
+++ b/crates/meilisearch/tests/settings/get_settings.rs
@@ -11,59 +11,62 @@ macro_rules! test_setting_routes {
 
                 #[actix_rt::test]
                 async fn get_unexisting_index() {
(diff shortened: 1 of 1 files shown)


crates/meilisearch/tests/settings/get_settings.rs
```diff
@@ -413,7 +414,7 @@ async fn reset_all_settings() {
 
     let (response, code) = index.add_documents(documents, None).await;
     assert_eq!(code, 202);
-    assert_eq!(response["taskUid"], 0);
+    assert!(response["taskUid"].as_i64().unwrap() >= 0);
```
ManyTheFish: This line should be removed
```suggestion
```

ManyTheFish (changes requested):

PR #5615, opened 2025-06-02
tests: Faster tasks::mod IT tests

martin-g:
# Pull Request

## Related issue
#4840

## What does this PR do?

### DIFF
--- a/crates/meilisearch/tests/tasks/mod.rs
+++ b/crates/meilisearch/tests/tasks/mod.rs
@@ -49,8 +49,9 @@ async fn get_task_status() {
 
 #[actix_rt::test]
 async fn list_tasks() {
-    let server = Server::new_shared();
-    let index = server.unique_index();
+    // Do not use a shared server because we want to assert stuff against the global list of tasks
+    let server = Server::new().await;
+    let index = server.index("test");
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
     index
--- a/crates/meilisearch/tests/tasks/mod.rs
+++ b/crates/meilisearch/tests/tasks/mod.rs
@@ -175,8 +175,9 @@ async fn list_tasks_status_filtered() {
 
 #[actix_rt::test]
 async fn list_tasks_type_filtered() {
-    let server = Server::new_shared();
-    let index = server.unique_index();
+    // Do not use a shared server because we want to assert stuff against the global list of tasks
+    let server = Server::new().await;
+    let index = server.index("test");
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
     index
--- a/crates/meilisearch/tests/tasks/mod.rs
+++ b/crates/meilisearch/tests/tasks/mod.rs
@@ -12,8 +12,6 @@ use crate::json;
 async fn error_get_unexisting_task_status() {
     let server = Server::new_shared();
     let index = server.unique_index();
-    let (task, _status_code) = index.create(None).await;
-    index.wait_task(task.uid()).await.succeeded();
     let (response, code) = index.get_task(u32::MAX as u64).await;
 
     let expected_response = json!({
--- a/crates/meilisearch/tests/tasks/mod.rs
+++ b/crates/meilisearch/tests/tasks/mod.rs
@@ -1,8 +1,8 @@
 mod errors;
 mod webhook;
 
-use meili_snap::insta::assert_json_snapshot;
-use meili_snap::snapshot;
+use insta::assert_json_snapshot;
+use meili_snap::{json_string, snapshot};
 use time::format_description::well_known::Rfc3339;
 use time::OffsetDateTime;
 
@@ -11,11 +11,11 @@ use crate::json;
 
 #[actix_rt::test]
 async fn error_get_unexisting_task_status() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
-    let (response, code) = index.get_task(1).await;
+    let (response, code) = index.get_task(u32::MAX as u64).await;
 
     let expected_response = json!({
         "message": "Task `1` not found.",
@@ -30,8 +30,8 @@ async fn error_get_unexisting_task_status() {
 
 #[actix_rt::test]
 async fn get_task_status() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
     let (create_task, _status_code) = index.create(None).await;
     let (add_task, _status_code) = index
         .add_documents(
@@ -50,8 +50,8 @@ async fn get_task_status() {
 
 #[actix_rt::test]
 async fn list_tasks() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
     index
@@ -64,14 +64,14 @@ async fn list_tasks() {
 
 #[actix_rt::test]
 async fn list_tasks_pagination_and_reverse() {
-    let server = Server::new().await;
+    let server = Server::new_shared();
     // First of all we want to create a lot of tasks very quickly. The fastest way is to delete a lot of unexisting indexes
     let mut last_task = None;
-    for i in 0..10 {
-        let index = server.index(format!("test-{i}"));
+    for _ in 0..10 {
+        let index = server.unique_index();
         last_task = Some(index.create(None).await.0.uid());
     }
-    server.wait_task(last_task.unwrap()).await;
+    server.wait_task(last_task.unwrap()).await.succeeded();
 
     let (response, code) = server.tasks_filter("limit=3").await;
     assert_eq!(code, 200);
@@ -102,14 +102,14 @@ async fn list_tasks_pagination_and_reverse() {
 
 #[actix_rt::test]
 async fn list_tasks_with_star_filters() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
     let (task, _code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
     index
         .add_documents(serde_json::from_str(include_str!("../assets/test_set.json")).unwrap(), None)
         .await;
-    let (response, code) = index.service.get("/tasks?indexUids=test").await;
+    let (response, code) = index.service.get(format!("/tasks?indexUids={}", index.uid)).await;
     assert_eq!(code, 200);
     assert_eq!(response["results"].as_array().unwrap().len(), 2);
 
@@ -127,50 +127,50 @@ async fn list_tasks_with_star_filters() {
 
     let (response, code) =
         index.service.get("/tasks?types=*,documentAdditionOrUpdate&statuses=*").await;
-    assert_eq!(code, 200, "{:?}", response);
+    assert_eq!(code, 200, "{response:?}");
     assert_eq!(response["results"].as_array().unwrap().len(), 2);
 
     let (response, code) = index
         .service
-        .get("/tasks?types=*,documentAdditionOrUpdate&statuses=*,failed&indexUids=test")
+        .get(format!("/tasks?types=*,documentAdditionOrUpdate&statuses=*,failed&indexUids={}", index.uid))
         .await;
-    assert_eq!(code, 200, "{:?}", response);
+    assert_eq!(code, 200, "{response:?}");
     assert_eq!(response["results"].as_array().unwrap().len(), 2);
 
     let (response, code) = index
         .service
         .get("/tasks?types=*,documentAdditionOrUpdate&statuses=*,failed&indexUids=test,*")
         .await;
-    assert_eq!(code, 200, "{:?}", response);
+    assert_eq!(code, 200, "{response:?}");
     assert_eq!(response["results"].as_array().unwrap().len(), 2);
 }
 
 #[actix_rt::test]
 async fn list_tasks_status_filtered() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.failed();
 
     let (response, code) = index.filtered_tasks(&[], &["succeeded"], &[]).await;
-    assert_eq!(code, 200, "{}", response);
+    assert_eq!(code, 200, "{response}");
     assert_eq!(response["results"].as_array().unwrap().len(), 1);
 
     let (response, code) = index.filtered_tasks(&[], &["succeeded"], &[]).await;
-    assert_eq!(code, 200, "{}", response);
+    assert_eq!(code, 200, "{response}");
     assert_eq!(response["results"].as_array().unwrap().len(), 1);
 
     let (response, code) = index.filtered_tasks(&[], &["succeeded", "failed"], &[]).await;
-    assert_eq!(code, 200, "{}", response);
+    assert_eq!(code, 200, "{response}");
     assert_eq!(response["results"].as_array().unwrap().len(), 2);
 }
 
 #[actix_rt::test]
 async fn list_tasks_type_filtered() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
     index
@@ -178,19 +178,19 @@ async fn list_tasks_type_filtered() {
         .await;
 
     let (response, code) = index.filtered_tasks(&["indexCreation"], &[], &[]).await;
-    assert_eq!(code, 200, "{}", response);
+    assert_eq!(code, 200, "{response}");
     assert_eq!(response["results"].as_array().unwrap().len(), 1);
 
     let (response, code) =
         index.filtered_tasks(&["indexCreation", "documentAdditionOrUpdate"], &[], &[]).await;
-    assert_eq!(code, 200, "{}", response);
+    assert_eq!(code, 200, "{response}");
     assert_eq!(response["results"].as_array().unwrap().len(), 2);
 }
 
 #[actix_rt::test]
 async fn list_tasks_invalid_canceled_by_filter() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
     index
@@ -198,14 +198,14 @@ async fn list_tasks_invalid_canceled_by_filter() {
         .await;
 
     let (response, code) = index.filtered_tasks(&[], &[], &["0"]).await;
-    assert_eq!(code, 200, "{}", response);
+    assert_eq!(code, 200, "{response}");
     assert_eq!(response["results"].as_array().unwrap().len(), 0);
 }
 
 #[actix_rt::test]
 async fn list_tasks_status_and_type_filtered() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
     index
@@ -213,7 +213,7 @@ async fn list_tasks_status_and_type_filtered() {
         .await;
 
     let (response, code) = index.filtered_tasks(&["indexCreation"], &["failed"], &[]).await;
-    assert_eq!(code, 200, "{}", response);
+    assert_eq!(code, 200, "{response}");
     assert_eq!(response["results"].as_array().unwrap().len(), 0);
 
     let (response, code) = index
@@ -223,12 +223,12 @@ async fn list_tasks_status_and_type_filtered() {
             &[],
         )
         .await;
-    assert_eq!(code, 200, "{}", response);
+    assert_eq!(code, 200, "{response}");
     assert_eq!(response["results"].as_array().unwrap().len(), 2);
 }
 
 macro_rules! assert_valid_summarized_task {
-    ($response:expr, $task_type:literal, $index:literal) => {{
+    ($response:expr, $task_type:literal, $index:tt) => {{
         assert_eq!($response.as_object().unwrap().len(), 5);
         assert!($response["taskUid"].as_u64().is_some());
         assert_eq!($response["indexUid"], $index);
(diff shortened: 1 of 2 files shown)


crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -11,14 +10,14 @@ use crate::json;
 
 #[actix_rt::test]
 async fn error_get_unexisting_task_status() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
```
irevoire:
```suggestion
```

I don't know why we were doing that here, but I guess we can get rid of it?

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -50,8 +49,8 @@ async fn get_task_status() {
 
 #[actix_rt::test]
 async fn list_tasks() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
```
irevoire:
```suggestion
    // Do not use a shared server because we want to assert stuff against the global list of tasks
    let server = Server::new().await;
    let index = server.index("test");
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -127,93 +128,96 @@ async fn list_tasks_with_star_filters() {
...
+    assert_eq!(code, 200, "{response:?}");
     assert_eq!(response["results"].as_array().unwrap().len(), 2);
 }
 
 #[actix_rt::test]
 async fn list_tasks_status_filtered() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
```
irevoire:
```suggestion
    // Do not use a shared server because we want to assert stuff against the global list of tasks
    let server = Server::new().await;
    let index = server.index("test");
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -127,93 +128,96 @@ async fn list_tasks_with_star_filters() {
...
+    assert_eq!(code, 200, "{response}");
     assert_eq!(response["results"].as_array().unwrap().len(), 2);
 }
 
 #[actix_rt::test]
 async fn list_tasks_type_filtered() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
```
irevoire:
```suggestion
    // Do not use a shared server because we want to assert stuff against the global list of tasks
    let server = Server::new().await;
    let index = server.index("test");
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -127,93 +128,96 @@ async fn list_tasks_with_star_filters() {
...
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
     index
         .add_documents(serde_json::from_str(include_str!("../assets/test_set.json")).unwrap(), None)
         .await;
 
     let (response, code) = index.filtered_tasks(&[], &[], &["0"]).await;
```
irevoire: The id is wrong here, we should retrieve the actual id of the task I believe

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -127,93 +128,96 @@ async fn list_tasks_with_star_filters() {
...
+    assert_eq!(code, 200, "{response}");
     assert_eq!(response["results"].as_array().unwrap().len(), 0);
 }
 
 #[actix_rt::test]
 async fn list_tasks_status_and_type_filtered() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
```
irevoire:
```suggestion
    // Do not use a shared server because we want to assert stuff against the global list of tasks
    let server = Server::new().await;
    let index = server.index("test");
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -242,49 +246,50 @@ macro_rules! assert_valid_summarized_task {
...
     let (task, _status_code) =
         index.add_documents(json!({ "id": 42, "content": "doggos & fluff" }), None).await;
     index.wait_task(task.uid()).await.succeeded();
-    let (task, _) = index.get_task(0).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    let (task, _) = index.get_task(task.uid()).await;
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -303,14 +308,14 @@ async fn test_summarized_document_addition_or_update() {
     let (task, _status_code) =
         index.add_documents(json!({ "id": 42, "content": "doggos & fluff" }), Some("id")).await;
     index.wait_task(task.uid()).await.succeeded();
-    let (task, _) = index.get_task(1).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    let (task, _) = index.get_task(task.uid()).await;
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -329,18 +334,23 @@ async fn test_summarized_document_addition_or_update() {
...
+    let (task, _status_code) = index
+        .delete_batch(vec![non_existing_task_id1, non_existing_task_id2, non_existing_task_id3])
+        .await;
     index.wait_task(task.uid()).await.failed();
-    let (task, _) = index.get_task(0).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    let (task, _) = index.get_task(task.uid()).await;
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -366,13 +376,13 @@ async fn test_summarized_delete_documents_by_batch() {
     let (del_task, _status_code) = index.delete_batch(vec![42]).await;
     index.wait_task(del_task.uid()).await.succeeded();
     let (task, _) = index.get_task(del_task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -392,20 +402,20 @@ async fn test_summarized_delete_documents_by_batch() {
...
+    let index = server.unique_index();
 
     let (task, _status_code) =
         index.delete_document_by_filter(json!({ "filter": "doggo = bernese" })).await;
     index.wait_task(task.uid()).await.failed();
     let (task, _) = index.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -432,13 +442,13 @@ async fn test_summarized_delete_documents_by_filter() {
...
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -465,13 +475,13 @@ async fn test_summarized_delete_documents_by_filter() {
         index.delete_document_by_filter(json!({ "filter": "doggo = bernese" })).await;
     index.wait_task(task.uid()).await.succeeded();
     let (task, _) = index.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -491,18 +501,18 @@ async fn test_summarized_delete_documents_by_filter() {
...
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
     let (task, _status_code) = index.delete_document(1).await;
     index.wait_task(task.uid()).await.failed();
     let (task, _) = index.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -528,13 +538,13 @@ async fn test_summarized_delete_document_by_id() {
     let (task, _status_code) = index.delete_document(42).await;
     index.wait_task(task.uid()).await.succeeded();
     let (task, _) = index.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -571,13 +581,13 @@ async fn test_summarized_settings_update() {
     let (task,_status_code) = index.update_settings(json!({ "displayedAttributes": ["doggos", "name"], "filterableAttributes": ["age", "nb_paw_pads"], "sortableAttributes": ["iq"] })).await;
     index.wait_task(task.uid()).await.succeeded();
     let (task, _) = index.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -605,18 +615,18 @@ async fn test_summarized_settings_update() {
...
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
     let (task, _) = index.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -634,21 +644,21 @@ async fn test_summarized_index_creation() {
     let (task, _status_code) = index.create(Some("doggos")).await;
     index.wait_task(task.uid()).await.failed();
     let (task, _) = index.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -772,27 +782,27 @@ async fn test_summarized_index_deletion() {
...
+    let server = Server::new_shared();
+    let index = server.unique_index();
     // If the index doesn't exist yet, we should get errors with or without the primary key.
     let (task, _status_code) = index.update(None).await;
     index.wait_task(task.uid()).await.failed();
     let (task, _) = index.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -807,21 +817,21 @@ async fn test_summarized_index_update() {
     let (task, _status_code) = index.update(Some("bones")).await;
     index.wait_task(task.uid()).await.failed();
     let (task, _) = index.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -839,13 +849,13 @@ async fn test_summarized_index_update() {
     let (task, _status_code) = index.update(None).await;
     index.wait_task(task.uid()).await.succeeded();
     let (task, _) = index.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -863,13 +873,13 @@ async fn test_summarized_index_update() {
     let (task, _status_code) = index.update(Some("bones")).await;
     index.wait_task(task.uid()).await.succeeded();
     let (task, _) = index.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -887,20 +897,20 @@ async fn test_summarized_index_update() {
...
         .index_swap(json!([
             { "indexes": ["doggos", "cattos"] }
         ]))
         .await;
     server.wait_task(task.uid()).await.failed();
     let (task, _) = server.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -1004,20 +1017,20 @@ async fn test_summarized_task_cancelation() {
...
     // to avoid being flaky we're only going to delete an already finished task :(
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
     let (task, _status_code) = server.delete_tasks("uids=0").await;
     index.wait_task(task.uid()).await.succeeded();
     let (task, _) = index.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -1038,16 +1051,16 @@ async fn test_summarized_task_deletion() {
...
 async fn test_summarized_dump_creation() {
-    let server = Server::new().await;
+    let server = Server::new_shared();
     let (task, _status_code) = server.create_dump().await;
     server.wait_task(task.uid()).await;
     let (task, _) = server.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".details.dumpUid" => "[dumpUid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(json_string!(task,
+        { ".uid" => "[uid]", ".batchUid" => "[batch_uid]", ".details.dumpUid" => "[dumpUid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" }),
```
irevoire:
```suggestion
    snapshot!(task,
```

irevoire (changes requested):
Awesome, we have a few issues at the beginning I believe when we want to list the tasks but then when testing the error it's awesome!

Also I noticed these tests were old and were not using the fact that the task type automatically redact itself so I left a few comments 😁

crates/meilisearch/tests/tasks/mod.rs
martin-g: How would that work ?
I do not see any static redactions for those in meili_snap crate.

martin-g: Most of the redactions work!
Just one fails now: https://github.com/meilisearch/meilisearch/actions/runs/15415538336/job/43377337947?pr=5615

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -1038,16 +1036,15 @@ async fn test_summarized_task_deletion() {
...
 #[actix_web::test]
 async fn test_summarized_dump_creation() {
-    let server = Server::new().await;
+    let server = Server::new_shared();
     let (task, _status_code) = server.create_dump().await;
     server.wait_task(task.uid()).await;
     let (task, _) = server.get_task(task.uid()).await;
-    assert_json_snapshot!(task,
-        { ".details.dumpUid" => "[dumpUid]", ".duration" => "[duration]", ".enqueuedAt" => "[date]", ".startedAt" => "[date]", ".finishedAt" => "[date]" },
+    snapshot!(task,
```
irevoire, line 1043: My bad, we cannot do it here for the dump 🙈 

Or you could update the Display implementation of our custom Value type to ignore the field `. details.dumpUid`

irevoire (changes requested):
Looks like there was a big network failure and all our tests failed for no reason.
But I think there is only one issue remaining

irevoire:
Aaah I think I got the issue by looking at the last two failures.

The tests are executing faster overall because they don't lose time waiting to create and delete threads.
But now they spend more time waiting for each other (which is not really an issue since Meilisearch is supposed to be as fast as possible processing the tasks).
Now the issue is that when an AI task must reach a remote URL and that takes too much time, it also fails other unrelated tests that were just waiting for their task to process.

I'll try to merge as-is, but overall we may want to increase the timeout time of `wait_task` 😩

irevoire (approved):

martin-g:
The same tests now break for https://github.com/meilisearch/meilisearch/pull/5622
https://github.com/meilisearch/meilisearch/actions/runs/15462318339/job/43526272581?pr=5622



> Now the issue is that when an AI task must reach a remote URL and that takes too much time, it also fails other unrelated tests that were just waiting for their task to process.

Do these tests need to use a remote URL ? Can we provide them mock data ?
I will take a closer look at the `vector` tests.

martin-g:
https://github.com/meilisearch/meilisearch/blob/1d02efeab9490e935b33c291f935aacf12370ee0/crates/meilisearch/tests/common/server.rs#L401-L405

The `url` always contains `/tasks/` ... The check could be removed.

Where is the (server ?!) logic that assigns `update_id` (task_uid) bigger than 1000 for vectors ?
Running the vector tests locally the task.uid() is always starting from 0 and increasing.

Where is the code doing the remote HTTP call ?

PR #5619, opened 2025-06-02
tests: Faster documents::delete_documents IT tests

martin-g:
# Pull Request

## Related issue
#4840

## What does this PR do?
- Assert .succeeded()/.failed() for the waited tasks

### DIFF
--- a/crates/meilisearch/tests/documents/delete_documents.rs
+++ b/crates/meilisearch/tests/documents/delete_documents.rs
@@ -5,35 +5,32 @@ use crate::json;
 
 #[actix_rt::test]
 async fn delete_one_document_unexisting_index() {
-    let server = Server::new().await;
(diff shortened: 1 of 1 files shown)


crates/meilisearch/tests/documents/delete_documents.rs
```diff
@@ -5,35 +5,32 @@ use crate::json;
 
 #[actix_rt::test]
 async fn delete_one_document_unexisting_index() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
```
irevoire: Here we can use the missing index directly instead of getting a unique index and not initializing it

irevoire (approved):
Perfect thanks 🐶

PR #5622, opened 2025-06-02
tests: Faster search::filters IT tests

martin-g:
# Pull Request

## Related issue
#4840

## What does this PR do?


crates/meilisearch/tests/search/filters.rs
```diff
@@ -10,62 +10,63 @@ use crate::{
...
+    snapshot!(code, @"202 Accepted");
 
     let documents = DOCUMENTS.clone();
     let (task, code) = index.add_documents(documents, None).await;
-    meili_snap::snapshot!(code, @"202 Accepted");
-    let res = index.wait_task(task.uid()).await;
-    meili_snap::snapshot!(res["status"], @r###""succeeded""###);
+    snapshot!(code, @"202 Accepted");
+    let res = index.wait_task(task.uid()).await.succeeded();
+    snapshot!(res["status"], @r###""succeeded""###);
```
irevoire: We can use the shared index with documents instead of creating a new index and indexing stuff in it here: https://github.com/meilisearch/meilisearch/blob/ea6bb4df1d76e8ae3880b60a99b85b1102891a68/crates/meilisearch/tests/common/mod.rs#L224

crates/meilisearch/tests/search/filters.rs
```diff
@@ -10,62 +10,63 @@ use crate::{
...
 
     let documents = NESTED_DOCUMENTS.clone();
-    let (task, code) = index.add_documents(documents, None).await;
-    meili_snap::snapshot!(code, @"202 Accepted");
-    let res = index.wait_task(task.uid()).await;
-    meili_snap::snapshot!(res["status"], @r###""succeeded""###);
+    let (task, code) = nested_index.add_documents(documents, None).await;
+    snapshot!(code, @"202 Accepted");
+    let res = nested_index.wait_task(task.uid()).await.succeeded();
+    snapshot!(res["status"], @r###""succeeded""###);
```
irevoire: And here we can use the index with nested documents https://github.com/meilisearch/meilisearch/blob/ea6bb4df1d76e8ae3880b60a99b85b1102891a68/crates/meilisearch/tests/common/mod.rs#L344

crates/meilisearch/tests/search/filters.rs
martin-g: There is a difference between the indices - the shared index has `{"filterableAttributes": ["id", "title"], "sortableAttributes": ["id", "title"]}` while the one in the test case is just `{"filterableAttributes": ["title"]}`

crates/meilisearch/tests/search/filters.rs
martin-g: Here the difference is - the shared index uses these settings `{"filterableAttributes": ["father", "doggos"], "sortableAttributes": ["doggos"]}` and the index in the test: `{"filterableAttributes": ["cattos", "doggos.age"]}`

crates/meilisearch/tests/search/filters.rs
irevoire: huum yeah, do you think we could update the `shared_index_with_documents`?
If a test was expecting the `id` to *not* be filterable, we could update it to use a different field than id anyway 🤔

martin-g: Using the shared index works without problems!

crates/meilisearch/tests/search/filters.rs
martin-g: Using the shared index with nested documents leads to a failure:
```
thread 'search::filters::search_with_filter_string_notation' panicked at crates/meilisearch/tests/search/filters.rs:49:17:
assertion `left == right` failed: {
  "message": "Index `SHARED_NESTED_DOCUMENTS`: Attribute `cattos` is not filterable. Available filterable attribute patterns are: `doggos`, `father`.\n1:7 cattos = pésti",
  "code": "invalid_search_filter",
  "type": "invalid_request",
  "link": "https://docs.meilisearch.com/errors#invalid_search_filter"
}
  left: 400
 right: 200
```

irevoire: And can't we add the `cattos` to the filterable attribute as well?

martin-g:
@irevoire https://github.com/meilisearch/meilisearch/pull/5622/commits/10028515acca965be2b37c6d488c2231edb15205
This fixes the issue with the timeouts for tasks and vector tests.
I'd suggest to delete Index::wait_task() and use Server::wait_task() in all IT tests.

crates/meilisearch/tests/common/server.rs
```diff
@@ -399,18 +399,9 @@ impl<State> Server<State> {
...
-        let max_attempts = if url.contains("/tasks/") {
-            if update_id > 1000 {
-                400 // 200 seconds for vector tests
-            } else {
-                100 // 50 seconds for other tests
-            }
-        } else {
-            100 // 50 seconds for other tests
-        };
+        let max_attempts = 100; // 50 seconds total, 0.5s per attempt
```
irevoire:
```suggestion
        let max_attempts = 400; // 200 seconds total, 0.5s per attempt
```

I think we want to conservatively use the maximum value we were using in the past (maybe even increase it) because it comes from actual network failures when Hugging Face is slow to answer

crates/meilisearch/tests/tasks/mod.rs
```diff
@@ -1036,7 +1038,8 @@ async fn test_summarized_task_deletion() {
 
 #[actix_web::test]
 async fn test_summarized_dump_creation() {
-    let server = Server::new_shared();
+    // Do not use a shared server because it takes too long to create a dump
+    let server = Server::new().await;
```
irevoire, line 1042: Oh, yes, excellent catch, that makes 100% sense.
It must have been pretty large on disk as well.
We probably want to do the same thing, the snapshot now that I think about it 🤔

martin-g: > We probably want to do the same thing, the snapshot now that I think about it 🤔

Should I change something more here ?!

irevoire: I thought we had a similar test for the snapshot but apparently not, thank you 🙏

PR #5625, opened 2025-06-02
tests: Faster search::hybrid IT tests

martin-g:
# Pull Request

## Related issue
#4840

## What does this PR do?

### DIFF
--- a/crates/meilisearch/tests/search/hybrid.rs
+++ b/crates/meilisearch/tests/search/hybrid.rs
@@ -582,7 +582,7 @@ async fn distinct_is_applied() {
 #[actix_rt::test]
 async fn retrieve_vectors() {
     let server = Server::new_shared();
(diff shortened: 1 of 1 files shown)


martin-g:
Did the merge fail due to the `Check db change labels` job ?

irevoire:
> Did the merge fail due to the `Check db change labels` job ?

No this one has a bug we don't understand and is always failed, but the PR can still merge.

It failed because of that:
image

I don't understand why it didn't trigger on your PR directly but only when trying to merge on main 🤔 
Anyway, I think the fix is to update `index_with_documents_user_provided` so it works with `Owned` indexes instead of the shared one

martin-g:
The reason was a "collision". This test case has been just added with 3c13feeb and it was using the old signature of the helper method.

PR #5626, opened 2025-06-02
tests: Faster batches:: IT tests

martin-g:
# Pull Request

## Related issue
#4840

## What does this PR do?
- Use shared server + unique indices where possible


martin-g:
@irevoire I will need some help with this one.
The tests use `index.get_batch(...)` passing hardcoded values like `0` and `1`.
But as far as I understand it the batches are global, i.e. per server, not per index.
This makes it very hard (impossible) to use shared server!

Am I correct ?

crates/meilisearch/tests/batches/mod.rs
```diff
@@ -27,18 +27,18 @@ async fn error_get_unexisting_batch_status() {
...
 #[actix_rt::test]
 async fn get_batch_status() {
-    let server = Server::new().await;
-    let index = server.index("test");
+    let server = Server::new_shared();
+    let index = server.unique_index();
     let (task, _status_code) = index.create(None).await;
     index.wait_task(task.uid()).await.succeeded();
-    let (_response, code) = index.get_batch(0).await;
+    let (_response, code) = index.get_batch(task.uid() as u32).await;
```
irevoire:
```suggestion
    let (_response, code) = index.get_batch(task.batch_uid()).await;
```

irevoire (changes requested):
I yeah I see the issue, these tests were a bit rushed it looks like.

The best solution IMO is to create a new method similar to this one: https://github.com/meilisearch/meilisearch/blob/ea6bb4df1d76e8ae3880b60a99b85b1102891a68/crates/meilisearch/tests/common/mod.rs#L26-L35

Except its `batch_uid` and it retrieve the `batchUid` field or panic.

Then the usage is that you must wait for the *task* to finish and once it's finished you can retrieve the associated batch with this new method from the task body

crates/meilisearch/tests/batches/mod.rs
martin-g: Hm.
It seems the `batch_uid` property is set only for `status=processing` tasks: https://github.com/meilisearch/meilisearch/blob/5c14a25d5a52af8f02cb0bdd41fc5ea41f002036/crates/index-scheduler/src/queue/tasks.rs#L525-L528

At the moment the new method fails with:
```
thread 'batches::test_summarized_document_addition_or_update' panicked at crates/meilisearch/tests/batches/mod.rs:276:43:
Didn't find any batch id in: {
  "taskUid": 19,
  "indexUid": "961f61b5-5ac7-4b03-adbd-5839519d886e",
  "status": "enqueued",
  "enqueuedAt": "[date]"
}
```

irevoire: It should also be set for all the finished tasks.
Basically, it should always be set except when the task is enqueued, iirc and that's why we have to wait for the task to finishes before calling this method

irevoire (approved):
Thank you that's perfect from what I see 😁

martin-g:
The merge conflicts have been resolved!

irevoire (approved):

PR #5635, opened 2025-06-05
Prompt for Email

Kerollmops:
This PR prompts for a contact email for support and to subscribe to the newsletter. It doesn't prompt if [the TTY is not a terminal](https://en.wikipedia.org/wiki/Tty_(Unix)). It can be disabled by specifying an email via this prompt, via the `contact-email` flag, or by leaving the parameter empty.

### Database Changes (db changes)

This feature creates a `skip-email` file in [global config directory](https://github.com/meilisearch/MeiliSearch/blob/main/crates/meilisearch/src/analytics/mod.rs#L44-L50) if an email (or a text) has been specified via the `--contact-email` parameter or the email prompt. This way, the user does not see the prompt when they start Meilisearch again. The file is not kept between dumps or snapshots, so Meilisearch instances starting from those will display the prompt again.

### DIFF
--- a/crates/meilisearch/src/main.rs
+++ b/crates/meilisearch/src/main.rs
@@ -28,6 +28,7 @@ use tracing::level_filters::LevelFilter;
 use tracing_subscriber::layer::SubscriberExt as _;
 use tracing_subscriber::Layer;
 
+const SKIP_EMAIL_FILENAME: &str = "skip-email";
 const PORTAL_ID: &str = "25945010";
 const FORM_GUID: &str = "991e2a09-77c2-4428-9242-ebf26bfc6c64";
 
@@ -99,22 +100,6 @@ async fn try_main() -> anyhow::Result<()> {
 
     std::panic::set_hook(Box::new(on_panic));
 
-    opt.contact_email = match opt.contact_email.as_ref().map(|email| email.as_deref()) {
-        Some(Some("false")) | None => prompt_for_contact_email().await.map(Some)?,
-        Some(Some("")) | Some(None) => None,
-        Some(Some(email)) => Some(Some(email.to_string())),
-    };
-
-    if let Some(Some(email)) = opt.contact_email.as_ref() {
-        let email = email.clone();
-        let _ = tokio::spawn(async move {
-            dbg!(&email);
-            if let Err(err) = register_contact_email(&email).await {
-                eprintln!("Failed to register email: {}", err);
-            }
-        });
-    }
-
     anyhow::ensure!(
         !(cfg!(windows) && opt.experimental_reduce_indexing_memory_usage),
         "The `experimental-reduce-indexing-memory-usage` flag is not supported on Windows"
@@ -145,6 +130,30 @@ async fn try_main() -> anyhow::Result<()> {
     }
 
     let (index_scheduler, auth_controller) = setup_meilisearch(&opt)?;
+
+    // We ask users their emails just after the data.ms is created
+    let skip_email_path = opt.db_path.join(SKIP_EMAIL_FILENAME);
+    opt.contact_email = match opt.contact_email.as_ref().map(|email| email.as_deref()) {
+        Some(Some("false")) | None if !skip_email_path.exists() => {
+            prompt_for_contact_email().await.map(Some)?
+        }
+        Some(Some(email)) if !skip_email_path.exists() => Some(Some(email.to_string())),
+        _otherwise => None,
+    };
+
+    if let Some(Some(email)) = opt.contact_email.as_ref() {
+        let email = email.clone();
+        // We spawn a task to register the email and create the skip email
+        // file to avoid blocking the Meilisearch launch further.
+        let _ = tokio::spawn(async move {
+            if let Err(e) = tokio::fs::File::create_new(skip_email_path).await {
+                eprintln!("Failed to create skip email file: {e}");
+            }
+            if let Err(err) = register_contact_email(&email).await {
+                eprintln!("Failed to register email: {}", err);
+            }
+        });
+    }
 
     let analytics =
         analytics::Analytics::new(&opt, index_scheduler.clone(), auth_controller.clone()).await;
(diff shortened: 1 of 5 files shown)


crates/meilisearch/src/main.rs
```diff
@@ -139,6 +146,27 @@ async fn try_main() -> anyhow::Result<()> {
...
+async fn prompt_for_contact_email() -> anyhow::Result<Option<String>> {
+    let stdin = tokio::io::stdin();
+
+    if !stdin.is_terminal() {
+        return Ok(None);
+    }
+
+    println!("Would you mind providing your contact email for support and news?");
+    println!("We will use it to contact you with news only.");
+    print!("contact email> ");
```
Kerollmops: Hey @guimachiavelli, and @tpayet, I need a better text to prompt users for their email. Something legally better and friendlier than that. Can you help me with this part, please?

irevoire:
Hey, I store it both in the data.ms and the "config path" (that changes depending on the OS, but basically it depends on the [`platform_dirs`](https://docs.rs/platform-dirs/latest/platform_dirs/struct.AppDirs.html) crate. You can find the related code [here](https://github.com/meilisearch/MeiliSearch/blob/main/crates/meilisearch/src/analytics/mod.rs#L44-L50).

For the analytics, I'm creating a `Meilisearch` (capitalized, yes, it's old) directory to avoid polluting the environment. Then, we're creating one file per database, not by user.

In your case, I think you'll want to create a unique file that disables your prompt if it exists, regardless of the database.

-----

Here's how I think you should do it, but I still think all of this is a very bad idea and a breaking change that should not be merged into Meilisearch.

crates/meilisearch/src/main.rs
```diff
@@ -139,6 +161,76 @@ async fn try_main() -> anyhow::Result<()> {
...
+/// Prompt the user about the contact email for support and news.
+/// It only displays the prompt if the input is an interactive terminal.
+async fn prompt_for_contact_email() -> anyhow::Result<Option<String>> {
+    let stdin = tokio::io::stdin();
+
+    if !stdin.is_terminal() {
+        return Ok(None);
+    }
+
+    println!("Enter your email to receive occasional updates and tips about Meilisearch.");
```
guimachiavelli: What exactly are we going to do with these emails? The way it's written reads very much like "we're going to spam you".

I would advise at least making it shorter and more direct.

```suggestion
    println!("Enter your email to receive updates about Meilisearch.");
```

guimachiavelli: You could also go with an even drier alternative:

```suggestion
    println!("Enter your email.");
```

This might raise questions of "what are you going to do with my email", but could also prevent problems with e.g. saying we're going to send people updates about Meilisearch and instead end up using their emails for something else.

Kerollmops: Can we find a friendlier version, then? Something like the following?
```suggestion
    println!("Want to know more about the latest Meilisearch news? Enter your email to receive occasional updates and tips.");
```

Kerollmops: We also have this one from the mini-dashboard. It's better formulated.
```suggestion
    println!("Stay up to date! Get monthly updates about new features and tips to get the most out of Meilisearch.");
    println!("You can use the `--contact-email` parameter to disable this prompt.");
    println!("Enter your email or leave blank to skip> ");
```

crates/meilisearch/src/main.rs
Kerollmops: Thank you for your help, I improved the prompting to use the one on the mini-dashboard.

Kerollmops:
Hey @nicolasvienot 👋

I created [a prototype available on Docker](https://github.com/meilisearch/meilisearch/actions/runs/15472364110) with the following tag/command. Can you please tell me if it works on the Cloud?
```
docker pull getmeili/meilisearch:prototype-email-prompt-2
```

crates/meilisearch/src/main.rs
```diff
@@ -139,6 +170,76 @@ async fn try_main() -> anyhow::Result<()> {
...
+async fn prompt_for_contact_email() -> anyhow::Result<Option<String>> {
+    let stdin = tokio::io::stdin();
+
+    if !stdin.is_terminal() {
+        return Ok(None);
+    }
+
+    println!("Stay up to date! Get monthly updates about new features and tips to get the most out of Meilisearch.");
+    println!("You can use the `--contact-email` parameter to disable this prompt.");
+    print!("Enter your email or leave blank to skip> ");
```
guimachiavelli: I would say that CLI communication should be very heavily skewed towards brevity and raw information. You don't usually fire up a console to have fun and browse around, you're there to accomplish a task.

Introducing the prompt will be an obstacle to that task, and asking for someone's email is particularly invasive and concerning, so a friendly tone is likely to make the experience even more aggravating.

I defer the final decision to you—I'm better at improving things for users, but in this case user experience is significantly less important than the company's needs.

Kerollmops: Yeah, I understand. Thank you @guimachiavelli. What do you suggest so that it's not too aggressive?
I want to remove the "Stay up to date!", what do you think?

guimachiavelli: Yeah, removing that bit at least takes away some of the noise. I think the `--contact-email` line could also be more direct.

For consistency, I changed `parameter` to `option` because that's the wording we use in the docs (instance options vs index settings vs API parameters). `argument` could also work here.

Oh, and I'm not sure `--contact-email` needs the backticks—will the markdown be formatted in the output?

```suggestion
    println!("Get monthly updates about new features and tips to get the most out of Meilisearch.");
    println!("Use the `--contact-email` option to disable this prompt.");
    print!("Enter your email or leave blank to skip> ");
```

tpayet (dismissed):

crates/meilisearch/src/main.rs
Kerollmops: You are right, it will not be formatted, so I'll remove this.

```suggestion
    println!("Get monthly updates about new features and tips to get the most out of Meilisearch.");
    println!("Use the --contact-email option to disable this prompt.");
    print!("Enter your email or leave blank to skip> ");
```

nicolasvienot:
> Hey @nicolasvienot 👋
> 
> I created [a prototype available on Docker](https://github.com/meilisearch/meilisearch/actions/runs/15472364110) with the following tag/command. Can you please tell me if it works on the Cloud?
> 
> ```
> docker pull getmeili/meilisearch:prototype-email-prompt-2
> ```

Confirmed working ✅

irevoire (approved):

PR #5637, opened 2025-06-05
chore: Explicitly depend on xyz_derive crates

martin-g:
# Pull Request

## Related issue
N/A

## What does this PR do?
This improves the full build time by building both xyz and xyz_derive crates in parallel.
By using a feature the `xyz`'s build will need to wait for `xyz_derive`'s build.

The incremental build times should not be affected.

Suggested-at: https://github.com/serde-rs/serde/issues/2907#issuecomment-2936782977


martin-g:
`cargo clean && time cargo build` on my dev machine:

Before:
Run 1
```
________________________________________________________
Executed in  343.09 secs    fish           external
   usr time   16.33 mins  865.00 micros   16.33 mins
   sys time    1.54 mins  950.00 micros    1.54 mins
```
Run 2
```
Executed in  341.28 secs    fish           external
   usr time   16.24 mins  630.00 micros   16.24 mins
   sys time    1.56 mins  890.00 micros    1.56 mins

```

After:

Run 1
```
________________________________________________________
Executed in  306.80 secs    fish           external
   usr time   15.97 mins  621.00 micros   15.97 mins
   sys time    1.50 mins  886.00 micros    1.50 mins
```
Run 2
```
________________________________________________________
Executed in  295.44 secs    fish           external
   usr time   15.23 mins    0.87 millis   15.23 mins
   sys time    1.46 mins    1.93 millis    1.46 mins

```

crates/fuzzers/src/bin/fuzz-indexing.rs
```diff
@@ -18,10 +18,10 @@ use milli::Index;
 use serde_json::Value;
 use tempfile::TempDir;
 
-#[derive(Debug, Arbitrary)]
+#[derive(Debug, derive_arbitrary::Arbitrary)]
```
irevoire: Is this really useful?
We're building the derive crate only once so we can use the derive version or the re-exported version the same and it should work right? 🤔

martin-g: Here is how it breaks with the old/current code:
```
error[E0433]: failed to resolve: could not find `Arbitrary` in `arbitrary`
  --> crates/fuzzers/src/bin/fuzz-indexing.rs:21:28
   |
21 | #[derive(Debug, arbitrary::Arbitrary)]
   |                            ^^^^^^^^^ could not find `Arbitrary` in `arbitrary`

error[E0599]: the function or associated item `arbitrary` exists for array `[Batch; 5]`, but its trait bounds were not satisfied
   --> crates/fuzzers/src/bin/fuzz-indexing.rs:79:49
    |
22  | struct Batch([Operation; 5]);
    | ------------ doesn't satisfy `Batch: Arbitrary<'_>`
...
79  |                     let batches = <[Batch; 5]>::arbitrary(&mut data).unwrap();
    |                                                 ^^^^^^^^^ function or associated item cannot be called on `[Batch; 5]` due to unsatisfied trait bounds
    |
    = note: the following trait bounds were not satisfied:
            `Batch: Arbitrary<'_>`
            which is required by `[Batch; 5]: Arbitrary<'_>`
note: the trait `Arbitrary` must be implemented
   --> /home/martin/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arbitrary-1.4.1/src/lib.rs:163:1
    |
163 | pub trait Arbitrary<'a>: Sized {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Since the `derive` feature is not enabled for the `arbitrary` crate some of its exports are not reachable:
- https://github.com/rust-fuzz/arbitrary/blob/334351140a0dcea63bc28dd5c57fab19762f6b1d/src/lib.rs#L35-L36
- https://github.com/rust-fuzz/arbitrary/blob/334351140a0dcea63bc28dd5c57fab19762f6b1d/derive/src/lib.rs#L19

irevoire: Aaah, yes, my bad, I didn't notice you also removed the feature.
Then, an additional question: shouldn't we be able to keep the derive feature in arbitrary and still get the compile time improvement because cargo can start compiling it before it takes on arbitrary (and then when arbitrary arrives, if it's already compiled, then there is nothing to do)?

martin-g: This should work too, I think!
Let me try it!

martin-g: Unfortunately this way the benefit is gone.

Run 1:
```
________________________________________________________
Executed in  330.66 secs    fish           external
   usr time   15.33 mins    1.15 millis   15.33 mins
   sys time    1.47 mins    0.04 millis    1.47 mins
```

Run 2:
```
________________________________________________________
Executed in  332.93 secs    fish           external
   usr time   15.46 mins    1.25 millis   15.46 mins
   sys time    1.49 mins    0.07 millis    1.49 mins
```

I have pushed the [change](https://github.com/meilisearch/meilisearch/pull/5637/commits/0effbe5e0322a1677d55f969c45a970d1545960c) for review but I think it should be reverted if this PR is going to be merged.

irevoire: Oooh that's so strange 😲

martin-g:
`cargo build --timings` produces the following (`compile-serde-derive-in-parallel` branch):

meili-cargo-build-timings

`serde_derive` is the first `derive` crate at position 52 with 4.9secs.

martin-g:
I am closing this PR for now.
If there is an interest to explore more we will reopen it!

PR #5660, opened 2025-06-12
Searchable fields aren't indexed when I add and remove them out of filterableAttributes

### LINKED ISSUE #5650: Searchable fields aren't indexed when I add and remove them out of filterableAttributes
**Describe the bug**
The searchable attributes on existing documents doesn't be indexed when I add and remove the same fields to list of filter attributes.

**To Reproduce**
1. Create index "products" with two fields: id (primary key) and name.
2. Add documents.
3. Perform a search for "product" text and everything is work.
4. Add the field "name" to list of filterable attributes and then remove it.
5. Perform a search for "product" text and see no results, although the field "name" is still a searchable field.

**Expected behavior**
The searchable fields don't be affected when I add and remove the same field to list of filterable attributes.

**Screenshots**
[Video](https://drive.google.com/file/d/1SlSKtsFVE0mopHjGk-W9DQBjP5om7oJE/view?usp=sharing)

**Meilisearch version:**
v1.14.0 - current

**Additional context**
- `searchableAttributes = ["*"] or ["name"]` also produces same results.

ManyTheFish:
# Pull Request

## Related issue
Fixes #5650

## What does this PR do?
- add a test reproducing the bug
- Skip indexing on settings update when possible,

> when removing a field from the filterable settings,
this will trigger a reindexing of the negative version of the document,
which removes the document from the searchable as well because the field was considered removed.

### DIFF
--- a/crates/meilisearch/tests/search/mod.rs
+++ b/crates/meilisearch/tests/search/mod.rs
@@ -2086,3 +2086,76 @@ async fn test_exact_typos_terms() {
     )
     .await;
 }
+
+#[actix_rt::test]
+async fn simple_search_changing_unrelated_settings() {
+    let server = Server::new_shared();
+    let index = server.unique_index();
+
+    let documents = DOCUMENTS.clone();
+    let (task, _status_code) = index.add_documents(documents, None).await;
+    index.wait_task(task.uid()).await.succeeded();
+
+    index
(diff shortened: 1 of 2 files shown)


irevoire (approved):
Thank you 🚗 🥇

PR #5661, opened 2025-06-12
Describe embedder settings in GH-flavored markdown

dureuill:
This allows to have a readable summary of all embedder settings generated from a single source of truth.

Can be linked to in the documentation, internal discussions, etc.

- Add meta information to embedder settings in `milli`
- Add a `milli` binary to format the meta information as a GH-flavored Markdown file.
- Add the output in the sources of the binary.

[Rendered](https://github.com/meilisearch/meilisearch/blob/71b5afa23bef92085db6b0303aa9e023990f80a7/docs/embedder_settings.md)


crates/milli/src/vector/settings.rs
```diff
@@ -1150,6 +1157,159 @@ impl MetaEmbeddingSetting {
...
+            Headers => "Additional headers to send to the remote embedder.",
+            SearchEmbedder => "Embedder settings for the embedder used at search time.",
+            IndexingEmbedder => "Embedder settings for the embedder used at indexing time.",
+            Distribution => "Affine transformation applied to the semantic score to make it more comparable to the ranking score.",
+            BinaryQuantized => r#"
+Whether to binary quantize the embeddings of this embedder.
+
+Binary quantized embeddings are smaller than regular embeddings, which improves
+disk usage and retrieval speed, at the cost of relevancy.
+            "#,
```
irevoire, line 1205: Maybe not the best sentence, but I think we should specify again that changing this parameter is a destructive operation, as it looks like a "simple trick you should try"

```suggestion
            BinaryQuantized => r#"
Whether to binary quantize the embeddings of this embedder.

Binary quantized embeddings are smaller than regular embeddings, which improves
disk usage and retrieval speed, at the cost of relevancy. Once enabled, it alters your vectors and cannot be disabled.
            "#,
```

crates/milli/src/vector/settings.rs
```diff
@@ -1150,6 +1157,159 @@ impl MetaEmbeddingSetting {
...
+    }
+
+    pub fn default_value(&self) -> &'static str {
+        use MetaEmbeddingSetting::*;
+        match self {
+            Source => r#""openAi""#,
+            Model => {
+                r#"
+- For source `openAi`, defaults to "text-embedding-3-small"
+- For source `huggingFace`, defaults to "BAAI/bge-base-en-v1.5"
```
irevoire, line 1240: Shouldn't we define a const holding these values, just to be sure we don't forget about it?
Or is it something that openAI decides on their side?

crates/milli/src/vector/settings.rs
```diff
@@ -1150,6 +1157,159 @@ impl MetaEmbeddingSetting {
...
+            "#
+            }
+            Revision => {
+                r#"
+- When `model` is set to default, defaults to "617ca489d9e86b49b8167676d8220688b99db36e"
```
irevoire, line 1245: Same question here

crates/milli/src/vector/settings.rs
```diff
@@ -1150,6 +1157,159 @@ impl MetaEmbeddingSetting {
...
+            IndexingEmbedder => ReindexOutcome::ReindexSometimes(
+                r#"
+- Embedding are regenerated when the setting modified in the indexing embedder require regeneration.
+            "#,
+            ),
+            Distribution => ReindexOutcome::NeverReindex,
+            BinaryQuantized => ReindexOutcome::ReindexSometimes(
+                r#"
+- Embeddings are not regenerated, but the binary quantization takes time during indexing.
+            "#,
```
irevoire, line 1309: Or maybe the comment I left on the bq above should be written here, it seems more appropriate 🤔

irevoire (changes requested):
Hey almost good I just had a few questions 😁 

Also, I find the place `milli/src/bin` a bit strange, wouldn't it make more sense to do a new crate? (and help compile time, not sure?)
I think I was kinda expecting that we would do a "doc" crate that contains both the codes, the "src", and all the expanded markdown if we add some one day.

Other question but -> what would be the process to keep the markdown updated? Should we run your binary in the CI and compare both file?
Or just tell ourselves to check it right before each release?

dureuill:
hey @irevoire 

where should the docs crate live?

1. in `crates/docs` (hard to find, but consistent with other crates)
2. in `docs` (easy to find, but inconsistent with other crates)

thanks for the review 👍

irevoire:
I think I  would shove it in `crates/docs`.
That's also where meilitools lives, it makes sense to me but I can be wrong 🤔

guimachiavelli:
Hi folks—we briefly discussed this during a squad meeting and were wondering who are we targeting with this? Are we addressing a customer complaint, or is this going to make it easier to provide proper support to paying customers?

dureuill:
Hi!

> who are we targeting with this?

internal docs. The need was reported internally.

> Are we addressing a customer complaint

No, the target was to make it easier to implement the Cloud frontend.

dureuill:
closing old PR

PR #5668, opened 2025-06-12
Various fixes to embedding regeneration

### LINKED ISSUE #5667: Enabling the regenerate options doesn't generate an embedding
**Describe the bug**

Enabling the regenerate option in the `_vectors` field of a document doesn't generate an embedding if the embedding was not set, i.e., `null` or `[]`.

**To Reproduce**

1. Setup an embedder with an external provider, e.g. OpenAI.
2. Send a document with the `_vectors` field set to `{ "regenerate": false }`.
3. Send the same document but with the `{ "regenerate": true }`.
4. Retrieve the document with the vectors using the retrieve vectors option enabled
5. See the document without an embedding.

**Expected behavior**
Meilisearch should generate an embedding when the regenerate option is set to true and there are no embedder set.

**Meilisearch version:**
At least in v1.15.0.

dureuill:
Embeddings will now be regenerated when:

- `regenerate` was previously `false` and became `true`
- rendering the old version of the docs failed

Additionally, when only `regenerate` is set in `_vectors` in the new version of the document, the embeddings will now be computed from the new version of the document instead of the old version.

## Related issue
Fixes #5667


Kerollmops (approved):
Easy and perfect ✅

PR #5672, opened 2025-06-16
docs: Recommend using a custom path for the benches' data

martin-g:
# Pull Request

## Related issue
N/A

## What does this PR do?
- This reduces the second build time of the `benchmarks` crate from ~220secs to 45secs (according to `cargo build --timings`) on my dev machine

Additionally I've introduced a parent folder for the Meili related cache paths - ~/.cache/meili


irevoire (approved):
Good catch, I thought that was already mentioned!

PR #5677, opened 2025-06-16
tests: Faster document::errors IT tests

martin-g:
# Pull Request

## Related issue
#4840

## What does this PR do?
* Add a call to .failed() for an awaited task
* Use Server::wait_task() instead of Index::wait_task() - it has better error checking

The tests already use the shared server!


irevoire (approved):
Awesome 📯

PR #5680, opened 2025-06-16
tests: Faster similar::mod IT tests

martin-g:
# Pull Request

## Related issue
#4840

## What does this PR do?
- Use shared server + unique indexes


irevoire (approved):
Thank you

PR #5687, opened 2025-06-18
Settings indexer edition 2024

ManyTheFish:
# Pull Request

## Related issue
Fixes #<issue_number>

## What does this PR do?
- ...

## PR checklist

### DIFF
--- a/crates/milli/src/update/new/document_change.rs
+++ b/crates/milli/src/update/new/document_change.rs
@@ -319,7 +319,7 @@ impl<'doc> SettingsChangeDocument<'doc> {
         self.docid
     }
 
-    pub fn external_document_id(&self) -> &'doc str {
+    pub fn external_docid(&self) -> &'doc str {
         self.external_document_id
     }
 
@@ -330,6 +330,18 @@ impl<'doc> SettingsChangeDocument<'doc> {
         mapper: &'a Mapper,
     ) -> Result<DocumentFromDb<'a, Mapper>> {
         Ok(DocumentFromDb::new(self.docid, rtxn, index, mapper)?.ok_or(
+            crate::error::UserError::UnknownInternalDocumentId { document_id: self.docid },
+        )?)
+    }
+
+    pub fn current_vectors<'a, Mapper: FieldIdMapper>(
+        &self,
+        rtxn: &'a RoTxn,
+        index: &'a Index,
+        mapper: &'a Mapper,
+        doc_alloc: &'a Bump,
+    ) -> Result<VectorDocumentFromDb<'a>> {
+        Ok(VectorDocumentFromDb::new(self.docid, index, rtxn, mapper, doc_alloc)?.ok_or(
             crate::error::UserError::UnknownInternalDocumentId { document_id: self.docid },
         )?)
     }
--- a/crates/milli/src/update/new/document_change.rs
+++ b/crates/milli/src/update/new/document_change.rs
@@ -37,6 +37,11 @@ pub struct Insertion<'doc> {
     new: Versions<'doc>,
 }
 
+pub struct SettingsChangeDocument<'doc> {
+    docid: DocumentId,
+    external_document_id: &'doc str,
+}
+
 impl<'doc> DocumentChange<'doc> {
     pub fn docid(&self) -> DocumentId {
         match &self {
@@ -302,5 +307,30 @@ impl<'doc> Update<'doc> {
                 embedders,
             )
         }
+    }
+}
+
+impl<'doc> SettingsChangeDocument<'doc> {
+    pub fn create(docid: DocumentId, external_document_id: &'doc str) -> Self {
+        Self { docid, external_document_id }
+    }
+
+    pub fn docid(&self) -> DocumentId {
+        self.docid
+    }
+
+    pub fn external_document_id(&self) -> &'doc str {
+        self.external_document_id
+    }
+
+    pub fn current<'a, Mapper: FieldIdMapper>(
+        &self,
+        rtxn: &'a RoTxn,
+        index: &'a Index,
+        mapper: &'a Mapper,
+    ) -> Result<DocumentFromDb<'a, Mapper>> {
+        Ok(DocumentFromDb::new(self.docid, rtxn, index, mapper)?.ok_or(
+            crate::error::UserError::UnknownInternalDocumentId { document_id: self.docid },
+        )?)
     }
 }
--- a/crates/milli/src/update/new/document_change.rs
+++ b/crates/milli/src/update/new/document_change.rs
@@ -14,14 +14,9 @@ use crate::vector::EmbeddingConfigs;
 use crate::{DocumentId, Index, InternalError, Result};
 
 pub enum DocumentChange<'doc> {
-    Deletion(Deletion<'doc>),
+    Deletion(DatabaseDocument<'doc>),
     Update(Update<'doc>),
     Insertion(Insertion<'doc>),
-}
-
-pub struct Deletion<'doc> {
-    docid: DocumentId,
-    external_document_id: &'doc str,
 }
 
 pub struct Update<'doc> {
@@ -37,7 +32,7 @@ pub struct Insertion<'doc> {
     new: Versions<'doc>,
 }
 
-pub struct SettingsChangeDocument<'doc> {
+pub struct DatabaseDocument<'doc> {
     docid: DocumentId,
     external_document_id: &'doc str,
 }
@@ -57,31 +52,6 @@ impl<'doc> DocumentChange<'doc> {
             DocumentChange::Update(update) => update.external_document_id(),
             DocumentChange::Insertion(insertion) => insertion.external_document_id(),
         }
-    }
-}
-
-impl<'doc> Deletion<'doc> {
-    pub fn create(docid: DocumentId, external_document_id: &'doc str) -> Self {
-        Self { docid, external_document_id }
-    }
-
-    pub fn docid(&self) -> DocumentId {
-        self.docid
-    }
-
-    pub fn external_document_id(&self) -> &'doc str {
-        self.external_document_id
-    }
-
-    pub fn current<'a, Mapper: FieldIdMapper>(
-        &self,
-        rtxn: &'a RoTxn,
-        index: &'a Index,
-        mapper: &'a Mapper,
-    ) -> Result<DocumentFromDb<'a, Mapper>> {
-        Ok(DocumentFromDb::new(self.docid, rtxn, index, mapper)?.ok_or(
-            crate::error::UserError::UnknownInternalDocumentId { document_id: self.docid },
-        )?)
     }
 }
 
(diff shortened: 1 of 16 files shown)


crates/milli/src/update/new/indexer/settings_changes.rs
```diff
@@ -14,6 +14,19 @@ use crate::update::new::parallel_iterator_ext::ParallelIteratorExt as _;
...
+/// An internal iterator (i.e. using `foreach`) of `DocumentChange`s
+pub trait SettingsChangeExtractor<'extractor>: Sync {
+    type Data: MostlySend;
+
+    fn init_data<'doc>(&'doc self, extractor_alloc: &'extractor Bump) -> Result<Self::Data>;
+
+    fn process<'doc>(
+        &'doc self,
+        changes: impl Iterator<Item = Result<DatabaseDocument<'doc>>>,
+        context: &'doc DocumentChangeContext<Self::Data>,
```
ManyTheFish: `DocumentChangeContext` -> `DocumentContext`

crates/milli/src/update/new/extract/documents.rs
```diff
@@ -158,3 +171,101 @@ impl<'extractor> Extractor<'extractor> for DocumentsExtractor<'_, '_> {
...
+    extractor_sender: &ExtractorBbqueueSender,
+    settings_delta: &SD,
+    extractor_allocs: &'extractor mut ThreadLocal<FullySend<Bump>>,
+) -> Result<()>
+where
+    MSP: Fn() -> bool + Sync,
+    SD: SettingsDelta,
+{
+    // skip if no embedder_actions
+    if settings_delta.embedder_actions().is_empty() {
```
ManyTheFish: skip if no WriteBack

crates/milli/src/update/new/extract/documents.rs
```diff
@@ -158,3 +171,101 @@ impl<'extractor> Extractor<'extractor> for DocumentsExtractor<'_, '_> {
...
+
+            let content = write_to_obkv(
+                &content,
+                Some(&vector_content),
+                &self.embedder_actions,
+                &mut new_fields_ids_map,
+                &mut document_buffer,
+            )?;
+
+            self.document_sender.uncompressed(document.docid(), external_docid, content).unwrap();
```
ManyTheFish, line 231: skip sending an unchanged document

ManyTheFish: chack if is_user_provided

crates/milli/src/update/new/document_change.rs
```diff
@@ -37,6 +32,11 @@ pub struct Insertion<'doc> {
     new: Versions<'doc>,
 }
 
+pub struct DatabaseDocument<'doc> {
```
dureuill: nitpick, but as this is describing the ids of a single document, it probably ought to live inside of `document.rs` rather than `document_change.rs`.

Also regarding naming it probably is better to have `DocumentIds` or something?

crates/milli/src/update/settings.rs
```diff
@@ -1418,6 +1422,106 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
...
+    pub fn execute<'indexer, MSP>(
+        mut self,
+        must_stop_processing: &'indexer MSP,
+        progress: &'indexer Progress,
+    ) -> Result<Option<ChannelCongestion>>
+    where
+        MSP: Fn() -> bool + Sync,
+    {
+        // force the old indexer if the environment says so
+        if std::env::var_os("MEILI_EXPERIMENTAL_NO_EDITION_2024_FOR_SETTINGS").is_some() {
```
dureuill: Sorry for bringing this when I'm the one who originally set to get the variable here, but I think it would be best if it was a function parameter, and if the env variable could be fetch from the meilisearch crate.

1. cleaner from a separation of concerns perspective (milli doesn't read the env variable itself)
2. makes it easier to test both setting indexers

crates/milli/src/update/settings.rs
```diff
@@ -1418,6 +1422,106 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
...
+            embedder_settings: _,
+            search_cutoff: Setting::NotSet,
+            localized_attributes_rules: Setting::NotSet,
+            prefix_search: Setting::NotSet,
+            facet_search: Setting::NotSet,
+            disable_on_numbers: Setting::NotSet,
+            chat: Setting::NotSet,
+            wtxn: _,
+            index: _,
+            indexer_config: _, // TODO: this is not used
```
dureuill: should this be used?

crates/milli/src/update/settings.rs
```diff
@@ -1418,6 +1422,106 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
...
+            self.index.set_updated_at(self.wtxn, &OffsetDateTime::now_utc())?;
+
+            let old_inner_settings = InnerIndexSettings::from_index(self.index, self.wtxn, None)?;
+
+            // Update index settings
+            let embedding_config_updates = self.update_embedding_configs()?;
+
+            let mut new_inner_settings =
+                InnerIndexSettings::from_index(self.index, self.wtxn, None)?;
+            new_inner_settings.recompute_searchables(self.wtxn, self.index)?;
```
dureuill: is this needed for regenerating the document templates in case the searchable changed (not possible right now but future-proofing)? if so, I would expect the result of this to influence the extraction

crates/milli/src/update/new/indexer/settings_changes.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// An internal iterator (i.e. using `foreach`) of `DocumentChange`s
+pub trait SettingsChangeExtractor<'extractor>: Sync {
+    type Data: MostlySend;
+
+    fn init_data<'doc>(&'doc self, extractor_alloc: &'extractor Bump) -> Result<Self::Data>;
+
+    fn process<'doc>(
+        &'doc self,
+        changes: impl Iterator<Item = Result<DatabaseDocument<'doc>>>,
```
dureuill: the parameter should probably not be called `changes` as the documents are not what change here. Maybe `documents`?

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -239,3 +202,240 @@ where
 
     Ok(congestion)
 }
+
+#[allow(clippy::too_many_arguments)] // clippy: 😝
```
dureuill: I think the allow might be unnecessary? 7 arguments

crates/milli/src/update/settings.rs
```diff
@@ -1418,6 +1422,106 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
...
+            let settings_update_only = true;
+            let inner_settings_diff = InnerIndexSettingsDiff::new(
+                old_inner_settings,
+                new_inner_settings,
+                primary_key_id,
+                embedding_config_updates,
+                settings_update_only,
+            );
+
+            if self.index.number_of_documents(self.wtxn)? > 0 {
```
dureuill, line 1511: is this safe to do? It means that all non-documents related side-effects must be performed outside of `reindex`.

crates/milli/src/update/new/extract/documents.rs
dureuill: to make it more future-proof, you could add a `let needs_change = ...` boolean that would now be a simple check on user provided, but might in the future be expanded if we change the document extractor.

crates/milli/src/update/new/extract/documents.rs
```diff
@@ -158,3 +171,101 @@ impl<'extractor> Extractor<'extractor> for DocumentsExtractor<'_, '_> {
...
+    pub fn new(
+        document_sender: DocumentsSender<'a, 'b>,
+        embedder_actions: &'a BTreeMap<String, EmbedderAction>,
+    ) -> Self {
+        Self { document_sender, embedder_actions }
+    }
+}
+
+impl<'extractor> SettingsChangeExtractor<'extractor> for SettingsChangeDocumentExtractor<'_, '_> {
+    type Data = FullySend<RefCell<DocumentExtractorData>>;
```
dureuill: I don't think this data type is actually used in the extractor?

crates/milli/src/update/new/indexer/extract.rs
```diff
@@ -312,6 +319,112 @@ where
...
+    )?;
+
+    'vectors: {
+        if settings_delta.embedder_actions().is_empty() {
+            break 'vectors;
+        }
+
+        let embedding_sender = extractor_sender.embeddings();
+
+        // extract the remaining embedders
```
dureuill:
```suggestion
        // extract the remaining embeddings
```

crates/milli/src/update/new/extract/vectors/mod.rs
```diff
@@ -290,6 +294,200 @@ impl<'extractor> Extractor<'extractor> for EmbeddingExtractor<'_, '_> {
...
+                        let old_rendered = old_prompt.render_document(
+                            document.external_document_id(),
+                            document.current(
+                                &context.rtxn,
+                                context.index,
+                                context.db_fields_ids_map,
+                            )?,
+                            context.new_fields_ids_map,
+                            &context.doc_alloc,
+                        )?;
```
dureuill, line 428: this should not be an error to be unable to render the document with the old embedder. See on main how the document change extractor now handles it (it changed recently)

crates/milli/src/update/new/extract/vectors/mod.rs
```diff
@@ -290,6 +294,200 @@ impl<'extractor> Extractor<'extractor> for EmbeddingExtractor<'_, '_> {
...
+
+        let mut all_chunks = BVec::with_capacity_in(embedders.len(), &context.doc_alloc);
+        for (embedder_name, (embedder, prompt, _is_quantized)) in embedders {
+            // if the embedder is not in the embedder_actions, we don't need to reindex.
+            if let Some((embedder_id, reindex_action)) =
+                self.embedder_actions.get(embedder_name).and_then(|action| {
+                    let embedder_id = self
+                        .embedder_category_id
+                        .get(embedder_name)
+                        .expect("embedder_category_id should be present");
```
dureuill: This works because we are taking our list of embedders from the new embedder configs. Otherwise it would also require filtering on actions that have a `reindex` set. This should probably be commented upon, or the code be updated for clarity

crates/milli/src/update/settings.rs
ManyTheFish: I think yes, it should be performed upper, when updating the settings.

crates/milli/src/update/settings.rs
ManyTheFish: I removed the comment, this is used to provide the threadpool and the grenad parameters to reindex

crates/milli/src/update/settings.rs
ManyTheFish: not needed, this is already done in `update_index` at the end of reindex

dureuill (approved):
Thank you 🌈

PR #5693, opened 2025-06-19
Add a Read-Only Admin API Key by default

Mubelotix:
# Pull Request

## Related issue
Fixes N/A

## What does this PR do?

This PR introduces a new kind of API key that will allow peeking into an instance without resorting to the master key, thus avoiding the risk of altering data.
This is a kind of "Read-Only Admin API Key."

## PR checklist

### DIFF
--- a/crates/meilisearch-types/src/keys.rs
+++ b/crates/meilisearch-types/src/keys.rs
@@ -131,7 +131,7 @@ pub struct Key {
 impl Key {
     pub fn default_admin() -> Self {
         let now = OffsetDateTime::now_utc();
-        let uid = Uuid::from_u128(0);
+        let uid = Uuid::new_v4();
         Self {
             name: Some("Default Admin API Key".to_string()),
             description: Some("Use it for anything that is not a search operation. Caution! Do not expose it on a public frontend".to_string()),
@@ -146,7 +146,7 @@ impl Key {
 
     pub fn default_management() -> Self {
         let now = OffsetDateTime::now_utc();
-        let uid = Uuid::from_u128(1);
+        let uid = Uuid::new_v4();
         Self {
             name: Some("Default Read-Only Admin API Key".to_string()),
             description: Some("Use it to peek into the instance in a read-only mode. Caution! Do not expose it on a public frontend. It would give access to all other keys".to_string()),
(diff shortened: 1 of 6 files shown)


crates/meilisearch-types/src/keys.rs
```diff
@@ -405,6 +468,7 @@ pub mod actions {
     use super::Action::*;
 
     pub(crate) const ALL: u8 = All.repr();
+    pub const ALL_READ: u8 = AllGet.repr();
```
martin-g: `read` vs `get` - is this naming inconsistency intentional ?

crates/meilisearch-types/src/keys.rs
```diff
@@ -144,6 +144,21 @@ impl Key {
...
+            name: Some("Default Read-Only Admin API Key".to_string()),
+            description: Some("Use it to peek into the instance in a read-only mode.".to_string()),
+            uid,
+            actions: vec![Action::AllGet, Action::KeysGet],
+            indexes: vec![IndexUidPattern::all()],
+            expires_at: None,
+            created_at: now,
+            updated_at: now,
+        }
+    }
```
irevoire: Huum, since I think we should name the key the same way internally and externally with the `Read-Only admin` term.

Also, the description should be clearer on the fact that you should not share this key more than the original admin key.

I wrote a small sentence, but @guimachiavelli, you may be interested in reviewing the description here.
Iirc it's exposed on the cloud as well with its description, so people are going to see it even if they don't read the documentation at all.

```suggestion
    pub fn default_read_only_admin_key() -> Self {
        let now = OffsetDateTime::now_utc();
        let uid = Uuid::new_v4();
        Self {
            name: Some("Default Read-Only Admin API Key".to_string()),
            description: Some("Use it to peek into the instance in a read-only mode. Caution: This key gives you access to all the other api keys. Do not expose it on a public frontend.".to_string()),
            uid,
            actions: vec![Action::AllGet, Action::KeysGet],
            indexes: vec![IndexUidPattern::all()],
            expires_at: None,
            created_at: now,
            updated_at: now,
        }
    }
```

crates/meilisearch-types/src/keys.rs
```diff
@@ -144,6 +144,21 @@ impl Key {
         }
     }
 
+    pub fn default_read_only_admin() -> Self {
+        let now = OffsetDateTime::now_utc();
+        let uid = Uuid::new_v4();
+        Self {
+            name: Some("Default Read-Only Admin API Key".to_string()),
+            description: Some("Use it to peek into the instance in a read-only mode. Caution: This key gives you access to all the other api keys. Do not expose it on a public frontend".to_string()),
```
guimachiavelli:
```suggestion
            description: Some("Use it to read information across the whole database. Caution! Do not expose this key on a public frontend".to_string()),
```

PR #5701, opened 2025-06-23
Fixes #5669 - Enable `chatCompletions` experimental feature via CLI and ENV

### LINKED ISSUE #5669: update config.toml file to reflect chatCompletion experimental feature
I'd like to enable some experimental features at startup, without dispatching an enabling request.

I assume there's some way to do that in the config.toml file, can you update it (commented out, like the others) to reflect that.

Thx.

martin-g:
# Pull Request

## Related issue
Fixes #5669 

## What does this PR do?
Currently the new `chatCompletions` feature could be enabled only via
```
curl \
  -X PATCH 'MEILISEARCH_URL/experimental-features/' \
  --data-binary '{
    "chatCompletions": true
  }'
```
https://www.meilisearch.com/docs/reference/api/chats#chat-completions-workspace-object 

Now it will be possible to use
1) command line argument: `--experimental_chat_completions` 
2) or environment variable: `export MEILI_EXPERIMENTAL_CHAT_COMPLETIONS=true`

## PR checklist
- [X] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [X] Have you read the contributing guidelines?
- [X] Have you made sure that the title is accurate and descriptive of the changes?

### DIFF
--- a/crates/index-scheduler/src/lib.rs
+++ b/crates/index-scheduler/src/lib.rs
@@ -45,6 +45,9 @@ use std::path::{Path, PathBuf};
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 
+use crate::features::FeatureData;
+use crate::index_mapper::IndexMapper;
+use crate::utils::clamp_to_page_size;
 use dump::Dump;
 pub use error::Error;
 pub use features::RoFeatures;
@@ -70,9 +73,6 @@ use roaring::RoaringBitmap;
 use scheduler::Scheduler;
 use time::OffsetDateTime;
 use versioning::Versioning;
-
-use crate::index_mapper::IndexMapper;
-use crate::utils::clamp_to_page_size;
 
 pub(crate) type BEI128 = I128<BE>;
 
@@ -136,6 +136,8 @@ pub struct IndexSchedulerOptions {
     pub embedding_cache_cap: usize,
     /// Snapshot compaction status.
     pub experimental_no_snapshot_compaction: bool,
+    /// Whether the chat completions are enabled or not.
+    pub experimental_chat_completions: bool,
 }
 
 /// Structure which holds meilisearch's indexes and schedules the tasks
@@ -273,7 +275,7 @@ impl IndexScheduler {
                 .open(&options.tasks_path)
         }?;
 
-        // We **must** starts by upgrading the version because it'll also upgrade the required database before we can open them
+        // We **must** start by upgrading the version because it'll also upgrade the required database before we can open them
         let version = versioning::Versioning::new(&env, from_db_version)?;
 
         let mut wtxn = env.write_txn()?;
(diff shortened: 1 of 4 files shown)


crates/index-scheduler/src/lib.rs
```diff
@@ -922,6 +927,22 @@ impl IndexScheduler {
...
+) -> Result<()> {
+    let current_features = features.features().runtime_features();
+
+    let new_features = RuntimeTogglableFeatures {
+        chat_completions: options.experimental_chat_completions,
+        ..current_features
+    };
+
+    let wtxn = env.write_txn().map_err(Error::HeedTransaction)?;
+    features.put_runtime_features(wtxn, new_features)
```
martin-g: This method accepts `mut`, not `&mut`, and this this is the reason to create a new `write_txn` instead of using the one at line 287.

martin-g:
There is an issue in the current implementation (this PR):

1. Enabling the feature via the REST API stores a setting in the DB (runtime features).  (this is old logic)
2. With the new code I also use the value read from the config file / env var / cli argument (the `config`).
3. So, if the feature is enabled via REST API, a restart of Meilisearch will overwrite the stored value with the one from the config, i.e. `false` ...

I do not see any support for Setting::NotSet for the values coming from the config. I'll dig deeper but any hints are welcome!

martin-g:
Adding the new setting to `InstanceTogglableFeatures` seems to fix the problem!

config.toml
```diff
@@ -132,3 +132,6 @@ experimental_reduce_indexing_memory_usage = false
 
 # Experimentally reduces the maximum number of tasks that will be processed at once, see: <https://github.com/orgs/meilisearch/discussions/713>
 # experimental_max_number_of_batched_tasks = 100
+
+# Experimental chat completions feature, see: <https://www.meilisearch.com/docs/reference/api/chats>
+experimental_chat_completions = false
```
irevoire, line 137: Oof I'm just realizing that we forgot so many features here 🥶

crates/meilisearch/src/analytics/segment_analytics.rs
```diff
@@ -281,6 +281,7 @@ impl Infos {
             indexer_options,
             config_file_path,
             no_analytics: _,
+            experimental_chat_completions: _,
```
irevoire: Actually, we want to use this one.
Could you move it above with all the other experimental features, please?

And then, below, let's say around line 327, you should do a `|` between the activation coming from the runtime feature and the one coming from the cli

crates/meilisearch/src/option.rs
```diff
@@ -475,6 +476,11 @@ pub struct Opt {
     /// Format must be TOML.
     #[clap(long)]
     pub config_file_path: Option<PathBuf>,
+
+    /// Enables the chat completions functionality.
+    #[serde(default)]
+    #[clap(long, env = MEILI_EXPERIMENTAL_CHAT_COMPLETIONS, default_value_t = false)]
+    pub experimental_chat_completions: bool,
```
irevoire:
```suggestion

    /// Enables the chat completions functionality.
    ///
    /// When enabled, Meilisearch will enable multiple new routes to dialog with external LLM.
    ///
    /// For more information, see <https://github.com/orgs/meilisearch/discussions/835>
    #[serde(default)]
    #[clap(long, env = MEILI_EXPERIMENTAL_CHAT_COMPLETIONS, default_value_t = false)]
    pub experimental_chat_completions: bool,
```

Also, could you move it up a bit with the other experimental features, please?

config.toml
martin-g: I may add them in a follow up PR if you like !

PR #5703, opened 2025-06-23
tests: Use Server::wait_task() instead of Index::wait_task()

martin-g:
# Pull Request

## Related issue
#4840 

## What does this PR do?
- The code is mostly duplicated. Server::wait_task() has better handling for errors and more retries.

### DIFF
--- a/crates/meilisearch/tests/common/server.rs
+++ b/crates/meilisearch/tests/common/server.rs
@@ -407,19 +407,12 @@ impl<State> Server<State> {
     }
 
     pub async fn wait_task(&self, update_id: u64) -> Value {
-        Server::<Shared>::_wait_task(async |url| self.service.get(url).await, update_id).await
-    }
-
-    pub(super) async fn _wait_task<F>(request_fn: F, update_id: u64) -> Value
-    where
-        F: AsyncFnOnce(String) -> (Value, StatusCode) + Copy,
-    {
         // try several times to get status, or panic to not wait forever
         let url = format!("/tasks/{update_id}");
         let max_attempts = 400; // 200 seconds in total, 0.5secs per attempt
 
         for i in 0..max_attempts {
-            let (response, status_code) = request_fn(url.clone()).await;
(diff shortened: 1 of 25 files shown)


crates/meilisearch/tests/common/server.rs
```diff
@@ -407,13 +407,20 @@ impl<State> Server<State> {
     }
 
     pub async fn wait_task(&self, update_id: u64) -> Value {
+        Server::<Shared>::_wait_task(async |url| self.service.get(url).await, update_id).await
+    }
+
+    pub(super) async fn _wait_task<F>(request_fn: F, update_id: u64) -> Value
```
martin-g: This method is needed because `common::index` needs to use it.
I guess this was the reason for the duplicated method in both files.

irevoire: But doesn't the common::Index contains a Server to make his request against?

martin-g: The Index looks like https://github.com/meilisearch/meilisearch/blob/bd2bd0f33baff650f475a78d1070ada296413135/crates/meilisearch/tests/common/index.rs#L15-L20
There is no `server` around to be used...

martin-g:
> why we're still stuck with this method in the Index

Few Index methods need to wait for tasks.
For example https://github.com/meilisearch/meilisearch/blob/bd2bd0f33baff650f475a78d1070ada296413135/crates/meilisearch/tests/common/index.rs#L36-L50
waits for the `test_set` JSON[NL] to be fully indexed before returning.

Another option:
These methods are helpers, so we could just pass `&server` as an additional argument and use it (`server.wait_task(...)`

martin-g:
> Another option:
> These methods are helpers, so we could just pass `&server` as an additional argument and use it (`server.wait_task(...)`

@irevoire Here is the alternative impl I proposed above - https://github.com/meilisearch/meilisearch/pull/5703/commits/ae912c4c3f2d0065f2251299e0c34cf588fc2e20

Issue #5684, opened 2025-06-17
Chat completions with gemini


scatto98:
Hi,

Setting up chat completions with Gemini, the POST /chats/{workspace}/chat/completions API returns this.

**Screenshots**

![Image](https://github.com/user-attachments/assets/cb3f34ea-599a-418a-8008-ed9f2a34976c)

**Meilisearch version:**
1.15.2

Kerollmops:
Hey @scatto98 👋

Thank you for reporting the issue. Can you please show me the parameters of your chat settings? It seems that Gemini is unfortunately incompatible with OpenAI, as [their documentation explains](https://ai.google.dev/gemini-api/docs/openai). I'll take some time next week to look into this.

However, if someone is willing to help with this issue, I will label it `good-first-issue`. I'm guessing the main problem is deserializing the type Gemini returns [in this part of the code](https://github.com/meilisearch/meilisearch/blob/9a729bf31d7a63478ef50f8858a46e827cbf3f5a/crates/meilisearch/src/routes/chats/chat_completions.rs#L499). We should probably use the byot feature of the `async-openai` SDK.

Have a nice day 🍽

scatto98:
Hi @Kerollmops,
I've attached the chat settings. 

![Image](https://github.com/user-attachments/assets/d8df3fa8-c672-4716-abb4-cef4cbe8f055)

Have a nice day 👍

diksipav:
I think one of the issues here is that the Gemini request is sent to "https://generativelanguage.googleapis.com/v1beta/openai//chat/completions" (openai//chat instead of just one "/"). I'm debugging.

I think I fixed it, just need to create the PR.

diksipav:
I believe the last slash should be removed from [here](https://github.com/diksipav/meilisearch/blob/main/crates/meilisearch-types/src/features.rs#L157).
With just that change I get the response as in the screenshot.

<img width="817" alt="Image" src="https://github.com/user-attachments/assets/68227ad5-8cd6-4cfb-8f6c-5b14df17536f" />

Kerollmops:
Hey 👋

Unfortunately, it's still not good on my side. By performing the following request to Meilisearch I still get an error. The issue is probably around different types and failing to be parsed by the async-openai library. We should [look at this example](https://github.com/64bit/async-openai/tree/main/examples/gemini-openai-compatibility) to make it work with more LLM providers.

```bash
curl -X POST 'http://localhost:7700/chats/default/chat/completions' \
[...]

Issue #5634, opened 2025-06-04
Chat route improvement and follow up


Kerollmops:
This is a tracking issue to follow up on [the chat completions route](<https://meilisearch.notion.site/Feature-Meilisearch-chat-route-usage-1f84b06b651f80eab264f8821e9f3a00>). We released the chat route in [#5556](<https://github.com/meilisearch/meilisearch/issues/5556>). I'll list the different concerns and fixes to introduce to strengthen the usage of the feature:

- [ ] Store chat workspace settings in the dumps
- [ ] Make sure we can rollback to pre-PR (chat settings in index settings)

### Support more LLM-providers

- [ ] Reintroduce the support of Gemini.
- [ ] Support Qwen3
- [ ] Support DeepSeek
- [ ] Support a generic interface

### Improve the documentation

Some users do not understand some parts of the documentation ([private link](<https://discord.com/channels/@me/1383127932788277308/1383338629123801118>)).

### Add more analytics

We should add enough analytics to tell who's using our feature and how often.

- [ ] Count the tokens and expose them in the stats (or somewhere else).
- [ ] Add analytics when changing the `/chats` workspace settings.
- [X] [https://github.com/meilisearch/meilisearch/pull/5709](<https://github.com/meilisearch/meilisearch/pull/5709>)

### Improve the chat completions parameters

By accepting any JSON object as input and only requiring the `model`, `messages`, `stream`, and `tools` fields, we can support any chat completions API as long as they have compatible outputs.

- [ ] Document the compatible APIs.
- [ ] (idea) [Accept any field](<https://serde.rs/attr-flatten.html>) as input and only care about the interesting ones.
- [ ] Explain the input and output (mirror the underlying LLM).
- [ ] Make the tool calls compatible with the `source` setting (OpenAI, Mistral, Ollama, or Anthropic)
- [X] Allow incremental changes of the chat hybrid search settings [https://github.com/meilisearch/meilisearch/pull/5556#discussion_r2137368486](<https://github.com/meilisearch/meilisearch/pull/5556#discussion_r2137368486>).
- [ ] List a subset of allowed models to be used.

### Improve the non-streamed chat completions route

The chat completions route is very good, but only the streamed version. We must work on the non-streamed version to improve it.

- [ ] Ensure correct tool support.
- [ ] Ensure correct error report.
- [ ] Enable the route again.

### Improve the index choice speed

[...]

PR #5708, opened 2025-06-24
Remove Gemini from the LLM-providers list

Kerollmops:
This PR removes Gemini from the list of supported providers. The main reason is that the function/tool call support and error reporting are incompatible with OpenAI. It is **very** different. You can see that in https://github.com/meilisearch/meilisearch/issues/5684#issuecomment-2996668744.

I prefer removing the support* than having something broken. We will [reintroduce it later](https://github.com/meilisearch/meilisearch/issues/5634).


dureuill (approved):
wise decision

PR #5709, opened 2025-06-24
Add analytics to the chat completions

Kerollmops:
This PR adds new analytics (that can be disabled) to the chat completions route. You can find an example of the analytics on [this page (internal)](https://app.amplitude.com/analytics/meili/project/256223/search/amplitude_id%3D571576966537?_source=direct%20link&eventId=bfd795c2-1e9d-4fd0-9e6f-9ed832cfdeae&eventHandle=Zeib%2FNd_1BQRD_-B_Zeib%2FNd).

```json
"event_properties": {
  "avg_messages_per_request": 2,
  "models_used": {
      "gpt-3.5-turbo": 1
  },
  "requests.total_received": 1,
  "streaming_ratio": 1,
  "time_spent.avg": 0,
  "time_spent.max": 0,
  "time_spent.min": 0,
  "total_messages": 2,
  "total_non_streamed_requests": 0,
  "total_received": 1,
  "total_streamed_requests": 1,
  "total_succeeded": 1,
  "user-agent": [
    "yaak"
  ]
}
```

### DIFF
--- a/crates/meilisearch/src/routes/chats/chat_completion_analytics.rs
+++ b/crates/meilisearch/src/routes/chats/chat_completion_analytics.rs
@@ -2,15 +2,10 @@ use std::collections::BinaryHeap;
 
 use serde_json::{json, Value};
 
-use crate::aggregate_methods;
-use crate::analytics::{Aggregate, AggregateMethod};
-
-aggregate_methods!(
-    ChatCompletionPOST => "Chat Completion POST",
-);
+use crate::analytics::Aggregate;
 
 #[derive(Default)]
-pub struct ChatCompletionAggregator<Method: AggregateMethod> {
+pub struct ChatCompletionAggregator {
     // requests
     total_received: usize,
     total_succeeded: usize,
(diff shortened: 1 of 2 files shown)


crates/meilisearch/src/routes/chats/chat_completion_analytics.rs
```diff
@@ -0,0 +1,139 @@
...
+use crate::aggregate_methods;
+use crate::analytics::{Aggregate, AggregateMethod};
+
+aggregate_methods!(
+    ChatCompletionPOST => "Chat Completion POST",
+);
+
+#[derive(Default)]
+pub struct ChatCompletionAggregator<Method: AggregateMethod> {
+    // requests
```
irevoire:
```suggestion
#[derive(Default)]
pub struct ChatCompletionAggregator {
    // requests
```

If we're not adding any other method, I think we can get rid of that and just do a normal Aggregate, no?

crates/meilisearch/src/routes/chats/chat_completion_analytics.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub fn succeed(&mut self, time_spent: std::time::Duration) {
+        self.total_succeeded += 1;
+        self.time_spent.push(time_spent.as_millis() as usize);
+    }
+}
+
+impl<Method: AggregateMethod> Aggregate for ChatCompletionAggregator<Method> {
+    fn event_name(&self) -> &'static str {
+        Method::event_name()
+    }
```
irevoire:
```suggestion
impl Aggregate for ChatCompletionAggregator {
    fn event_name(&self) -> &'static str {
        "Chat Completion POST"
    }
```

irevoire (dismissed):
Hey, I don't think we need to use the complex version of the aggregator if we're not going to have multiple implementations for the same payload. What do you think?

crates/meilisearch/src/routes/chats/chat_completion_analytics.rs
Kerollmops: Ho yeah, you're right! Thank you 🙏

PR #5710, opened 2025-06-24
Introduce filters in the chat completions

Kerollmops:
This PR adds filter support in the chat completions route. It is very basic and driven only by a small documentation of the filter syntax. I want to improve the support by adding a list of the filterable attributes and some example values.

## To Do
- [x] [Add examples of filters in the parameter description](https://learnprompting.org/fr/docs/basics/few_shot)
- [ ] (optional) Use templating in the parameter descriptions


crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -280,14 +300,23 @@ async fn process_search_request(
...
 
-        search_from_kind(index_uid, search_kind, search)
-            .map(|(search_results, _)| (rtxn, search_results))
-            .map_err(ResponseError::from)
+        match search_from_kind(index_uid, search_kind, search) {
+            Ok((search_results, _)) => Ok((rtxn, Ok(search_results))),
+            Err(MeilisearchHttpError::Milli {
+                error: meilisearch_types::milli::Error::UserError(user_error),
+                index_name: _,
+            }) => Ok((rtxn, Err(user_error))),
```
ManyTheFish, line 308: This allows the LLM to get the error and correct the query, am I right?

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -801,4 +834,42 @@ struct SearchInIndexParameters {
...
+    let filterable_attributes = fields_ids_map
+        .names()
+        .filter(|name| rules.iter().any(|rule| matches!(rule.match_str(name), PatternMatch::Match)))
+        .map(|name| (name, OrderBy::Count));
+    let facets_distribution = index
+        .facets_distribution(rtxn)
+        .max_values_per_facet(max_values_per_facet)
+        .candidates(universe)
+        .facets(filterable_attributes)
+        .execute()?;
```
ManyTheFish, line 858: is computing the facet count necessary ?

crates/meilisearch/src/routes/chats/chat_completions.rs
```diff
@@ -801,4 +834,42 @@ struct SearchInIndexParameters {
...
+        let total_entries = entries.len();
+        for (i, (value, _count)) in entries.into_iter().enumerate() {
+            let _ = if total_entries.saturating_sub(1) == i {
+                write!(&mut output, "{value}.")
+            } else {
+                write!(&mut output, "{value}, ")
+            };
+        }
+        let _ = writeln!(&mut output);
+    }
```
ManyTheFish, line 872: would it make sense to only retrieve the exhaustive list for facet string?
For facet number we may want to print the min and the max as we do in the [facetStats](https://www.meilisearch.com/docs/reference/api/search#facets)

ManyTheFish (commented):
I have some question before merging the PR ☺️

crates/meilisearch/src/routes/chats/chat_completions.rs
Kerollmops: It's important to have the most important ones first in description.

crates/meilisearch/src/routes/chats/chat_completions.rs
Kerollmops: For facet strings it could be too much for the LLM context.

And for the numbers, yes, you are absolutely right!

PR #5714, opened 2025-06-26
Use a patch version of arroy to disable SIMD

Kerollmops:
A prototype to disable SIMD in arroy in v1.14.0. There is a prototype available for this:

```
docker pull getmeili/meilisearch:prototype-no-simd-x86-arroy-0
```


Kerollmops:
Closing as do not seems to make any difference.

PR #5718, opened 2025-06-27
Add `/indexes/{indexUID}/fields` route to get all field names

qdequele:
# Summary
This PR adds a new dedicated route `/indexes/{indexUID}/fields` that returns a complete list of all field names in an index, including nested fields. This provides a lightweight alternative to the stats route for clients that only need field information without the overhead of field distribution calculations.

# Changes Made

## 1. Route Details
- **Endpoint**: `GET /indexes/{indexUid}/fields`
- **Response**: Array of strings containing all field names (e.g., `["id", "title", "user.name", "user.email"]`)
- **Authentication**: Requires `indexes.get` permission
- **Error Handling**: Returns 404 for non-existent indexes with proper error format

## 2. Implementation Details
- Uses `FieldsIdsMap::names()` to get all field names including nested fields
- Returns field names in dot notation format (e.g., `user.profile.age`)
- No performance impact on indexing as it only reads existing field metadata
- Includes both top-level and deeply nested fields

## 3. Testing
- **Added** `fields()` method to test utilities in `crates/meilisearch/tests/common/index.rs`
- **Added** comprehensive test suite in `crates/meilisearch/tests/index/stats.rs`:
  - Basic functionality test with simple nested fields
  - Complex nested fields test with up to 6 levels of nesting
  - Error handling test for non-existent indexes
  - Verification of field name format and uniqueness

# Example Usage
```bash
# Get all fields from an index
GET /indexes/movies/fields

# Response
[
  "id",
  "title", 
  "user.name",
  "user.email",
  "metadata.category",
  "metadata.author.name"
]
```

# Breaking Changes
None - this is a new route that doesn't affect existing functionality.

### DIFF
--- a/crates/meilisearch/src/routes/indexes/mod.rs
+++ b/crates/meilisearch/src/routes/indexes/mod.rs
@@ -73,6 +73,7 @@ pub fn configure(cfg: &mut web::ServiceConfig) {
                     .route(web::delete().to(SeqHandler(delete_index))),
             )
             .service(web::resource("/stats").route(web::get().to(SeqHandler(get_index_stats))))
+            .service(web::resource("/fields").route(web::get().to(SeqHandler(get_index_fields))))
             .service(web::scope("/documents").configure(documents::configure))
             .service(web::scope("/search").configure(search::configure))
             .service(web::scope("/facet-search").configure(facet_search::configure))
@@ -579,4 +580,48 @@ pub async fn get_index_stats(
 
     debug!(returns = ?stats, "Get index stats");
     Ok(HttpResponse::Ok().json(stats))
+}
+
+/// Get fields of index
+///
+/// Get all field names in the index (including nested fields).
+#[utoipa::path(
+    get,
+    path = "/{indexUid}/fields",
+    tag = "Indexes",
+    security(("Bearer" = ["indexes.get", "indexes.*", "*"])),
+    params(("indexUid", example = "movies", description = "Index Unique Identifier", nullable = false)),
+    responses(
+        (status = OK, description = "The fields of the index", body = Vec<String>, content_type = "application/json", example = json!(
+            ["id", "title", "user.name", "user.email", "metadata.category"]
+        )),
+        (status = 404, description = "Index not found", body = ResponseError, content_type = "application/json", example = json!(
+            {
+                "message": "Index `movies` not found.",
+                "code": "index_not_found",
+                "type": "invalid_request",
+                "link": "https://docs.meilisearch.com/errors#index_not_found"
+            }
+        )),
+        (status = 401, description = "The authorization header is missing", body = ResponseError, content_type = "application/json", example = json!(
+            {
+                "message": "The Authorization header is missing. It must use the bearer authorization method.",
+                "code": "missing_authorization_header",
+                "type": "auth",
+                "link": "https://docs.meilisearch.com/errors#missing_authorization_header"
+            }
+        )),
+    )
+)]
+pub async fn get_index_fields(
+    index_scheduler: GuardedData<ActionPolicy<{ actions::INDEXES_GET }>, Data<IndexScheduler>>,
+    index_uid: web::Path<String>,
+) -> Result<HttpResponse, ResponseError> {
+    let index_uid = IndexUid::try_from(index_uid.into_inner())?;
+    let index = index_scheduler.index(&index_uid)?;
+    let rtxn = index.read_txn()?;
+    let fields: Vec<String> = index.fields_ids_map(&rtxn)?.names().map(|s| s.to_string()).collect();
+
+    debug!(returns = ?fields, "Get index fields");
+    Ok(HttpResponse::Ok().json(fields))
 }
--- a/crates/meilisearch/tests/index/stats.rs
+++ b/crates/meilisearch/tests/index/stats.rs
@@ -60,3 +60,265 @@ async fn error_get_stats_unexisting_index() {
     assert_eq!(response, expected_response);
     assert_eq!(code, 404);
 }
+
+#[actix_rt::test]
+async fn fields() {
+    let server = Server::new_shared();
+    let index = server.unique_index();
+    let (task, code) = index.create(Some("id")).await;
+
+    assert_eq!(code, 202);
+    server.wait_task(task.uid()).await.succeeded();
+
+    // Test empty index
+    let (response, code) = index.fields().await;
+    assert_eq!(code, 200);
+    assert!(response.as_array().unwrap().is_empty());
+
+    // Test with documents containing nested fields
+    let documents = json!([
+        {
+            "id": 1,
+            "name": "John",
+            "user": {
+                "email": "john@example.com",
+                "profile": {
+                    "age": 30,
+                    "location": "Paris"
+                }
+            },
+            "tags": ["developer", "rust"]
+        },
+        {
+            "id": 2,
+            "title": "Article",
+            "metadata": {
+                "category": "tech",
+                "author": {
+                    "name": "Jane",
+                    "id": 123
+                }
+            }
+        }
+    ]);
+
+    let (response, code) = index.add_documents(documents, None).await;
+    assert_eq!(code, 202);
+    server.wait_task(response.uid()).await.succeeded();
+
+    // Test fields including nested fields
+    let (response, code) = index.fields().await;
+    assert_eq!(code, 200);
+    
+    let fields = response.as_array().unwrap();
+    let field_names: Vec<&str> = fields.iter()
+        .map(|f| f.as_str().unwrap())
+        .collect();
+    
+    // Check that all expected fields are present (including nested fields)
+    assert!(field_names.contains(&"id"));
+    assert!(field_names.contains(&"name"));
+    assert!(field_names.contains(&"title"));
+    assert!(field_names.contains(&"user.email"));
+    assert!(field_names.contains(&"user.profile.age"));
+    assert!(field_names.contains(&"user.profile.location"));
+    assert!(field_names.contains(&"tags"));
+    assert!(field_names.contains(&"metadata.category"));
+    assert!(field_names.contains(&"metadata.author.name"));
+    assert!(field_names.contains(&"metadata.author.id"));
+    
+    // Verify the response is a simple array of strings
+    for field in fields {
+        assert!(field.is_string());
+    }
+}
+
+#[actix_rt::test]
+async fn fields_nested_complex() {
+    let server = Server::new_shared();
+    let index = server.unique_index();
+    let (task, code) = index.create(Some("id")).await;
+
+    assert_eq!(code, 202);
+    server.wait_task(task.uid()).await.succeeded();
+
+    // Test with complex deeply nested structures
+    let documents = json!([
+        {
+            "id": 1,
+            "product": {
+                "name": "Laptop",
+                "specifications": {
+                    "hardware": {
+                        "cpu": {
+                            "brand": "Intel",
+                            "model": "i7-12700K",
+                            "cores": {
+                                "physical": 12,
+                                "logical": 20
+                            }
+                        },
+                        "memory": {
+                            "ram": 16,
+                            "type": "DDR4"
+                        },
+                        "storage": {
+                            "primary": {
+                                "type": "SSD",
+                                "capacity": 512
+                            },
+                            "secondary": {
+                                "type": "HDD",
+                                "capacity": 1000
+                            }
+                        }
+                    },
+                    "software": {
+                        "os": "Windows 11",
+                        "applications": ["Chrome", "VS Code", "Docker"]
+                    }
+                },
+                "pricing": {
+                    "base": 1299.99,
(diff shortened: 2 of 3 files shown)


crates/meilisearch/src/routes/indexes/mod.rs
```diff
@@ -580,3 +581,47 @@ pub async fn get_index_stats(
...
+    )
+)]
+pub async fn get_index_fields(
+    index_scheduler: GuardedData<ActionPolicy<{ actions::INDEXES_GET }>, Data<IndexScheduler>>,
+    index_uid: web::Path<String>,
+) -> Result<HttpResponse, ResponseError> {
+    let index_uid = IndexUid::try_from(index_uid.into_inner())?;
+    let index = index_scheduler.index(&index_uid)?;
+    let rtxn = index.read_txn()?;
+    let fields: Vec<String> = index.fields_ids_map(&rtxn)?.names().map(|s| s.to_string()).collect();
```
ManyTheFish: I suggest using a [FieldIdMapWithMetaData](https://github.com/meilisearch/meilisearch/blob/3aa6c3c7505561684d5df47b657579c945d0c439/crates/milli/src/fields_ids_map/metadata.rs#L15-L38), this struct will provide all the metadata linked to a field. We'll just have to display them, see the below code to build it:
https://github.com/meilisearch/meilisearch/blob/3aa6c3c7505561684d5df47b657579c945d0c439/crates/meilisearch/src/routes/chats/utils.rs#L223-L225

Then to get the related filterable attributes rule or localized attribute rule, you'll have to fetch both structs from the index like:
https://github.com/meilisearch/meilisearch/blob/3aa6c3c7505561684d5df47b657579c945d0c439/crates/milli/src/update/new/indexer/post_processing.rs#L168-L169

ManyTheFish:
Hey @qdequele,

After discussing with the team, we could accept a minimal implementation that doesn't contain the field Metadata.
However, for future proofing, we'd like the below changes before merging the PR:
1) the simple array of string must be transformed into an array of object containing a `name` field like `[{"name": "title"}, ..]`, this will allow us to add metadata related to the field in the future.
2) instead of a GET query, we'd prefer having a POST query allowing us to extend the request body easily.
3) the response should be paginated, we can have up to 16535 fields in a single index, fortunatelly it's a rare case, but it remains possible ☺️ 

The keys API use an [autopagination helper](https://github.com/meilisearch/meilisearch/blob/e8b2bb3ea67b8c53c8d821234d58005872b9e5df/crates/meilisearch/src/routes/mod.rs#L208-L264), see below:
https://github.com/meilisearch/meilisearch/blob/7dfb2071b5720563da82b54b629d9d8f78a5688c/crates/meilisearch/src/routes/api_key.rs#L189-L193

crates/meilisearch/src/routes/indexes/mod.rs
ManyTheFish: to ignore in favor of https://github.com/meilisearch/meilisearch/pull/5718#issuecomment-3023156207

crates/meilisearch/src/routes/indexes/fields.rs
```diff
@@ -0,0 +1,328 @@
...
+use serde::Serialize;
+use utoipa::{IntoParams, ToSchema};
+
+use crate::extractors::authentication::policies::*;
+use crate::extractors::authentication::GuardedData;
+
+use super::{Pagination, PAGINATION_DEFAULT_LIMIT};
+
+/// Field configuration for a specific field in the index
+#[derive(Debug, Serialize, Clone, ToSchema)]
```
Mubelotix, line 27: All the structures you added with `ToSchema` need to be added to the big list of structures [in the main OpenApi structure](https://github.com/meilisearch/meilisearch/blob/f2185438eed60fa32d25b15480c5ee064f6fba4a/crates/meilisearch/src/routes/mod.rs#L88).

crates/meilisearch/src/routes/indexes/fields.rs
```diff
@@ -0,0 +1,328 @@
...
+#[serde(rename_all = "camelCase")]
```
Mubelotix, line 28: Also to be added on all other schema-deriving structures

```suggestion
#[serde(rename_all = "camelCase")]
#[schema(rename_all = "camelCase")]
```

crates/meilisearch/src/routes/indexes/fields.rs
```diff
@@ -0,0 +1,328 @@
...
+    pub limit: Param<usize>,
+    #[param(value_type = Option<String>, example = "displayed.enabled = true && localized.locales : fra")]
+    #[deserr(default, error = DeserrQueryParamError<InvalidIndexFieldsFilter>)]
+    pub filter: Option<String>,
+}
+
+impl ListFields {
+    fn into_pagination(self) -> Pagination {
+        Pagination { offset: self.offset.0, limit: self.limit.0 }
+    }
```
Mubelotix, line 104: No reason to take ownership here, it forces you to clone the fields later on

```suggestion
    fn to_pagination(&self) -> Pagination {
        Pagination { offset: *self.offset.0, limit: *self.limit.0 }
    }
```

crates/meilisearch/src/routes/indexes/fields.rs
```diff
@@ -0,0 +1,328 @@
...
+            let path: Vec<String> = lhs.trim().split('.').map(|s| s.trim().to_string()).collect();
+            let value = matches!(rhs.trim(), "true" | "True" | "TRUE");
+            conditions.push(FilterCond::BoolEq { path, value });
+        } else if let Some((lhs, rhs)) = cond_str.split_once(':') {
+            let path: Vec<String> = lhs.trim().split('.').map(|s| s.trim().to_string()).collect();
+            conditions.push(FilterCond::Contains { path, value: rhs.trim().to_string() });
+        }
+    }
+    conditions
+}
```
Mubelotix, line 130: Can't fields contain ampersands, points or colons? If they can we need to provide a way to escape the path. I had this exact issue with embedder names last week and switching from a splitted iterator to a closure yielding the next potentially-escaped part was very easy https://github.com/meilisearch/meilisearch/blob/6113831eb517e620313f6f945770a679ef446b5b/crates/meilisearch/src/routes/indexes/render.rs#L400-L417

crates/meilisearch/src/routes/indexes/fields.rs
```diff
@@ -0,0 +1,328 @@
...
+
+fn field_satisfies(field: &Field, conds: &[FilterCond]) -> bool {
+    let field_value = serde_json::to_value(field).unwrap_or_default();
```
Mubelotix, line 133: Not fond of having to serialize every field metadata. Something like that would be more efficient

```rust
fn field_satisfies(field: &Field, conds: &[FilterCond]) -> bool {
    conds.iter().all(|cond| -> bool {
        match cond {
            FilterCond::BoolEq { path, value } => {
                match path.join(".").as_str() {
                    "displayed.enabled" => field.displayed.enabled == *value,
                    "searchable.enabled" => field.searchable.enabled == *value,
                    "distinct.enabled" => field.distinct.enabled == *value,
                    "filterable.enabled" => field.filterable.enabled == *value,
                    "filterable.filter.equality" => field.filterable.filter.equality == *value,
                    "filterable.filter.comparison" => field.filterable.filter.comparison == *value,
                    _ => false,
                }
            }
            FilterCond::Contains { path, value } => {
                match path.join(".").as_str() {
                    "localized.locales" => field.localized.locales.iter().any(|l| l.contains(value)),
                    "filterable.facet_search.sort_by" => field.filterable.facet_search.sort_by.contains(value),
                    "name" => field.name.contains(value),
                    locale if locale.starts_with("localized.locales.") => {
                        let locale = locale.strip_prefix("localized.locales.").unwrap();
                        field.localized.locales.iter().find(|l| l == locale).is_some_and(|l| l.contains(value))
                    }
                    _ => false,
                }
            }
        }
    })
}
```

We could add a test to ensure that we don't forget to add new fields here

crates/meilisearch/src/routes/indexes/fields.rs
```diff
@@ -0,0 +1,328 @@
...
+    if let Some(stripped) = search_pattern.strip_suffix('*') {
+        field_name.starts_with(stripped)
+    } else if let Some(stripped) = search_pattern.strip_prefix('*') {
+        field_name.ends_with(stripped)
+    } else if search_pattern.contains('*') {
+        let parts: Vec<&str> = search_pattern.split('*').collect();
+        if parts.len() == 2 {
+            field_name.starts_with(parts[0]) && field_name.ends_with(parts[1])
+        } else {
+            field_name.contains(&search_pattern.replace('*', ""))
```
Mubelotix, line 190: Seems like this should be an error

crates/meilisearch/src/routes/indexes/fields.rs
```diff
@@ -0,0 +1,328 @@
...
+        Setting::Set(r) => r.iter().cloned().map(|r| r.into()).collect(),
+        _ => Vec::new(),
+    };
+    let facet_sort_config = match &settings.faceting {
+        Setting::Set(f) => match &f.sort_facet_values_by {
+            Setting::Set(c) => c.clone(),
+            _ => BTreeMap::new(),
+        },
+        _ => BTreeMap::new(),
+    };
```
Mubelotix, line 291: This doesn't seem reliable. It's just guessing which fields are what, but the actual behavior in the engine may differ from this logic.

The data actually exists and can be retrieved easily by using `index.fields_ids_map_with_metadata` instead of `index.fields_ids_map`. There should be everything in it [Metadata](https://github.com/meilisearch/meilisearch/blob/c37ed05f49e0b091794e6e951d20638926bafddc/crates/milli/src/fields_ids_map/metadata.rs#L16) and if anything is missing, there must be a way to access it directly from the index scheduler

crates/meilisearch/src/routes/indexes/fields.rs
```diff
@@ -0,0 +1,328 @@
...
+        })
+        .collect();
+
+    if let Some(expr) = &params.filter {
+        let conds = parse_filter(expr);
+        enriched_fields.retain(|f| field_satisfies(f, &conds));
+    }
+
+    let total = enriched_fields.len();
+    let pagination = (*params).clone().into_pagination();
```
Mubelotix, line 320: Remove the clone once into_pagination is turned to to_pagination

crates/meilisearch/src/routes/indexes/fields.rs
```diff
@@ -0,0 +1,328 @@
...
+                &localized_rules,
+                &facet_sort_config,
+            )
+        })
+        .collect();
+
+    if let Some(expr) = &params.filter {
+        let conds = parse_filter(expr);
+        enriched_fields.retain(|f| field_satisfies(f, &conds));
+    }
```
Mubelotix, line 317: Turning the iterator to a vec is a missed opportunity. Use filter on the iterator instead of retain on a vec, you will benefit from `skip` and `limit` from the pagination. The filter will be evaluated lazily and not all fields will be processed. Since filter isn't always set, you may need to use itertools' `Either`

crates/meilisearch/src/routes/indexes/mod.rs
```diff
@@ -125,7 +129,7 @@ pub struct ListIndexes {
 }
 
 impl ListIndexes {
-    fn as_pagination(self) -> Pagination {
+    fn into_pagination(self) -> Pagination {
```
Mubelotix, line 132: There is no need to take ownership. According to my interpretation of [Rust's naming conventions](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv), it should be `to_pagination(&self)`

crates/meilisearch/src/routes/indexes/mod.rs
```diff
@@ -580,3 +584,5 @@ pub async fn get_index_stats(
     debug!(returns = ?stats, "Get index stats");
     Ok(HttpResponse::Ok().json(stats))
 }
+
+// Field-related structs, helpers and the `get_index_fields` handler have been moved to `fields.rs`.
```
Mubelotix, line 588: Not sure what this is about

crates/meilisearch/tests/index/fields.rs
Mubelotix, line 1: Every test in this file should use `snapshot` instead of asserts. It will greatly simplify them

crates/meilisearch/tests/index/stats.rs
Mubelotix, line 1: Every test in this file should use snapshot instead of asserts. It will greatly simplify them

crates/meilisearch/src/routes/indexes/fields.rs
qdequele: What do you think in term of readability vs performance, because I don't have the impression that performance is needed in that case.

Mubelotix: Not sure what's best ; both ideas make sense

YoEight:
Replaced in favor of #6082

PR #5729, opened 2025-07-02
Take into account the allowed max memory of the container

### LINKED ISSUE #5616: Memory spike on deployment
**Describe the bug**
I've upgraded from 1.8 to 1.14, and hit a really nasty performance issue, which I can only contribute to the newer versions, not my own indexes, as nothing changed and I never seen this before.

My memory consumption is very low, under 2GB normally and was the case on re-deploys as well. But now, when I deploy the instance, it hits 4GB of memory and then goes to under 1GB. This is a huge problem, as I am unable to deploy to a smaller instance, because it just fails when going over the memory (let's say 2GB).

Is this issue known? I will try to rollback version by version to figure out when this happened, but it's really frustrating, as this is the issue on a production server, 

**To Reproduce**
I am trying to figure out the cause, but it could be due to how big the index is, specifically storing vectors

**Expected behavior**

**Screenshots**
Image
Image

**Meilisearch version:**
1.14.0

**Additional context**
This happens on [Render](https://render.com/) deploy

martin-g:
# Pull Request

## Related issue
Fixes #5616

## What does this PR do?
- When Meilisearch runs inside a container (e.g. Docker or Kubernetes) it may run with less max memory than the available on the host, e.g. `docker run --memory 1G ...`

## PR checklist
- [X] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [X] Have you read the contributing guidelines?
- [X] Have you made sure that the title is accurate and descriptive of the changes?


martin-g:
The test application I used to verify the changes: https://github.com/martin-g/meili-5616-memtest

mlh758:
Thank you @martin-g

mlh758:
I tried creating a Docker image based on this patch: https://hub.docker.com/r/osiris2918/meilisearch

~I deployed it to our k8s cluster with a 1 GB limit on memory and reindexed those documents I shared originally. It still got OOM killed.~

I'm dumb, forgot to reset the branch after I cloned.

mlh758:
Okay, rebuilt it from the correct branch and replaced the tag but it is still showing the issue. Do you have a k8s cluster to try it in?

martin-g:
Can you reproduce it with minikube, I.e. single node?
If YES then please share the steps and configuration files!

mlh758:
I'll see what I can do. I'll also try again with a new tag on my image in case GKE cached it instead of honoring the `imagePullPolicy: Always` setting. Replacing an image like that wouldn't be the first time it's caused me problems.

mlh758:
@martin-g A lesson in not mutating tags, sorry for that. GKE was indeed ignoring the image pull policy. I built a fresh one from your branch and tagged it and we re-tested the flow. Pods didn't OOM this time. Thank you for the fix.

emiliofernandes:
Is this going to be merged for 1.16 ?
I have the feeling that it won't be :-(

As explained at https://github.com/meilisearch/meilisearch/issues/5616 it is a real issue for containerized deployments!

mlh758:
We're currently deployed with a patched Meilisearch that incorporates this PR. I'd love to to be able to switch back so we can upgrade to 1.16 and now apparently 1.18 is out. @Kerollmops I saw in the original issue you recommended people upgrade to latest but without this PR in place the behavior isn't improved.

Kerollmops (approved):

Kerollmops:
Hey @mlh758, you are absolutely right. Thanks, @martin-g, for this PR. I'm sorry this one slipped from my radar.

martin-g:
Welcome!

mlh758:
Thank you both!

PR #5731, opened 2025-07-03
Export and import chat completions workspace settings in dumps

Kerollmops:
This PR adds support for the chat completions settings in the dumps. Before this PR, the support was partial, and the chat completions settings (workspace settings) were skipped. With this PR, everything is correctly exported and imported.

### DIFF
--- a/crates/index-scheduler/src/scheduler/process_dump_creation.rs
+++ b/crates/index-scheduler/src/scheduler/process_dump_creation.rs
@@ -43,7 +43,16 @@ impl IndexScheduler {
 
         let rtxn = self.env.read_txn()?;
 
-        // 2. dump the tasks
+        // 2. dump the chat completion settings
+        // TODO should I skip the export if the chat completion has been disabled?
+        progress.update_progress(DumpCreationProgress::DumpTheChatCompletionSettings);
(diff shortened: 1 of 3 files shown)


crates/index-scheduler/src/scheduler/process_dump_creation.rs
```diff
@@ -43,7 +43,16 @@ impl IndexScheduler {
 
         let rtxn = self.env.read_txn()?;
 
-        // 2. dump the tasks
+        // 2. dump the chat completion settings
+        // TODO should I skip the export if the chat completion has been disabled?
```
Kerollmops, line 47: @irevoire I am wondering if I should skip creating this folder when the feature is disabled. If I remove this folder I need to make sure, during the import, that a missing folder still works.

Kerollmops: Ok, I fixed the error we were getting when the `chat-completions-settings` folder was not found.

irevoire: Since it's going to stay, I would say it's not an issue to create the file even if it's irrelevant when the feature is disabled.
We could even get rid of it later in a non-breaking manner, so it's ok to me 👍

PR #5733, opened 2025-07-04
Improve the analytics of the `/export` route

Kerollmops:
This PR exposes the host of the export URL in the analytics. It extracts it using [the `Url::host_str` method](https://docs.rs/url/latest/url/struct.Url.html#method.host_str).


dureuill (dismissed):
:+1:

dureuill:
clippy's not happy

PR #5596, opened 2025-05-27
Request fragments

dureuill:
WIP

PR #5734, opened 2025-07-04
Tests for multimodal

Mubelotix:
Follow-up for #5596


crates/meilisearch/tests/vector/fragments.rs
```diff
@@ -0,0 +1,2248 @@
...
+                    "justName": {"value": "{{ media.name }} is a dog"},
+                    "query": {"value": "Some pre-prompt for query {{ q }}"},
+                }
+            },
+        },
+    });
+    let (response, code) = index.update_settings(settings.clone()).await;
+    snapshot!(code, @"202 Accepted");
+
+    let task = server.wait_task(response.uid()).await;
```
martin-g:
```suggestion
    let task = server.wait_task(response.uid()).await.succeeded();
```
and remove the next line

crates/meilisearch/tests/vector/fragments.rs
```diff
@@ -0,0 +1,2248 @@
...
+    snapshot!(json_string!(value["embedders"], {
+        ".rest.url" => "[url]",
+    }), @r#"
+    {
+      "rest": {
+        "source": "rest",
+        "dimensions": 3,
+        "url": "[url]",
+        "indexingFragments": {
+          "withBreed": {
```
martin-g, line 946: What happens here ? Why `withBreed` is still here ?

martin-g (commented):
&nbsp;

crates/meilisearch/tests/common/server.rs
```diff
@@ -327,7 +327,7 @@ impl<State> Server<State> {
         self.service.get(url).await
     }
 
-    pub(super) fn _index(&self, uid: impl AsRef<str>) -> Index<'_> {
+    pub fn _index(&self, uid: impl AsRef<str>) -> Index<'_> {
```
dureuill: @irevoire not sure if this is intended, making an underscore-starting method public is tingling my spider sense

Mubelotix: Didn't like it too but removing it forces code to be moved to `common` which I didn't like either

dureuill (approved):
Excellent work, thank you!

PR #5737, opened 2025-07-07
Fix the dumpless upgrade from v1.15 to v1.16 for request fragments

dureuill:
- Add dumpless upgrade support for request fragments:
- Add compile-time checks to make sure that the last package can be upgraded to itself + that no index is out of bounds.

- [x] test that missing the last package entry in the `start` function causes a compilation error
- [x] test that referencing an out of bound index in `function_index!` macro causes a compilation error
- [x] test dumpless upgrade from a v1.15 db with some user provided embeddings
- [x] test a "long" dumpless upgrade from a v1.15 db to check that search is functional during  the upgrade.


crates/milli/src/update/upgrade/v1_15.rs
```diff
@@ -26,3 +28,14 @@ impl UpgradeIndex for Latest_V1_14_To_Latest_V1_15 {
...
+/// Parts of v1.15 `IndexingEmbeddingConfig` that are relevant for upgrade to v1.16
+///
+/// # Warning
+///
+/// This object should not be rewritten to the DB, only read to get the name and `user_provided` roaring.
+#[derive(Debug, Deserialize, Serialize)]
+pub struct IndexEmbeddingConfig {
+    pub name: String,
+    pub user_provided: RoaringBitmap,
+}
```
irevoire:
```suggestion
/// Parts of v1.15 `IndexingEmbeddingConfig` that are relevant for upgrade to v1.16
///
/// # Warning
///
/// This object should not be rewritten to the DB, only read to get the name and `user_provided` roaring.
#[derive(Debug, Deserialize)]
pub struct IndexEmbeddingConfig {
    pub name: String,
    pub user_provided: RoaringBitmap,
}
```

If we don't want to write the object to the db can we avoid making it serializable?
That's what I'm doing in the dump where I can, and it has helped a few times

dureuill: oh that's nice, making the change

dureuill: done

irevoire (approved):

Issue #4704, opened 2024-06-17
Pagination issue when using Semantic Search


runabol:
**Describe the bug**
The pagination results behavior changes when I use semantic search. Specifically, it seems to get offsets `0-10` when I specify `offset: 5`, `limit: 5`.

**To Reproduce**

In the first two `curl` commands I turn off semantic search and get the expected results. 

In the following ones I turn it on and get the unexpected behavior.

```
# curl -s -X POST -H "content-type:application/json" -d '{"q":"car","offset":0,"limit":5,"hybrid":{"embedder":"default","semanticRatio":0}}'  'http://localhost:7700/indexes/books/search'  | jq '.hits[].id'
"658c8e3c8e2163f80f97a743"
"658c8e1e8e2163f80f97a2e8"
"658c8fc1372f01b94b520b8e"
"658c8d938e2163f80f978ffe"
"658c8df48e2163f80f979c02"


# curl -s -X POST -H "content-type:application/json" -d '{"q":"car","offset":5,"limit":5,"hybrid":{"embedder":"default","semanticRatio":0}}' 'http://localhost:7700/indexes/books/search'  | jq '.hits[].id'
"658c8ec3372f01b94b51ea56"
"658c8dd58e2163f80f979a1a"
"658c913a372f01b94b523bc7"
"658c9153372f01b94b523f2d"
"658c9205372f01b94b525476"

# curl -s -X POST -H "content-type:application/json" -d '{"q":"car","offset":0,"limit":5,"hybrid":{"embedder":"default","semanticRatio":0.5}}'  'http://localhost:7700/indexes/books/search'  | jq '.hits[].id'
"658c8e3c8e2163f80f97a743"
"658c8e1e8e2163f80f97a2e8"
"658c8fc1372f01b94b520b8e"
"658c8d938e2163f80f978ffe"
"658c8df48e2163f80f979c02"

# curl -s -X POST -H "content-type:application/json" -d '{"q":"car","offset":5,"limit":5,"hybrid":{"embedder":"default","semanticRatio":0.5}}'  'http://localhost:7700/indexes/books/search'  | jq '.hits[].id'
"658c8e3c8e2163f80f97a743"
"658c8e1e8e2163f80f97a2e8"
"658c8fc1372f01b94b520b8e"
"658c8d938e2163f80f978ffe"
"658c8df48e2163f80f979c02"
"658c8ec3372f01b94b51ea56"
"658c8dd58e2163f80f979a1a"
[...]

PR #5738, opened 2025-07-07
Set `search_k` to `max_hits * n_trees` when using finite pagination

Mubelotix:
# Pull Request

## Related issue
Related to #4704

## What does this PR do?

In theory, setting arroy's `search_k` to `max_hits * n_trees` would ensure there are no duplicated results over different paginated pages, nor missing results.


Kerollmops (commented):
I don't know what to think about this but we had some customers who set the max total hits to very large values, i.e. 1000000+ 🤔

PR #5741, opened 2025-07-08
Add Vector filters

### LINKED ISSUE #5747: Vector filters
Related product team resources: [PRD](https://www.notion.so/meilisearch/42fb4e7135a54efdb484273107b84bf5?v=8d4fb6cee38048c99e14bcc5018abce0&p=2294b06b651f809b9befc3bea470ee26&pm=s) (_internal only_)
Related product discussion:

## Motivation

This introduces a new filter that can select documents that have embeddings for a specific embedder or even a specific fragment.

## Usage

https://meilisearch.notion.site/Usage-for-vector-filters-2294b06b651f807ab568fe429170e8d3

Mubelotix:
# Pull Request

## Related issue
Fixes #5747

## What does this PR do?

Adds a new filter. See [the usage page](https://meilisearch.notion.site/Usage-for-vector-filters-2294b06b651f807ab568fe429170e8d3).


dureuill:
Hello!

Please link to the public usage page instead of repeating it here, it'll be easier to maintain if it changes (only one location to update).

Sorry for the late feedback, but looking at the usage you outlined, I see 3 changes I would like done before starting the review proper:

1. Change what `_vectors.{embedder_name}.documentTemplate EXISTS` does. In your description it matches all vectors if the embedder is configured to use document templates, but actually even when the embedder is configured to use document templates, a particular document might opt-out from the document template by containing a user-provided vector. So the correct description should be: Matches all documents containing an embedder generated from the document template. Please make sure that the implementation matches this correct description
2. Remove `_vectors.{embedder_name}.documentTemplate.userProvided EXISTS`. This one feels confusing to me, and can be obtained from the static knowledge of how the embedder is configured + another filter.
3. Add `_vectors.{embedder_name}.regenerate EXISTS`: Matches all documents whose `regenerate` value is set to `true`.

Available if you have any question about these changes. Thank you!

Mubelotix:
The full list of valid formats is now 

- `_vectors`
- `_vectors.{embedder_name}`
- `_vectors.{embedder_name}.regenerate`
- `_vectors.{embedder_name}.userProvided`
- `_vectors.{embedder_name}.userProvided.regenerate`
- `_vectors.{embedder_name}.documentTemplate`
- `_vectors.{embedder_name}.documentTemplate.regenerate`
- `_vectors.{embedder_name}.fragments.{fragment_name}`
- `_vectors.{embedder_name}.fragments.{fragment_name}.regenerate`

dureuill:
Sorry to ask for changes again, but can we remove:

`_vectors.{embedder_name}.userProvided.regenerate`,
`_vectors.{embedder_name}.documentTemplate.regenerate`
&
`_vectors.{embedder_name}.fragments.{fragment_name}.regenerate`

only keeping `_vectors.{embedder_name}.regenerate`

as these can easily be expressed as e.g. `_vectors.{embedder_name}.regenerate AND _vectors.{embedder_name}.userProvided`

crates/milli/src/search/facet/filter_vector.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// - `_vectors.{embedder_name}`
+    /// - `_vectors.{embedder_name}.regenerate`
+    /// - `_vectors.{embedder_name}.userProvided`
+    /// - `_vectors.{embedder_name}.userProvided.regenerate`
+    /// - `_vectors.{embedder_name}.documentTemplate`
+    /// - `_vectors.{embedder_name}.documentTemplate.regenerate`
+    /// - `_vectors.{embedder_name}.fragments.{fragment_name}`
+    /// - `_vectors.{embedder_name}.fragments.{fragment_name}.regenerate`
+    pub(super) fn parse(s: &'a Token<'a>) -> Result<Self, VectorFilterError<'a>> {
+        let mut split = s.split(".").peekable();
```
irevoire: Sorry for being that late on the question, but don't we have an issue with the dot here?
From what I know, most client name their embedder after the name of their model, which sometimes also includes **their version**.
Like `openAI 3.1` or stuff like that?

Maybe I'm wrong thought, do we have any limitation on the name of the embedder @dureuill?

crates/milli/src/search/facet/filter_vector.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// - `_vectors`
+    /// - `_vectors.{embedder_name}`
+    /// - `_vectors.{embedder_name}.regenerate`
+    /// - `_vectors.{embedder_name}.userProvided`
+    /// - `_vectors.{embedder_name}.userProvided.regenerate`
+    /// - `_vectors.{embedder_name}.documentTemplate`
+    /// - `_vectors.{embedder_name}.documentTemplate.regenerate`
+    /// - `_vectors.{embedder_name}.fragments.{fragment_name}`
+    /// - `_vectors.{embedder_name}.fragments.{fragment_name}.regenerate`
+    pub(super) fn parse(s: &'a Token<'a>) -> Result<Self, VectorFilterError<'a>> {
```
irevoire: Seeing what we're doing here, I must say I was kinda expecting this code to be in the filter parser directly.
Something like => when parsing a field name, we now return an enum that contains the whole selector thingy, making sure it makes sense or a normal field name.

When we go up in the chain of parser, it also gives us the opportunity to quickly return a parsing error if someone used a filter incompatible with the vector field.
What do you think?

crates/milli/src/search/facet/filter_vector.rs
dureuill: no limitation on the name of the emedder tamo. I would expect us to be able to parse "<embedder_name>" even if it contains `.`

crates/filter-parser/src/lib.rs
```diff
@@ -500,8 +535,7 @@ fn parse_primary(input: Span, depth: usize) -> IResult<FilterCondition> {
         parse_is_not_null,
         parse_is_empty,
         parse_is_not_empty,
-        parse_exists,
-        parse_not_exists,
+        alt((parse_vectors_exists, parse_vectors_not_exists, parse_exists, parse_not_exists)),
```
Mubelotix: I had to nest alt's because otherwise the tuple would be too big and wouldn't implement the nom's parser trait

crates/milli/src/search/facet/filter_vector.rs
```diff
@@ -0,0 +1,152 @@
...
+                    fragment,
+                    available: embedding_config
+                        .fragments
+                        .as_slice()
+                        .iter()
+                        .map(|f| f.name.clone())
+                        .collect(),
+                })?;
+
+            arroy_wrapper.items_in_store(rtxn, fragment_config.id, |bitmap| bitmap.clone())?
```
dureuill: you need to subtract the user provided docids from the roaring here.

crates/milli/src/search/facet/filter_vector.rs
```diff
@@ -0,0 +1,152 @@
...
+        }
+        VectorFilter::DocumentTemplate => {
+            if !embedding_config.fragments.as_slice().is_empty() {
+                return Ok(RoaringBitmap::new());
+            }
+
+            let user_provided_docsids = embedder_info.embedding_status.user_provided_docids();
+            let mut stats = ArroyStats::default();
+            arroy_wrapper.aggregate_stats(rtxn, &mut stats)?;
```
dureuill, line 136: not directly related to this PR, but I see a mistake in `aggregate_stats`. Could you change the `break`s there into `continue`s?

crates/milli/src/search/facet/filter_vector.rs
```diff
@@ -0,0 +1,152 @@
...
+                })?;
+
+            arroy_wrapper.items_in_store(rtxn, fragment_config.id, |bitmap| bitmap.clone())?
+        }
+        VectorFilter::DocumentTemplate => {
+            if !embedding_config.fragments.as_slice().is_empty() {
+                return Ok(RoaringBitmap::new());
+            }
+
+            let user_provided_docsids = embedder_info.embedding_status.user_provided_docids();
```
dureuill:
```suggestion
            let user_provided_docids = embedder_info.embedding_status.user_provided_docids();
```

crates/milli/src/search/facet/filter_vector.rs
```diff
@@ -0,0 +1,152 @@
...
+            let mut stats = ArroyStats::default();
+            arroy_wrapper.aggregate_stats(rtxn, &mut stats)?;
+            stats.documents - user_provided_docsids.clone()
+        }
+        VectorFilter::UserProvided => {
+            let user_provided_docsids = embedder_info.embedding_status.user_provided_docids();
```
dureuill:
```suggestion
            let user_provided_docids = embedder_info.embedding_status.user_provided_docids();
```

crates/filter-parser/src/lib.rs
```diff
@@ -136,13 +139,23 @@ impl<'a> From<&'a str> for Token<'a> {
...
+}
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum FilterCondition<'a> {
     Not(Box<Self>),
     Condition { fid: Token<'a>, op: Condition<'a> },
     In { fid: Token<'a>, els: Vec<Token<'a>> },
     Or(Vec<Self>),
     And(Vec<Self>),
+    VectorExists { fid: Token<'a>, embedder: Option<Token<'a>>, filter: VectorFilter<'a> },
```
dureuill, line 158: do we use the `fid` in this variant? Like in error messages or the like?

crates/filter-parser/src/lib.rs
dureuill: unfortunate, but OK :+1:

crates/milli/src/search/facet/filter_vector.rs
Mubelotix: Then I will just remove this block, it has no effect as bitOr will just be an empty loop and len is 0

crates/meilisearch/tests/search/filters.rs
```diff
@@ -731,3 +731,470 @@ async fn test_filterable_attributes_priority() {
...
+      "message": "The vector filter is missing a fragment name.\n24:31 _vectors.rest.fragments EXISTS",
+      "code": "invalid_search_filter",
+      "type": "invalid_request",
+      "link": "https://docs.meilisearch.com/errors#invalid_search_filter"
+    }
+    "#);
+}
+
+#[actix_rt::test]
+async fn vector_filter_non_existant_embedder() {
```
dureuill:
```suggestion
async fn vector_filter_nonexistent_embedder() {
```

crates/milli/src/search/facet/filter_vector.rs
dureuill: that works as well

crates/milli/src/search/facet/filter_vector.rs
dureuill: for context, user provided vectors are stored in any store id starting from 0, so if you don't account for them you may or may not accidentally retrieve user provided vectors with `items_in_store`

crates/filter-parser/src/lib.rs
Mubelotix: Yes [here](https://github.com/meilisearch/meilisearch/blob/6c3dd83ae528a8b82614880c0206cd7886a187c7/crates/filter-parser/src/lib.rs#L206). It's only used for the message to display when trying to use vector filters without enabling the multimodal feature. We will be able to remove it in the future

dureuill (approved):

dureuill:
To land on `main` after:

- @irevoire gives final approval on the filter side
- v1.16.0 is released
- v1.16.0 changes are ported back to `main`

crates/filter-parser/src/lib.rs
irevoire: ahah I had to do the same thing in my cellulite integration with the _geoPolygon filter

crates/filter-parser/src/lib.rs
```diff
@@ -630,6 +670,9 @@ pub mod tests {
         insta::assert_snapshot!(p(r"title = 'foo\\\\\\\\'"), @r#"{title} = {foo\\\\}"#);
         // but it also works with other sequences
         insta::assert_snapshot!(p(r#"title = 'foo\x20\n\t\"\'"'"#), @"{title} = {foo \n\t\"\'\"}");
+
+        insta::assert_snapshot!(p(r#"_vectors." valid.name  ".fragments."also.. valid! " EXISTS"#), @r#"_vectors." valid.name  ".fragments."also.. valid! " EXISTS"#);
+        insta::assert_snapshot!(p("_vectors.\"\n\t\r\\\"\" EXISTS"), @r#"_vectors."\n\t\r\"" EXISTS"#);
```
irevoire, line 675: Very cursed, very nice

crates/filter-parser/src/lib.rs
```diff
@@ -946,6 +1001,59 @@ pub mod tests {
...
         );
 
+        insta::assert_snapshot!(p(r#"_vectors _vectors EXISTS"#), @r"
+        Was expecting an operation `=`, `!=`, `>=`, `>`, `<=`, `<`, `IN`, `NOT IN`, `TO`, `EXISTS`, `NOT EXISTS`, `IS NULL`, `IS NOT NULL`, `IS EMPTY`, `IS NOT EMPTY`, `CONTAINS`, `NOT CONTAINS`, `STARTS WITH`, `NOT STARTS WITH`, `_geoRadius`, or `_geoBoundingBox` at `_vectors _vectors EXISTS`.
+        1:25 _vectors _vectors EXISTS
+        ");
+        insta::assert_snapshot!(p(r#"_vectors. embedderName EXISTS"#), @r"
+        The vector filter's embedder is invalid.
+        10:30 _vectors. embedderName EXISTS
+        ");
```
irevoire: Shouldn't the error message be "Was expecting embedder name but found nothing" and point to the space?

crates/filter-parser/src/lib.rs
```diff
@@ -946,6 +1001,59 @@ pub mod tests {
...
+        insta::assert_snapshot!(p(r#"_vectors .embedderName EXISTS"#), @r"
+        Was expecting an operation `=`, `!=`, `>=`, `>`, `<=`, `<`, `IN`, `NOT IN`, `TO`, `EXISTS`, `NOT EXISTS`, `IS NULL`, `IS NOT NULL`, `IS EMPTY`, `IS NOT EMPTY`, `CONTAINS`, `NOT CONTAINS`, `STARTS WITH`, `NOT STARTS WITH`, `_geoRadius`, or `_geoBoundingBox` at `_vectors .embedderName EXISTS`.
+        1:30 _vectors .embedderName EXISTS
+        ");
```
irevoire: If `_vectors` is valid as-is why does the error message says the error happened at: `_vectors .embedderName EXISTS`
It should be at `.embedderName EXISTS`
Not even sure about the EXISTS part, but at least it should not contain the `_vectors` as part of the error

crates/filter-parser/src/lib.rs
```diff
@@ -946,6 +1001,59 @@ pub mod tests {
...
+        insta::assert_snapshot!(p(r#"_vectors.embedderName. EXISTS"#), @r"
+        The vector filter has leftover tokens.
+        22:30 _vectors.embedderName. EXISTS
+        ");
```
irevoire: Huuum, it may be hard to do since you're re-using the code for the embedder name and the last token, and the error message is already ok-ish, but do you think it would make sense to list what is possible in this specific case instead of saying it's a leftover token?

Like:
```
The vector filter has leftover tokens. Was expecting one of `.fragment`, `.userProvided`, ... or nothing but instead got `.`
```

Because currently, if I write `_vectors.embedderName.fargment`, I'll get the leftover tokens error, which can be hard to understand when I just made a small typo 😭 
In deserr we even give you a `did you mean ".fragment"` when the entered value is very close to one of the possible value.
Since the filter API is a bit hard to understand and document it would be awesome if you feel like doing it 💯 

That's how we do it in deserr:
- https://docs.rs/deserr/latest/src/deserr/errors/helpers.rs.html
- https://docs.rs/deserr/latest/src/deserr/errors/json.rs.html#167-177

crates/filter-parser/src/lib.rs
```diff
@@ -946,6 +1001,59 @@ pub mod tests {
...
+        insta::assert_snapshot!(p(r#"_vectors."embedderName EXISTS"#), @r#"
+        The vector filter's embedder is invalid.
+        30:30 _vectors."embedderName EXISTS
+        "#);
```
irevoire: Two comments on this one. First of all I think for all of these error we should write "The vector filter's emedder **name** is invalid."
Because otherwise it's not immediately obvious which part is wrong here 🤔 

And second comment, I struggled to see what was wrong because we had a closing delimiter. But it was the one of rust, not the one of the embedder name.
People will struggle the same with their filter in a json value so maybe we could instead return a "Missing closing delimitor `"` in embedder name" or something?
What do you think? I'm not sure that's possible with the `delimited` function of nom, though, so you tell me. We won't rewrite the whole parser for that

crates/filter-parser/src/value.rs
```diff
@@ -80,6 +80,39 @@ pub fn word_exact<'a, 'b: 'a>(tag: &'b str) -> impl Fn(Span<'a>) -> IResult<'a,
...
+/// vector_value          = ( non_dot_word | singleQuoted | doubleQuoted)
+pub fn parse_vector_value(input: Span) -> IResult<Token> {
+    pub fn non_dot_word(input: Span) -> IResult<Token> {
+        let (input, word) = take_while1(|c| is_value_component(c) && c != '.')(input)?;
+        Ok((input, word.into()))
+    }
+
+    let (input, value) = alt((
+        delimited(char('\''), cut(|input| quoted_by('\'', input)), cut(char('\''))),
+        delimited(char('"'), cut(|input| quoted_by('"', input)), cut(char('"'))),
```
irevoire, line 92: I guess that's where we can shove our nice error message. Not sure what the span will be though 😬 

```
        delimited(char('\''), cut(|input| quoted_by('\'', input)), cut_with_err(char('\''), expected closing delimiter ')),
        delimited(char('"'), cut(|input| quoted_by('"', input)), cut_with_err(char('"'), expected closing delimiter ")),
```

crates/milli/src/search/facet/filter_vector.rs
```diff
@@ -0,0 +1,155 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum VectorFilterError<'a> {
+    #[error("The embedder `{}` does not exist. {}", embedder.value(), {
+        if available.is_empty() {
+            String::from("This index does not have any configured embedders.")
+        } else {
+            let mut available = available.clone();
+            available.sort_unstable();
+            format!("Available embedders are: {}.", available.iter().map(|e| format!("`{e}`")).collect::<Vec<_>>().join(", "))
```
irevoire: Here we also have the opportunity to write a did you mean message 👀

irevoire (changes requested):
Hey, I think we're good in terms of correctness, but as a noob that doesn't know the API I would like the error message to be way more helpful.
Having a DSL for our filter means it's very hard to understand what's possible or not, and reading the documentation about it is not obvious either so we should really strive to help the user actually understand and discover what's possible through the error messages!

I left a bunch of things I would like to see, but if you have other/better ideas, feel free to propose/discuss/implement them! 😁 
We should really take the place of a client who doesn't really read the doc and will try to write a valid filter through the error message. Every time they have to open the doc to understand what they did wrong, they may give up because it's hard to understand what they did wrong

crates/filter-parser/src/lib.rs
Mubelotix: It was possible using error handling and it works like a charm!

PR #5751, opened 2025-07-10
Fix: Preserve order of searchable attributes when modified

ManyTheFish:
## Bug Description

When modifying the `searchableAttributes` setting through the settings API, the order of attributes was not being preserved. The system would incorrectly detect that no changes had occurred when only the order was modified, causing the update to be skipped.

**Example:**
- Set `searchableAttributes: ["title", "overview"]`
- Modify to `searchableAttributes: ["overview", "title"]` 
- The system would not detect this as a change and keep the original order

## Root Cause

The bug was in the change detection logic within `update_user_defined_searchable_attributes()` in `crates/milli/src/update/settings.rs`. The method was comparing the new user-defined order against the computed `searchable_fields` instead of the `user_defined_searchable_fields`.

```rust
// BUGGY: Comparing against computed fields
let old_fields = self.index.searchable_fields(self.wtxn)?;
let did_change = new_fields != old_fields;
```

Since `searchable_fields` returns computed fields that may have a different order than the user-defined ones, order changes were not being detected.

## Fix

Changed the comparison to use `user_defined_searchable_fields` instead:

```rust
// FIXED: Compare against user-defined fields
let old_fields = self.index.user_defined_searchable_fields(self.wtxn)?;
let did_change = old_fields.map(|old| new_fields != old).unwrap_or(true);
```

This ensures that any change in the user-defined order is properly detected and applied.

### DIFF
--- a/crates/milli/src/update/settings.rs
+++ b/crates/milli/src/update/settings.rs
@@ -557,7 +557,7 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
                 let old_fields = self.index.user_defined_searchable_fields(self.wtxn)?;
                 let did_change = {
                     let new_fields = fields.iter().map(String::as_str).collect::<Vec<_>>();
-                    old_fields.map(|old| new_fields != old).unwrap_or(true)
+                    old_fields.is_none_or(|old| new_fields != old)
                 };
                 if !did_change {
                     return Ok(false);


crates/milli/src/update/settings.rs
```diff
@@ -554,10 +554,10 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
         match self.searchable_fields {
             Setting::Set(ref fields) => {
                 // Check to see if the searchable fields changed before doing anything else
-                let old_fields = self.index.searchable_fields(self.wtxn)?;
+                let old_fields = self.index.user_defined_searchable_fields(self.wtxn)?;
                 let did_change = {
                     let new_fields = fields.iter().map(String::as_str).collect::<Vec<_>>();
-                    new_fields != old_fields
+                    old_fields.map(|old| new_fields != old).unwrap_or(true)
```
Kerollmops:
```suggestion
                    old_fields.is_none_or(|old| new_fields != old)
```

Kerollmops (changes requested):

Kerollmops (approved):

PR #5753, opened 2025-07-11
Various fixes on the export route

Mubelotix:
# Pull Request

## Related issues

Fixes anonymous issues:
- Default payload size being over the cloud's default
- Payload size error message not being formatted with the best unit
- Export errors being silently dropped making the task succeed even when documents can't be sent
- Bearer being reconstructed too many times
- Document step being overestimated (fix untested)

### DIFF
--- a/crates/meilisearch/src/error.rs
+++ b/crates/meilisearch/src/error.rs
@@ -49,7 +49,7 @@ pub enum MeilisearchHttpError {
     TooManySearchRequests(usize),
     #[error("Internal error: Search limiter is down.")]
     SearchLimiterIsDown,
-    #[error("The provided payload reached the size limit. The maximum accepted payload size is {}.",  Byte::from_u64(*.0 as u64).get_appropriate_unit(UnitType::Binary))]
+    #[error("The provided payload reached the size limit. The maximum accepted payload size is {}.", Byte::from_u64(dbg!(*.0 as u64)).get_appropriate_unit(if *.0 % 1024 == 0 { UnitType::Binary } else { UnitType::Decimal }))]
     PayloadTooLarge(usize),
     #[error("Two indexes must be given for each swap. The list `[{}]` contains {} indexes.",
         .0.iter().map(|uid| format!("\"{uid}\"")).collect::<Vec<_>>().join(", "), .0.len()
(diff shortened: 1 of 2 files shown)


crates/meilisearch/src/error.rs
```diff
@@ -49,7 +49,7 @@ pub enum MeilisearchHttpError {
     TooManySearchRequests(usize),
     #[error("Internal error: Search limiter is down.")]
     SearchLimiterIsDown,
-    #[error("The provided payload reached the size limit. The maximum accepted payload size is {}.",  Byte::from_u64(*.0 as u64).get_appropriate_unit(UnitType::Binary))]
+    #[error("The provided payload reached the size limit. The maximum accepted payload size is {}.", Byte::from_u64(*.0 as u64).get_appropriate_unit(if *.0 % 1024 == 0 { UnitType::Binary } else { UnitType::Decimal }))]
```
Kerollmops, line 52:
```suggestion
    #[error("The provided payload reached the size limit. The maximum accepted payload size is {}.", Byte::from_u64(*.0 as u64).get_appropriate_unit(UnitType::Both))]
```

Mubelotix: I tried that but I still had "4.88 KiB" instead of "5 KB"

Mubelotix:
```rust
#[test]
fn test() {
    println!("{}", Byte::from_u64(5_000).get_appropriate_unit(UnitType::Both));
    println!("{}", Byte::from_u64(5_000_000).get_appropriate_unit(UnitType::Both));
    println!("{}", Byte::from_u64(5_000_000_000).get_appropriate_unit(UnitType::Both));
}
```

Outputs:

```
4.8828125 KiB
4.76837158203125 MiB
4.656612873077393 GiB
```

After reading the code I determined that what `get_appropriate_unit` does is just take the largest unit that is right under the value. Unfortunately, binary units are allways just a little over decimal units. The result is that the binary unit gets selected almost always (in >95% of cases).

This is bad.

Unfortunately, it's used at 6 locations in our codebase. If you had like, I could add a little simple method that's better at detecting the unit and replace all calls to `get_appropriate_unit`.

PR #5758, opened 2025-07-15
Cellulite integration

irevoire:
Just a draft to see how hard it would be to make a proper integration

### DIFF
--- a/crates/meilisearch-types/src/error.rs
+++ b/crates/meilisearch-types/src/error.rs
@@ -484,7 +484,9 @@ impl ErrorCode for milli::Error {
                         Code::InvalidFacetSearchFacetName
                     }
                     UserError::CriterionError(_) => Code::InvalidSettingsRankingRules,
-                    UserError::InvalidGeoField { .. } => Code::InvalidDocumentGeoField,
+                    UserError::InvalidGeoField { .. } | UserError::GeoJsonError(_) => {
+                        Code::InvalidDocumentGeoField
+                    }
                     UserError::InvalidVectorDimensions { .. }
                     | UserError::InvalidIndexingVectorDimensions { .. } => {
                         Code::InvalidVectorDimensions
--- a/crates/milli/src/index.rs
+++ b/crates/milli/src/index.rs
@@ -114,6 +114,7 @@ pub mod db_name {
     pub const FIELD_ID_DOCID_FACET_STRINGS: &str = "field-id-docid-facet-strings";
     pub const VECTOR_EMBEDDER_CATEGORY_ID: &str = "vector-embedder-category-id";
     pub const VECTOR_ARROY: &str = "vector-arroy";
+    pub const CELLULITE: &str = "cellulite";
     pub const DOCUMENTS: &str = "documents";
 }
 const NUMBER_OF_DBS: u32 = 25;
@@ -182,6 +183,9 @@ pub struct Index {
     /// Vector store based on arroy™.
     pub vector_arroy: arroy::Database<Unspecified>,
 
+    /// Geo store based on cellulite™.
+    pub cellulite: cellulite::Database,
+
     /// Maps the document id to the document as an obkv store.
     pub(crate) documents: Database<BEU32, ObkvCodec>,
 }
@@ -238,6 +242,7 @@ impl Index {
         let embedder_category_id =
             env.create_database(&mut wtxn, Some(VECTOR_EMBEDDER_CATEGORY_ID))?;
         let vector_arroy = env.create_database(&mut wtxn, Some(VECTOR_ARROY))?;
+        let cellulite = env.create_database(&mut wtxn, Some(CELLULITE))?;
 
         let documents = env.create_database(&mut wtxn, Some(DOCUMENTS))?;
 
@@ -266,6 +271,7 @@ impl Index {
             field_id_docid_facet_strings,
             vector_arroy,
             embedder_category_id,
+            cellulite,
             documents,
         };
         if this.get_version(&wtxn)?.is_none() && creation {
@@ -1021,6 +1027,13 @@ impl Index {
         let geo_filter =
             self.filterable_attributes_rules(rtxn)?.iter().any(|field| field.has_geo());
         Ok(geo_filter)
+    }
+
+    /// Returns true if the geo sorting feature is enabled.
+    pub fn is_geojson_enabled(&self, rtxn: &RoTxn<'_>) -> Result<bool> {
+        let geojson_filter =
+            self.filterable_attributes_rules(rtxn)?.iter().any(|field| field.has_geojson());
+        Ok(geojson_filter)
     }
 
     pub fn asc_desc_fields(&self, rtxn: &RoTxn<'_>) -> Result<HashSet<String>> {
@@ -1842,6 +1855,7 @@ impl Index {
             field_id_docid_facet_strings,
             vector_arroy,
             embedder_category_id,
+            cellulite: _,
             documents,
         } = self;
 
--- a/crates/milli/src/lib.rs
+++ b/crates/milli/src/lib.rs
@@ -85,7 +85,7 @@ pub use self::search::{
 };
 pub use self::update::ChannelCongestion;
 
-pub type Result<T> = std::result::Result<T, error::Error>;
+pub type Result<T, E = error::Error> = std::result::Result<T, E>;
 
 pub type Attribute = u32;
 pub type BEU16 = heed::types::U16<heed::byteorder::BE>;
--- a/crates/milli/src/update/index_documents/extract/extract_geo_points.rs
+++ b/crates/milli/src/update/index_documents/extract/extract_geo_points.rs
@@ -2,6 +2,7 @@ use std::fs::File;
 use std::io::{self, BufReader};
 
 use concat_arrays::concat_arrays;
+use geojson::GeoJson;
 use serde_json::Value;
 
 use super::helpers::{create_writer, writer_into_reader, GrenadParameters};
@@ -103,6 +104,80 @@ fn extract_lat_lng(
             .map_err(|lng| GeoError::BadLongitude { document_id: document_id(), value: lng })
             .map_err(Box::new)?;
             Ok(Some([lat, lng]))
+        }
+        None => Ok(None),
+    }
+}
+
+/// Extracts the geographical coordinates contained in each document under the `_geo` field.
+///
+/// Returns the generated grenad reader containing the docid as key associated to the (latitude, longitude)
+#[tracing::instrument(level = "trace", skip_all, target = "indexing::extract")]
+pub fn extract_geojson<R: io::Read + io::Seek>(
+    obkv_documents: grenad::Reader<R>,
+    indexer: GrenadParameters,
+    primary_key_id: FieldId,
+    settings_diff: &InnerIndexSettingsDiff,
+) -> Result<grenad::Reader<BufReader<File>>> {
+    let mut writer = create_writer(
+        indexer.chunk_compression_type,
+        indexer.chunk_compression_level,
+        tempfile::tempfile()?,
+    );
+
+    tracing::info!("Extracting one geojson");
+
+    let mut cursor = obkv_documents.into_cursor()?;
+    while let Some((docid_bytes, value)) = cursor.move_on_next()? {
+        let obkv = obkv::KvReader::from_slice(value);
+        // since we only need the primary key when we throw an error
+        // we create this getter to lazily get it when needed
+        let document_id = || -> Value {
+            let reader = KvReaderDelAdd::from_slice(obkv.get(primary_key_id).unwrap());
+            let document_id =
+                reader.get(DelAdd::Deletion).or(reader.get(DelAdd::Addition)).unwrap();
+            serde_json::from_slice(document_id).unwrap()
+        };
+
+        // extract old version
+        let del_geojson =
+            extract_geojson_field(obkv, &settings_diff.old, DelAdd::Deletion, document_id)?;
+        // extract new version
+        let add_geojson =
+            extract_geojson_field(obkv, &settings_diff.new, DelAdd::Addition, document_id)?;
+
+        if del_geojson != add_geojson {
+            let mut obkv = KvWriterDelAdd::memory();
+            if del_geojson.is_some() {
+                #[allow(clippy::drop_non_drop)]
+                // We don't need to store the geojson, we'll just delete it by id
+                obkv.insert(DelAdd::Deletion, [])?;
+            }
+            if let Some(geojson) = add_geojson {
+                #[allow(clippy::drop_non_drop)]
+                obkv.insert(DelAdd::Addition, geojson.to_string().as_bytes())?;
+            }
+            let bytes = obkv.into_inner()?;
+            writer.insert(docid_bytes, bytes)?;
+        }
+    }
+
+    writer_into_reader(writer)
+}
+
+fn extract_geojson_field(
+    obkv: &obkv::KvReader<FieldId>,
+    settings: &InnerIndexSettings,
+    deladd: DelAdd,
+    document_id: impl Fn() -> Value,
+) -> Result<Option<GeoJson>> {
+    match settings.geojson_fid {
+        Some(fid) => {
+            let value = obkv.get(fid).map(KvReaderDelAdd::from_slice).and_then(|r| r.get(deladd));
+            // TODO: That's a user error, not an internal error
+            Ok(value
+                .map(|v| serde_json::from_slice(v).map_err(InternalError::SerdeJson))
+                .transpose()?)
         }
         None => Ok(None),
     }
--- a/crates/milli/src/update/index_documents/extract/mod.rs
+++ b/crates/milli/src/update/index_documents/extract/mod.rs
@@ -31,6 +31,7 @@ use self::extract_word_position_docids::extract_word_position_docids;
 use super::helpers::{as_cloneable_grenad, CursorClonableMmap, GrenadParameters};
 use super::{helpers, TypedChunk};
 use crate::progress::EmbedderStats;
+use crate::update::index_documents::extract::extract_geo_points::extract_geojson;
 use crate::update::index_documents::extract::extract_vector_points::extract_embeddings_from_fragments;
 use crate::update::settings::InnerIndexSettingsDiff;
 use crate::vector::db::EmbedderInfo;
@@ -62,6 +63,7 @@ pub(crate) fn data_from_obkv_documents(
                         original_documents_chunk,
                         indexer,
                         lmdb_writer_sx.clone(),
+                        primary_key_id,
                         settings_diff.clone(),
                         embedder_info.clone(),
                         possible_embedding_mistakes.clone(),
@@ -232,6 +234,7 @@ fn send_original_documents_data(
     original_documents_chunk: Result<grenad::Reader<BufReader<File>>>,
     indexer: GrenadParameters,
     lmdb_writer_sx: Sender<Result<TypedChunk>>,
+    primary_key_id: FieldId,
     settings_diff: Arc<InnerIndexSettingsDiff>,
     embedder_info: Arc<Vec<(String, EmbedderInfo)>>,
     possible_embedding_mistakes: Arc<PossibleEmbeddingMistakes>,
@@ -239,6 +242,22 @@ fn send_original_documents_data(
 ) -> Result<()> {
     let original_documents_chunk =
         original_documents_chunk.and_then(|c| unsafe { as_cloneable_grenad(&c) })?;
+
+    tracing::info!("Do we have a geojson");
+    if settings_diff.run_geojson_indexing() {
+        tracing::info!("Yes we do");
+        let documents_chunk_cloned = original_documents_chunk.clone();
+        let lmdb_writer_sx_cloned = lmdb_writer_sx.clone();
+        let settings_diff = settings_diff.clone();
+        rayon::spawn(move || {
+            let result =
+                extract_geojson(documents_chunk_cloned, indexer, primary_key_id, &settings_diff);
+            let _ = match result {
+                Ok(geojson) => lmdb_writer_sx_cloned.send(Ok(TypedChunk::GeoJson(geojson))),
+                Err(error) => lmdb_writer_sx_cloned.send(Err(error)),
+            };
+        });
+    }
 
     let index_vectors = (settings_diff.reindex_vectors() || !settings_diff.settings_update_only())
         // no point in indexing vectors without embedders
--- a/crates/milli/src/update/new/document.rs
+++ b/crates/milli/src/update/new/document.rs
@@ -10,7 +10,9 @@ use serde_json::value::RawValue;
 
 use super::vector_document::VectorDocument;
 use super::{KvReaderFieldId, KvWriterFieldId};
-use crate::constants::{RESERVED_GEO_FIELD_NAME, RESERVED_VECTORS_FIELD_NAME};
+use crate::constants::{
+    RESERVED_GEOJSON_FIELD_NAME, RESERVED_GEO_FIELD_NAME, RESERVED_VECTORS_FIELD_NAME,
+};
 use crate::documents::FieldIdMapper;
 use crate::update::del_add::KvReaderDelAdd;
 use crate::update::new::thread_local::{FullySend, MostlySend, ThreadLocal};
@@ -26,6 +28,7 @@ use crate::{
 ///
 /// The 'doc lifetime is meant to live sufficiently for the document to be handled by the extractors.
 pub trait Document<'doc> {
+    fn geojson_field(&self) -> Result<Option<&'doc RawValue>>;
     /// Iterate over all **top-level** fields of the document, returning their name and raw JSON value.
     ///
     /// - The returned values *may* contain nested fields.
@@ -113,6 +116,10 @@ impl<'t, Mapper: FieldIdMapper> Document<'t> for DocumentFromDb<'t, Mapper> {
         self.field(RESERVED_GEO_FIELD_NAME)
     }
 
+    fn geojson_field(&self) -> Result<Option<&'t RawValue>> {
+        self.field(RESERVED_GEOJSON_FIELD_NAME)
+    }
+
     fn top_level_fields_count(&self) -> usize {
         let has_vectors_field = self.vectors_field().unwrap_or(None).is_some();
         let has_geo_field = self.geo_field().unwrap_or(None).is_some();
@@ -175,6 +182,10 @@ impl<'doc> Document<'doc> for DocumentFromVersions<'_, 'doc> {
 
     fn geo_field(&self) -> Result<Option<&'doc RawValue>> {
         Ok(self.versions.geo_field())
+    }
+
+    fn geojson_field(&self) -> Result<Option<&'doc RawValue>> {
+        Ok(self.versions.geojson_field())
     }
 
     fn top_level_fields_count(&self) -> usize {
@@ -265,6 +276,16 @@ impl<'d, 'doc: 'd, 't: 'd, Mapper: FieldIdMapper> Document<'d>
         db.geo_field()
     }
 
+    fn geojson_field(&self) -> Result<Option<&'d RawValue>> {
+        if let Some(geojson) = self.new_doc.geojson_field()? {
+            return Ok(Some(geojson));
+        }
+
+        let Some(db) = self.db else { return Ok(None) };
+
+        db.geojson_field()
+    }
+
     fn top_level_fields_count(&self) -> usize {
         self.iter_top_level_fields().count()
     }
@@ -294,6 +315,10 @@ where
 
     fn geo_field(&self) -> Result<Option<&'doc RawValue>> {
         D::geo_field(self)
+    }
+
+    fn geojson_field(&self) -> Result<Option<&'doc RawValue>> {
+        D::geojson_field(self)
     }
 
     fn top_level_fields_count(&self) -> usize {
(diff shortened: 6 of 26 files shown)


gmourier:
## Next Steps & Deadline

Following today’s discussion with @tamo, we’ve defined the final steps needed to reach a version we consider done and ready to merge into main.

Target date: branch ready for merge before August 22 (2 weeks from now).

# ⚠️ Compatibility Note

Currently, there is little integration between the current geo database and this new one, which means the _geoPoint and _geoRadius operators are not applied to this database and will only works with what you added in the _geo field.

_geoBoundingBox is supported and will work over both the previous and new geo databases

## To be added

- [x] _geoBoundingBox support - https://github.com/meilisearch/meilisearch/issues/5818
- [x] _geoRadius support - https://github.com/meilisearch/meilisearch/issues/5819

## Won't be added for now

- _geoPoint sorting - we don't know how yet how to do it - https://github.com/irevoire/cellulite/issues/2

# GeoJSON Support

To fully match the GeoJSON specification, we still need to implement:

- Geometry Objects:
    - [x] [Point](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.2)
    - [x] [MultiPoint](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.3)
    - [x] [LineString](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.4) - https://github.com/irevoire/cellulite/issues/9
    - [x] [MultiLineString](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.5) - https://github.com/irevoire/cellulite/issues/11
    - [x] [Polygon](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6)
    - [x] [MultiPolygon](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.7)
    - [x] [GeometryCollection](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.8) - https://github.com/irevoire/cellulite/issues/12
- [x] [Feature](https://datatracker.ietf.org/doc/html/rfc7946#section-3.2)
- [x] [FeatureCollection](https://datatracker.ietf.org/doc/html/rfc7946#section-3.3) - https://github.com/irevoire/cellulite/issues/13


# Identified bug to fix

- [x] Antimeridian: storing/retrieving near ±180° longitude returns incorrect/random documents - https://github.com/irevoire/cellulite/issues/3
- [x] Lat/Lng inversion in _geoPolygon filter - https://github.com/meilisearch/meilisearch/issues/5820
- [x] Error type mismatch: filtering errors return geoBoundingBox instead of geoPolygon - https://github.com/meilisearch/meilisearch/issues/5821

# Error handling
- [ ] Ensure all geo-related error reporting matches our current error definitions

# Performance Improvements
- [x] In extractor: points removed are still serialized as GeoJSON - https://github.com/meilisearch/meilisearch/issues/5822 -> crates/milli/src/update/new/extract/geo/cellulite.rs
- [x] After extracting and while sending the data to the writer: crates/milli/src/update/new/channel.rs:617
- [ ] Indexing is not parallelized (bonus improvement, as the observed performance are considered good)

# Miscellaneous

- [x] Verify rollback logic in milli::index::rollback @irevoire 
- [x] Add missing database size reporting in milli::index::database_sizes @irevoire 
- [x] Check geo_faceted_docids handling in milli::update::index_documents::typed_chunks (search for GeoJson or RTree) @irevoire -> We don't have the sort so we don't need that currently
- [x] Store the version in cellulite directly for future dumpless upgrade

dureuill:
Review of the cellulite code:

1. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/reader.rs#L86 hardcoded-constant, could maybe be a `const`.
2. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/reader.rs#L87 I'm having trouble understanding the `too_large` algorithm. It allows for as many cells as possible, but then toggles the `too_large` mode whenever more than 3 cells are added at once, and then it remains in "too large mode" for the rest of the search? Can you elaborate on that?
3. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/metadata/mod.rs#L25 it looks like the current behavior is to return `false` as soon as the version changes? This doesn't look consistent with the comment right above that states that there is no breaking version right now? I would expect the code to read `true` with this comment.
4. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/builder.rs#L25 iteration over the full DB is proving problematic when using slower I/O and doesn't scale well with the number of items. Also, one Zerometry is 152 bytes (according to rust-analyzer), so having e.g. 10M of them to retrieve would cause an allocation of 1.56GB just for the frozen items with their u32 key (excluding any IntMap overhead)
5. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/lib.rs#L61 Why is `Cellulite::cell` a DB with `CellIndexCodec` as value, when it is always used as a DB holding a roarings
6. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/lib.rs#L64 Given that the `UpdateDb` is associating a U32 with a kind of update, could it be represented by two roaring bitmaps in RAM? Even if it means committing the transaction without calling `build` would break the DB. Or if we insist on keeping them persistent, maybe store them as two roaring in the UpdateDB? one for deletion and one for additions
7. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/builder.rs#L110 do you think there could be a way to find the list of cells of deleted entries and avoid a full db scan here? Like we did in the differential indexing refactor
8. TBC

Thank you for your hard work, it is a beautiful library ❤️

dureuill:
Continued.

8. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/builder.rs#L220: I think explode_level_zero should *really* take its output vectors by &mut ref instead of doing a fresh allocation everytime. The vectors should be initialized once per thread and cleared before the call at this line (but not inside of the recursive calls for a zollection), or maybe the iterator should be a `drain(..)`. hoisting allocations in this way will relieve a bit of pressure on the allocator.
9. all the `put` in level0 could benefit from being asynchronous, such as having the writer be on a different thread and a pipe between the two. it is not necessary to wait for each individual write to complete before starting the next one.
10. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/builder.rs#L484 unclear comment, are we referring to another `get_children_cells` function? if not, what can this function not be used for?
11. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/builder.rs#L491 not super clear why this is all the children cell?
12. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/builder.rs#L401 I'm having a hard time reconciling `strict_contained: false` with the check on `relation.strict_contains`. Looks to me like `strict_contains` should always be `None` if it wasn't requested?
13. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/builder.rs#L391 I feel like the shadowing is dangerous here. I would have the argument to the function be called `parent_cell` and the loop variable be called `child_cell`.
14. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/builder.rs#L100 I'm not sure I understand why we are inserting `bitmap` rather than its intersection with the freshly `inserted_items`
15. https://github.com/irevoire/cellulite/blob/2f359235ca69176f49c9b5fca01f0fd2dc1e5ca9/src/builder.rs#L90 re: we cannot detail, could we maybe count the number of base cells and use that with an atomic counter? (see `BaseCell::count`)

That's all for cellulite from me :-)

crates/meilisearch/tests/common/mod.rs
```diff
@@ -522,6 +522,26 @@ pub async fn shared_index_with_geo_documents() -> &'static Index<'static, Shared
         .await
 }
 
+pub async fn shared_index_geojson_documents() -> &'static Index<'static, Shared> {
+    static INDEX: OnceCell<Index<'static, Shared>> = OnceCell::const_new();
+    INDEX
+        .get_or_init(|| async {
+            // Retrieved from https://gitlab-forge.din.developpement-durable.gouv.fr/pub/geomatique/descartes/d-map/-/blob/main/demo/examples/commons/countries.geojson?ref_type=heads
```
dureuill, line 529: 👍

crates/meilisearch-types/src/error.rs
```diff
@@ -524,6 +526,7 @@ impl ErrorCode for milli::Error {
                     | UserError::DocumentEditionCompilationError(_) => {
                         Code::EditDocumentsByFunctionError
                     }
+                    UserError::CelluliteError(_) => Code::Internal,
```
dureuill: I'm a bit surprised that we would send a `UserError` to an `Intenral` error code

crates/milli/src/index.rs
```diff
@@ -182,6 +184,9 @@ pub struct Index {
     /// Vector store based on arroy™.
     pub vector_arroy: arroy::Database<Unspecified>,
 
+    /// Geo store based on cellulite™.
+    pub cellulite: Cellulite,
```
dureuill, line 189: no reference here

crates/milli/src/index.rs
```diff
@@ -1917,6 +1932,16 @@ impl Index {
         sizes.insert("embedder_category_id", embedder_category_id.stat(rtxn).map(compute_size)?);
         sizes.insert("documents", documents.stat(rtxn).map(compute_size)?);
 
+        // Cellulite
+        const _CELLULITE_DB_CHECK: () = {
+            if Cellulite::nb_dbs() != 3 {
```
dureuill: sounds like the referenced cellulite version only has 2 DBs

crates/milli/src/lib.rs
```diff
@@ -86,7 +86,7 @@ pub use self::search::{
 };
 pub use self::update::ChannelCongestion;
 
-pub type Result<T> = std::result::Result<T, error::Error>;
+pub type Result<T, E = error::Error> = std::result::Result<T, E>;
```
dureuill, line 90: 👀

crates/milli/src/index.rs
```diff
@@ -1917,6 +1932,16 @@ impl Index {
...
+                panic!("Cellulite database count has changed, please update the code accordingly.")
+            }
+        };
+        sizes.insert("cellulite_item", self.cellulite.item_db_stats(rtxn).map(compute_size)?);
```
dureuill: method doesn"'t seem to exist in the referenced cellulite version

crates/meilisearch-types/src/error.rs
irevoire: Yes, I've logged an issue saying I should double-check all the errors to see what is internal or not! For the first prototype, I just sent everything as internal 😬

irevoire:
Hey,

> I'm having trouble understanding the too_large algorithm. It allows for as many cells as possible, but then toggles the too_large mode whenever more than 3 cells are added at once, and then it remains in "too large mode" for the rest of the search? Can you elaborate on that?

Yes, you understood what it's doing perfectly. The reason for that is that we want to reduce as much as possible the number of lookups in the DB.
I wanted to review this number for a long time and just did it, it's now set to 60.

<details>
<summary>And that's why:</summary>

Let's say I'm looking for points in le vigan, so I make my shape.

## The first step is to cover my shape with res0 cells:

image

And after that we have a choice:
1. We cover our shape at the res1
2. We look into the children of our cell

If we do the 1️⃣ we're looking at 1 extra cell:
image

If we do the 2️⃣ we're looking at 18 extra cells:
image

## We continue with the 1️⃣ algorithm until we generate more than 3 cells

image

At res 4 and 5 we used three cells to cover le vigan because it falls exactly in between all of them.
But at res 6 we had to use 4 because the shape is simply too large:
image

That's the point we were using to say it's actually better to start looking into our children than to cover our shape again.
Except it's not; every time we look at the children of a cell, we're actually looking at 20 more cells.
So when we have 3-4 cells, we'll generate a maximum of 60 cells lookup on the next iteration. That's why I think 60 is a better threshold number, and experimentally, it seems like everything between 50 and 65 is a good fit. When we're above or below, it increases our number of lookups. (It's hard to select a "perfect" number because it depends of the shape)

## We continue until we reach the too-large point

image

That's the last resolution where we can cover our shape and realize we had too many cells.
Now, let's get back to our previous choice:
- Either we cover the whole shape with cells with the resolution +1, which adds about 600 db lookups
- Or we only retrieve the few cells inside of all the blue ,cells (only one here) which adds about 20 db lookups

image

</details>

I did a "big" revamp of the search in this commit: https://github.com/irevoire/cellulite/commit/d7e533090bee5943a907f2196cd2a8dca9581ecf

> it looks like the current behavior is to return false as soon as the version changes? This doesn't look consistent with the comment right above that states that there is no breaking version right now? I would expect the code to read true with this comment.

Since there is only one version of cellulite currently, I just want it to break as soon as we update the version. Then we'll be able to decide how we want to update 🤔 
Do you think we should ignore the patch version or something?

> iteration over the full DB is proving problematic when using slower I/O and doesn't scale well with the number of items. Also, one Zerometry is 152 bytes (according to rust-analyzer), so having e.g. 10M of them to retrieve would cause an allocation of 1.56GB just for the frozen items with their u32 key (excluding any IntMap overhead)

Real question: should we try to fix that right now, or let it ship like that and wait for the first performance issues?
Because we don't have any actual easy-to-use solution right now, right?
Or did you finish the heed-reader crate you told me about since then?

> Why is Cellulite::cell a DB with CellIndexCodec as value, when it is always used as a DB holding a roarings

Totally a reminiscence of the past, fixed! Good catch, thanks for noticing! 😁 

> Given that the UpdateDb is associating a U32 with a kind of update, could it be represented by two roaring bitmaps in RAM? Even if it means committing the transaction without calling build would break the DB.

That would be the best tbh, but I would be afraid of losing stuff at some point 🤔 
But I would love to chat about it, it comes from arroy and I think it's just something we were both afraid of doing with kero and then never thought about it again.

> Or if we insist on keeping them persistent, maybe store them as two roaring in the UpdateDB? one for deletion and one for additions

That's what we were doing initially in arroy, but we noticed that in a lot of cases it was taking a _lot_ of time, and then stopped in favor of one key per item.
I think it was deserializing + serializing + finding a place in lmdb that was taking a ton of time vs finding the place for a single key with no value

> do you think there could be a way to find the list of cells of deleted entries and avoid a full db scan here? Like we did in the differential indexing refactor

I think we could try something where we make a query like "a search" with the fusions of everything that was removed, and remove the elements that need to be removed at the same time.
Also, same question as above, not sure we want to do that in the first iteration.

> I think explode_level_zero should really take its output vectors by &mut ref instead of doing a fresh allocation everytime.

Done in https://github.com/irevoire/cellulite/commit/5d4f9dddf96e884d505ee60fd3caf5139807f5ef

> all the put in level0 could benefit from being asynchronous, such as having the writer be on a different thread and a pipe between the two. it is not necessary to wait for each individual write to complete before starting the next one.

Huum not sure I understand, are you talking about that part?
```rust
        for (cell, items) in to_insert {
            let mut bitmap = self
                .cell_db()
                .get(wtxn, &Key::Cell(cell))?
                .unwrap_or_default();
            bitmap |= items;
            self.cell_db().put(wtxn, &Key::Cell(cell), &bitmap)?;
            atomic.fetch_add(1, Ordering::Relaxed);
        }
```

You would like to spawn a thread that does the read and then sends the result to write it to the main DB?
In order to do that, we would need a rtxn per thread somewhere, right?

> unclear comment, are we referring to another get_children_cells function? if not, what can this function not be used for?

I'll update the comment, its a reference of the [`children`](https://docs.rs/h3o/latest/h3o/struct.CellIndex.html#method.children) method of h3o that was called `get_children_cells` in the previous version.
It was only returning the cells whose center is contained inside of the specified cells.
That means, for this cell in Italy all the red cell children cells containing the red parts would be missing:
Screenshot 2025-09-15 at 15 56 56

> not super clear why this is all the children cell?

With the picture above, I think it's easier to understand.
We start by selecting the middle cells at the res+1 with `center_child`.
Then, if we have a grid_disk of 0, we're only selecting ourselves, the middle cell.
With a grid_disk of 1, we select all the cells of the screenshot above, and with a grid_disk of 2, we select all the neighboring cells of the one we already selected.

> I'm having a hard time reconciling strict_contained: false with the check on relation.strict_contains. Looks to me like strict_contains should always be None if it wasn't requested?

But `strict_contains` was requested through the `..Relation::all()` at the end. It's `strict_contained` that wasn't requested.
In other words, we only want to know if we fully contain the cell, but we don't need to know if the cell fully contains us.

> I feel like the shadowing is dangerous here. I would have the argument to the function be called parent_cell and the loop variable be called child_cell.

Done in https://github.com/irevoire/cellulite/commit/cd9bc0d6c92703df1efcc21c4b3d5420a8678bf1

> I'm not sure I understand why we are inserting bitmap rather than its intersection with the freshly inserted_items

No reason, it works like that, but we're just doing extra work for no reason 😬 

It was not as trivial as expected but it's done in https://github.com/irevoire/cellulite/commit/198a3c58e34bd366989602f61e34e9c8cc4cc868

> we cannot detail, could we maybe count the number of base cells and use that with an atomic counter? (see BaseCell::count)

We can, but for France, that's only three cells, for example, I'm not sure it'll really help 🤔 
Since the iterator yield them ordered by proximity, it also means we'll immediately jump to let's say 60/122 and then stay stuck here for 2h and then jump immediately to 122/122.
I think it's even more confusing than having nothing, but I'm open to changing it because I'm not happy either with what I have currently 😭 
The other solution would be to update the max while iterating, but then the percentage may decrease, which is also awful. Maybe we could find something in the middle where every time we process a base_cell it counts for 16 (the max resolution + 1), and then 15, 14 etc, I'm not sure it would really fix the issue though, it's probably just a mitigation 🤔

crates/meilisearch-types/src/error.rs
irevoire: Should be fixed

crates/milli/src/index.rs
irevoire: Fixed by updating cellulite to the latest version

crates/milli/src/index.rs
irevoire: Fixed by updating cellulite to the latest version

.github/workflows/test-suite.yml
```diff
@@ -184,7 +184,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
-      - uses: dtolnay/rust-toolchain@1.85
+      - uses: dtolnay/rust-toolchain@1.89
         with:
           profile: minimal
           toolchain: nightly-2024-07-09
```
dureuill, line 190: not new in this PR, but you can't set a toolchain on dtolay's action, so this just gets ignored

irevoire: Apparently no because rustfmt failed when I didn't set it to the right version, this is the error I got: https://github.com/meilisearch/meilisearch/actions/runs/17770898550/job/50506396649
You can read more on this PR: https://github.com/meilisearch/meilisearch/pull/5892

crates/milli/src/index.rs
```diff
@@ -115,9 +116,10 @@ pub mod db_name {
     pub const FIELD_ID_DOCID_FACET_STRINGS: &str = "field-id-docid-facet-strings";
     pub const VECTOR_EMBEDDER_CATEGORY_ID: &str = "vector-embedder-category-id";
     pub const VECTOR_STORE: &str = "vector-arroy";
+    pub const CELLULITE: &str = "cellulite";
```
dureuill, line 119: unused key?

crates/milli/src/update/new/document.rs
```diff
@@ -26,6 +28,7 @@ use crate::{
 ///
 /// The 'doc lifetime is meant to live sufficiently for the document to be handled by the extractors.
 pub trait Document<'doc> {
+    fn geojson_field(&self) -> Result<Option<&'doc RawValue>>;
```
dureuill: nitpick: can you add this to the bottom of the trait 😇?

and maybe add some token documentation

crates/milli/src/update/new/document.rs
```diff
@@ -26,6 +28,7 @@ use crate::{
...
     /// Iterate over all **top-level** fields of the document, returning their name and raw JSON value.
     ///
     /// - The returned values *may* contain nested fields.
```
dureuill, line 33: is _geoJSON also ignored by this function or not?

dureuill: should it?

crates/milli/src/update/index_documents/transform.rs
```diff
@@ -820,19 +820,20 @@ impl<'a, 'i> Transform<'a, 'i> {
...
-                self.indexer_settings.chunk_compression_level,
-                self.indexer_settings.max_nb_chunks,
-                self.indexer_settings.max_memory.map(|mem| mem / 2),
-                true,
-            ))
-        } else {
-            None
-        };
+        let mut original_sorter =
+            if settings_diff.reindex_vectors() || settings_diff.reindex_geojson() {
```
dureuill, line 824: this is the part where we take all searchables when the vectors must be reindexed. Why is this necessary in this case?

crates/milli/src/update/index_documents/extract/extract_geo_points.rs
```diff
@@ -107,3 +108,76 @@ fn extract_lat_lng(
...
+    primary_key_id: FieldId,
+    settings_diff: &InnerIndexSettingsDiff,
+) -> Result<grenad::Reader<BufReader<File>>> {
+    let mut writer = create_writer(
+        indexer.chunk_compression_type,
+        indexer.chunk_compression_level,
+        tempfile::tempfile()?,
+    );
+
+    tracing::info!("Extracting one geojson");
```
dureuill: should maybe be moved to `trace` level?

crates/milli/src/update/index_documents/extract/extract_geo_points.rs
```diff
@@ -107,3 +108,76 @@ fn extract_lat_lng(
         None => Ok(None),
     }
 }
+
+/// Extracts the geographical coordinates contained in each document under the `_geo` field.
```
dureuill:
```suggestion
/// Extracts the geographical coordinates contained in each document under the `_geojson` field.
```

crates/milli/src/update/index_documents/extract/extract_geo_points.rs
```diff
@@ -107,3 +108,76 @@ fn extract_lat_lng(
...
+        let del_geojson =
+            extract_geojson_field(obkv, &settings_diff.old, DelAdd::Deletion, document_id)?;
+        // extract new version
+        let add_geojson =
+            extract_geojson_field(obkv, &settings_diff.new, DelAdd::Addition, document_id)?;
+
+        if del_geojson != add_geojson {
+            let mut obkv = KvWriterDelAdd::memory();
+            if del_geojson.is_some() {
+                #[allow(clippy::drop_non_drop)]
```
dureuill: no `drop` though?

crates/milli/src/update/index_documents/extract/extract_geo_points.rs
```diff
@@ -107,3 +108,76 @@ fn extract_lat_lng(
...
+}
+
+fn extract_geojson_field(
+    obkv: &obkv::KvReader<FieldId>,
+    settings: &InnerIndexSettings,
+    deladd: DelAdd,
+    _document_id: impl Fn() -> Value,
+) -> Result<Option<GeoJson>> {
+    match settings.geojson_fid {
+        Some(fid) if settings.filterable_attributes_rules.iter().any(|rule| rule.has_geojson()) => {
```
dureuill, line 171: I feel the old_fid and new_fid should probably be hoisted in the caller rather than recomputed for each document.

crates/milli/src/update/index_documents/extract/mod.rs
```diff
@@ -62,6 +63,7 @@ pub(crate) fn data_from_obkv_documents(
                         original_documents_chunk,
                         indexer,
                         lmdb_writer_sx.clone(),
+                        primary_key_id,
```
dureuill, line 66: why is this change necessary?

dureuill: for computing the external docid, nevermind

crates/milli/src/update/new/indexer/de.rs
```diff
@@ -235,28 +233,6 @@ impl<'de, 'a, Mapper: MutFieldIdMapper> Visitor<'de> for MutFieldIdMapVisitor<'a
     }
 }
```
dureuill, line 235: surprised we removed this. Was it unused?

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -187,6 +187,14 @@ where
             facet_field_ids_delta,
         )?;
 
+        println!("Building geojson");
```
dureuill:
```suggestion
```

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -187,6 +187,14 @@ where
...
+        indexing_context.progress.update_progress(IndexingStep::BuildingGeoJson);
+        index.cellulite.build(
```
dureuill, line 191: should we guard this behind the `index.has_geojson`?

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -317,6 +325,13 @@ where
         })
         .unwrap()?;
 
+        indexing_context.progress.update_progress(IndexingStep::BuildingGeoJson);
+        index.cellulite.build(
```
dureuill, line 328: same, guard?

crates/milli/src/index.rs
irevoire: Oops, yes, my bad. I wanted to use it as a prefix for the cellulite DBs, done now ✅

crates/milli/src/update/new/document.rs
irevoire: We decided that they should be

crates/milli/src/update/index_documents/extract/extract_geo_points.rs
irevoire: I removed it entirely, I don't think we ever want to flood that much 👍

crates/milli/src/update/new/indexer/de.rs
irevoire: Yeah, it was a warning I got when I updated the msrv on my PR.
What I'm surprised by is that I could update the msrv on main without getting the warning again 🤔 
But yes, it's unused, and it still builds and cargo test without these lines 👍

crates/milli/src/update/index_documents/extract/extract_geo_points.rs
irevoire: right, that was an outdated line from when I had no deletion support I think 🤔

crates/milli/src/update/new/indexer/mod.rs
irevoire: Huum I tried to do it but I don't see any way to call `has_geojson`.
The method is defined on the settings, not the index and I can't find any way to access the settings from here 🤔 

I think I'll just implement it inside of cellulite directly. We should be able to get it down to two reads:
- one for the version
- one for the updated items

And if there is nothing we early exit

irevoire: Done here https://github.com/meilisearch/cellulite/commit/5451a30e502a4a702708b696054d9939b332ddff

crates/milli/src/update/new/indexer/mod.rs
irevoire: Done in cellulite instead since we don't have access to the `has_geojson` method in this context: https://github.com/meilisearch/cellulite/commit/5451a30e502a4a702708b696054d9939b332ddff

spointecker:
Has anyone a working example for this kind of geo-filtering?

Kerollmops:
Hey @spointecker 👋  I think you can see [some examples for GeoJSON in the doc](https://www.meilisearch.com/docs/learn/filtering_and_sorting/geosearch#preparing-documents-for-location-based-search) 👀

spointecker:
> Hey @spointecker 👋 I think you can see [some examples for GeoJSON in the doc](https://www.meilisearch.com/docs/learn/filtering_and_sorting/geosearch#preparing-documents-for-location-based-search) 👀

Yes but unfortunately I am getting this error from the endpoint:
```
{
"message":"Index `dev_voucher`: Attribute `_geojson` is not filterable. Available filterable attribute patterns are: `_geojson`, `app`, `category.sub`, `company.permalink`, `country`, `elastic_type`, `is_expired`, `is_plus`, `is_sticky`, `keywords`. 14:31 _geoPolygon([48.22400189946506, 16.33361608906227],[48.22400189946506, 16.33361608906227],[48.22400189946506, 16.33361608906227])"
"code":"invalid_search_filter"
"type":"invalid_request"
"link":"https://docs.meilisearch.com/errors#invalid_search_filter"
}
```

using this query:
`{"filter":["_geoPolygon([48.22400189946506, 16.33361608906227],[48.22400189946506, 16.33361608906227],[48.22400189946506, 16.33361608906227])"]}`

Kerollmops:
Thanks for the report @spointecker, However, would it be possible to report this issue [on our Discord](https://discord.meilisearch.com), please? You'll receive more assistance, and the Meilisearch team will be able to assist you there instead of getting deeply involved in this PR. Have a nice day 🌵

PR #5761, opened 2025-07-17
Fix chat settings dumpless upgrade

Kerollmops:
Fix a bug when using the dumpless upgrade when chat settings were already present in the index. This bug was introduced in https://github.com/meilisearch/meilisearch/pull/5710, and I mislabeled the PR as `no db change` when, in reality, it was.


irevoire (approved):
Thank you perfect

PR #5762, opened 2025-07-17
Use the edition 2024 documents indexer in the dumps

Kerollmops:
This PR makes the dump import use the edition-2024 document indexer. It also provides an experimental CLI option and an env variable to disable this and keep using the pre-edition-2024 document indexer. You can take a look at [the new experimental feature page](https://github.com/orgs/meilisearch/discussions/851).


crates/dump/src/reader/mod.rs
```diff
@@ -192,6 +192,13 @@ impl DumpIndexReader {
         }
     }
 
+    pub fn documents_file(&self) -> &File {
```
dureuill, line 196:
```suggestion
    /// A reference to a file in the NDJSON format containing all the documents of the index
    pub fn documents_file(&self) -> &File {
```

crates/index-scheduler/src/lib.rs
```diff
@@ -139,6 +139,8 @@ pub struct IndexSchedulerOptions {
     pub embedding_cache_cap: usize,
     /// Snapshot compaction status.
     pub experimental_no_snapshot_compaction: bool,
+    /// Whether dump import use the old document indexer or the new one.
```
dureuill:
```suggestion
    /// Whether dump import uses the old document indexer or the new one.
```

crates/meilisearch/Cargo.toml
```diff
@@ -50,6 +50,7 @@ jsonwebtoken = "9.3.1"
 lazy_static = "1.5.0"
 meilisearch-auth = { path = "../meilisearch-auth" }
 meilisearch-types = { path = "../meilisearch-types" }
+memmap2 = "0.9.7"
```
dureuill, line 53:
```suggestion
memmap2 = "0.9.5"
```

The rest of the code is using this version, I don't think we need to have both

Kerollmops: The Cargo.lock has been updated, I'll update the rest of the toml to switch to the latest version.

crates/meilisearch/src/option.rs
```diff
@@ -467,6 +469,15 @@ pub struct Opt {
...
     pub experimental_no_snapshot_compaction: bool,
 
+    /// Experimental make dump imports use the old document indexer.
+    ///
+    /// When enabled, Meilisearch will use the old document indexer when importing dumps.
+    ///
+    /// For more information, see <https://github.com/orgs/meilisearch/discussions/851>.
+    #[clap(long, env = MEILI_EXPERIMENTAL_NO_EDITION_2024_FOR_DUMPS)]
+    #[serde(default)]
+    pub experimental_no_edition_2024_for_dumps: bool,
```
dureuill: should that be in the indexer opts instead? Really not sure so just asking

dureuill:
nitpick: any option to display the progress of the document import in the logs?

PR #5763, opened 2025-07-17
Regenerate all fragments when coming from a user provided vector

dureuill:
This PR makes the following changes:

1. This changes the indexers in order to regenerate all fragments if the embeddings for that document were previously marked as user-provided. 
2. Modifies the dump creation (+ meilitool) to inject vectors with `regenerate` set to `false` when fragments are involved. This is to avoid regenerating too many embeddings at once when fragments are involved.

This is important especially in the context of dumps, as otherwise documents would have their embeddings removed unexpectedly without any new embedding being generated.


crates/meilisearch/db.snapshot
Kerollmops, line 1: Nice 👍

crates/milli/src/update/new/extract/vectors/mod.rs
```diff
@@ -669,13 +684,16 @@ impl<'a, 'b, 'extractor> Chunks<'a, 'b, 'extractor> {
                         Result::Ok(())
                     },
                 )?;
-                self.set_status(docid, old_is_user_provided, true, false, true);
             }
             ChunkType::DocumentTemplate { document_template, session } => {
                 let doc_alloc = session.doc_alloc();
 
+                if old_is_user_provided | full_reindex {
```
dureuill: `| => ||`

Kerollmops (approved):

PR #5767, opened 2025-07-21
Add index setting to switch from arroy to hannoy

Kerollmops:
This PR adds an experimental feature an index setting that allow migration from arroy to hannoy (and back as needs be). 

- [Experimental feature description](https://github.com/orgs/meilisearch/discussions/860)
- [Linear page (internal)](https://linear.app/meilisearch/project/arroy-becomes-hannoy-de1a9dd1005e/issues)

- [x] Detail the change to the DB format and why they are forward compatible: see section below
- [ ] Forward-compatibility: A database created before this PR and using the features touched by this PR was able to be opened by a Meilisearch produced by the code of this PR.

### Detail the changes to the DB format

1. A new experimental feature `vectorStoreSetting` has been added. Adding and removing experimental features is both forward and backward-compatible.
2. A new index setting `vectorStore` has been added. It has a default value so is forward-compatible.
3. When using the new index setting to set it to the `"experimental"` value, then the vector database is moved from arroy to hannoy. As the default value for the `vectorStore` setting is `"stable"`, existing databases don't require any migration, making the change forward-compatible.

---

There is a [running prototype](https://github.com/meilisearch/meilisearch/actions/runs/16719828947) to generate a Docker Image. The name of the Docker image is `prototype-arroy-becomes-hannoy-x` where `x` is an incrementing number.


crates/meilitool/src/upgrade/v1_11.rs
```diff
@@ -79,15 +79,16 @@ pub fn v1_10_to_v1_11(
...
+            try_opening_poly_database(&index_env, &index_wtxn, db_name::VECTOR_HANNOY)
                 .with_context(|| format!("while updating date format for index `{uid}`"))?;
 
-        meilisearch_types::milli::arroy::upgrade::cosine_from_0_4_to_0_5(
-            &index_rtxn,
-            index_read_database.remap_types(),
-            &mut index_wtxn,
-            index_write_database.remap_types(),
-        )?;
+        // meilisearch_types::milli::hannoy::upgrade::cosine_from_0_4_to_0_5(
```
Kerollmops: When moving from v1.11 we will keep arroy to do the upgrade.

crates/milli/Cargo.toml
```diff
@@ -87,7 +87,7 @@ rhai = { version = "1.22.2", features = [
     "no_time",
     "sync",
 ] }
-arroy = "0.6.1"
+hannoy = { git = "https://github.com/kerollmops/hannoy", branch = "basic-and-ugly-filtering" }
```
Kerollmops: @nnethercott Can you release a version 0.0.1 on crates.io, please? It would be safer to use on our end 🌵

nnethercott: Yo are the `hannoy_stats` defined on your end or are you using the build stats from hannoy for the `IndexStats`

Kerollmops: Hey @nnethercott, we are using the build stats of Hannoy to basically retrieve the number of embedded documents.

crates/milli/src/index.rs
```diff
@@ -113,7 +113,7 @@ pub mod db_name {
     pub const FIELD_ID_DOCID_FACET_F64S: &str = "field-id-docid-facet-f64s";
     pub const FIELD_ID_DOCID_FACET_STRINGS: &str = "field-id-docid-facet-strings";
     pub const VECTOR_EMBEDDER_CATEGORY_ID: &str = "vector-embedder-category-id";
-    pub const VECTOR_ARROY: &str = "vector-arroy";
+    pub const VECTOR_HANNOY: &str = "vector-hannoy";
```
Kerollmops:
```suggestion
    pub const VECTOR_STORE: &str = "vector-arroy";
```

We will need to use the same database as we plan to keep the same format for the raw vectors and do the dumpless upgrade by "simply" deleting all the arroy tree nodes and ask Hannoy to generate the links by using the raw vectors. cc @nnethercott

crates/milli/src/index.rs
```diff
@@ -177,10 +177,10 @@ pub struct Index {
     /// Maps the document id, the facet field id and the strings.
     pub field_id_docid_facet_strings: Database<FieldDocIdFacetStringCodec, Str>,
 
-    /// Maps an embedder name to its id in the arroy store.
+    /// Maps an embedder name to its id in the hannoy store.
     pub(crate) embedder_category_id: Database<Unspecified, Unspecified>,
-    /// Vector store based on arroy™.
-    pub vector_arroy: arroy::Database<Unspecified>,
+    /// Vector store based on hannoy™.
+    pub vector_hannoy: hannoy::Database<Unspecified>,
```
Kerollmops: To discuss, but we should only keep the hannoy one and remap the types when we need. We can do that in the `VectorStore` wrapper ex-`ArroyWrapper`.

crates/milli/src/progress.rs
```diff
@@ -98,12 +98,12 @@ impl Progress {
     }
 
     // TODO: ideally we should expose the progress in a way that let arroy use it directly
-    pub(crate) fn update_progress_from_arroy(&self, progress: arroy::WriterProgress) {
```
Kerollmops: In progress https://github.com/nnethercott/hannoy/pull/16

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -66,7 +66,7 @@ where
     let mut bbbuffers = Vec::new();
     let finished_extraction = AtomicBool::new(false);
 
-    let arroy_memory = grenad_parameters.max_memory;
+    let hannoy_memory = grenad_parameters.max_memory;
```
Kerollmops: Hey @nnethercott, do you use the max memory in Hannoy to build the tree?

crates/milli/src/update/upgrade/v1_14.rs
```diff
@@ -25,12 +25,13 @@ impl UpgradeIndex for Latest_V1_13_To_Latest_V1_14 {
         progress.update_progress(VectorStore::UpdateInternalVersions);
 
         let rtxn = index.read_txn()?;
-        arroy::upgrade::from_0_5_to_0_6::<Cosine>(
```
Kerollmops, line 28: We need to keep that to be able to move from v1.13 to v1.14.

crates/milli/src/vector/mod.rs
```diff
@@ -41,15 +41,15 @@ pub type Embedding = Vec<f32>;
...
-pub struct ArroyWrapper {
+pub struct HannoyWrapper {
     quantized: bool,
     embedder_index: u8,
-    database: arroy::Database<Unspecified>,
+    database: hannoy::Database<Unspecified>,
 }
 
-impl ArroyWrapper {
+impl HannoyWrapper {
```
Kerollmops: We should name this `VectorStore` instead

crates/milli/src/vector/mod.rs
```diff
@@ -140,39 +140,41 @@ impl ArroyWrapper {
...
-    ) -> Result<(), arroy::Error> {
-        for index in arroy_store_range_for_embedder(self.embedder_index) {
+    ) -> Result<(), hannoy::Error> {
+        for index in hannoy_store_range_for_embedder(self.embedder_index) {
             if self.quantized {
-                let writer = arroy::Writer::new(self.quantized_db(), index, dimension);
+                let writer = hannoy::Writer::new(self.quantized_db(), index, dimension);
                 if writer.need_build(wtxn)? {
-                    writer.builder(rng).build(wtxn)?
+                    writer.builder(rng).ef_construction(48).build::<16, 32>(wtxn)?
```
Kerollmops: @nnethercott we were talking about these parameters with @dureuill and were thinking about providing some presets for different tradeoffs:
- Indexing vs search performance
- Based on the different number of dimensions

However, we must determine how many presets we want to propose. What do you think?

crates/milli/src/vector/mod.rs
```diff
@@ -140,39 +140,41 @@ impl ArroyWrapper {
...
                         .builder(rng)
-                        .available_memory(arroy_memory.unwrap_or(usize::MAX))
-                        .progress(|step| progress.update_progress_from_arroy(step))
-                        .cancel(cancel)
-                        .build(wtxn)?;
+                        .available_memory(hannoy_memory.unwrap_or(usize::MAX))
+                        // .progress(|step| progress.update_progress_from_hannoy(step))
+                        // .cancel(cancel)
+                        .ef_construction(48)
+                        .build::<16, 32>(wtxn)?;
```
Kerollmops: I'll put these variable into consts.

crates/milli/src/vector/mod.rs
```diff
@@ -335,71 +337,68 @@ impl ArroyWrapper {
...
 
-        for index in arroy_store_range_for_embedder(self.embedder_index) {
-            let writer = arroy::Writer::new(db, index, dimension);
-            let Some(candidate) = writer.item_vector(wtxn, item_id)? else {
-                continue;
-            };
-            if candidate == vector {
+        for index in hannoy_store_range_for_embedder(self.embedder_index) {
+            let writer = hannoy::Writer::new(db, index, dimension);
+            if writer.contains_item(wtxn, item_id)? {
```
Kerollmops: I need to reintroduce the `hannoy::Writer::item_vector`.

crates/milli/src/vector/mod.rs
```diff
@@ -442,27 +441,27 @@ impl ArroyWrapper {
...
         limit: usize,
         filter: Option<&RoaringBitmap>,
-    ) -> Result<Vec<(ItemId, f32)>, arroy::Error> {
+    ) -> Result<Vec<(ItemId, f32)>, hannoy::Error> {
         let mut results = Vec::new();
 
         for reader in self.readers(rtxn, db) {
             let reader = reader?;
-            let mut searcher = reader.nns(limit);
+            let mut searcher = reader.nns(limit, limit * 2); // TODO find better ef
```
Kerollmops, line 464: @nnethercott What would be the best default value we could put, and should we base it on the limit?

Kerollmops: @dureuill tells me that we should base this on the `maxTotalHits`. The problem is that even if we ask for one doc we will pay the cost.

crates/milli/src/vector/mod.rs
```diff
@@ -484,27 +483,27 @@ impl ArroyWrapper {
...
```
Kerollmops: Same here.

crates/milli/src/update/new/indexer/mod.rs
nnethercott: Not currently, lemme know if you guys are running into issues later though & we can integrate what's in arroy here

crates/milli/src/vector/mod.rs
nnethercott: For `M` I'd go with like 8,12,16,32 (and `M0` is thus 16,24,32,64) and for `ef_construction` you need it to be bigger than the M you selected. [Here's an interactive graph](https://ann-benchmarks.com/weaviate.html) for weaviate's hnsw implementation showing how recall and query speed vary with these params to give you an idea.

Taking `M` and `ef_construction` big can really slow down indexing times (slower than arroy), especially on large dimensional datasets. 

M=16 is a good default for most datasets. Another parameter you need to consider is `ef_search` which is used in the reader.nns (i'm gonna turn this into a builder method later). This controls how long you search for before stopping.

crates/milli/src/vector/mod.rs
```diff
@@ -225,7 +225,7 @@ impl VectorStore {
                 // only happens once in the life of an embedder, it's not very performances
                 // sensitive.
                 if quantizing && !self.quantized {
-                    let writer = writer.prepare_changing_distance::<BinaryQuantizedCosine>(wtxn)?;
+                    let writer = writer.prepare_changing_distance::<Hamming>(wtxn)?;
```
nnethercott: ayyyyyy 😎

crates/milli/src/update/index_documents/extract/extract_word_docids.rs
```diff
@@ -201,45 +200,3 @@ fn words_into_sorter(
 
     Ok(())
 }
-
-#[tracing::instrument(level = "trace", skip_all, target = "indexing::extract")]
-fn docids_into_writers<W>(
```
dureuill, line 206: I wager @Kerollmops removed this function because it was unused

crates/milli/src/vector/embedder/mod.rs
dureuill, line 1: Content of this file moved from `vector/mod.rs`

crates/milli/src/vector/distribution.rs
dureuill, line 1: Content of this file moved from `vector/mod.rs`

crates/milli/src/vector/embeddings.rs
dureuill, line 1: Content of this file moved from `vector/mod.rs`

crates/milli/src/vector/runtime.rs
dureuill, line 1: Content of this file moved from `vector/mod.rs`

crates/milli/src/vector/store.rs
dureuill, line 1: > [!WARNING]
> Content of this file heavily modified to support both hannoy and arroy after moving from `vector/mod.rs`. Review with care.

crates/milli/src/progress.rs
```diff
@@ -277,6 +277,30 @@ impl<U: Send + Sync + 'static> Step for VariableNameStep<U> {
...
+
+    fn current(&self) -> u32 {
+        self.0.current().try_into().unwrap_or(u32::MAX)
+    }
+
+    fn total(&self) -> u32 {
+        self.0.total().try_into().unwrap_or(u32::MAX)
+    }
+}
+
```
dureuill, line 303: not sure why this is part of this PR tbh

crates/meilisearch-types/src/settings.rs
```diff
@@ -922,6 +940,9 @@ pub fn settings(
             (name, SettingEmbeddingSettings { inner: Setting::Set(config.into()) })
         })
         .collect();
+
+    let vector_store = index.get_vector_store(rtxn)?;
```
dureuill, line 944: do not lose the default value here

crates/meilisearch/src/routes/indexes/settings.rs
```diff
@@ -770,5 +789,9 @@ fn validate_settings(
         features.check_chat_completions("setting `chat` in the index settings")?;
     }
 
+    if let Setting::Set(_) = &settings.vector_store {
+        features.check_vector_store_setting("setting `vectorStore`  in the index settings")?;
```
dureuill:
```suggestion
        features.check_vector_store_setting("setting `vectorStore` in the index settings")?;
```

crates/meilisearch/tests/vector/binary_quantized.rs
```diff
@@ -335,3 +335,191 @@ async fn binary_quantize_clear_documents() {
...
+            }
+          }
+        }
+      ],
+      "offset": 0,
+      "limit": 20,
+      "total": 2
+    }
+    "###);
+}
```
dureuill, line 525: test backend change without bq

crates/milli/src/index.rs
```diff
@@ -454,6 +455,35 @@ impl Index {
...
+        backend: VectorStoreBackend,
+    ) -> Result<()> {
+        Ok(self.main.remap_types::<Str, SerdeJson<VectorStoreBackend>>().put(
+            wtxn,
+            main_key::VECTOR_STORE_BACKEND,
+            &backend,
+        )?)
+    }
+
+    pub fn get_vector_store(&self, rtxn: &RoTxn<'_>) -> Result<VectorStoreBackend> {
```
dureuill: `Result<Option<VectorStoreBackend>>`

crates/milli/src/progress.rs
dureuill: Discussed with tamo

- Implement the steppe trait for our version of progress
- in steppe we are using u64 rather than u32, otherwise it's pretty similar

crates/milli/src/vector/store.rs
```diff
@@ -0,0 +1,1176 @@
...
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    Eq,
+    Default,
+    Serialize,
+    Deserialize,
+    deserr::Deserr,
+    utoipa::ToSchema,
```
dureuill, line 26: add to the mega list of types in utoipa

crates/milli/src/vector/store.rs
```diff
@@ -0,0 +1,1176 @@
...
+        item: ItemId,
+        limit: usize,
+        filter: Option<&RoaringBitmap>,
+    ) -> Result<Vec<(ItemId, f32)>, hannoy::Error> {
+        let mut results = Vec::new();
+
+        for reader in self._hannoy_readers(rtxn, db) {
+            let reader = reader?;
+            let mut searcher = reader.nns(limit);
+            searcher.ef_search((limit * 10).max(100)); // TODO find better ef
```
dureuill, line 955:
```suggestion
            searcher.ef_search((limit * 10).max(limit)); // TODO find better ef
```

?

Should we have  at least `ef == limit`?

crates/milli/src/vector/store.rs
```diff
@@ -0,0 +1,1176 @@
...
+        vector: &[f32],
+        limit: usize,
+        filter: Option<&RoaringBitmap>,
+    ) -> Result<Vec<(ItemId, f32)>, hannoy::Error> {
+        let mut results = Vec::new();
+
+        for reader in self._hannoy_readers(rtxn, db) {
+            let reader = reader?;
+            let mut searcher = reader.nns(limit);
+            searcher.ef_search((limit * 10).max(100)); // TODO find better ef
```
dureuill, line 1009: same as above?

crates/milli/src/vector/store.rs
dureuill: done

crates/meilisearch-types/src/settings.rs
dureuill: done

crates/milli/src/index.rs
dureuill: done

crates/milli/src/vector/store.rs
dureuill: see https://github.com/meilisearch/meilisearch/issues/5877

crates/milli/src/vector/store.rs
dureuill: see https://github.com/meilisearch/meilisearch/issues/5877

crates/meilisearch/tests/vector/binary_quantized.rs
dureuill: done

crates/milli/src/vector/store.rs
```diff
@@ -0,0 +1,1176 @@
...
+        rtxn: &RoTxn,
+        wtxn: &mut RwTxn,
+        progress: Progress,
+        must_stop_processing: &MSP,
+        available_memory: Option<usize>,
+    ) -> crate::Result<()>
+    where
+        MSP: Fn() -> bool + Sync,
+    {
+        let mut rng = rand::rngs::StdRng::from_entropy();
```
irevoire, line 133: Nice, I didn't know that existed 😲

crates/milli/src/vector/store.rs
```diff
@@ -0,0 +1,1176 @@
...
+            let hannoy_reader: hannoy::Reader<HD> =
+                match hannoy::Reader::open(hannoy_rtxn, index, self.database.remap_types()) {
+                    Ok(reader) => reader,
+                    Err(hannoy::Error::MissingMetadata(_)) => continue,
+                    Err(err) => return Err(err.into()),
+                };
+            let dimensions = hannoy_reader.dimensions();
+            let arroy_writer: arroy::Writer<AD> =
+                arroy::Writer::new(self.database.remap_types(), index, dimensions);
+            arroy_writer.clear(arroy_wtxn)?;
```
irevoire, line 1098:
```suggestion
            // Since the bq mode of arroy and hannoy are not compatible, we have to clear and re-insert everything
            arroy_writer.clear(arroy_wtxn)?;
```

crates/milli/src/vector/store.rs
```diff
@@ -0,0 +1,1176 @@
...
+            let arroy_reader: arroy::Reader<AD> =
+                match arroy::Reader::open(arroy_rtxn, index, self.database.remap_types()) {
+                    Ok(reader) => reader,
+                    Err(arroy::Error::MissingMetadata(_)) => continue,
+                    Err(err) => return Err(err.into()),
+                };
+            let dimensions = arroy_reader.dimensions();
+            let hannoy_writer: hannoy::Writer<HD> =
+                hannoy::Writer::new(self.database.remap_types(), index, dimensions);
+            hannoy_writer.clear(hannoy_wtxn)?;
```
irevoire, line 1060:
```suggestion
            // Since the bq mode of arroy and hannoy are not compatible, we have to clear and re-insert everything
            hannoy_writer.clear(hannoy_wtxn)?;
```

irevoire (approved):
Hey I think we're good!
I think it would be nice to add a small comment on the why we're clearing the db when it's binary quantized wdyt 🤔

crates/milli/src/vector/store.rs
dureuill: I didn't add that though, it's from kero I think

irevoire (approved):

PR #5773, opened 2025-07-22
Fix snapshotCreation task being included in snapshot

### LINKED ISSUE #4653: snapshotCreation task included in snapshot
Hi!

We have about 30G index with something like 2M documents here. When we are creating snapshot there (it takes about 30 minutes or so, just as an info), and when we are importing it to another instance, new instance was taking 100% CPU a while after start.

Then we found out, that even if we start this new instance without any snapshot parameter, there is still snapshotCreation -task there, with status `processing`.

Of course, we can cancel the tasks (seems to take a while) but it would be best, that that task would not be there in snapshot waiting to be launched on new instance.

Or, are we probably doing something wrong here?

We are launching snapshot via API and then importing it via `--import-snapshot`.

Mubelotix:
# Pull Request

## Related issue
Fixes #4653

### DIFF
--- a/crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
+++ b/crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
@@ -15,7 +15,7 @@ use crate::{Error, IndexScheduler, Result};
 /// # Safety
 ///
 /// See [`EnvOpenOptions::open`].
-unsafe fn mark_tasks_as_succeeded(
+unsafe fn remove_tasks(
     tasks: &[Task],
     dst: &std::path::Path,
     index_base_map_size: usize,
@@ -112,7 +112,7 @@ impl IndexScheduler {
         };
         self.env.copy_to_path(dst.join("data.mdb"), compaction_option)?;
 
-        // 2.2 Mark the current snapshot tasks as succeeded in the newly created env
+        // 2.2 Remove the current snapshot tasks
         //
         // This is done to ensure that the tasks are not processed again when the snapshot is imported
         //
@@ -121,7 +121,7 @@ impl IndexScheduler {
         // This is safe because we open the env file we just created in a temporary directory.
         // We are sure it's not being used by any other process nor thread.
         unsafe {
-            mark_tasks_as_succeeded(&tasks, &dst, self.index_mapper.index_base_map_size)?;
+            remove_tasks(&tasks, &dst, self.index_mapper.index_base_map_size)?;
         }
 
         // 2.3 Create a read transaction on the index-scheduler


Mubelotix:
I'm still wondering if it's a problem for a succeeded task to lack `finished_at` and `duration` values. Those can't be determined in advance as the task isn't actually finished, but I could add dummy values for this one

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -7,9 +7,33 @@ use meilisearch_types::milli::progress::{Progress, VariableNameStep};
...
+    let env = env_options.max_dbs(nb_db).map_size(index_base_map_size).open(dst)?;
+    let mut wtxn = env.write_txn()?;
+    let task_queue = TaskQueue::new(&env, &mut wtxn)?;
+    for mut task in tasks.iter().cloned() {
+        task.status = Status::Succeeded;
+        task_queue.update_task(&mut wtxn, &task)?;
+    }
+    wtxn.commit()?;
+    Ok(())
+}
```
irevoire:
```suggestion
unsafe fn mark_tasks_as_succeeded(
    tasks: &[Task],
    dst: &std::path::Path,
    index_base_map_size: usize,
) -> Result<()> {
    let env_options = EnvOpenOptions::new();
    let mut env_options = env_options.read_txn_without_tls();
    let env = env_options.max_dbs(TaskQueue::nb_db()).map_size(index_base_map_size).open(dst)?;
    let mut wtxn = env.write_txn()?;
    let task_queue = TaskQueue::new(&env, &mut wtxn)?;
    for mut task in tasks.iter().cloned() {
        task.status = Status::Succeeded;
        task_queue.update_task(&mut wtxn, &task)?;
    }
    wtxn.commit()?;
    Ok(())
}
```

Hey, here we can use the [provided number of DB](https://github.com/meilisearch/meilisearch/blob/421a23ee3d0952a9d719f3d7434f6e418e36f4f0/crates/index-scheduler/src/queue/tasks.rs#L67) directly instead of receiving it from somewhere and taking the `Self::nb_db` which contains enough space for _all_ db instead of just the task queue db.

crates/meilisearch/tests/snapshot/mod.rs
```diff
@@ -203,3 +199,52 @@ async fn perform_on_demand_snapshot() {
...
+    let options = Opt { import_snapshot: Some(snapshot_path), ..default_settings(temp.path()) };
+    let snapshot_server = Server::new_with_options(options).await.unwrap();
+
+    // The snapshot creation task should NOT be spawned again => task is succeeded
+    let (task, code) = snapshot_server.get_task(task.uid()).await;
+    snapshot!(code, @"200 OK");
+    snapshot!(json_string!(task, { ".enqueuedAt" => "[date]" }), @r#"
+    {
+      "uid": 0,
+      "batchUid": 0,
```
irevoire: I don't understand where this batch_id comes from.

To be sure we did everything properly, could you also snapshot the associated batch please?

crates/meilisearch/tests/snapshot/mod.rs
```diff
@@ -203,3 +199,52 @@ async fn perform_on_demand_snapshot() {
...
+      "type": "snapshotCreation",
+      "canceledBy": null,
+      "error": null,
+      "duration": null,
+      "enqueuedAt": "[date]",
+      "startedAt": null,
+      "finishedAt": null
+    }
+    "#);
+}
```
irevoire, line 268: A bit boring, sorry, but since the tasks also contain metadata, could you also make sure you can retrieve the task and batch while using filter please?

Usually what I do is at least one request on the status or kind + at least one request on the startedAt, finishedAt or enqueuedAt since the dates are usually stored elsewhere.
Same for the batch.

For these tests, there is no need to be smart or precise. Don't even snapshot the results; just make sure that it returns 1 task. We know the index is empty and it'll be ours 👍

irevoire (changes requested):
Hey that's a great solution you found! It's safe and won't lead to any corruption ever 👍 

> I'm still wondering if it's a problem for a succeeded task to lack finished_at and duration values. Those can't be determined in advance as the task isn't actually finished, but I could add dummy values for this one

About that yes, I would highly prefer we set the duration properly as the rest of the codebase expect a finished task to contains it.

Something I don't understand as well is where does your batch_uid comes from?
The current batch processing is not written in the database so in your test at the end it should be marked as null instead of 0 🤔 
Anyway just saying that you should also update the batch_uid and create a dummy batch.

We don't really care if the date are slightly off don't worry

crates/meilisearch/tests/snapshot/mod.rs
Mubelotix: You were right, tasks refer to a Batch that does not exist in the db. It seems way too complicated to insert the batch and make sure all its metadata is up to date, and also set timing in the tasks. I'll just remove the tasks so that the snapshot reflects the state before the snapshot was enqueued

Mubelotix:
It seems way too complicated to insert the batch and make sure all its metadata is up to date, and also set timing in the tasks. I chose to just remove the tasks so that the snapshot reflects the state before the snapshot was enqueued

irevoire (approved):
Hey, yeah it's not as cool as initially planned but it works, we don't need the info and it's easy to maintain, well done 👍  
Thanks for finding a solution 🔥

PR #5775, opened 2025-07-23
Experimental search personalization

ManyTheFish:
## Summary

Linear (intern): [link](https://linear.app/meilisearch/issue/PRO-755/develop-prototype-branch-on-meilisearch)
notion (intern): [link](https://www.notion.so/meilisearch/Personalized-Search-2334b06b651f8034a587e7ebd8c086a1)
experimental feature discussion: https://github.com/orgs/meilisearch/discussions/866

Enhance search results by incorporating user context using Cohere's reranking API.

> Experimental feature: this feature is deactivated by default and can only be activated at Meilisearch start using CLI or an environment variable. Moreover, it has been implemented only on `indexes/<index>/search/`route and not multi-search 

## Setup

1. Get a Cohere API key from [cohere.ai](https://cohere.ai)
2. Set the environment variable:

```bash
export MEILI_EXPERIMENTAL_PERSONALIZATION_API_KEY="your_cohere_api_key_here"
```

## Usage

Add a `personalize` object to your search requests:

```json
{
  "q": "laptop",
  "personalize": {
    "userContext": "User is a graphic designer who needs high-performance hardware for video editing"
  }
}
```

### DIFF
--- a/crates/meilisearch/tests/search/errors.rs
+++ b/crates/meilisearch/tests/search/errors.rs
@@ -1326,6 +1326,7 @@ async fn search_with_contains_without_enabling_the_feature() {
 }
 
 #[actix_rt::test]
+#[ignore]
 async fn search_with_personalization_invalid_api_key() {
     // Create a server with a fake personalization API key
     let dir = TempDir::new().unwrap();


crates/index-scheduler/src/features.rs
```diff
@@ -234,6 +241,10 @@ impl FeatureData {
         RoFeatures::new(self)
     }
 
+    pub fn experimental_personalization_api_key(&self) -> Option<&String> {
+        self.experimental_personalization_api_key.as_ref()
+    }
```
dureuill:
```suggestion
    pub fn experimental_personalization_api_key(&self) -> Option<&str> {
        self.experimental_personalization_api_key.as_deref()
    }
```

crates/index-scheduler/src/lib.rs
```diff
@@ -886,6 +887,10 @@ impl IndexScheduler {
         self.features.features()
     }
 
+    pub fn experimental_personalization_api_key(&self) -> Option<&String> {
+        self.features.experimental_personalization_api_key()
+    }
```
dureuill:
```suggestion
    pub fn experimental_personalization_api_key(&self) -> Option<&str> {
        self.features.experimental_personalization_api_key()
    }
```

crates/meilisearch-types/src/error.rs
```diff
@@ -315,6 +315,8 @@ InvalidSearchShowRankingScoreDetails           , InvalidRequest       , BAD_REQU
 InvalidSimilarShowRankingScoreDetails          , InvalidRequest       , BAD_REQUEST ;
 InvalidSearchSort                              , InvalidRequest       , BAD_REQUEST ;
 InvalidSearchDistinct                          , InvalidRequest       , BAD_REQUEST ;
+InvalidSearchPersonalize                        , InvalidRequest       , BAD_REQUEST ;
+InvalidSearchPersonalizeUserContext             , InvalidRequest       , BAD_REQUEST ;
```
dureuill:
```suggestion
InvalidSearchPersonalize                       , InvalidRequest       , BAD_REQUEST ;
InvalidSearchPersonalizeUserContext            , InvalidRequest       , BAD_REQUEST ;
```

crates/meilisearch/src/personalization/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+                // Reorder the hits based on Cohere's reranking
+                let mut reranked_hits = Vec::new();
+                for index in reranked_indices.iter() {
+                    reranked_hits.push(search_result.hits[*index].clone());
+                }
+
+                Ok(SearchResult { hits: reranked_hits, ..search_result })
+            }
+            Err(e) => {
+                error!("Cohere rerank failed with model EnglishV3: {}", e);
```
dureuill: should we surface the error in the response somehow? Having it only in logs means that it will be impossible to support from a Cloud user's PoV

crates/meilisearch-types/src/features.rs
```diff
@@ -24,11 +24,12 @@ pub struct RuntimeTogglableFeatures {
     pub vector_store_setting: bool,
 }
 
-#[derive(Default, Debug, Clone, Copy)]
+#[derive(Default, Debug, Clone)]
 pub struct InstanceTogglableFeatures {
     pub metrics: bool,
     pub logs_route: bool,
     pub contains_filter: bool,
+    pub experimental_personalization_api_key: Option<String>,
```
dureuill: I don't think we ought to keep the key here since it is already inserted in the AppData of the personalization service.

Instead, I would have `personalization_api_key: bool` in this struct (retain its `Copy` trait)

crates/meilisearch/src/lib.rs
```diff
@@ -684,12 +685,19 @@ pub fn configure_data(
     (logs_route, logs_stderr): (LogRouteHandle, LogStderrHandle),
     analytics: Data<Analytics>,
 ) {
+    // Create personalization service with API key from options
+    let personalization_service = index_scheduler
+        .experimental_personalization_api_key()
+        .cloned()
+        .map(personalization::PersonalizationService::cohere)
+        .unwrap_or_else(personalization::PersonalizationService::uninitialized);
```
dureuill: the key doesn't need to be stored in the index scheduler here, as it can be recovered from `Opt`.

crates/meilisearch/src/personalization/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+impl CohereService {
+    pub fn new(api_key: String) -> Self {
+        info!("Personalization service initialized with Cohere API");
+        Self { cohere: Cohere::new("https://api.cohere.ai", api_key) }
+    }
+
+    pub async fn rerank_search_results(
+        &self,
+        search_result: SearchResult,
+        personalize: Option<&Personalize>,
```
dureuill: I find it would make more sense to have `&Personalize` be passed without option here, and have the caller only make the call when the option is some.

crates/meilisearch/src/routes/indexes/search.rs
```diff
@@ -461,6 +480,10 @@ pub async fn search_with_post(
 
     let include_metadata = parse_include_metadata_header(&req);
 
+    // Extract personalization and query string before moving query
+    let personalize = query.personalize.clone();
```
dureuill: we need to stop in error here if the experimental feature is not enabled and `personalize` is Some, otherwise we're making the feature *de-facto* stable.

crates/meilisearch/src/routes/indexes/search.rs
```diff
@@ -345,6 +353,11 @@ pub async fn search_with_url_query(
     let search_kind =
         search_kind(&query, index_scheduler.get_ref(), index_uid.to_string(), &index)?;
     let retrieve_vector = RetrieveVectors::new(query.retrieve_vectors);
+
+    // Extract personalization and query string before moving query
+    let personalize = query.personalize.clone();
```
dureuill: we need to stop in error here if the experimental feature is not enabled and `personalize` is Some, otherwise we're making the feature *de-facto* stable.

crates/meilisearch/Cargo.toml
```diff
@@ -95,6 +95,7 @@ uuid = { version = "1.18.0", features = ["serde", "v4", "v7"] }
 serde_urlencoded = "0.7.1"
 termcolor = "1.4.1"
 url = { version = "2.5.4", features = ["serde"] }
+cohere-rust = "0.6.0"
```
dureuill: not sure about adding this dependency here:

1. we're only using a small part of cohere's API, we could use the retry code + reqwest to make the call ourselves
2. we don't know anything about their retry strategy (exponential backoff? looking quickly there seems to be none) and timeouts (looks like 240s), and how graceful their error handling is.
3. this is a one-person project

crates/meilisearch/src/lib.rs
```diff
@@ -684,12 +685,19 @@ pub fn configure_data(
     (logs_route, logs_stderr): (LogRouteHandle, LogStderrHandle),
     analytics: Data<Analytics>,
 ) {
+    // Create personalization service with API key from options
+    let personalization_service = opt
+        .clone()
+        .experimental_personalization_api_key
```
dureuill:
```suggestion
        .experimental_personalization_api_key
        .clone()
```

Only clone what we need

crates/meilisearch/src/personalization/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+    #[error("Personalization service: Deadline exceeded")]
+    DeadlineExceeded,
+    #[error(
+        "{disabled_action} requires enabling the `{feature}` experimental feature. See {issue_link}"
+    )]
+    FeatureNotEnabled {
+        disabled_action: &'static str,
+        feature: &'static str,
+        issue_link: &'static str,
+    },
```
dureuill: Could we avoid duplicating this error?

crates/meilisearch/src/personalization/mod.rs
ManyTheFish: I've added a bunch of errors that make the search fail, they are mainly forwarding the rerankibng service errors

crates/meilisearch/src/personalization/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+}
+
+pub struct CohereService {
+    client: Client,
+    api_key: String,
+}
+
+impl CohereService {
+    pub fn new(api_key: String) -> Self {
+        info!("Personalization service initialized with Cohere API");
```
dureuill, line 65: that log appears to be duplicated
Capture d’écran 2025-11-05 à 11 54 09

crates/meilisearch/src/routes/indexes/search.rs
ManyTheFish: indeed my bad, for this one the error return is made inside the personalization service.
For now, the `feature_not_enabled` error is returned for any `RuntimeTogglableFeatures`, however, the personalization feature can't be activated at runtime and so have no place in the `RuntimeTogglableFeatures` struct.
I chose to put it inside the service for both simplicity and locality.

crates/meilisearch/src/lib.rs
ManyTheFish: I completelly removed the personalization from the index scheduler

crates/meilisearch/Cargo.toml
ManyTheFish: The cohere dependency have been removed and I copy pasted the retry strategy used in the emeddings processing

crates/meilisearch/src/analytics/segment_analytics.rs
```diff
@@ -285,6 +285,7 @@ impl Infos {
             indexer_options,
             config_file_path,
             no_analytics: _,
+            experimental_personalization_api_key: _,
```
dureuill: shouldn't the presence of an API key (feature enabled) be added to the analytics?

crates/meilisearch/src/personalization/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+    }
+}
+
+#[derive(Serialize)]
+struct CohereRerankRequest {
+    query: String,
+    documents: Vec<String>,
+    model: String,
+    top_n: Option<usize>,
+    #[serde(rename = "return_documents")]
```
dureuill: I don't get why the rename is needed, can you tell me?

crates/meilisearch/src/personalization/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+    Disabled,
+}
+
+impl PersonalizationService {
+    pub fn cohere(api_key: String) -> Self {
+        Self::Cohere(CohereService::new(api_key))
+    }
+
+    pub fn disabled() -> Self {
+        debug!("Personalization service disabled");
```
dureuill, line 339: this log seems to be duplicated
Capture d’écran 2025-11-05 à 11 55 19

crates/meilisearch/src/personalization/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+        match self {
+            Self::Cohere(cohere_service) => {
+                cohere_service
+                    .rerank_search_results(search_result, personalize, query, deadline)
+                    .await
+            }
+            Self::Disabled => Err(PersonalizationError::FeatureNotEnabled {
+                disabled_action: "reranking search results",
+                feature: "personalization",
+                issue_link: "https://github.com/orgs/meilisearch/discussions/866",
```
dureuill: maybe we should plan on adding Cloud specific instructions to this issue? Cloud customers won't be able to enable the feature by adding an environment variable

crates/meilisearch/src/routes/indexes/search.rs
```diff
@@ -455,12 +486,27 @@ pub async fn search_with_post(
 
     let index = index_scheduler.index(&index_uid)?;
 
+    // Extract personalization and query string before moving query
+    let personalize = query.personalize.clone();
```
dureuill:
```suggestion
    let personalize = query.personalize.take();
```

crates/meilisearch/src/routes/indexes/search.rs
```diff
@@ -455,12 +486,27 @@ pub async fn search_with_post(
...
+        None
+    };
+
     let search_kind =
         search_kind(&query, index_scheduler.get_ref(), index_uid.to_string(), &index)?;
     let retrieve_vectors = RetrieveVectors::new(query.retrieve_vectors);
 
     let include_metadata = parse_include_metadata_header(&req);
 
+    let query_str = query.q.clone();
```
dureuill:
```suggestion
    let query_str = personalize.is_some().then(|| query.q.clone());
```

crates/meilisearch/src/personalization/mod.rs
dureuill: So about this, we have about 3 options:

1. (previous option) log the issues but ignore them. It has the pro than a cohere downtime doesn't become a Meilisearch downtime, and I think long term this is an important property as personalization is not a core feature of the search so if it punctually doesn't work it should cause the prod to fail. The con is that currently logging anything makes it impossible for Cloud users (and difficult for other users) to become aware of the issue, which is an issue like if they mistype or revoke their API key for instance.
2. (current option) return an error and fail the search. It has the pro of surfacing the error, but the con of causing full search downtime whenever there is personalization downtime
3. (preferred option, can be implemented later) return the search results, but add a field to the search response that contains the error message from cohere. This has the pro of surfacing the error to the users, without the con of provider downtime become Meilisearch downtime. The real con is that it requires extending the API though. Remote federated search is implemented in this way (see `remoteErrors` here: https://www.meilisearch.com/docs/reference/api/multi_search#federated-multi-search-requests)

Option (3) can be implemented in a future PR.

crates/meilisearch/src/personalization/mod.rs
ManyTheFish: Yes, but it is due to Actix configuring twice per worker 🤔 :
Capture d’écran 2025-11-05 à 15 00 33

crates/meilisearch/src/personalization/mod.rs
ManyTheFish: same issue as the cohere initialization

crates/meilisearch/src/analytics/segment_analytics.rs
ManyTheFish: You're right, I've converted it to a boolean and added it in the Infos

crates/meilisearch/src/personalization/mod.rs
ManyTheFish: modified the issue by adding a cloud and an OSS sections

crates/meilisearch/src/personalization/mod.rs
ManyTheFish: No, hallucinated while generating the `send_rerank_request` function, my fault, I am sorry

crates/meilisearch/src/personalization/mod.rs
ManyTheFish: created a linear issue on the personalization project https://linear.app/meilisearch/issue/EXP-690/stabilize-the-reranking-personalization

crates/meilisearch/src/personalization/mod.rs
dureuill: init the service outside rather than in the workers then pass that to the app data?

crates/meilisearch/src/routes/indexes/search.rs
dureuill: if you want to avoid an unwrap down the line, you could even have a single option containing both `personalize` and the cloned query

crates/meilisearch/src/personalization/mod.rs
ManyTheFish: yup, so the following code calls `create_app` twice per actix worker:
https://github.com/meilisearch/meilisearch/blob/1a21cc1a17970d792fe9b9e611636beaf94f30ca/crates/meilisearch/src/main.rs#L166

ManyTheFish: Ok, I moved the code, I removed some extra clones I am not sure they was necessary 🤔

crates/meilisearch/src/lib.rs
```diff
@@ -148,12 +151,13 @@ pub fn create_app(
         .configure(|s| {
             configure_data(
                 s,
-                index_scheduler.clone(),
-                auth_controller.clone(),
-                search_queue.clone(),
+                index_scheduler,
+                auth_controller,
+                search_queue,
+                personalization_service,
```
ManyTheFish: @dureuill I removed some extra clones here while moving out the personalization service. Am I wrong?

crates/meilisearch/src/personalization/mod.rs
dureuill: thanks

crates/meilisearch/src/lib.rs
dureuill: indeed, the nice thing is that if it compiles it works (I checked that the clone had no side effect)

dureuill:
need to fix compilation and make clippy happy

crates/meilisearch/src/personalization/mod.rs
dureuill: thank you

crates/meilisearch/src/personalization/mod.rs
dureuill: solved, then

crates/meilisearch/src/routes/indexes/search.rs
```diff
@@ -342,9 +349,25 @@ pub async fn search_with_url_query(
...
+        })
+    } else {
+        None
+    };
+
     let search_kind =
         search_kind(&query, index_scheduler.get_ref(), index_uid.to_string(), &index)?;
     let retrieve_vector = RetrieveVectors::new(query.retrieve_vectors);
+
+    let query_str = query.q.clone();
```
dureuill: same as below?

crates/meilisearch/src/search/mod.rs
```diff
@@ -543,6 +558,9 @@ pub struct SearchQueryWithIndex {
     pub ranking_score_threshold: Option<RankingScoreThreshold>,
     #[deserr(default, error = DeserrJsonError<InvalidSearchLocales>, default)]
     pub locales: Option<Vec<Locale>>,
+    #[deserr(default, error = DeserrJsonError<InvalidSearchPersonalize>, default)]
+    #[serde(skip)]
+    pub personalize: Option<Personalize>,
```
dureuill, line 563: We should make sure that multi-search requests using `personalize` return an error

Currently:
```
curl http://localhost:7700/multi-search  -H 'Content-Type: application/json' -d' {"queries": [{"indexUid": "mieli"}, {"indexUid": "mieli", "personalize": {"userContext": "toto"}}]}'
```

works (without ever calling cohere)

crates/meilisearch/src/personalization/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+    }
+}
+
+pub struct CohereService {
+    client: Client,
+    api_key: String,
+}
+
+impl CohereService {
+    pub fn new(api_key: String) -> Self {
```
dureuill, line 64: could we call `Self::disabled()` when `api_key` is empty or whitespace?

crates/meilisearch/src/routes/indexes/search.rs
ManyTheFish: I renamed the variable and added a comment

crates/meilisearch/src/personalization/mod.rs
ManyTheFish: done in PersonalizationService

crates/meilisearch/src/search/mod.rs
ManyTheFish: I made the personalization work with multi search, and an error is returned on federation

dureuill: nice

dureuill (dismissed):
thank you 💯

PR #5777, opened 2025-07-24
feat: support multiple task webhook URLs

qdequele:
## Summary
- allow specifying multiple webhook URLs via `MEILI_TASK_WEBHOOK_URL`
- propagate the list of URLs to index scheduler
- send task updates to each configured endpoint
- test multi‑webhook behaviour

## Testing
- `cargo test --locked test_basic_webhook -- --nocapture` *(fails: Failed to download a valid file from all sources)*

### DIFF
--- a/crates/index-scheduler/src/lib.rs
+++ b/crates/index-scheduler/src/lib.rs
@@ -98,8 +98,8 @@ pub struct IndexSchedulerOptions {
     pub snapshots_path: PathBuf,
     /// The path to the folder containing the dumps.
     pub dumps_path: PathBuf,
-    /// The URL on which we must send the tasks statuses
-    pub webhook_url: Option<String>,
+    /// The URLs on which we must send the tasks statuses
+    pub webhook_urls: Vec<String>,
     /// The value we will send into the Authorization HTTP header on the webhook URL
     pub webhook_authorization_header: Option<String>,
     /// The maximum size, in bytes, of the task index.
@@ -168,8 +168,8 @@ pub struct IndexScheduler {
     /// Whether we should automatically cleanup the task queue or not.
     pub(crate) cleanup_enabled: bool,
 
-    /// The webhook url we should send tasks to after processing every batches.
-    pub(crate) webhook_url: Option<String>,
+    /// The webhook urls we should send tasks to after processing every batches.
+    pub(crate) webhook_urls: Vec<String>,
     /// The Authorization header to send to the webhook URL.
     pub(crate) webhook_authorization_header: Option<String>,
 
@@ -210,7 +210,7 @@ impl IndexScheduler {
 
             index_mapper: self.index_mapper.clone(),
             cleanup_enabled: self.cleanup_enabled,
-            webhook_url: self.webhook_url.clone(),
+            webhook_urls: self.webhook_urls.clone(),
             webhook_authorization_header: self.webhook_authorization_header.clone(),
             embedders: self.embedders.clone(),
             #[cfg(test)]
@@ -296,7 +296,7 @@ impl IndexScheduler {
             index_mapper,
             env,
             cleanup_enabled: options.cleanup_enabled,
-            webhook_url: options.webhook_url,
+            webhook_urls: options.webhook_urls,
             webhook_authorization_header: options.webhook_authorization_header,
             embedders: Default::default(),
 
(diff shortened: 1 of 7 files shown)


crates/meilisearch/src/option.rs
```diff
@@ -204,8 +204,9 @@ pub struct Opt {
     pub env: String,
 
     /// Called whenever a task finishes so a third party can be notified.
-    #[clap(long, env = MEILI_TASK_WEBHOOK_URL)]
-    pub task_webhook_url: Option<Url>,
+    /// Multiple URLs can be specified separated by a space or by repeating the option.
+    #[clap(long, env = MEILI_TASK_WEBHOOK_URL, value_delimiter = ' ')]
+    pub task_webhook_url: Vec<Url>,
```
irevoire:
```suggestion
    #[clap(long, env = MEILI_TASK_WEBHOOK_URL, value_delimiter = ',')]
    pub task_webhook_url: Vec<Url>,
```

I don't understand how space could work. The space should split the argument into multiple arguments instead right? 🤔 
Did you try it by hand on bash? If yes, I'm very curious to see how clap handles that

crates/index-scheduler/src/lib.rs
```diff
@@ -795,27 +795,29 @@ impl IndexScheduler {
...
-            };
-
-            if let Err(e) = request.send(reader) {
-                tracing::error!("While sending data to the webhook: {e}");
+            for url in &self.webhook_urls {
+                let mut task_reader = TaskReader {
+                    rtxn: &rtxn,
+                    index_scheduler: self,
+                    tasks: updated.iter(),
+                    buffer: Vec::with_capacity(50), // on average a task is around ~100 bytes
```
irevoire, line 803: That was some old outdated code 😱 , tasks are way larger today 
```suggestion
                    buffer: Vec::with_capacity(1024),
```

Or maybe even better if we use a [`page_size`](https://docs.rs/page_size/latest/page_size/) because once we want to send it we'll have to allocate a full page size anyway.
```suggestion
                    buffer: Vec::with_capacity(page_size()),
```

crates/meilisearch/src/option.rs
qdequele: What do you mean, it will just fo a split_by ' '. This should work. @Kerollmops explicitly asked for a space separated instead of coma separated. I do agree than coma separated would be more suited.

qdequele:
Let me review your comments and the ones from Copilot :)

crates/meilisearch/src/option.rs
```diff
@@ -204,8 +204,9 @@ pub struct Opt {
     pub env: String,
 
     /// Called whenever a task finishes so a third party can be notified.
-    #[clap(long, env = MEILI_TASK_WEBHOOK_URL)]
-    pub task_webhook_url: Option<Url>,
+    /// Multiple URLs can be specified separated by a space or by repeating the option.
```
Mubelotix, line 207:
```suggestion
    /// Multiple URLs can be specified separated by a comma or by repeating the option.
```

crates/index-scheduler/src/lib.rs
```diff
@@ -795,27 +795,29 @@ impl IndexScheduler {
...
+                    written: 0,
+                };
+
+                let reader =
+                    GzEncoder::new(BufReader::new(&mut task_reader), Compression::default());
+                let request = ureq::post(url)
+                    .timeout(Duration::from_secs(30))
+                    .set("Content-Encoding", "gzip")
+                    .set("Content-Type", "application/x-ndjson");
+                let request = match &self.webhook_authorization_header {
```
Mubelotix, line 813: Different services are not going to use the same authorization so I feel like it should be possible to have one authorization header for each URL. We need to either turn `webhook_authorization_header` into a list or create a nice little API for webhooks

crates/index-scheduler/src/lib.rs
```diff
@@ -795,27 +795,29 @@ impl IndexScheduler {
...
+            for url in &self.webhook_urls {
+                let mut task_reader = TaskReader {
+                    rtxn: &rtxn,
+                    index_scheduler: self,
+                    tasks: updated.iter(),
+                    buffer: Vec::with_capacity(50), // on average a task is around ~100 bytes
+                    written: 0,
+                };
+
+                let reader =
```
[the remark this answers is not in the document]
Mubelotix: Copilot is right. When there are multiple urls, it would be more efficient to collect all the data into a vec so that it can be reused

Mubelotix:
Hello, we have designed a new API to solve this issue, as it's not trivial to handle different optional authorization headers as cli arguments, and an API will allow us creating an UI in the cloud https://github.com/meilisearch/meilisearch/pull/5785

PR #5778, opened 2025-07-25
Return `queryVector` in the search response when using `retrieveVectors`

### LINKED ISSUE #5779: Return query vector
Related product team resources: [PRD](https://www.notion.so/meilisearch/Query-vectors-23b4b06b651f80c19502c0c4d714c905) (_internal only_)
Related product discussion:

## Motivation

Also return the query vector when retrieveVectors is specified on the search route.

## Usage

https://www.notion.so/meilisearch/Usage-for-query-vectors-23b4b06b651f80acabb0c018b32077ac

@meilisearch/docs-team

Mubelotix:
# Pull Request

## Related issue
Fixes #5779 

## What does this PR do?

When `retrieveVectors` is true, search should also return the vector of the query if it was embedded automatically.


crates/meilisearch/tests/search/multi/mod.rs
```diff
@@ -3911,9 +3911,20 @@ async fn federation_vector_two_indexes() {
...
+        "0": [
+          1.0,
+          0.0,
+          0.5
+        ],
+        "1": [
+          0.8,
+          0.6
+        ]
+      },
```
Mubelotix, line 3924: Shouldn't be here, they are provided in the query 🤔

Mubelotix:
I previously wanted to only return the query vector when it was embedded locally, not when it was provided as-is in the request, but that was too complex. It's much more reliable to always return the query vector (when `retrieveVectors` is set of course)

crates/meilisearch/tests/search/multi/proxy.rs
irevoire, line 1: Huu, I think you were not supposed to push this file?

irevoire (changes requested):
Hey, everything looks good to me, but I think a file was not intended to be pushed?

crates/meilisearch/tests/search/multi/proxy.rs
Mubelotix: It was intentional. While the new test seems to be about sharding, it tests the effects of `queryVectors` on federated results. The cases it includes were not covered by any other test

irevoire: Ah yes nice thanks 👍

irevoire (approved):
Awesome thank you!

PR #5781, opened 2025-07-28
Make updating chat settings override previous value

### LINKED ISSUE #5772: Index chat settings PUT should override values
**Describe the bug**
I think the behavior of the route `PUT /indexes/{index_uid}/settings/chat` is wrong, let's take an example 🤔

**To Reproduce**
Current state with `GET /indexes/{index_uid}/settings/chat`:
```json
{
    "description": "test!",
    "documentTemplate": "{% for field in fields %}{% if field.is_searchable and field.value != nil %}{{ field.name }}: {{ field.value }}\n{% endif %}{% endfor %}",
    "documentTemplateMaxBytes": 400,
    "searchParameters": {
        "limit": 15,
        "sort": [],
        "attributesToSearchOn": []
    }
}
```

Then I do a `PUT /indexes/{index_uid}/settings/chat` with:
```json
{
  "description": "test!",
  "documentTemplate": "{% for field in fields %}{% if field.is_searchable and field.value != nil %}{{ field.name }}: {{ field.value }}\n{% endif %}{% endfor %}",
  "documentTemplateMaxBytes": 400,
  "searchParameters": {
    "limit": 16
  }
}
```

New `GET /indexes/{index_uid}/settings/chat`:
```json
{
    "description": "test!",
    "documentTemplate": "{% for field in fields %}{% if field.is_searchable and field.value != nil %}{{ field.name }}: {{ field.value }}\n{% endif %}{% endfor %}",
    "documentTemplateMaxBytes": 400,
    "searchParameters": {
        "limit": 16,
        "sort": [],
        "attributesToSearchOn": []
    }
}
```

**Expected behavior**
I would expect my last `GET /indexes/{index_uid}/settings/chat` to be:
```json
{
    "description": "test!",
    "documentTemplate": "{% for field in fields %}{% if field.is_searchable and field.value != nil %}{{ field.name }}: {{ field.value }}\n{% endif %}{% endfor %}",
    "documentTemplateMaxBytes": 400,
    "searchParameters": {
        "limit": 16,
    }
}
```

**Meilisearch version:**
v1.15.2

Mubelotix:
# Pull Request

## Related issue
Fixes #5772

## What does this PR do?

Refactor chat settings so that the `PUT`method doesn't behave like a `PATCH` method. It now correctly overrides everything


Mubelotix:
Not sure whether it should land on 1.16 or 1.17

Mubelotix:
`milli::update::chat::ChatSettings` now looks almost exactly like `milli::index::ChatConfig`, and `milli::update::chat::ChatSearchParams` is almost a copy of `milli::update::SearchParameters`. I could merge both but the original types already implement serde without renameCase so I decided to leave them separated. Tell me if merging them would be better

crates/milli/src/update/chat.rs
```diff
@@ -65,62 +57,104 @@ impl From<ChatConfig> for ChatSettings {
...
+        ChatConfig { description, prompt, search_parameters }
+    }
+}
+
 #[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Deserr, ToSchema)]
 #[serde(deny_unknown_fields, rename_all = "camelCase")]
 #[deserr(error = JsonError, deny_unknown_fields, rename_all = camelCase)]
 pub struct ChatSearchParams {
-    #[serde(default, skip_serializing_if = "Setting::is_not_set")]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
Kerollmops, line 124: By changing all those `Setting` to `Option` will the `null` or missing behavior be the expected one?
- `null`: I want to reset to the default value
- missing: I don't want to touch this parameter

Kerollmops (changes requested):

crates/milli/src/update/chat.rs
Mubelotix: Because it's a `PUT` method the value gets replaced completely. There cannot be partial modification: all parameters get changed. We always used `PATCH` for such complex values before, I can make it a `PATCH` if you had like

Kerollmops: Please make it the same as the other nested settings we have. The most important is the consistency here 😊

PR #5783, opened 2025-07-28
Optimize the starts_with filter

Mubelotix:
# Pull Request

Previously, the `STARTS_WITH` filter performed a deep prefix scan on the facet tree, starting from the leaf level (level 0). While this approach works, it’s inefficient in cases where a higher-level node already matches the prefix and contains all relevant document IDs. In such cases, scanning each leaf node individually is unnecessary and costly.

## Related issue

Fixes https://github.com/meilisearch/meilisearch/issues/5801

## What does this PR do?

This PR introduces the use of a recursive algorithm to optimize the `STARTS_WITH` filter by converting it into a range search. Specifically, instead of performing a deep scan, the filter now searches for facets within a prefix-based range. For example, querying facets starting with `baba` will now look for facets in the range `baba..babb`. This doesn't change anything for users.

This is a simple yet effective optimization that significantly reduces scan time in cases where the prefix is common across many entries.

## Benchmarks

The new approach offers substantial speed improvements for short or common prefixes. As expected, the previous implementation remains faster for unique prefixes, but the new algorithm still performs very well in those cases and offers a considerable net gain overall.

 Prefix length | Original | New | Speedup
-- | -- | -- | --
1 letter | 1168.2 | 263.84 | 4.4x (faster)
2 letters | 198.68 | 49.542 | 4.0x (faster)
3 letters | 136.64 | 40.057 | 3.4x (faster)
6 letters | 23.567 | 27.907 | 0.8x (slower)

## Design Considerations

While it’s slightly slower for highly specific prefixes (e.g., 6+ characters), the performance hit is minimal. I considered enabling the new logic only for shorter prefixes (e.g., ≤6 characters), but this could lead to confusing behavior and inconsistent performance. Especially in datasets with uneven value distributions, where adding a single character could drastically change query speed.

Ultimately, I believe the simplicity and overall performance gains of the new approach outweigh the downsides, and it offers a more robust experience for the majority of use cases.


crates/milli/src/search/facet/filter.rs
```diff
@@ -416,7 +417,41 @@ impl<'a> Filter<'a> {
...
+                //
+                // - The second one looks directly at level 0 of the facet group database.
+                //   This pessimistic approach is more efficient when the value is unique.
+                //   It's used as a fallback.
+
                 let value = crate::normalize_facet(word.value());
+
+                let mut value2 = value.as_bytes().to_owned();
+                if let Some(last) = value2.last_mut() {
+                    if *last != 255 {
```
irevoire: Since we're working with utf-8 we know it won't be a 255, [it's one of the forbidden byte](https://en.wikipedia.org/wiki/UTF-8#Error_handling)

crates/milli/src/search/facet/filter.rs
```diff
@@ -416,7 +417,41 @@ impl<'a> Filter<'a> {
...
+                        *last += 1;
+                        if let Ok(value2) = String::from_utf8(value2) {
```
irevoire: Since we increased a random byte, our string won't always be UTF-8 here.

I agree that LMDB works over random bytes, but this can definitely fail and should not force us to change the algorithm to something way slower on a large dataset.

I guess we should update `find_docids_of_facet_within_bounds` so it works with bytes or something 🤔

Mubelotix:
You were right, and it turned out `find_docids_of_facet_within_bounds` already accepted Bytes.

crates/milli/src/search/facet/filter.rs
```diff
@@ -416,20 +416,53 @@ impl<'a> Filter<'a> {
...
+                        return index
+                            .exists_faceted_documents_ids(rtxn, field_id)
+                            .map_err(|e| e.into());
+                    }
+                };
+
+                if *last == 255 {
+                    // The prefix is invalid utf8, so no documents will match anyway
+                    return Ok(RoaringBitmap::new());
+                }
```
irevoire:
```suggestion
                if *last == u8::MAX {
                    // u8::MAX is a forbidden UTF-8 byte, we're guaranteed it cannot be sent through a filter to meilisearch, but just in case, we're going to return something
                    tracing::warn!("Found non utf-8 character in filter. That shouldn't be possible");
                    return Ok(RoaringBitmap::new());
                }
```

If we're going to check an impossible property, let's log when it happens, just so we'll notice if it happens one day

irevoire (approved):

irevoire (approved):

Mubelotix:
@meilisearch/docs-team This PR stabilizes the `STARTS_WITH` filter.

Also, this should probably be stated somewhere on https://github.com/orgs/meilisearch/discussions/763

Mubelotix:
Why is everything so flaky today?

PR #5784, opened 2025-07-29
Sharding and EE license

dureuill:
[Internal tracker reference](https://linear.app/meilisearch/issue/SCA-4/shard-documents-by-splitting-them-before-processing)

This PR introduces the following changes:

- EE license for Meilisearch
- New `network.sharding` to enable the "sharding mode".
    - In sharding mode,  Meilisearch will proxy any document modification requests to all members of the network.
    - In sharding mode, each member of the network will only process the portion designated for them after rendezvous hashing on the remote names and the external document id of the document. On average, in a network with `N` nodes for a payload of `K` documents, each node will process `K/N` documents.
- Each `remote` object now exposes a `writeApiKey` used when proxying tasks. That key should at least have `documents.add` and `documents.delete` permissions and might require more in the future.


LICENSE-EE
```diff
@@ -0,0 +1,67 @@
+Business Source License 1.1 – Adapted for Meili SAS
+This license is based on the Business Source License version 1.1, as published by MariaDB Corporation Ab.
+
+Parameters
+
+Licensor: Meili SAS
+
+Licensed Work: Any file explicitly marked as “Enterprise Edition (EE)” or “governed by the Business Source License”.
```
curquiza: After discussion with @Kerollmops, since we cannot afford to add comments about BUSL license in every added line in "MIT" files, we should be more accurate here and mention the sharding/network feature as EE. What do you think @dureuill?

dureuill:
Rebased on `main`

LICENSE-EE
dureuill: I'll move the EE modules as submodules of `enterprise_edition` modules for clarity.

PR #5785, opened 2025-07-30
Webhook api

### LINKED ISSUE #5792: Webhook API
Related product team resources: [PRD](https://www.notion.so/meilisearch/Webhook-API-23f4b06b651f80518280d41803feab4d) (_internal only_)
Related product discussion:

## Motivation

We need a way to specify multiple webhook URLs, each having its own optional authorization header. Having an API to define webhook would allow creating a nice interface to add, edit or remove webhooks in the cloud.

## Usage

https://www.notion.so/meilisearch/Usage-for-webhook-API-23f4b06b651f8054b6fed3ca02641770

- [x] Update the openAPI file with utoipa:
  - [x] If a new module has been introduced, create a new structure deriving [the OpenAPI proc-macro](https://docs.rs/utoipa/latest/utoipa/derive.OpenApi.html) and nest it in the main [openAPI structure](https://github.com/meilisearch/meilisearch/blob/f2185438eed60fa32d25b15480c5ee064f6fba4a/crates/meilisearch/src/routes/mod.rs#L64-L78).
  - [x] If a new route has been introduced, add the [path decorator](https://docs.rs/utoipa/latest/utoipa/attr.path.html) to it and add the route at the top of the file in its openAPI structure.
  - [x] If a structure which is deserialized or serialized in the API has been introduced or modified, it must derive the [`schema`](https://docs.rs/utoipa/latest/utoipa/macro.schema.html) or the [`IntoParams`](https://docs.rs/utoipa/latest/utoipa/derive.IntoParams.html) proc-macro.
  - [x] Once everything is done, start Meilisearch with the swagger flag: `cargo run --features swagger`, open `http://localhost:7700/scalar` on your browser, and ensure everything works as expected.

## Impacted teams

@meilisearch/docs-team

Mubelotix:
# Pull Request

## `prototype-webhooks-0`

[**Usage**](https://www.notion.so/meilisearch/Usage-for-webhook-API-23f4b06b651f8054b6fed3ca02641770)

## Related issue
Fixes #5792

## What does this PR do?

Adds an API to add, edit or remove new webhooks. Webhooks get sent the tasks once they complete, and could previously only be set using a command line argument

## Db-breaking

_This part was written by @irevoire_

> Detail the change to the DB format and why they are forward compatible

This PR introduces a new database in the index-scheduler.
We're retrieving the `webhooks` key from this database; if it doesn't exist, we just initialize the webhooks as empty so it's 100% forward compatible.


crates/index-scheduler/src/lib.rs
```diff
@@ -754,80 +768,120 @@ impl IndexScheduler {
...
-            };
+        // When there is more than one webhook, cache the data in memory
+        let mut data;
+        let mut reader = match webhooks.webhooks.len() {
+            1 => EitherRead::Other(Some(reader)),
+            _ => {
+                data = Vec::new();
+                reader.read_to_end(&mut data)?;
+                EitherRead::Data(&data)
+            }
```
irevoire: It's not ok to `read_to_end` here as the user sometimes processes 40k tasks together

crates/meilisearch-types/src/keys.rs
```diff
@@ -365,6 +365,12 @@ pub enum Action {
     #[serde(rename = "*.get")]
     #[deserr(rename = "*.get")]
     AllGet,
+    #[serde(rename = "webhooks.get")]
+    #[deserr(rename = "webhooks.get")]
+    WebhooksGet,
+    #[serde(rename = "webhooks.update")]
+    #[deserr(rename = "webhooks.update")]
+    WebhooksUpdate,
```
irevoire, line 373: Inconsistency between the PRD and the implementation https://www.notion.so/meilisearch/Usage-for-webhook-API-23f4b06b651f8054b6fed3ca02641770

Should we update the code or the PRD?

crates/meilisearch/src/routes/webhooks.rs
```diff
@@ -0,0 +1,502 @@
...
+#[serde(rename_all = "camelCase")]
+pub(super) struct WebhookResults {
+    results: Vec<WebhookWithMetadata>,
+}
+
+#[utoipa::path(
+    get,
+    path = "",
+    tag = "Webhooks",
+    security(("Bearer" = ["webhooks.get", "*.get", "*"])),
```
irevoire: If we change the right don't forget to change it here

irevoire (changes requested):

crates/index-scheduler/src/lib.rs
```diff
@@ -752,82 +766,75 @@ impl IndexScheduler {
...
-            let rtxn = self.env.read_txn()?;
+        let rtxn = self.env.read_txn()?;
 
+        for (uuid, Webhook { url, headers }) in webhooks.webhooks.iter() {
             let task_reader = TaskReader {
                 rtxn: &rtxn,
                 index_scheduler: self,
                 tasks: &mut updated.into_iter(),
-                buffer: Vec::with_capacity(50), // on average a task is around ~100 bytes
+                buffer: Vec::with_capacity(800), // on average a task is around ~600 bytes
```
irevoire: Oh, I thought about that, but it doesn't make sense to put anything less than a cache page since that's the minimal amount of bytes we'll have to read.
Instead of 800 we should make a call to [`page_size::get`](https://docs.rs/page_size/latest/page_size/fn.get.html)

crates/meilisearch-types/src/error.rs
```diff
@@ -418,7 +418,16 @@ InvalidChatCompletionSearchDescriptionPrompt   , InvalidRequest       , BAD_REQU
 InvalidChatCompletionSearchQueryParamPrompt    , InvalidRequest       , BAD_REQUEST ;
 InvalidChatCompletionSearchFilterParamPrompt   , InvalidRequest       , BAD_REQUEST ;
 InvalidChatCompletionSearchIndexUidParamPrompt , InvalidRequest       , BAD_REQUEST ;
-InvalidChatCompletionPreQueryPrompt            , InvalidRequest       , BAD_REQUEST
+InvalidChatCompletionPreQueryPrompt            , InvalidRequest       , BAD_REQUEST ;
+// Webhooks
+InvalidWebhooks                                , InvalidRequest       , BAD_REQUEST ;
+InvalidWebhooksUrl                             , InvalidRequest       , BAD_REQUEST ;
+InvalidWebhooksHeaders                         , InvalidRequest       , BAD_REQUEST ;
+ReservedWebhook                                , InvalidRequest       , BAD_REQUEST ;
```
irevoire:
```suggestion
ImmutableWebhook                                , InvalidRequest       , BAD_REQUEST ;
```

After discussing with @gmourier we decided this name would be better suited

crates/meilisearch-types/src/keys.rs
```diff
@@ -416,6 +431,8 @@ impl Action {
             NETWORK_GET => Some(Self::NetworkGet),
             NETWORK_UPDATE => Some(Self::NetworkUpdate),
             ALL_GET => Some(Self::AllGet),
+            WEBHOOKS_GET => Some(Self::WebhooksGet),
+            WEBHOOKS_UPDATE => Some(Self::WebhooksUpdate),
```
irevoire, line 435:
```suggestion
            WEBHOOKS_CREATE => Some(Self::WebhooksCreate),
            WEBHOOKS_GET => Some(Self::WebhooksGet),
            WEBHOOKS_UPDATE => Some(Self::WebhooksUpdate),
            WEBHOOKS_DELETE => Some(Self::WebhooksDelete),
```

crates/meilisearch/src/lib.rs
```diff
@@ -327,6 +328,30 @@ pub fn setup_meilisearch(opt: &Opt) -> anyhow::Result<(Arc<IndexScheduler>, Arc<
...
+        match cli_webhook {
+            Some(webhook) => {
+                webhooks.webhooks.insert(Uuid::nil(), webhook);
+            }
+            None => {
+                webhooks.webhooks.remove(&Uuid::nil());
+            }
+        }
+        index_scheduler.put_webhooks(webhooks)?;
+    }
```
irevoire: Hey sorry that's going to be a very boring request as I believe your code work as-is.

But at meilisearch, we discovered through repeated bugs that the best way to not have unwanted data in our database is to never insert it in the database in the first place.

I would really prefer that we get rid of this and just load the webhooks like we would do. But add a new field called `cli_webhook` or something in the `WebHooks` struct that contains the CLI defined one.
This way we don't have to store anything and reduce our bug surface.

crates/meilisearch/src/routes/webhooks.rs
```diff
@@ -0,0 +1,481 @@
...
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum WebhooksError {
+    #[error("The URL for the webhook `{0}` is missing.")]
+    MissingUrl(Uuid),
+    #[error("Defining too many webhooks would crush the server. Please limit the number of webhooks to 20. You may use a third-party proxy server to dispatch events to more than 20 endpoints.")]
+    TooManyWebhooks,
+    #[error("Too many headers for the webhook `{0}`. Please limit the number of headers to 200.")]
```
irevoire: Since we can edit webhook through a patch we might not know how to remove headers

```suggestion
    TooManyWebhooks,
    #[error("Too many headers for the webhook `{0}`. Please limit the number of headers to 200. Hint: To remove an already defined header set its value to `null`")]
```

crates/meilisearch/src/routes/webhooks.rs
```diff
@@ -0,0 +1,481 @@
...
+    old_webhook: Option<Webhook>,
+    new_webhook: WebhookSettings,
+) -> Result<Webhook, WebhooksError> {
+    let (old_url, mut headers) =
+        old_webhook.map(|w| (Some(w.url), w.headers)).unwrap_or((None, BTreeMap::new()));
+
+    let url = match new_webhook.url {
+        Setting::Set(url) => url,
+        Setting::NotSet => old_url.ok_or_else(|| MissingUrl(uuid.to_owned()))?,
+        Setting::Reset => return Err(MissingUrl(uuid.to_owned())),
```
irevoire, line 248: In this case, a reset means the user actually specified a `null` instead of a string, which should be caught by `deserr` now that I think about it.
Since we only want the `url` to be set or not be set but never be `null` we should use an `Option` instead of a `Setting` I think

crates/meilisearch/src/routes/webhooks.rs
```diff
@@ -0,0 +1,481 @@
...
+    let webhook_settings = webhook_settings.into_inner();
+    debug!(parameters = ?webhook_settings, "Post webhook");
+
+    let uuid = Uuid::new_v4();
+    if webhook_settings.headers.as_ref().set().is_some_and(|h| h.len() > 200) {
+        return Err(TooManyHeaders(uuid).into());
+    }
+
+    let mut webhooks = index_scheduler.webhooks();
+    if dbg!(webhooks.webhooks.len() >= 20) {
```
irevoire:
```suggestion
    if webhooks.webhooks.len() >= 20 {
```

crates/meilisearch/tests/auth/api_keys.rs
```diff
@@ -421,7 +421,7 @@ async fn error_add_api_key_invalid_parameters_actions() {
...
+      "message": "Unknown value `doc.add` at `.actions[0]`: expected one of `*`, `search`, `documents.*`, `documents.add`, `documents.get`, `documents.delete`, `indexes.*`, `indexes.create`, `indexes.get`, `indexes.update`, `indexes.delete`, `indexes.swap`, `tasks.*`, `tasks.cancel`, `tasks.delete`, `tasks.get`, `settings.*`, `settings.get`, `settings.update`, `stats.*`, `stats.get`, `metrics.*`, `metrics.get`, `dumps.*`, `dumps.create`, `snapshots.*`, `snapshots.create`, `version`, `keys.create`, `keys.get`, `keys.update`, `keys.delete`, `experimental.get`, `experimental.update`, `export`, `network.get`, `network.update`, `chatCompletions`, `chats.*`, `chats.get`, `chats.delete`, `chatsSettings.*`, `chatsSettings.get`, `chatsSettings.update`, `*.get`, `webhooks.get`, `webhooks.update`",
```
irevoire: huum we're missing `create` and `delete` here right?

crates/meilisearch/tests/auth/errors.rs
```diff
@@ -93,7 +93,7 @@ async fn create_api_key_bad_actions() {
...
+      "message": "Unknown value `doggo` at `.actions[0]`: expected one of `*`, `search`, `documents.*`, `documents.add`, `documents.get`, `documents.delete`, `indexes.*`, `indexes.create`, `indexes.get`, `indexes.update`, `indexes.delete`, `indexes.swap`, `tasks.*`, `tasks.cancel`, `tasks.delete`, `tasks.get`, `settings.*`, `settings.get`, `settings.update`, `stats.*`, `stats.get`, `metrics.*`, `metrics.get`, `dumps.*`, `dumps.create`, `snapshots.*`, `snapshots.create`, `version`, `keys.create`, `keys.get`, `keys.update`, `keys.delete`, `experimental.get`, `experimental.update`, `export`, `network.get`, `network.update`, `chatCompletions`, `chats.*`, `chats.get`, `chats.delete`, `chatsSettings.*`, `chatsSettings.get`, `chatsSettings.update`, `*.get`, `webhooks.get`, `webhooks.update`",
```
irevoire: same

crates/meilisearch/tests/tasks/webhook.rs
```diff
@@ -127,3 +163,484 @@ async fn test_basic_webhook() {
...
+#[actix_web::test]
+async fn multiple_receive_data() {
+    let server = Server::new().await;
+
+    let WebhookHandle { server_handle: handle1, url: url1, receiver: mut receiver1 } =
+        create_webhook_server().await;
+    let WebhookHandle { server_handle: handle2, url: url2, receiver: mut receiver2 } =
+        create_webhook_server().await;
+    let WebhookHandle { server_handle: handle3, url: url3, receiver: mut receiver3 } =
+        create_webhook_server().await;
```
irevoire: It would be nice to add a fourth one that comes from the CLI, just to be sure because we don't really test it above

crates/index-scheduler/src/lib.rs
```diff
@@ -171,10 +176,11 @@ pub struct IndexScheduler {
...
 
-    /// The webhook url we should send tasks to after processing every batches.
-    pub(crate) webhook_url: Option<String>,
-    /// The Authorization header to send to the webhook URL.
-    pub(crate) webhook_authorization_header: Option<String>,
+    /// A database to store single-keyed data that is persisted across restarts.
+    persisted: Database<Str, Str>,
+
+    /// Webhook
+    cached_webhooks: Arc<RwLock<Webhooks>>,
```
irevoire: Here instead of storing all the webhooks like they were created equal I think we coud store something like that:
```suggestion
    cached_webhooks: Arc<(Option<Webhook>, RwLock<Webhooks>)>,
```

Or make another object that stores and names the content of the tuple.

This way we're sure we'll never write the cli defined webhook to the database.
Another solution could be to just skip the uuid 0 everytime we're writing stuff to the database but that means we cannot use serde to automatically serialize the data for us right?

crates/meilisearch/src/routes/webhooks.rs
Mubelotix: I tried but it makes sending a patch with `{"url":null}` return 200 OK without the url being actually removed. I think it's important to keep returning an error in this case

crates/index-scheduler/src/lib.rs
```diff
@@ -284,10 +304,17 @@ impl IndexScheduler {
...
         let features = features::FeatureData::new(&env, &mut wtxn, options.instance_features)?;
         let queue = Queue::new(&env, &mut wtxn, &options)?;
         let index_mapper = IndexMapper::new(&env, &mut wtxn, &options, budget)?;
-        let chat_settings = env.create_database(&mut wtxn, Some(CHAT_SETTINGS_DB_NAME))?;
+        let chat_settings = env.create_database(&mut wtxn, Some(db_name::CHAT_SETTINGS))?;
+
+        let persisted = env.create_database(&mut wtxn, Some(db_name::PERSISTED))?;
+        let webhooks_db = persisted.remap_data_type::<SerdeJson<Webhooks>>();
+        let mut webhooks = webhooks_db.get(&wtxn, db_keys::WEBHOOKS)?.unwrap_or_default();
+        webhooks.webhooks.remove(&Uuid::nil()); // remove the cli webhook if it was saved by mistake
```
irevoire:
```suggestion
```

We don't have to do that since we never save it ever now

crates/index-scheduler/src/lib.rs
```diff
@@ -862,6 +886,29 @@ impl IndexScheduler {
         self.features.network()
     }
 
+    pub fn put_webhooks(&self, mut webhooks: Webhooks) -> Result<()> {
+        let mut wtxn = self.env.write_txn()?;
+        let webhooks_db = self.persisted.remap_data_type::<SerdeJson<Webhooks>>();
+        webhooks.webhooks.remove(&Uuid::nil()); // the cli webhook must not be saved
```
irevoire:
```suggestion
```

But we should not store it here anymore

crates/meilisearch/src/routes/webhooks.rs
irevoire: Ooooh I thought deserr would catch it my bad

irevoire (approved):
I edited a few things just to be 100% sure we'll never write the cli webhook by mistake in the database.
I basically wrapped the webhooks in a dedicated structure made to be written to the database (and hidden for crates outside of the index-scheduler), one wrapper for the dump and another for the web api.

We're good to merge! Thanks for your work

PR #5790, opened 2025-07-31
Adapt Go CI to recent change in the Go repo

curquiza:
CI launched to prove it works (only go to watch): https://github.com/meilisearch/meilisearch/actions/runs/16650255986

### DIFF
--- a/.github/workflows/sdks-tests.yml
+++ b/.github/workflows/sdks-tests.yml
@@ -114,7 +114,7 @@ jobs:
             dep ensure
           fi
       - name: Run integration tests
-        run: go test -v ./...
+        run: go test --race -v ./integration
 
   meilisearch-java-tests:
     needs: define-docker-image


.github/workflows/sdks-tests.yml
```diff
@@ -114,7 +114,7 @@ jobs:
             dep ensure
           fi
       - name: Run integration tests
-        run: go test -v ./...
+        run: go test --race -v ./integration
```
Kerollmops, line 117: Wow! What is this --race parameter?

Kerollmops (approved):

.github/workflows/sdks-tests.yml
curquiza: Tu check any concurrency issues that could happened in the go code base, as far as I understand (the main maintainer of Go added it)

PR #5805, opened 2025-08-05
Add Index Rename Feature

qdequele:
## What does this PR do?
This PR introduces the ability to rename indexes through the PATCH `/indexes/{indexUid}` endpoint by adding a new `uid` field to the update request body.

## Motivation

Currently, to rename an index, users must:
1. Create a new index with the desired name
2. Swap index

This process is not very convenient and neither super clear. This PR provides a native rename operation that is cleaner and more intuitive.

### API Changes

- Added optional uid field to the PATCH `/indexes/{indexUid}` request body
- When provided, this triggers a IndexUpdate task, the same that where used before to change the `primary_key`
- When looking on the tasks, no the `IndexUpdate` task has a new field `new_index_uid` containing the optional new index for renaming. The old index name could still be found on the `indexUid` of the task.

### Core Changes

- Integrated into existing IndexUpdate task: Rather than creating a new task type, the rename functionality is added as an optional `new_index_uid `field to the existing IndexUpdate task
- Index Mapper: Added rename() method to handle the atomic rename operation
- Stats: Preserves and transfers index statistics to the renamed index

## Key Benefits

- Atomic operation: The rename happens in a single transaction
- Preserves everything: Documents, settings, stats, and search functionality remain intact
- Efficient: No data copying required, just reference updates
- Not breaking: Existing IndexUpdate operations continue to work unchanged however this is not backward compatible 

## Testing
Comprehensive test coverage added in `crates/meilisearch/tests/index/rename_index.rs`

## Example Usage

Rename an index from "movies" to "films":
```sh
  curl -X PATCH 'http://localhost:7700/indexes/movies' \
    --data-binary '{
      "uid": "films"
    }'
```

Rename and update primary key simultaneously
```sh
  curl -X PATCH 'http://localhost:7700/indexes/products' \
    --data-binary '{
      "uid": "inventory",
      "primaryKey": "sku"
    }'
```

## Breaking Changes
- [x] API: No breaking change but new fields available.
- [x] BD: No breaking changes but not backward compatible.

### DIFF
--- a/crates/index-scheduler/src/upgrade/mod.rs
+++ b/crates/index-scheduler/src/upgrade/mod.rs
@@ -39,6 +39,7 @@ pub fn upgrade_index_scheduler(
         (1, 13, _) => 0,
         (1, 14, _) => 0,
         (1, 15, _) => 0,
+        (1, 16, _) => 0,
         (major, minor, patch) => {
             if major > current_major
                 || (major == current_major && minor > current_minor)


irevoire:
Hey, I didn't have the time to review the code yet but just a few comments from reading the description:

> This process is cumbersome, time-consuming, and resource-intensive, especially for large indexes.

This is wrong, currently, to rename an index, what our users do is:
- Create a new index with the name you want
- Swap the old and new index
- Delete the old index

I'm not sure you can be more efficient than that 🤔 

> Task Queue: Updates all task references from the old index name to the new one

You should be careful with this one, I don't know if it's the ai generated description that skipped the details but you must **not** update the tasks that were enqueued **after** your index update.
If I push:
1. add document to A
2. add document to A
3. rename A to B
4. add document to A

I don't want the task `4` to be applied to `B`, it should continue to be applied to the "old" `A`.
But, once the rename happens, I want to find the tasks `1` and `2` by filtering on the task queue with the following filter: `indexes=A`.
From the description, I think that's done.

> None. This is a backward-compatible addition to the existing API.

This is wrong as well, since you're modifying a task stored in the database, it is a DB-breaking.
We should take care that the we can still read the old database from the new meilisearch (should be easy since the uid is optional).
We should also update the dump so that it contains these new fields and can forward them to the new versions.

irevoire:
Also I was thinking about it again and I remember we had the idea and made the API in a way that lets you customize the swap.
Initially, we were planning to add a flag that would delete the other index.
But maybe we could add a flag that allows swapping with a non-existing index, what do you think?

Doing it in the swap gives us the opportunity to do it on multiple indexes at the same time in an atomic way, which is what you seem to want here

qdequele:
> This is wrong, currently, to rename an index, what our users do is:
> 
> * Create a new index with the name you want
> * Swap the old and new index
> * Delete the old index
> 
> I'm not sure you can be more efficient than that 🤔

Ok, it's perhaps exaggerated; it's absolutely not intuitive to do this process to update the name of an index, even more so when we have a PATCH on the `/indexes/{indexUid}` that is made in the REST to update a resource, like the uid of an index at that time.

> > Task Queue: Updates all task references from the old index name to the new one
> 
> You should be careful with this one, I don't know if it's the ai generated description that skipped the details but you must **not** update the tasks that were enqueued **after** your index update. If I push:
> 
> 1. add document to A
> 2. add document to A
> 3. rename A to B
> 4. add document to A

Perhaps on that point, it's a misunderstanding of the behavior in the code. Let me check by adding some tests. So if I understand correctly **we do not change the other tasks data to rename the indexUid**. Got it! 

> > None. This is a backward-compatible addition to the existing API.
> 
> This is wrong as well, since you're modifying a task stored in the database, it is a DB-breaking. We should take care that the we can still read the old database from the new meilisearch (should be easy since the uid is optional). We should also update the dump so that it contains these new fields and can forward them to the new versions.

Not sure to understand, if it's optional it mean that it will not be breaking and will be compatible no?

Ok, breaking change. I mean the API, not the database. But is it breaking on the DB if it's an optional field? If your answer is yes, I can write it down.





> Also I was thinking about it again and I remember we had the idea and made the API in a way that lets you customize the swap. Initially, we were planning to add a flag that would delete the other index. But maybe we could add a flag that allows swapping with a non-existing index, what do you think?
> 
> Doing it in the swap gives us the opportunity to do it on multiple indexes at the same time in an atomic way, which is what you seem to want here

For me it make no logic to use a swap for renaming an index when we can use the existing PATCH on the `/indexes/{indexUid}`.

Doing this would be simpler to discover, to understand, to do.

irevoire:
> Ok, breaking change. I mean the API, not the database.

No, on the API side, we should be good 🤔 

> But is it breaking on the DB if it's an optional field? If your answer is yes, I can write it down.

Yes BUT in a forward compatible way as you said:
- When we enqueue this new kind of task we push a new kind of object that cannot be deserialized anymore by the previous version of meilisearch

Overall, since your field is optional, we should not have to do anything in the dumpless upgrade; we can always read the previous version of update index tasks. But we still keep track of these changes with the "Db change" label

This quick explanation + the label is enough for us to track the breaking so nothing else to do on your side 👍 

> Doing this would be simpler to discover, to understand, to do.

Yeah, about that, and especially about the discoverability I think you're 100% right 💯 
But if we add support for the swap, then internally, when we receive the `UpadeIndex` task, we can convert it to an index swap completely transparently for the end user 🤔 

> Perhaps on that point, it's a misunderstanding of the behavior in the code. Let me check by adding some tests. So if I understand correctly we do not change the other tasks data to rename the indexUid. Got it!

Maybe it was just an issue within the description of the PR, I didn't take a look at the code yet 😬

qdequele:
>

> But if we add support for the swap, then internally, when we receive the `UpadeIndex` task, we can convert it to an index swap completely transparently for the end user 🤔

I was thinking about this at some point, something @Kerollmops told me but we should be able also like before to update the primary_key, and also to change both together, what is not possible in an index_swap. 


> > Ok, breaking change. I mean the API, not the database.
> 
> No, on the API side, we should be good 🤔
> 
> > But is it breaking on the DB if it's an optional field? If your answer is yes, I can write it down.
> 
> Yes BUT in a forward compatible way as you said:
> 
> * When we enqueue this new kind of task we push a new kind of object that cannot be deserialized anymore by the previous version of meilisearch
> 
> Overall, since your field is optional, we should not have to do anything in the dumpless upgrade; we can always read the previous version of update index tasks. But we still keep track of these changes with the "Db change" label

From what I understand, we should highlight any breaking changes in the API and on the Disk, and make a specific comment if it's backward compatible. My bad indeed; in my case, it's not backward compatible. I will update that.


> This quick explanation + the label is enough for us to track the breaking so nothing else to do on your side 👍

❤️❤️


I will review the code again and update the description if necessary regarding the last point about whether the UpdateIndex changes the indexUid in the queue.

qdequele:
> I will review the code again and update the description if necessary regarding the last point about whether the UpdateIndex changes the indexUid in the queue.

Nope it doesn't, I updated the description of the PR to be more accurate.

irevoire:
> I was thinking about this at some point, something @Kerollmops told me but we should be able also like before to update the primary_key, and also to change both together, what is not possible in an index_swap.

That still works, take a look at the document addition, we call the update index function to set the primary key properly, _and then_, **but on the same transaction** we add the documents.
You could call the index swap with the same transaction as well I think and that would also avoid replicating some of the complex code we wrote with the index swap.

Also, sub-question but is there really any use case with updating the primary key?
You can only do it while your index is empty, so it's not like it's something people crave, I think 😅 

> From what I understand, we should highlight any breaking changes in the API and on the Disk, and make a specific comment if it's backward compatible. My bad indeed; in my case, it's not backward compatible. I will update that.

Yes, basically the process is:
- If it's breaking on the API: we won't merge it, we're stable since the v1.0 (that's why I think I didn't understand your comment initially, it's not a question I asked myself for years now 😂)
- If it doesn't impact the database at all => we add the `no db change` label.
- If it's breaking, we check:
  - if it's a forward-compatible change (like you), then it's all cool, and we have nothing to do
  - if it's not, then we have to write a function for the dumpless upgrade that makes sure we can read the previous version of the database and update to the new version of the database

And finally, we cannot merge your PR without one of the two labels; it's just to make sure we don't forget to ask ourselves if the DB is breaking or not even if there is nothing to do in the end.

qdequele:
> That still works, take a look at the document addition, we call the update index function to set the primary key properly, _and then_, **but on the same transaction** we add the documents.
> You could call the index swap with the same transaction as well I think and that would also avoid replicating some of the complex code we wrote with the index swap.

Is it a request to handle the rename this way or is it just for saying?



> Also, sub-question but is there really any use case with updating the primary key?
> You can only do it while your index is empty, so it's not like it's something people crave, I think 😅

To be honest, I think it's not the most used feature, but it is here. So, it should be taken in consideration.

irevoire:
TODO:
- The task queue is not properly updated when renaming an index => should be fixed once I re-use the complex logic of swap
- We need a test to ensure this part works

----
We need to ping doc&integration team for:
## API change

### swap route
The index swap route now lets you specify a `rename` boolean parameter.
When inputted, the second index **must not exist**. If it exists, an error will be thrown.

__Note for the future, we could provide an `override` boolean that, when set to yes, lets you delete the index you're swapping with__

### index update
When updating an index we can specify its `uid` again to change it to something else.
If the uid is not valid an error is thrown.
If the uid is already taken at the time of **processing** the task an error is thrown as well.

New error codes:
- `invalid_swap_rename` when something is wrong with the `rename` field

irevoire:
Huuu, I think I had a bug with github-mobile and it closed the PR while I was closing an issue on a completely unrelated repo, I don't get why 🤔 
Will re-open tomorrow and we'll be able to merge 🎉

PR #5810, opened 2025-08-07
Add category to release draft

curquiza:
- Add categories in changelogs: helps for writing the changelog
- Remove the bump of major when using `breaking change` label because we use it mostly for exp feature. Can be risky to let it if the human doing the release does not notice it


Kerollmops (approved):
Peeeeerfect 👍

PR #5846, opened 2025-08-21
Update Arroy v0.6.2

ManyTheFish:
The new version of arroy contains a search optimization when there is few input candidates compared to the number of documents in the database.


ManyTheFish:
@dureuill I changed the base :D

PR #5861, opened 2025-08-26
Declarative tests

Mubelotix:
Add a new `cargo xtask test` command, able to run declarative tests.

This is especially useful for testing version upgrades.

- [x] Implementation
- [x] [Documentation](https://github.com/meilisearch/meilisearch/blob/upgrade-tests/TESTING.md)
- [x] Run tests on the CI
- [x] Add test for search on movies
- [x] Add test for API keys
- [x] Add test for embeddings

### DIFF
--- a/workloads/tests/hf-embed-hannoy.json
+++ b/workloads/tests/hf-embed-hannoy.json
@@ -1,0 +1,445 @@
+{
+  "type": "test",
+  "name": "hf-embed-hannoy",
+  "binary": {
+    "source": "release",
+    "edition": "community",
+    "version": "1.21.0"
+  },
+  "assets": {
+    "movies-100.json": {
+      "local_location": null,
+      "remote_location": "https://milli-benchmarks.fra1.digitaloceanspaces.com/bench/datasets/movies-100.json",
+      "sha256": "d215e395e4240f12f03b8f1f68901eac82d9e7ded5b462cbf4a6b8efde76c6c6"
+    }
+  },
+  "commands": [
+    {
+      "route": "experimental-features",
+      "method": "PATCH",
+      "body": {
+        "inline": {
+          "vectorStoreSetting": true
+        }
+      },
+      "expectedStatus": 200,
+      "expectedResponse": {
+        "chatCompletions": false,
+        "compositeEmbedders": false,
+        "containsFilter": false,
+        "editDocumentsByFunction": false,
+        "getTaskDocumentsRoute": false,
+        "logsRoute": true,
+        "metrics": false,
+        "multimodal": false,
+        "network": false,
+        "vectorStoreSetting": true
+      },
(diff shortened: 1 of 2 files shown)


crates/xtask/src/test/versions.rs
```diff
@@ -0,0 +1,197 @@
...
+        .with_context(|| format!("asset {asset_name} has no digest"))?;
+
+    let sha256 =
+        digest.strip_prefix("sha256:").map(|s| s.to_string()).context("invalid sha256 format")?;
+
+    Ok(sha256)
+}
+
+pub fn get_arch() -> anyhow::Result<&'static str> {
+    let arch;
```
dureuill: initialize `arch` in case none of the `cfg` match

crates/xtask/Cargo.toml
```diff
@@ -42,3 +42,5 @@ tracing = "0.1.41"
 tracing-subscriber = "0.3.19"
 tracing-trace = { version = "0.1.0", path = "../tracing-trace" }
 uuid = { version = "1.17.0", features = ["v7", "serde"] }
+similar-asserts = "1.7.0"
+chrono = "0.4"
```
dureuill: we're more of a `time` team at meili

crates/xtask/src/test/workload.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+/// A test workload.
+/// Not to be confused with [a bench workload](crate::bench::workload::Workload).
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct TestWorkload {
+    pub name: String,
+    pub initial_version: Version,
```
dureuill: Allow `latest` here. Maybe, make the field optional and assume `latest` if missing 🤔

dureuill:
After discussing with @irevoire: 

1. It would be nice to have the option to test the search while the upgrade is ongoing
2. To avoid making this racy, tamo suggested it would be nice to be able to start the new version of Meilisearch with the `--experimental-max-number-of-batched-tasks=0`, which acts as a read-only mode.
3. To implement this, I think we should generalize the `upgrade` command to a `changeBinary` command that would allow starting a new binary (from a version, the latest, or at a specified path) with specified flags.
4. In line with this I feel like the `initialVersion` should specify a `binary` instead (again, version, latest or specified path)

crates/xtask/Cargo.toml
dureuill: removed

crates/xtask/src/test/versions.rs
dureuill: fixed

dureuill (approved):

dureuill (approved):

PR #5867, opened 2025-08-28
Fix links to update and migration

Rattone:
I don't know if this solves the problem, but I've seen it done in other files in the project. The problem is that the link currently contains a dot, resulting in a 307 redirect to the documentation home page.

image


curquiza:
@Rattone can you remove the dot instead? So that we are sure it works for everyone

Kerollmops:
Hey @Rattone 👋 Sorry, but this PR is stalled. I'd be happy to review a new one starting from an up-to-date main branch. Have a nice day 🌵

PR #5870, opened 2025-08-30
optimizing the search time by running it in parallel for fast search 

itsamine27:
## Related issue

Fixes #...

## Requirements


dureuill:
hello,

thank you for your interest in meilisearch.

However, we will not accept this PR.

1. The search requests are not parallelized on purpose, to allow for concurrent search requests to occur and to avoid having a single client hogging all the CPUs of the machine
2. The change disregards a property of the indexes, which is that not many indexes can be opened simultaneously.

I'm closing this PR. Thank you again.

PR #5896, opened 2025-09-18
Document template: Correctly render when indexing first item in array

dureuill:
In a fragment or document template, Meilisearch would not render the template/fragment when it referenced index 0 of an array.

Content: [internal slack channel](https://meilisearch.slack.com/archives/CD7Q2UKGB/p1757948589965599)


Kerollmops (approved):
Impressive changes!

PR #5905, opened 2025-09-25
Override "default" source base url with the config one to support AWS bedrock OpenAI API

dureuill:
[Internal discussion](https://linear.app/meilisearch/issue/ENGPROD-2114/making-the-chat-route-compatible-with-aws-bedrock-openai-route#comment-39313caf)

## Breaking change

Technically a breaking change, because the source URL from the config is no longer ignored for the OpenAI source

## Draft

Requires updating the async-openai dependency after landing https://github.com/meilisearch/async-openai/pull/1

### DIFF
--- a/crates/meilisearch/src/routes/chats/config.rs
+++ b/crates/meilisearch/src/routes/chats/config.rs
@@ -24,8 +24,9 @@ impl Config {
                 if let Some(api_key) = chat_settings.api_key.as_ref() {
                     config = config.with_api_key(api_key);
                 }
+                // config base url always overrides the one from the source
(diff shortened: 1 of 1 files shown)


dbrain:
@dureuill Anything blocking this now? Prevents users (me, hi!) from wiring in geminis openai endpoint for conversational search (i.e. https://generativelanguage.googleapis.com/v1beta/openai/) if I understand it correctly.

dureuill:
had to rebase to get the CI to run 🤦

dbrain:
Thanks!

PR #5906, opened 2025-09-25
Delete oldest tasks first

dureuill:
I'm not sure if we had a reason to use `rev` here @irevoire?

### DIFF
--- a/crates/index-scheduler/src/queue/mod.rs
+++ b/crates/index-scheduler/src/queue/mod.rs
@@ -310,7 +310,7 @@ impl Queue {
             | self.tasks.status.get(wtxn, &Status::Failed)?.unwrap_or_default()
             | self.tasks.status.get(wtxn, &Status::Canceled)?.unwrap_or_default();
 
-        let to_delete = RoaringBitmap::from_iter(finished.into_iter().rev().take(100_000));
+        let to_delete = RoaringBitmap::from_iter(finished.into_iter().take(100_000));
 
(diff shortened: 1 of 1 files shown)


Kerollmops (dismissed):
Looks good to me, but I need @irevoire's confirmation 👀

crates/index-scheduler/src/queue/mod.rs
```diff
@@ -310,7 +310,7 @@ impl Queue {
             | self.tasks.status.get(wtxn, &Status::Failed)?.unwrap_or_default()
             | self.tasks.status.get(wtxn, &Status::Canceled)?.unwrap_or_default();
 
-        let to_delete = RoaringBitmap::from_iter(finished.into_iter().rev().take(100_000));
+        let to_delete = RoaringBitmap::from_iter(finished.into_iter().take(100_000));
```
irevoire:
```suggestion
        let to_delete = RoaringBitmap::from_sorted_iter(finished.into_iter().take(100_000)).unwrap();
```

irevoire (dismissed):
Yep that was a bug

crates/index-scheduler/src/queue/mod.rs
Kerollmops: As `finished` comes from a `RoaringBitmap` this unwrap looks good to me.

irevoire (approved):
re-approve after running rustfmt

PR #5908, opened 2025-09-25
Update version

irevoire:
We're waiting for https://github.com/meilisearch/meilisearch/pull/5906 and https://github.com/meilisearch/meilisearch/pull/5907 to land before merging

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -589,7 +589,7 @@ source = "git+https://github.com/meilisearch/bbqueue#cbb87cc707b5af415ef203bdaf2
 
 [[package]]
 name = "benchmarks"
-version = "1.22.0"
+version = "1.22.1"
 dependencies = [
  "anyhow",
  "bumpalo",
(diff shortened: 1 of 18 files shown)


crates/meilisearch/tests/documents/geojson/mod.rs
```diff
@@ -418,3 +418,36 @@ async fn geo_radius() {
     }
     "#);
 }
+
+#[actix_rt::test]
```
dureuill, line 422: this should not be part of this PR?

crates/milli/Cargo.toml
```diff
@@ -119,7 +119,7 @@ twox-hash = { version = "2.1.1", default-features = false, features = [
     "xxhash64",
 ] }
 geo-types = "0.7.16"
-zerometry = "0.1.0"
+zerometry = "0.3.0"
```
dureuill, line 122: same

crates/meilisearch/tests/documents/geojson/mod.rs
irevoire: Yes, that's why we're waiting for #5907 to land before merging this one 😁 
Since I updated the Cargo.lock on both PR they'll probably have a conflict, so I based this one on #5907

dureuill (approved):

PR #5919, opened 2025-09-30
Raise remote search timeout to 30s

dureuill:
In a remote federated search, mark remotes as timeout after 30s rather than 5.


dureuill:
> That's nice, next time it could be great to set those timeouts by using an optional env variable, don't you think?

I'm very afraid of the explosion of env variables :(

maybe as a request parameter inside of `federation`? But again I wanted to be conservative here

PR #5926, opened 2025-10-02
Search metadata

ManyTheFish:
# Search Metadata Feature Documentation

## Overview
Added opt-in metadata to search responses for query traceability. Metadata is only included when the `Meili-Include-Metadata` header is present, ensuring backward compatibility.

### New Header
- **Header**: `Meili-Include-Metadata`
- **Value**: Any value (presence indicates opt-in)
- **Default**: Metadata not included (backward compatible)

#### Single Search (`/indexes/{indexUid}/search`)
```json
{
  "hits": [...],
  "metadata": {
    "queryUid": "0199a41a-8186-70b3-b6e1-90e8cb582f35",
    "indexUid": "movies",
    "primaryKey": "id"
  }
}
```

#### Multi-Search (`/multi-search`)
```json
{
  "results": [
    {
      "indexUid": "movies",
      "hits": [...],
      "metadata": {
        "queryUid": "0199a41a-8186-70b3-b6e1-90e8cb582f35",
        "indexUid": "movies", 
        "primaryKey": "id"
      }
    }
  ]
}
```

#### Federated Search (`/multi-search` with federation)
```json
{
  "hits": [...],
  "metadata": [
    {
      "queryUid": "0199a41a-8186-70b3-b6e1-90e8cb582f35",
      "indexUid": "movies",
      "primaryKey": "id"
    },
    {
      "queryUid": "0199a41a-8186-70b3-b6e1-90e8cb582f36", 
      "indexUid": "remote_movies",
      "primaryKey": "id",
      "remote": "remote-1"
    }
  ]
}
```

## Metadata Fields

| Field | Type | Description | When Present |
|-------|------|-------------|--------------|
| `queryUid` | UUID v7 | Unique identifier for the query | Always |
| `indexUid` | String | Index identifier | Always |
| `primaryKey` | String | Primary key field name | None if the index has no primary key |
| `remote` | String | Remote instance name | Remote queries only |

### Basic Search with Metadata

**Request:**
```bash
curl -X POST 'http://localhost:7700/indexes/movies/search' \
  -H 'Authorization: Bearer masterKey' \
  -H 'Meili-Include-Metadata: true' \
  -d '{"q": "batman"}'
```

**Response:**
```json
{
  "hits": [
    {
      "id": 1,
      "title": "Batman Begins",
      "overview": "After training with his mentor, Batman begins his fight to free crime-ridden Gotham City from corruption."
    }
  ],
  "query": "batman",
  "processingTimeMs": 2,
  "limit": 20,
  "offset": 0,
  "estimatedTotalHits": 1,
  "requestUid": "0199a41a-60e2-7b91-9843-9ff0f0e36a3d",
  "metadata": {
    "queryUid": "0199a41a-8186-70b3-b6e1-90e8cb582f35",
    "indexUid": "movies",
    "primaryKey": "id"
  }
}
```

### Multi-Search with Metadata

**Request:**
```bash
curl -X POST 'http://localhost:7700/multi-search' \
  -H 'Authorization: Bearer masterKey' \
  -H 'Meili-Include-Metadata: true' \
  -d '{
    "queries": [
      {"indexUid": "movies", "q": "batman"},
      {"indexUid": "books", "q": "harry"}
    ]
  }'
```

**Response:**
```json
{
  "results": [
    {
      "indexUid": "movies",
      "hits": [
        {
          "id": 1,
          "title": "Batman Begins",
          "overview": "After training with his mentor, Batman begins his fight to free crime-ridden Gotham City from corruption."
        }
      ],
      "query": "batman",
      "processingTimeMs": 0,
      "limit": 20,
      "offset": 0,
      "estimatedTotalHits": 1,
      "requestUid": "0199a41a-cbf3-78e1-a04a-a93c3cca0f06",
      "metadata": {
        "queryUid": "0199a41a-cbf4-73d1-a80a-0aa43823cae9",
        "indexUid": "movies",
        "primaryKey": "id"
      }
    },
    {
      "indexUid": "books",
      "hits": [],
      "query": "harry",
      "processingTimeMs": 0,
      "limit": 20,
      "offset": 0,
      "estimatedTotalHits": 0,
      "requestUid": "0199a41a-cbf3-78e1-a04a-a93c3cca0f06",
      "metadata": {
        "queryUid": "0199a41a-cbf4-73d1-a80a-0ab74a5eae99",
        "indexUid": "books",
        "primaryKey": "id"
      }
    }
  ]
}
```

### Federated Search with Metadata

**Request:**
```bash
curl -X POST 'http://localhost:7700/multi-search' \
  -H 'Authorization: Bearer masterKey' \
  -H 'Meili-Include-Metadata: true' \
  -d '{
    "queries": [
      {"indexUid": "movies", "q": "batman"},
      {"indexUid": "remote_movies", "q": "matrix", "federationOptions": {"remote": "remote-1"}}
    ],
    "federation": {"limit": 5}
  }'
```

**Response:**
```json
{
  "hits": [
    {
      "id": 1,
      "title": "Batman Begins",
      "overview": "After training with his mentor, Batman begins his fight to free crime-ridden Gotham City from corruption.",
      "_federation": {
        "indexUid": "movies",
        "queriesPosition": 0,
        "weightedRankingScore": 0.9848484848484849,
        "remote": null
      }
    },
    {
      "id": 101,
      "title": "The Matrix",
      "overview": "A computer hacker learns from mysterious rebels about the true nature of his reality and his role in the war against its controllers.",
      "_federation": {
        "indexUid": "remote_movies",
        "queriesPosition": 1,
        "weightedRankingScore": 0.9242424242424242,
        "remote": "remote-1"
      }
    }
  ],
  "processingTimeMs": 256,
  "limit": 5,
  "offset": 0,
  "estimatedTotalHits": 2,
  "requestUid": "0199a428-54d6-7a73-91bb-cb2f7e8c2d5a",
  "metadata": [
    {
      "queryUid": "0199a428-56d2-7873-a505-f88093d8d68e",
      "indexUid": "movies",
      "primaryKey": "id"
    },
    {
      "queryUid": "0199a428-56d2-7873-a505-f89683a44df9",
      "indexUid": "remote_movies",
      "primaryKey": "id",
      "remote": "remote-1"
    }
  ],
  "remoteErrors": {}
}
```

## Implementation Notes

- **Backward Compatible**: No metadata included by default
- **Header Forwarding**: Remote requests automatically forward the header
- **Query Ordering**: Federated search metadata array matches query order

### DIFF
--- a/crates/meilisearch/src/routes/multi_search.rs
+++ b/crates/meilisearch/src/routes/multi_search.rs
@@ -9,6 +9,7 @@ use meilisearch_types::keys::actions;
 use serde::Serialize;
 use tracing::debug;
 use utoipa::{OpenApi, ToSchema};
+use uuid::Uuid;
 
 use super::multi_search_analytics::MultiSearchAggregator;
 use crate::analytics::Analytics;
@@ -151,6 +152,7 @@ pub async fn multi_search_with_post(
     // Since we don't want to process half of the search requests and then get a permit refused
     // we're going to get one permit for the whole duration of the multi-search request.
     let permit = search_queue.try_get_search_permit().await?;
+    let request_uid = Uuid::now_v7();
 
     let federated_search = params.into_inner();
 
@@ -188,14 +190,27 @@ pub async fn multi_search_with_post(
 
     let response = match federation {
         Some(federation) => {
+            debug!(
+                request_uid = ?request_uid,
+                federation = ?federation,
+                parameters = ?queries,
+                "Federated-search"
+            );
+
             // check remote header
             let is_proxy = req
                 .headers()
                 .get(PROXY_SEARCH_HEADER)
                 .is_some_and(|value| value.as_bytes() == PROXY_SEARCH_HEADER_VALUE.as_bytes());
-            let search_result =
-                perform_federated_search(&index_scheduler, queries, federation, features, is_proxy)
-                    .await;
+            let search_result = perform_federated_search(
+                &index_scheduler,
+                queries,
+                federation,
+                features,
+                is_proxy,
+                request_uid,
+            )
+            .await;
             permit.drop().await;
 
             if search_result.is_ok() {
@@ -203,6 +218,13 @@ pub async fn multi_search_with_post(
             }
 
             analytics.publish(multi_aggregate, &req);
+
+            debug!(
+                request_uid = ?request_uid,
+                returns = ?search_result,
+                "Federated-search"
+            );
+
             HttpResponse::Ok().json(search_result?)
         }
         None => {
@@ -216,7 +238,12 @@ pub async fn multi_search_with_post(
                     .map(SearchQueryWithIndex::into_index_query_federation)
                     .enumerate()
                 {
-                    debug!(on_index = query_index, parameters = ?query, "Multi-search");
+                    debug!(
+                        request_uid = ?request_uid,
+                        on_index = query_index,
+                        parameters = ?query,
+                        "Multi-search"
+                    );
 
                     if federation_options.is_some() {
                         return Err((
@@ -258,6 +285,7 @@ pub async fn multi_search_with_post(
                             search_kind,
                             retrieve_vector,
                             features,
+                            request_uid,
                         )
                     })
                     .await
@@ -286,7 +314,11 @@ pub async fn multi_search_with_post(
                 err
             })?;
 
-            debug!(returns = ?search_results, "Multi-search");
+            debug!(
+                request_uid = ?request_uid,
+                returns = ?search_results,
+                "Multi-search"
+            );
 
             HttpResponse::Ok().json(SearchResults { results: search_results })
         }
--- a/crates/meilisearch/src/routes/multi_search.rs
+++ b/crates/meilisearch/src/routes/multi_search.rs
@@ -21,7 +21,7 @@ use crate::routes::indexes::search::search_kind;
 use crate::search::{
     add_search_rules, perform_federated_search, perform_search, FederatedSearch,
     FederatedSearchResult, RetrieveVectors, SearchQueryWithIndex, SearchResultWithIndex,
-    PROXY_SEARCH_HEADER, PROXY_SEARCH_HEADER_VALUE,
+    PROXY_SEARCH_HEADER, PROXY_SEARCH_HEADER_VALUE, INCLUDE_METADATA_HEADER,
 };
 use crate::search_queue::SearchQueue;
 
@@ -202,6 +202,10 @@ pub async fn multi_search_with_post(
                 .headers()
                 .get(PROXY_SEARCH_HEADER)
                 .is_some_and(|value| value.as_bytes() == PROXY_SEARCH_HEADER_VALUE.as_bytes());
+            let include_metadata = req
+                .headers()
+                .get(INCLUDE_METADATA_HEADER)
+                .is_some();
             let search_result = perform_federated_search(
                 &index_scheduler,
                 queries,
@@ -209,6 +213,7 @@ pub async fn multi_search_with_post(
                 features,
                 is_proxy,
                 request_uid,
+                include_metadata,
             )
             .await;
             permit.drop().await;
@@ -228,6 +233,11 @@ pub async fn multi_search_with_post(
             HttpResponse::Ok().json(search_result?)
         }
         None => {
+            let include_metadata = req
+                .headers()
+                .get(INCLUDE_METADATA_HEADER)
+                .is_some();
+
             // Explicitly expect a `(ResponseError, usize)` for the error type rather than `ResponseError` only,
             // so that `?` doesn't work if it doesn't use `with_index`, ensuring that it is not forgotten in case of code
             // changes.
@@ -286,6 +296,7 @@ pub async fn multi_search_with_post(
                             retrieve_vector,
                             features,
                             request_uid,
+                            include_metadata,
                         )
                     })
                     .await
--- a/crates/meilisearch/src/search/federated/perform.rs
+++ b/crates/meilisearch/src/search/federated/perform.rs
@@ -17,6 +17,7 @@ use meilisearch_types::milli::vector::Embedding;
 use meilisearch_types::milli::{self, DocumentId, OrderBy, TimeBudget, DEFAULT_VALUES_PER_FACET};
 use roaring::RoaringBitmap;
 use tokio::task::JoinHandle;
+use uuid::Uuid;
 
 use super::super::ranking_rules::{self, RankingRules};
 use super::super::{
@@ -39,6 +40,7 @@ pub async fn perform_federated_search(
     federation: Federation,
     features: RoFeatures,
     is_proxy: bool,
+    request_uid: Uuid,
 ) -> Result<FederatedSearchResult, ResponseError> {
     if is_proxy {
         features.check_network("Performing a remote federated search")?;
@@ -170,6 +172,7 @@ pub async fn perform_federated_search(
         facet_stats,
         facets_by_index,
         remote_errors: partitioned_queries.has_remote.then_some(remote_errors),
+        request_uid: Some(request_uid),
     })
 }
 
@@ -439,6 +442,7 @@ fn merge_metadata(
         degraded: degraded_for_host,
         used_negative_operator: host_used_negative_operator,
         remote_errors: _,
+        request_uid: _,
     } in remote_results
     {
         let this_remote_duration = Duration::from_millis(*processing_time_ms as u64);
--- a/crates/meilisearch/src/search/federated/perform.rs
+++ b/crates/meilisearch/src/search/federated/perform.rs
@@ -20,6 +20,7 @@ use tokio::task::JoinHandle;
 use uuid::Uuid;
 
 use super::super::ranking_rules::{self, RankingRules};
+use super::super::SearchMetadata;
 use super::super::{
     compute_facet_distribution_stats, prepare_search, AttributesFormat, ComputedFacets, HitMaker,
     HitsInfo, RetrieveVectors, SearchHit, SearchKind, SearchQuery, SearchQueryWithIndex,
@@ -59,7 +60,10 @@ pub async fn perform_federated_search(
 
     // 1. partition queries by host and index
     let mut partitioned_queries = PartitionedQueries::new();
+    let mut query_metadata = Vec::new();
     for (query_index, federated_query) in queries.into_iter().enumerate() {
+        let query_uid = Uuid::now_v7();
+        query_metadata.push(SearchMetadata { query_uid });
         partitioned_queries.partition(federated_query, query_index, &network, features)?
     }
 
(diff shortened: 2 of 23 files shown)


crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -112,6 +122,56 @@ pub async fn perform_federated_search(
...
+            let remote = federation_options.and_then(|options| options.remote);
+
+            // Get primary key for this index
+            let mut primary_key = None;
+
+            if remote.is_some() {
+                // For remote queries, try to get primary key from remote results
+                if let Some(remote_meta) = remote_results_by_index.get(&index_uid) {
+                    primary_key = remote_meta.primary_key.clone();
+                }
```
ManyTheFish: The current implementation suffers from a caveat, when several remotes have the same index UID but different `primary_key`, then the same primary key would be associated to all remotes.

crates/meilisearch/src/search/federated/proxy.rs
```diff
@@ -143,6 +155,8 @@ async fn try_proxy_search(
         request
     };
     let request = request.header(PROXY_SEARCH_HEADER, PROXY_SEARCH_HEADER_VALUE);
+    let request =
+        if include_metadata { request.header(INCLUDE_METADATA_HEADER, "true") } else { request };
```
dureuill, line 163: parsing of the header is not the same as in the single-index search, we could probably factor it.

crates/meilisearch/src/search/federated/proxy.rs
```diff
@@ -11,6 +11,7 @@ use crate::search::SearchQueryWithIndex;
 
 pub const PROXY_SEARCH_HEADER: &str = "Meili-Proxy-Search";
 pub const PROXY_SEARCH_HEADER_VALUE: &str = "true";
+pub const INCLUDE_METADATA_HEADER: &str = "Meili-Include-Metadata";
```
dureuill: nitpick: code-organization-wise, it is a bit unexpected to have this header in `federated::proxy` for both the single search and the (federated or not) multi search.

I would expect the header to be defined at the root of the search module, along with a parsing routine, and then imported by all its users

crates/meilisearch/tests/common/index.rs
```diff
@@ -516,6 +516,33 @@ impl<State> Index<'_, State> {
...
+    ) -> (Value, StatusCode) {
+        let url = format!("/indexes/{}/search", urlencode(self.uid.as_ref()));
+        let body = serde_json::to_string(&query).unwrap();
+        let mut all_headers = vec![("content-type", "application/json")];
+        all_headers.extend(headers);
+        self.service.post_str(url, body, all_headers).await
+    }
+
+    pub async fn multi_search_post(&self, queries: Value) -> (Value, StatusCode) {
+        self.service.post("/multi-search", queries).await
```
dureuill: FYI this already exists, but is for some reason declared on the test index type: https://github.com/meilisearch/meilisearch/blob/a0ef8008d6eb6549ed3f2a7675686e69befbdeef/crates/meilisearch/tests/common/index.rs#L532

crates/meilisearch/src/search/mod.rs
```diff
@@ -1150,7 +1181,20 @@ pub fn perform_search(
...
+    ) = search_from_kind(index_uid.clone(), search_kind, search)?;
+
+    let metadata = if include_metadata {
+        let query_uid = Uuid::now_v7();
+        let primary_key = index.primary_key(&rtxn)?.map(|pk| pk.to_string());
+        Some(SearchMetadata {
+            query_uid,
+            index_uid: index_uid_for_metadata,
+            primary_key,
+            remote: None, // Local searches don't have a remote
```
dureuill, line 1193: that's not true in the hits: if at least one search is remote, then the remote is computed even for local search queries, with the goal of making it possible to easily compare hits irrespective of the host the multi-search request was sent to.

crates/meilisearch/src/routes/multi_search.rs
```diff
@@ -202,13 +202,20 @@ pub async fn multi_search_with_post(
                 .headers()
                 .get(PROXY_SEARCH_HEADER)
                 .is_some_and(|value| value.as_bytes() == PROXY_SEARCH_HEADER_VALUE.as_bytes());
+            let include_metadata = req
+                .headers()
+                .get(INCLUDE_METADATA_HEADER)
+                .and_then(|h| h.to_str().ok())
+                .map(|v| matches!(v.to_lowercase().as_str(), "true" | "1"))
+                .unwrap_or(false);
```
dureuill: this, and line 238-244 can be computed once before the match as it doesn't depend on whether we're federating or not.

crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -59,14 +61,22 @@ pub async fn perform_federated_search(
 
     // 1. partition queries by host and index
     let mut partitioned_queries = PartitionedQueries::new();
+
+    // Store the original queries order for later metadata building
+    let original_queries = queries.clone();
```
dureuill: I'd rather we avoid this clone if possible (queries can be large), and work to populate the metadata from the results of the individual search queries.

crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -59,14 +61,22 @@ pub async fn perform_federated_search(
...
     for (query_index, federated_query) in queries.into_iter().enumerate() {
         partitioned_queries.partition(federated_query, query_index, &network, features)?
     }
 
     // 2. perform queries, merge and make hits index by index
     // 2.1. start remote queries
-    let remote_search =
-        RemoteSearch::start(partitioned_queries.remote_queries_by_host, &federation, deadline);
+    let remote_search = RemoteSearch::start(
+        partitioned_queries.remote_queries_by_host.clone(),
```
dureuill: same here, we'd like to avoid a clone here

crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -108,10 +118,68 @@ pub async fn perform_federated_search(
...
+                        let key = (remote_name, index_uid);
+                        remote_primary_keys.insert(key, primary_key);
+                    }
+                }
+            }
+        }
+
+        // Build metadata in the same order as the original queries
+        for original_query in original_queries {
+            let query_uid = Uuid::now_v7();
```
dureuill: should we reuse the remote query uid for remote queries?

crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -108,10 +118,68 @@ pub async fn perform_federated_search(
...
+            let primary_key = match &remote {
+                Some(remote_name) => {
+                    // For remote queries, try to get primary key from remote results
+                    // Use composite key (remote, index_uid) to avoid collisions
+                    let lookup_key = (remote_name, &index_uid);
+                    remote_primary_keys.get(&lookup_key).map(|pk| pk.to_string())
+                }
+                None => {
+                    // For local queries, get primary key from local index
+                    index_scheduler.index(&index_uid).ok().and_then(|index| {
```
dureuill: this implementation of federated search tries very hard to avoid opening indexes more than once.
Normally indexes are opened in `SearchByIndex::execute`

crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -590,7 +665,10 @@ impl RemoteSearch {
                     // never merge distant facets
                     proxy_federation.merge_facets = None;
                     let params = params.clone();
-                    async move { proxy_search(&node, queries, proxy_federation, &params).await }
+                    async move {
+                        proxy_search(&node, queries, proxy_federation, &params, include_metadata)
```
dureuill, line 698: beware that the query index is overridden in the case of a remote search, so care should be taken to interpret the array of metadata in response to the remote search

crates/meilisearch/src/search/mod.rs
dureuill: it shouldn't be done in this module though 😓

crates/meilisearch/src/search/federated/perform.rs
ManyTheFish: I found it too complicated; it would imply retrieving which metadata corresponds to which query ID.

The only solution I found was to collect all the query IDs in one vector per remote and reassign them to the metadata by popping each vector.
But the result was really ugly and not straightforward :S

crates/meilisearch/src/search/federated/proxy.rs
ManyTheFish: we are not parsing the header here

crates/meilisearch/tests/common/index.rs
ManyTheFish: I don't understand

dureuill: Sorry was mistaken for some reason; the issue is that this PR add this function to `index` and `server`,  but it only really makes sense in `server`

crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -108,11 +123,63 @@ pub async fn perform_federated_search(
...
+        for local_meta in &results_by_index {
+            if let SearchResultByIndex { index, primary_key: Some(primary_key), .. } = &local_meta {
+                let key = (None, index);
+                primary_key_per_index.insert(key, primary_key);
+            }
+        }
+
+        // if there are remote results, set the local remote name
+        let local_remote_name =
+            (!remote_results.is_empty()).then_some(network.local.clone()).flatten();
```
dureuill: there are 2 subtle bugs here:

1. the remote results can be empty if there were remote names but they were all equal to the local name. In that case, the general behavior is to do the same as if there was a remote, to be independent of the remote that the search was sent to.
2. the remote results can also be empty if there were actually remote names, but they all failed to answer for some reason.

I suggest looking (or even reusing) `PartitionedQueries::has_remote`

crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -108,11 +123,63 @@ pub async fn perform_federated_search(
     let after_waiting_remote_results = std::time::Instant::now();
 
     // 3. merge hits and metadata across indexes and hosts
-    // 3.1. merge metadata
+
+    // 3.1. Build metadata in the same order as the original queries
+    let query_metadata = if include_metadata {
```
dureuill: consider moving the code to a `build_query_metadata` function?

crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -59,14 +61,27 @@ pub async fn perform_federated_search(
...
     let mut partitioned_queries = PartitionedQueries::new();
+
+    // Preconstruct metadata keeping the original queries order for later metadata building
+    let precomputed_query_metadata: Vec<_> = queries
+        .iter()
+        .map(|q| {
+            (q.index_uid.to_string(), q.federation_options.as_ref().and_then(|o| o.remote.clone()))
+        })
+        .collect();
+
```
dureuill: A nitpick, but could we skip populating this vector when `!include_metadata`?

The way we could do this:
1. Store an option of that vector depending on `include_metadata` (maybe we should call it `index_remote_queries` btw)
2. Instead of testing `include_metadata` again on step (3.1), `if let` on the vector to "carry" the boolean.

PR #5932, opened 2025-10-06
Make remote federated search timeout configurable with an env variable

dureuill:
Use `MEILI_EXPERIMENTAL_REMOTE_SEARCH_TIMEOUT_SECONDS` to override the timeout of the remote federated search.


Kerollmops (approved):
LGTM 🌵

PR #5933, opened 2025-10-06
Fix ranking score bug when sort is present

ManyTheFish:
### Problem
When using `sort` criteria in search queries, the `_rankingScore` was always set to `1.0` instead of reflecting the actual search relevance score.

### Root Cause
The `global_score` function in `score_details.rs` was not properly handling the case where both search relevance scores and sort details were present. It was taking the first score value (which was always 1.0 for word ranking rule score when only one word was present) instead of computing the actual search relevance score.

### Solution
- **Prioritize semantic scores**: When vector/embedding scores are available, use them as the ranking score
- **Fall back to ranking scores**: For traditional text searches, properly calculate and merge ranking scores using `Rank::global_score()`
- **Exclude sort details**: Filter out sort and geo sort details that don't contribute to relevance scoring

### Changes
- Modified `global_score` function to properly handle semantic vs ranking scores
- Added test case to verify the fix

### Before
```json
{
  "hits": [
    {"title": "Tumbler Cup", "_rankingScore": 1.0},
    {"title": "Stainless Steel Tumbler", "_rankingScore": 1.0}
  ]
}
```

### After
```json
{
  "hits": [
    {"title": "Tumbler Cup", "_rankingScore": 0.9848484848484848},
    {"title": "Stainless Steel Tumbler", "_rankingScore": 0.8787878787878788}
  ]
}
```

Fixes the issue where `_rankingScore` was always 1.0 when sort criteria were present in search queries.

### DIFF
--- a/crates/milli/src/score_details.rs
+++ b/crates/milli/src/score_details.rs
@@ -66,6 +66,12 @@ impl ScoreDetails {
         }
     }
 
+    /// Calculate the global score of the details.
+    ///
+    /// It is computed from the ranks of the ranking rules, excluding the sort/geo sort rules.
+    /// If the details contain a semantic score (ScoreDetails::Vector), it is used instead of the ranking score.
+    ///
+    /// note: this function expects a maximum of one semantic score, otherwise only the last one will be used.
     pub fn global_score<'a>(details: impl Iterator<Item = &'a Self> + 'a) -> f64 {
         // Filter out only the ranking scores (Rank values) and exclude sort/geo sort
         let mut semantic_score = None;


crates/milli/src/score_details.rs
```diff
@@ -67,14 +67,22 @@ impl ScoreDetails {
...
-                };
-                Some(score)
-            })
-            .unwrap_or(1.0f64)
+        // Filter out only the ranking scores (Rank values) and exclude sort/geo sort
+        let mut semantic_score = None;
+        let ranking_ranks = details.filter_map(|detail| match detail.rank_or_value() {
+            RankOrValue::Rank(rank) => Some(rank),
+            RankOrValue::Score(score) => {
+                semantic_score = Some(score);
```
dureuill, line 81: note: this works because we can guarantee that there is only 0 or 1 `RankOrValue::Score` in `details`.

Maybe worth putting as a precondition to this function?

crates/milli/src/score_details.rs
```diff
@@ -67,14 +67,22 @@ impl ScoreDetails {
...
-        Self::score_values(details)
-            .find_map(|x| {
-                let ScoreValue::Score(score) = x else {
-                    return None;
-                };
-                Some(score)
-            })
-            .unwrap_or(1.0f64)
+        // Filter out only the ranking scores (Rank values) and exclude sort/geo sort
+        let mut semantic_score = None;
```
dureuill, line 77: this is very smart

PR #5945, opened 2025-10-15
Search cutoff vector store

dureuill:
Improves the interaction of vector store and the `searchCutoffMs` when using the `vectorStore: experimental`.

1. The vector store will try to honor the deadline set by the cutoff.
2. The cutoff will add results that are ready to the list without discarding them.
3. During federated search, the cutoff is shared between all search queries for an index rather than reset on each query

# Changelog

- feature_name: search cutoff
- feature_description: Set a duration after which a search request should finish as fast as possible.
- change_description: Improves the interaction of vector store and the `searchCutoffMs` when using the `vectorStore: experimental`.
- feature_documentation: https://www.meilisearch.com/docs/reference/api/settings#search-cutoff

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2758,9 +2758,9 @@ dependencies = [
 
 [[package]]
 name = "hannoy"
-version = "0.0.9-nested-rtxns"
+version = "0.0.9-nested-rtxns-2"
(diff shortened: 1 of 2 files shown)


Kerollmops (approved):
This PR is very cool, let's merge it!

PR #5946, opened 2025-10-15
Improve compaction behaviors

Kerollmops:
This PR ensures the engine:
- Reuses the same file when compacting an index
- Deletes the potentially large and incomplete compacted file from disk when the compaction is canceled
- Prioritizes the compaction tasks so that it is easier to predict when they'll be processed

### DIFF
--- a/crates/index-scheduler/src/scheduler/process_batch.rs
+++ b/crates/index-scheduler/src/scheduler/process_batch.rs
@@ -998,7 +998,7 @@ impl IndexScheduler {
 
             let index_path = match self.index_mapper.index_mapping.get(rtxn, &index_uid)? {
                 Some(index_uuid) => self.index_mapper.index_path(index_uuid),
-                None => return Err(Error::IndexNotFound(index_uid)),
+                None => continue,
             };
 
             if let Err(e) = remove_file(index_path.join(DATA_MDB_COPY_NAME)) {


crates/index-scheduler/src/scheduler/process_batch.rs
```diff
@@ -979,7 +983,33 @@ impl IndexScheduler {
...
+                pre_compaction_size: _,
+                post_compaction_size: _,
+            }) = task.details
+            else {
+                unreachable!("wrong details for compaction task {compaction_task}")
+            };
+
+            let index_path = match self.index_mapper.index_mapping.get(rtxn, &index_uid)? {
+                Some(index_uuid) => self.index_mapper.index_path(index_uuid),
+                None => return Err(Error::IndexNotFound(index_uid)),
```
dureuill: should that be an error? Like, if we're cancelling a task that is for an index that no longer exists, I think no cleanup should be necessary, but no error either

crates/index-scheduler/src/scheduler/process_batch.rs
```diff
@@ -979,7 +983,33 @@ impl IndexScheduler {
...
+            };
+
+            if let Err(e) = remove_file(index_path.join(DATA_MDB_COPY_NAME)) {
+                match e.kind() {
+                    ErrorKind::NotFound => (),
+                    _ => return Err(Error::IoError(e)),
```
dureuill, line 1007: nitpick: should we return an error here, or maybe just log it?

crates/index-scheduler/src/scheduler/create_batch.rs
```diff
@@ -525,7 +525,20 @@ impl IndexScheduler {
...
 
-        // 3. we batch the export.
+        // 3. we get the next task to compact
+        let to_compact = self.queue.tasks.get_kind(rtxn, Kind::IndexCompaction)? & enqueued;
+        if let Some(task_id) = to_compact.min() {
+            let mut task =
+                self.queue.tasks.get_task(rtxn, task_id)?.ok_or(Error::CorruptedTaskQueue)?;
+            current_batch.processing(Some(&mut task));
+            current_batch
+                .reason(BatchStopReason::TaskKindCannotBeBatched { kind: Kind::IndexCompaction });
```
dureuill: nitpick: for the stop reason you want `TaskCannotBeBatched` so that the batch stop reason message reads:

> created batch containing only task with id {id} of type `{kind}` that cannot be batched with any other task.

rather than:

> stopped after the last task of type `{kind}` because they cannot be batched with tasks of any other type.

(the former is happening rather than the later)

dureuill:
I'm wondering, should we remove the `AutobatchKind::IndexCompaction` since compaction tasks are now batched first?

Normally there is no AutobatchKind for prioritary tasks, but this is because so far they didn't act on one index.

https://github.com/meilisearch/meilisearch/blob/fix-compaction-issues/crates/index-scheduler/src/scheduler/autobatcher.rs#L28

what do you think?

crates/index-scheduler/src/scheduler/process_batch.rs
Kerollmops: Ok, let's go for a `continue` 👍

crates/index-scheduler/src/scheduler/process_batch.rs
Kerollmops: I prefer to keep the error.

PR #5947, opened 2025-10-16
Support full or sharded replication

### LINKED ISSUE #3494: About replicating Meilisearch
This morning I had a meeting with @dureuill, and we discussed the different solutions to replicate Meilisearch we could implement and the pros and cons of each.

First, we must define what we want to achieve. Meilisearch is very fast to boot. It doesn't need to process anything before being able to serve requests, not even if a crash occurs. The reason is that it uses LMDB and not RocksDB or SQLite, which are [WAL-based embedded databases](https://en.wikipedia.org/wiki/Write-ahead_logging). The high-availability feature of the Cloud highly depends on this feature, but it doesn't apply in situations where an entire cluster is down.

These different solutions can fix the weaknesses of the current design. They are more or less complex to implement.

So that you know, no single solution described below straightforwardly manages task cancellation, and therefore task cancellation will be disabled for the prototypes.

## A Single Writer Broadcasts its Task Queue to multiple Readers

This solution is the easiest to implement by far. The principle is that we broadcast the tasks received by the single writer to the other readers. There are different quirks to think about, but the current Meilisearch codebase rests unchanged.

A Meilisearch server would be allowed to receive user write requests. Every time it successfully writes it to its task queue, it broadcasts it to all the previously registered Meilisearch Readers. The Meilisearch Readers also store this task, but only if the highest task id follows the previous one. If not, it asks for the set of missing tasks to the Meilisearch Writer. The Meilisearch Writer will generate a dump or send the list of tasks depending on whether the task content files are available (we delete the task content files when a task is finished).

Pros:
 - Seems easy enough to create a prototype.
 - Elegant solution where we keep the engine as is.
 - Highly available on the read side.

Cons:
 - Highly available on the read side only.

## The Task Queue is Replicated using a Raft or Paxos Consensus Protocol

This solution looks to be the ideal one. Multiple Meilisearch servers can synchronize the task queue together. Sending the tasks to each other and committing changes at the same time. Unfortunately, it is not the easiest to build as consensus protocols are [hard to implement right](https://jepsen.io/).

Here is a list of available Rust replication libraries: [tikv/raft-rs](https://github.com/tikv/raft-rs), [drmingdrmer/openraft](https://github.com/drmingdrmer/openraft), [benschulz/paxakos](https://github.com/benschulz/paxakos). And here is a list of some C/C++ libraries we can probably wrap: [baidu/braft](https://github.com/baidu/braft), [canonical/raft](https://github.com/canonical/raft), [willemt/raft](https://github.com/willemt/raft), [logcabin/logcabin](https://github.com/logcabin/logcabin).

Pros:
 - Highly available on reads and writes.
 - The replication library manages the replication.

Cons:
 - [It doesn't seem easy to create a prototype](https://github.com/Kerollmops/canonical-raft/issues/4).
 - It can be hard to debug network and consistency issues.

## The Task Queue is a Message Broker

A [message broker](https://en.wikipedia.org/wiki/Message_broker) is a, most of the time, distributed event store in which we can send events that every listener will read.

In this solution, all Meilisearch servers were listening to the same broker and subscribed simultaneously. When a written request is sent to any Meilisearch server, this server will store it in the broker, and all the cluster members will receive it and start processing it.

When a Meilisearch server is lacking, it must start from a dump or process the tasks in the order from the start, which can take a lot of time. I didn't take the time to think more about that, but there surely be a communication between two Meilisearch servers to ask for a dump at a moment which can complexify the solution.

Pros:
 - The broker manages replication and high availability.
 - Highly available on reads and writes.

Cons:
 - Depending on the broker, the user must set up and monitor another program.
 - We must maintain another task queue that lies in the broker queue.

darthhexx:
## Related issue

Fixes #3494

## Description

Using the network sharding feature results in the data being distributed over the nodes using the [must_process](https://github.com/meilisearch/meilisearch/blob/c45172a4bf2eb514856c597f213fa0a19904632f/crates/milli/src/update/new/indexer/enterprise_edition/sharding.rs#L14) check and this results in performance gains. However, it comes at the cost of multipoint failure resilience, i.e. if either node fails, the data becomes inconsistent.

In setups where it is preferable to have full replicas, we should allow for full replication configurations. This PR modifies the check for when to replicate data to only condition on when there are remotes configured. Although it maintains the sharding logic, allowing for both fully or sharded replication configurations.

## Requirements

### DIFF
--- a/crates/meilisearch/src/routes/indexes/documents.rs
+++ b/crates/meilisearch/src/routes/indexes/documents.rs
@@ -362,7 +362,7 @@ pub async fn delete_document(
         tokio::task::spawn_blocking(move || index_scheduler.register(task, uid, dry_run)).await??
     };
 
-    if network.remotes.len() > 0 && !dry_run {
+    if !network.remotes.is_empty() && !dry_run {
         proxy(&index_scheduler, &index_uid, &req, network, Body::none(), &task).await?;
     }
 
@@ -1075,7 +1075,7 @@ async fn document_addition(
         }
     };
 
-    if network.remotes.len() > 0 {
+    if !network.remotes.is_empty() {
         if let Some(file) = file {
             proxy(
                 &index_scheduler,
@@ -1193,7 +1193,7 @@ pub async fn delete_documents_batch(
         tokio::task::spawn_blocking(move || index_scheduler.register(task, uid, dry_run)).await??
     };
 
-    if network.remotes.len() > 0 && !dry_run {
+    if !network.remotes.is_empty() && !dry_run {
         proxy(&index_scheduler, &index_uid, &req, network, Body::Inline(body), &task).await?;
     }
 
@@ -1285,7 +1285,7 @@ pub async fn delete_documents_by_filter(
         tokio::task::spawn_blocking(move || index_scheduler.register(task, uid, dry_run)).await??
     };
 
-    if network.remotes.len() > 0 && !dry_run {
+    if !network.remotes.is_empty() && !dry_run {
         proxy(&index_scheduler, &index_uid, &req, network, Body::Inline(filter), &task).await?;
     }
 
@@ -1435,7 +1435,7 @@ pub async fn edit_documents_by_function(
         tokio::task::spawn_blocking(move || index_scheduler.register(task, uid, dry_run)).await??
     };
 
-    if network.remotes.len() > 0 && !dry_run {
+    if !network.remotes.is_empty() && !dry_run {
         proxy(&index_scheduler, &index_uid, &req, network, Body::Inline(params), &task).await?;
     }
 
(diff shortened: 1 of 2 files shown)


dureuill:
Hello,

This PR introduces full replication, that is, all nodes have the same content.

It makes it so that when `sharding = false`, then the tasks are replicated (as soon as there is at least one node), but `must_process` always answers `true`.

It is an economical implementation, however:
- we are working on shard-based replication which will allow sharded replication including full replication
- the PR touches the behavior of users with `sharding: false`, there's no way to disable both sharding and replication.

The introduction of sharded replication will change a lot of that code, so it is not practical for us to accept this PR at the time.

Thank you for your contribution, and for your patience while we add proper replication.

PR #5948, opened 2025-10-16
Upload snapshot tarballs to S3

Kerollmops:
This PR implements a new way to upload snapshots directly to S3. It has many advantages like being about to stream the whole process and make good use of the multipart technology to send chunks of data in parallel.

The is a working [prototype available (or building) on Docker](https://github.com/meilisearch/meilisearch/actions/runs/19134675619) named: `prototype-v1.24.0.s3-snapshots-3`. To try this new S3 method with this prototype you'll have to specify some env variables, if you don't the default behavior of snapshoting in a folder will be used.

- `MEILI_S3_BUCKET_URL`: The URL in the format `https://s3.<region>.amazonaws.com`.
- `MEILI_S3_BUCKET_REGION`: The region in the format us-east-1.
- `MEILI_S3_BUCKET_NAME`: The bucket name.
- `MEILI_S3_SNAPSHOT_PREFIX`: The prefix path where to put the snapshot, uses normal slashes.
- `MEILI_S3_ACCESS_KEY`: The S3 access key.
- `MEILI_S3_SECRET_KEY`: The S3 secret key.

Additional but optional environment variables are available to configure further the way the snapshots are made. You can read the documentation of each of those parameters [on the dedicated page](https://github.com/orgs/meilisearch/discussions/869).
- `MEILI_EXPERIMENTAL_S3_MAX_IN_FLIGHT_PARTS`: The number of in-flight multipart requests to send to S3.
- `MEILI_EXPERIMENTAL_S3_COMPRESSION_LEVEL`: The compression level to use for the snapshot tarball. Defaults to 0: no compression.
- `MEILI_EXPERIMENTAL_S3_SIGNATURE_DURATION_SECONDS`: The signature duration for the multipart upload. Defaults to 8 hours.
- `MEILI_EXPERIMENTAL_S3_MULTIPART_PART_SIZE`: The size of the the multipart parts. Defaults to 375 MiB, enabling databases of up to 3.5 TiB.

## To do
- [x] Clean up the code.
- [x] Correctly manage errors.
- [x] Keep the name of the database.
- [x] Talk to the documentation.
- [x] Talk to the Cloud team.
- [x] Support cancellation.
- [x] Crash at startup if S3 Snapshots are used on Windows.
- [x] Add exponential backoff when sending multiparts.
- [x] Should I sign parts for 24 hours or less?
- [x] Expose the sign duration and the multipart part sizes.
- [x] Make the options CLI arguments, too.

### DIFF
--- a/crates/meilisearch/src/option.rs
+++ b/crates/meilisearch/src/option.rs
@@ -83,10 +83,11 @@ const MEILI_S3_BUCKET_NAME: &str = "MEILI_S3_BUCKET_NAME";
 const MEILI_S3_SNAPSHOT_PREFIX: &str = "MEILI_S3_SNAPSHOT_PREFIX";
 const MEILI_S3_ACCESS_KEY: &str = "MEILI_S3_ACCESS_KEY";
 const MEILI_S3_SECRET_KEY: &str = "MEILI_S3_SECRET_KEY";
-const MEILI_S3_MAX_IN_FLIGHT_PARTS: &str = "MEILI_S3_MAX_IN_FLIGHT_PARTS";
-const MEILI_S3_COMPRESSION_LEVEL: &str = "MEILI_S3_COMPRESSION_LEVEL";
-const MEILI_S3_SIGNATURE_DURATION_SECONDS: &str = "MEILI_S3_SIGNATURE_DURATION_SECONDS";
-const MEILI_S3_MULTIPART_PART_SIZE: &str = "MEILI_S3_MULTIPART_PART_SIZE";
+const MEILI_EXPERIMENTAL_S3_MAX_IN_FLIGHT_PARTS: &str = "MEILI_EXPERIMENTAL_S3_MAX_IN_FLIGHT_PARTS";
+const MEILI_EXPERIMENTAL_S3_COMPRESSION_LEVEL: &str = "MEILI_EXPERIMENTAL_S3_COMPRESSION_LEVEL";
+const MEILI_EXPERIMENTAL_S3_SIGNATURE_DURATION_SECONDS: &str =
+    "MEILI_EXPERIMENTAL_S3_SIGNATURE_DURATION_SECONDS";
+const MEILI_EXPERIMENTAL_S3_MULTIPART_PART_SIZE: &str = "MEILI_EXPERIMENTAL_S3_MULTIPART_PART_SIZE";
 
 const DEFAULT_CONFIG_FILE_PATH: &str = "./config.toml";
 const DEFAULT_DB_PATH: &str = "./data.ms";
@@ -954,28 +955,36 @@ pub struct S3SnapshotOpts {
     pub s3_secret_key: String,
 
     /// The maximum number of parts that can be uploaded in parallel.
-    #[clap(long, env = MEILI_S3_MAX_IN_FLIGHT_PARTS, default_value_t = default_s3_snapshot_max_in_flight_parts())]
-    #[serde(default = "default_s3_snapshot_max_in_flight_parts")]
-    pub s3_max_in_flight_parts: NonZeroUsize,
+    ///
+    /// For more information, see <https://github.com/orgs/meilisearch/discussions/869>.
+    #[clap(long, env = MEILI_EXPERIMENTAL_S3_MAX_IN_FLIGHT_PARTS, default_value_t = default_experimental_s3_snapshot_max_in_flight_parts())]
+    #[serde(default = "default_experimental_s3_snapshot_max_in_flight_parts")]
+    pub experimental_s3_max_in_flight_parts: NonZeroUsize,
 
     /// The compression level. Defaults to no compression (0).
-    #[clap(long, env = MEILI_S3_COMPRESSION_LEVEL, default_value_t = default_s3_snapshot_compression_level())]
-    #[serde(default = "default_s3_snapshot_compression_level")]
-    pub s3_compression_level: u32,
+    ///
+    /// For more information, see <https://github.com/orgs/meilisearch/discussions/869>.
+    #[clap(long, env = MEILI_EXPERIMENTAL_S3_COMPRESSION_LEVEL, default_value_t = default_experimental_s3_snapshot_compression_level())]
+    #[serde(default = "default_experimental_s3_snapshot_compression_level")]
+    pub experimental_s3_compression_level: u32,
 
     /// The signature duration for the multipart upload.
-    #[clap(long, env = MEILI_S3_SIGNATURE_DURATION_SECONDS, default_value_t = default_s3_snapshot_signature_duration_seconds())]
-    #[serde(default = "default_s3_snapshot_signature_duration_seconds")]
-    pub s3_signature_duration_seconds: u64,
+    ///
+    /// For more information, see <https://github.com/orgs/meilisearch/discussions/869>.
+    #[clap(long, env = MEILI_EXPERIMENTAL_S3_SIGNATURE_DURATION_SECONDS, default_value_t = default_experimental_s3_snapshot_signature_duration_seconds())]
+    #[serde(default = "default_experimental_s3_snapshot_signature_duration_seconds")]
+    pub experimental_s3_signature_duration_seconds: u64,
 
     /// The size of the the multipart parts.
     ///
     /// Must not be less than 10MiB and larger than 8GiB. Yes,
     /// twice the boundaries of the AWS S3 multipart upload
     /// because we use it a bit differently internally.
-    #[clap(long, env = MEILI_S3_MULTIPART_PART_SIZE, default_value_t = default_s3_snapshot_multipart_part_size())]
-    #[serde(default = "default_s3_snapshot_multipart_part_size")]
-    pub s3_multipart_part_size: Byte,
+    ///
+    /// For more information, see <https://github.com/orgs/meilisearch/discussions/869>.
+    #[clap(long, env = MEILI_EXPERIMENTAL_S3_MULTIPART_PART_SIZE, default_value_t = default_experimental_s3_snapshot_multipart_part_size())]
+    #[serde(default = "default_experimental_s3_snapshot_multipart_part_size")]
+    pub experimental_s3_multipart_part_size: Byte,
 }
 
 impl S3SnapshotOpts {
@@ -988,10 +997,10 @@ impl S3SnapshotOpts {
             s3_snapshot_prefix,
             s3_access_key,
             s3_secret_key,
-            s3_max_in_flight_parts,
-            s3_compression_level,
-            s3_signature_duration_seconds,
-            s3_multipart_part_size,
+            experimental_s3_max_in_flight_parts,
+            experimental_s3_compression_level,
+            experimental_s3_signature_duration_seconds,
+            experimental_s3_multipart_part_size,
         } = self;
 
         export_to_env_if_not_present(MEILI_S3_BUCKET_URL, s3_bucket_url);
@@ -1001,17 +1010,20 @@ impl S3SnapshotOpts {
         export_to_env_if_not_present(MEILI_S3_ACCESS_KEY, s3_access_key);
         export_to_env_if_not_present(MEILI_S3_SECRET_KEY, s3_secret_key);
         export_to_env_if_not_present(
-            MEILI_S3_MAX_IN_FLIGHT_PARTS,
-            s3_max_in_flight_parts.to_string(),
-        );
-        export_to_env_if_not_present(MEILI_S3_COMPRESSION_LEVEL, s3_compression_level.to_string());
-        export_to_env_if_not_present(
-            MEILI_S3_SIGNATURE_DURATION_SECONDS,
-            s3_signature_duration_seconds.to_string(),
+            MEILI_EXPERIMENTAL_S3_MAX_IN_FLIGHT_PARTS,
+            experimental_s3_max_in_flight_parts.to_string(),
         );
         export_to_env_if_not_present(
-            MEILI_S3_MULTIPART_PART_SIZE,
-            s3_multipart_part_size.to_string(),
+            MEILI_EXPERIMENTAL_S3_COMPRESSION_LEVEL,
+            experimental_s3_compression_level.to_string(),
+        );
+        export_to_env_if_not_present(
+            MEILI_EXPERIMENTAL_S3_SIGNATURE_DURATION_SECONDS,
+            experimental_s3_signature_duration_seconds.to_string(),
+        );
+        export_to_env_if_not_present(
+            MEILI_EXPERIMENTAL_S3_MULTIPART_PART_SIZE,
+            experimental_s3_multipart_part_size.to_string(),
         );
     }
 }
@@ -1027,10 +1039,10 @@ impl TryFrom<S3SnapshotOpts> for S3SnapshotOptions {
             s3_snapshot_prefix,
             s3_access_key,
             s3_secret_key,
-            s3_max_in_flight_parts,
-            s3_compression_level,
-            s3_signature_duration_seconds,
-            s3_multipart_part_size,
+            experimental_s3_max_in_flight_parts,
+            experimental_s3_compression_level,
+            experimental_s3_signature_duration_seconds,
+            experimental_s3_multipart_part_size,
         } = other;
 
         Ok(S3SnapshotOptions {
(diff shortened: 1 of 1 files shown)


dureuill:
Hello kero,

1. should we mark the variables as experimental? `MEILI_EXPERIMENTAL_blabla`? Just in case we decide to no longer support this snapshot option in the future? Or do we want it to be insta-stable?
2. Windows CI is failing

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,340 @@ impl IndexScheduler {
 
         Ok(tasks)
     }
+
+    #[cfg(unix)]
+    #[allow(clippy::too_many_arguments)]
```
dureuill:
```suggestion
```

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -4,6 +4,8 @@ use std::sync::atomic::Ordering;
 
 use meilisearch_types::heed::CompactionOption;
 use meilisearch_types::milli::progress::{Progress, VariableNameStep};
+#[cfg(unix)]
+use meilisearch_types::milli::update::S3SnapshotOptions;
```
dureuill: possible to add inside of the function like the other uses?

Kerollmops: Nope, impossible because it's in the arguments of the function. I did my best here 😅

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,340 @@ impl IndexScheduler {
...
+
+            let s3_snapshot_prefix = PathBuf::from_slash(s3_snapshot_prefix);
+            let url = s3_bucket_url
+                .parse()
+                .map_err(BucketError::ParseError)
+                .map_err(Error::S3BucketError)?;
+            let bucket = Bucket::new(url, UrlStyle::Path, s3_bucket_name, s3_bucket_region)
+                .map_err(Error::S3BucketError)?;
+            let credential = Credentials::new(s3_access_key, s3_secret_key);
+            // Note for the future: use with_added_extension, it's prettier
```
dureuill:
```suggestion
            // Note for the future (rust 1.91+): use with_added_extension, it's prettier
```

👍

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,340 @@ impl IndexScheduler {
...
+            let object_path = s3_snapshot_prefix.join(format!("{db_name}.snapshot"));
+            let object = object_path.to_slash().expect("Invalid UTF-8 path").into_owned();
```
dureuill: could we perhaps remove `to_slash` since we don't support windows?

Kerollmops: Hum... Actually that's a good idea 🤔 And the associated crate, too!

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,340 @@ impl IndexScheduler {
...
+                        Ok(response) => response,
+                        Err(_) => {
+                            return Err(Error::S3Error {
+                                status: resp.status(),
+                                body: resp.text().await.unwrap_or_default(),
+                            });
+                        }
+                    };
+                    let etag =
+                        resp.headers().get(ETAG).expect("every UploadPart request returns an Etag");
```
dureuill: Since this depends of S3 rather than our code, I feel like this should be an Error rather than a panic

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,340 @@ impl IndexScheduler {
...
+                    &object,
+                    part_number,
+                    multipart.upload_id(),
+                );
+                let url = part_upload.sign(s3_signature_duration);
+
+                // Wait for a buffer to be ready if there are in-flight parts that landed
+                let mut buffer = if in_flight.len() >= s3_max_in_flight_parts.get() {
+                    let (request, buffer) = in_flight.pop_front().unwrap();
+                    let resp = request.await.unwrap().map_err(Error::S3HttpError)?;
```
dureuill: can you comment on the `unwrap`? When does it happen?

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,340 @@ impl IndexScheduler {
...
+                            body: resp.text().await.unwrap_or_default(),
+                        })
+                    }
+                };
+                let etag =
+                    resp.headers().get(ETAG).expect("every UploadPart request returns an Etag");
+                etags.push(bump.alloc_str(etag.to_str().unwrap()));
+            }
+
+            tracing::trace!("Finalizing the multipart upload");
```
dureuill: the log above was debug, should this be debug as well, or should the log above be trace as well?

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,339 @@ impl IndexScheduler {
...
+        let must_stop_processing = self.scheduler.must_stop_processing.clone();
+        let retry_backoff = backoff::ExponentialBackoff::default();
+        let db_name = {
+            let mut base_path = self.env.path().to_owned();
+            base_path.pop();
+            base_path.file_name().and_then(OsStr::to_str).unwrap_or("data.ms").to_string()
+        };
+
+        let (reader, writer) = std::io::pipe()?;
+        let uploader_task = tokio::spawn(async move {
```
dureuill: do you think you could move that future to a function?

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,339 @@ impl IndexScheduler {
...
+
+            if status.is_success() {
+                Ok(())
+            } else {
+                Err(Error::S3Error { status, body })
+            }
+        });
+
+        let index_scheduler = IndexScheduler::private_clone(self);
+        let builder_task = tokio::task::spawn_blocking(move || {
```
dureuill: same, moving the closure to a function would help with readability

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,339 @@ impl IndexScheduler {
...
+                VERSION_FILE_NAME,
+            )?;
+
+            // 2. Snapshot the index scheduler LMDB env
+            progress.update_progress(SnapshotCreationProgress::SnapshotTheIndexScheduler);
+            let mut tasks_env_file = index_scheduler.env.try_clone_inner_file()?;
+            // Note: Seeking to the start of the file is necessary to ensure
+            //       the tarball reads the file from the beginning. I still
+            //       don't know why the read cursor is not at the beginning.
+            tasks_env_file.seek(SeekFrom::Start(0))?;
```
dureuill: this is concerning 🤔

do we know for sure that parallel reads of the DB won't result in seeks?

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,339 @@ impl IndexScheduler {
...
+
+            // 2.4 Create the update files directory
+            //     And only copy the update files of the enqueued tasks
+            progress.update_progress(SnapshotCreationProgress::SnapshotTheUpdateFiles);
+            let enqueued = index_scheduler.queue.tasks.get_status(&rtxn, Status::Enqueued)?;
+            let (atomic, update_file_progress) = AtomicUpdateFileStep::new(enqueued.len() as u32);
+            progress.update_progress(update_file_progress);
+            // TODO I need to create the update files directory (even if empty)
+            //      I should probably simply use the append_dir_all method
+            //      but I'll loose the progression.
```
dureuill: is this todo in this PR?

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,339 @@ impl IndexScheduler {
...
+            // It's prettier to use a for loop instead of the IndexMapper::try_for_each_index
+            // method, especially when we need to access the UUID, local path and index number.
+            for (i, (name, uuid)) in indexes_references.into_iter().enumerate() {
+                progress.update_progress(VariableNameStep::<SnapshotCreationProgress>::new(
+                    &name, i as u32, nb_indexes,
+                ));
+                let path = indexes_dir.join(uuid.to_string()).join("data.mdb");
+                let index = index_scheduler.index_mapper.index(&rtxn, &name)?;
+                let mut index_file = index.try_clone_inner_file().unwrap();
+                index_file.seek(SeekFrom::Start(0))?;
```
dureuill: same question, does this support e.g., parallel reads to the index file?

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,339 @@ impl IndexScheduler {
...
+
+            drop(rtxn);
+
+            // 4. Snapshot the auth LMDB env
+            progress.update_progress(SnapshotCreationProgress::SnapshotTheApiKeys);
+            let mut auth_env_file =
+                index_scheduler.scheduler.auth_env.try_clone_inner_file().unwrap();
+            auth_env_file.seek(SeekFrom::Start(0))?;
+            let path = Path::new("auth").join("data.mdb");
+            tarball.append_file(path, &mut auth_env_file)?;
```
dureuill: maybe factor this into a `append_env(&mut tarball, base_path, env)` that gets the file from the env, seeks it, creates the path by appending the data.mdb, and append to the tarball

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,339 @@ impl IndexScheduler {
...
+                                status: resp.status(),
+                                body: resp.text().await.unwrap_or_default(),
+                            });
+                        }
+                    };
+                    let etag =
+                        resp.headers().get(ETAG).expect("every UploadPart request returns an Etag");
+                    let mut buffer = match buffer.try_into_mut() {
+                        Ok(buffer) => buffer,
+                        Err(_) => unreachable!("Impossible to convert into BytesMut"),
```
dureuill:
```suggestion
                        Err(_) => unreachable!("All bytes references were consumed in the task"),
```

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -206,4 +230,339 @@ impl IndexScheduler {
...
+                    Err(_) => {
+                        return Err(Error::S3Error {
+                            status: resp.status(),
+                            body: resp.text().await.unwrap_or_default(),
+                        })
+                    }
+                };
+                let etag =
+                    resp.headers().get(ETAG).expect("every UploadPart request returns an Etag");
+                etags.push(bump.alloc_str(etag.to_str().unwrap()));
```
dureuill: could we factor part of the logic here with the case above where there are too many in flight requests and so we are processing a request

dureuill (commented):
I'm not sold on adding a lot of permanent (non-experimental) CLI flags that are very 
technical

- MEILI_S3_MAX_IN_FLIGHT_PARTS: The impact of this parameter is hard to measure. I imagine it is a RAM/concurrency tradeoff, but the tradeoff is hard to articulate for users.
- MEILI_S3_COMPRESSION_LEVEL: would be better as a general snapshot parameter (S3 or not) when we get to it.
- MEILI_S3_SIGNATURE_DURATION_SECONDS: This does not really communicate well that this defines the maximum duration a snapshot can take to be processed. Why not choosing the largest available value? Why should we choose a lower value when cancellation works?
- MEILI_S3_MULTIPART_PART_SIZE: The user can choose a value that makes snapshotting of their DB impossible, and the user must check that the chosen size will translate to a good limit for their D. Ideally, rather than exposing this technical parameter, Meili would set it to a "reasonable value", checking itself if the chosen value is big enough to snapshot the DB.

I'm not opposed to exposing the parameters for this iteration, but I think they should at least be marked as experimental, and possibly removed in a future version

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
Kerollmops: Yes, LMDB doesn't rely on the seeking cursor. Readers use the memory-mapped bytes to access data, which is why we can have multiple search requests/readers operating simultaneously.

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
Kerollmops: I was wondering the same. It seems feasible. I'll try, thanks for pointing me to this.

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
Kerollmops: Yes, readers are not using the read(2) method but rather the memory-mapped bytes and will not be impacted by the seeking file cursor.  The [only reference to the pread(2) function](https://github.com/LMDB/lmdb/blob/e13f05211f8c0869fe1475cd632ba59d4f957656/libraries/liblmdb/mdb.c#L4244) is before the env is mapped into memory, so before any read txn could be instantiated. Plus it is [safe to use pread from different threads simultaneously](https://stackoverflow.com/a/1687302/1941280):
> The atomicity of pread enables processes or threads that share file descriptors to read from a shared file at a particular offset without using a locking mechanism that would be necessary to achieve the same result in separate lseek and read system calls. Atomicity is required as the file pointer is shared and one thread might move the pointer using lseek after another process completes an lseek but prior to the read.

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
Kerollmops: I can, it would be much cleaner 👌 Thanks.

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
Kerollmops: I use debug levels everywhere, now! Thanks.

crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
Kerollmops: Unfortunately, I had to do two functions to factorize the code because of the `&bumpalo::Bump` type not being `Send` and therefore impossible to move between two await points.

Kerollmops:
> I'm not sold on adding a lot of permanent (non-experimental) CLI flags that are very technical [..]

Ok, I'll rename them into `MEILI_EXPERIMENTAL_S3_MAX_IN_FLIGHT_PARTS`, `MEILI_EXPERIMENTAL_S3_COMPRESSION_LEVEL`, `MEILI_EXPERIMENTAL_S3_SIGNATURE_DURATION_SECONDS`, and `MEILI_EXPERIMENTAL_S3_MULTIPART_PART_SIZE`.

crates/meilisearch/src/option.rs
```diff
@@ -83,6 +98,10 @@ const DEFAULT_SNAPSHOT_DIR: &str = "snapshots/";
 const DEFAULT_SNAPSHOT_INTERVAL_SEC: u64 = 86400;
 const DEFAULT_SNAPSHOT_INTERVAL_SEC_STR: &str = "86400";
 const DEFAULT_DUMP_DIR: &str = "dumps/";
+const DEFAULT_S3_SNAPSHOT_MAX_IN_FLIGHT_PARTS: NonZeroUsize = NonZeroUsize::new(10).unwrap();
+const DEFAULT_S3_SNAPSHOT_COMPRESSION_LEVEL: u32 = 0;
+const DEFAULT_S3_SNAPSHOT_SIGNATURE_DURATION_SECONDS: u64 = 8 * 3600; // 8 hours
```
dureuill, line 103: might be insufficient for larger customers unfortunately 😬

Kerollmops: Yup, I agree. That's why I added the possibility to change them through an env variable so we can iterate.

PR #5951, opened 2025-10-20
Ignore embedding fails

dureuill:
`prototype-v1.24.0.ignore-embedding-failures-2` contains the following changes:

1. Ignore document template rendering errors
2. Do not fail a batch when the embedder fails
3. Add `MEILI_EXPERIMENTAL_REST_EMBEDDER_TIMEOUT_SECONDS` to configure the duration of the REST client timeout.
4. Add `MEILI_EXPERIMENTAL_REST_EMBEDDER_MAX_RETRY_DURATION_SECONDS` to configure the max wait time of a REST client between two attempts.

# Draft status

Created to push prototypes `prototype-v1.24.0.ignore-embedding-failures-x`

### DIFF
--- a/crates/milli/src/vector/embedder/rest.rs
+++ b/crates/milli/src/vector/embedder/rest.rs
@@ -91,6 +91,7 @@ struct EmbedderData {
     request: RequestData,
     response: Response,
     configuration_source: ConfigurationSource,
+    max_retry_duration: std::time::Duration,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


crates/milli/src/vector/session.rs
```diff
@@ -108,60 +108,66 @@ impl<'doc, C: OnEmbed<'doc>, I: Input> EmbedSession<'doc, C, I> {
...
+                        }
+                        Err(err) => {
+                            tracing::warn!(
+                                docid = metadata.external_docid,
+                                %err,
+                                "error embedding document"
+                            );
+                        }
+                    }
+                }
```
Kerollmops, line 166: At the end, when you received errors, you decide to send single-document rendered templates to the embedder, as a last resort, right?

Kerollmops (approved):

crates/milli/src/vector/session.rs
dureuill: Correct, this serves multiple purposes:

1. If the failure is caused by the rendered text of one document, only this document will be in error, and not the unrelated documents in the same batch
2. If the failure is caused by the total payload being too large, sending document by document will make it easier on the remote REST embedder

Note that these attempts happen *after* the exponential backoff and the multiple retries.

dureuill:
Superseded by https://github.com/meilisearch/meilisearch/pull/5984

PR #5955, opened 2025-10-24
Update heed and avoid memory leaks

Kerollmops:
This PR is related to #5307 and mostly fixes a memory leak in the LMDB patch. The LMDB patch has also been improved to ensure correct integration in upstream LMDB 🤞


Kerollmops:
Duplicate of #5967...

PR #5959, opened 2025-10-27
Parallelize the word prefix docids

Kerollmops:
This PR improves the word prefix docids by utilizing the forked version of LMDB to read uncommitted entries in parallel.

### DIFF
--- a/crates/milli/src/update/new/words_prefix_docids.rs
+++ b/crates/milli/src/update/new/words_prefix_docids.rs
@@ -4,7 +4,7 @@ use std::io::{BufReader, BufWriter, Read, Seek, Write};
 use std::iter;
 
 use hashbrown::HashMap;
-use heed::types::{Bytes, DecodeIgnore};
+use heed::types::{Bytes, DecodeIgnore, Str};
 use heed::{BytesDecode, Database, Error, RoTxn, RwTxn};
 use rayon::iter::{IndexedParallelIterator as _, IntoParallelIterator, ParallelIterator as _};
 use roaring::MultiOps;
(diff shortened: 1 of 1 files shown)


crates/milli/src/update/new/words_prefix_docids.rs
```diff
@@ -42,7 +49,77 @@ impl WordPrefixDocids {
...
+                        .prefix_iter(&rtxn, prefix.as_bytes())?
+                        .remap_types::<Str, CboRoaringBitmapCodec>()
+                        .map(|result| result.map(|(_word, bitmap)| bitmap))
+                        .union()?;
+
+                    buffer.clear();
+                    CboRoaringBitmapCodec::serialize_into_vec(&output, &mut buffer);
+                    indexes.push(PrefixEntry { prefix, serialized_length: buffer.len() });
+                    file.write_all(&buffer)?;
+                }
```
ManyTheFish, line 98: It works for me.

However, prefixes have several sizes and can share the same ranges of words; for instance, the range of `hel` is a subset of the range of 'h'.
I don't know if it would improve the performance if we tried to iterate once on each word `O(words)` instead of iterating on each word for each prefix `O(words*3)`, 3 represents the maximum number of prefixes for a word, for example, `hello` can have `hel`, `he` and `h`.

ManyTheFish (approved):
LGTM, I suggested another optimisation but it can be tested in another PR.

crates/milli/src/update/new/words_prefix_docids.rs
Kerollmops: I get it.

However, I am wondering how many entries are skipped when iterating on the prefixes to compute instead of iterating on every single word in the database. I am pretty sure that avoiding full scans is the keep to performance.

On another hand, I think we could find a way to iterate on the prefixes with common prefixes, this means that we could iterate on all the words starting with `h` and collect the prefixes for `h`, `he` and `hel`. That's maybe what you are proposing there 🤔

I created an issue to [track this optimization in linear (internal)](https://linear.app/meilisearch/issue/SCA-264/implement-another-optimization-to-compute-prefixes).

PR #5963, opened 2025-10-30
Allow to attach `customMetadata` in the document addition or update tasks

dureuill:
[Linear issue (internal)](https://linear.app/meilisearch/issue/ENGPROD-2128/allow-attaching-user-defined-metadata-to-tasks-and-return-them-in)

This PR adds a new optional string query parameter `customMetadata` to the following routes:

```
POST /indexes/{indexUid}/documents
PUT /indexes/{indexUid}/documents
DELETE /indexes/{indexUid}/documents/{documentId}
POST /indexes/{indexUid}/documents/delete-batch
POST /indexes/{indexUid}/documents/delete
POST /indexes/{indexUid}/documents/edit
DELETE /indexes/{indexUid}/documents
```

If the parameter is provided:
- it will be added to the created task object,
- The task view returned by these routes also echoes the `customMetadata` parameter.
- The payload sent to the webhooks upon task completion will also contain the parameter with its value.

The parameter can be used when sending a task to identify it with a custom user-provided identifier.

# Tests

This PR adds the following tests:

1. Test that checks that webhooks properly add the metadata to the payload
2. Test that checks that when using sharding, tasks are correctly replicated (not related to this PR, but was missing)
3. Test that checks that when using sharding, custom metadata is correctly replicated in the replicated tasks

## This PR makes breaking changes

*Breaking changes are changes to the database such that databases created in an older version of Meilisearch need changes to remain valid in the new version of Meilisearch. This typically happens when the way to store the data changed (change of database, new required key, etc). This can also happen due to breaking changes in the API of an experimental feature. ⚠️ This kind of changes are more difficult to achieve safely, so proceed with caution and test dumpless upgrade right before merging the PR.*

- [x] Detail the change to the DB format and why they are forward compatible
    - Tasks on disk (`KindWithContent`) now contain a new `customMetadata` field. This is forward compatible because the field is optional, and if missing its default value (`null`) will be used instead
- [ ] Forward-compatibility: A database created before this PR and using the features touched by this PR was able to be opened by a Meilisearch produced by the code of this PR.

# Changelog

- To make it easier to keep track of which documents were processed by Meilisearch, it is now possible to attach an arbitrary string to all routes that create document-related tasks.
- Tasks created with this custom metadata will display the passed metadata when accessed by the tasks route or sent in webhooks.
- To use this feature, add the `customMetadata` query parameter to any supported route:
```
POST /indexes/{indexUid}/documents?customMetadata=my-metadata-for-the-task
```
- Note that, as usual for query parameters, the value of the parameter must be URL-encoded.
- List of supported routes:
```
POST /indexes/{indexUid}/documents
PUT /indexes/{indexUid}/documents
DELETE /indexes/{indexUid}/documents/{documentId}
POST /indexes/{indexUid}/documents/delete-batch
POST /indexes/{indexUid}/documents/delete
POST /indexes/{indexUid}/documents/edit
DELETE /indexes/{indexUid}/documents
```
- Sample output of `GET /tasks` for tasks with metadata:
```json
{
  "results": [
    {
      "uid": 37,
      "batchUid": 37,
      "indexUid": "mieli",
      "status": "succeeded",
      "type": "documentDeletion",
      "canceledBy": null,
      "details": {
        "deletedDocuments": 31944
      },
      "error": null,
      "duration": "PT0.511099S",
      "enqueuedAt": "2025-11-06T16:33:37.816237Z",
      "startedAt": "2025-11-06T16:33:37.821591Z",
      "finishedAt": "2025-11-06T16:33:38.33269Z",
      "customMetadata": "removeall"
    },
    {
      "uid": 36,
      "batchUid": 36,
      "indexUid": "movies",
      "status": "succeeded",
      "canceledBy": null,
      "details": {
        "receivedDocuments": 31968,
        "indexedDocuments": 31944
      },
      "error": null,
      "duration": "PT3.192271S",
      "enqueuedAt": "2025-10-30T10:31:12.896073Z",
      "startedAt": "2025-10-30T10:31:12.911905Z",
      "finishedAt": "2025-10-30T10:31:16.104176Z",
      "customMetadata": "foo"
    }
  ],
  "total": 38,
  "limit": 2,
  "from": 36,
  "next": 35
}
```

### DIFF
--- a/crates/meilisearch/src/routes/indexes/documents.rs
+++ b/crates/meilisearch/src/routes/indexes/documents.rs
@@ -1153,7 +1153,7 @@ async fn copy_body_to_file(
 /// Delete a set of documents based on an array of document ids.
 #[utoipa::path(
     post,
-    path = "{indexUid}/delete-batch",
+    path = "{indexUid}/documents/delete-batch",
     tag = "Documents",
     security(("Bearer" = ["documents.delete", "documents.*", "*"])),
     params(


Kerollmops (commented):
LGTM, but we decided to use a query parameter and not a header. Why not a header? Is it more complex, stranger? Should we accept both? How do query parameters behave with JSON, and isn't it easier to put JSON in headers?

Shouldn't we add this parameter to the documents deletion routes as well?

dureuill:
@Kerollmops 

>  Why not a header?

1. We don't have prior examples of Meilisearch APIs using headers
2. I'm not sure that we have anything in place to document route headers in the API reference
3. Adding headers is more complicated in most HTTP clients (having to pass a dictionary of headers, the naming convention for header is different than the one for query parameters and body parameters, some clients may remove headers that are otherwise automatically added such as Content-Type and Authorization, etc)
4. There was already a `primaryKey` query parameter on the modified routes, so this is "just" adding another query parameter.

> How do query parameters behave with JSON

The `customMetadata` parameter is meant to be a string, not an arbitrary JSON value. A user may use a JSON value as a string in the query parameter (as long as it is properly url-encoded), but Meilisearch will still display it as a string.

> and isn't it easier to put JSON in headers?

I don't think so, they need to be url encoded in both cases

> Shouldn't we add this parameter to the documents deletion routes as well?

We may, I didn't because the relative implementation complexity wasn't the same, as these routes currently don't have any query parameter.

Kerollmops (approved):
A lot of tests, fairly easy to implement, easy to review, perfect 👌

PR #5968, opened 2025-11-03
Redact Authorization header in webhooks

dureuill:
[Linear issue (internal)](https://linear.app/meilisearch/issue/ENGPROD-2116/privilege-escalation-from-webhook-api-key-to-master-key)

The value of the `Authorization` header is now redacted when getting webhook, getting webhooks, or in the object returned when posting a new webhook or deleting a webhook.

- Impact on Cloud: Similarly to embedder API, the UI should make sure to never send the `Authorization` header back after it has been redacted
- Technically a **breaking change**, as users could previously get the key value back when getting the webhook, and that will no longer be possible

### DIFF
--- a/crates/meilisearch-types/src/settings.rs
+++ b/crates/meilisearch-types/src/settings.rs
@@ -351,6 +351,7 @@ impl<T> Settings<T> {
     }
 }
 
+/// Redact a secret string, starting from the `secret_offset`th byte.
 pub fn hide_secret(secret: &mut String, secret_offset: usize) {
     match secret.len().checked_sub(secret_offset) {
         None => (),


crates/meilisearch-types/src/webhooks.rs
```diff
@@ -11,6 +11,24 @@ pub struct Webhook {
...
+    pub fn redact_authorization_header(&mut self) {
+        // headers are case insensitive, so to make the redaction robust we iterate over qualifying headers
+        // rather than getting one canonical `Authorization` header.
+        for value in self
+            .headers
+            .iter_mut()
+            .filter_map(|(name, value)| name.eq_ignore_ascii_case("authorization").then_some(value))
+        {
+            if value.starts_with("Bearer ") {
+                crate::settings::hide_secret(value, "Bearer ".len());
```
Kerollmops, line 24: Now I understand the secret offset 👀

Kerollmops (dismissed):

crates/meilisearch-types/src/webhooks.rs
dureuill: is the parameter badly named? I can change it

dureuill:
@Kerollmops : pointing out this is a breaking change

crates/meilisearch-types/src/settings.rs
```diff
@@ -346,24 +346,25 @@ impl<T> Settings<T> {
...
-            x if x < 20 => {
-                secret.replace_range(2.., "XXXX...");
-            }
-            x if x < 30 => {
-                secret.replace_range(3.., "XXXXX...");
-            }
-            _x => {
-                secret.replace_range(5.., "XXXXXX...");
-            }
+pub fn hide_secret(secret: &mut String, secret_offset: usize) {
```
dureuill, line 355:
```suggestion
/// Redact a secret string, starting from the `secret_offset`th byte.
pub fn hide_secret(secret: &mut String, secret_offset: usize) {
```

PR #5977, opened 2025-11-06
Fix error that rust-analyzer reports because it is compiling all code with the `test` cfg

dureuill:
If using rust-analyzer, you might have noticed false positives in terms of reported errors:

Capture d’écran 2025-11-06 à 18 20 10

The cause was that `IndexScheduler::new` has a different signature (4 or 6 arguments) depending on whether the cfg is `test` or not.

Rust analyzer compiles all code with the `test` configuration, causing the non-test code to error on missing arguments.

This PR solves this small hurdle by separating the constructors for index scheduler between the test and non-test variant, and calling the test one in test contexts.

As much as the code as possible was factored between both constructors, although one must note that a tiny bit of duplication was introduced (calling `IndexScheduler::run`) on the created index scheduler) and the test parameters are set using mutation.

### DIFF
--- a/crates/index-scheduler/src/lib.rs
+++ b/crates/index-scheduler/src/lib.rs
@@ -259,14 +259,23 @@ impl IndexScheduler {
     }
 
     /// Create an index scheduler and start its run loop.
-    #[allow(private_interfaces)] // because test_utils is private
     pub fn new(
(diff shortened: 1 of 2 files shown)


Kerollmops (approved):
Thank you very much, no more single remaining error while compiling 👌

PR #5980, opened 2025-11-09
Support ModernBERT architecture on `huggingface` embedder

hayatosc:
This PR is that I forked and created for my personal project. Please feel free to close it if you feel it is not needed.

As a Japanese developer, I'm planning to use hybrid search with [Ruri v3](https://huggingface.co/cl-nagoya/ruri-v3-30m) , which is extremely powerful embedding model that beats OpenAI/text-embedding-3-large in Japanese benchmarks with just 30m parameters.

But it doesn't work because meilisearch does not support any non-BERT models on `huggingface` embedder.

This PR is adding support for ModernBERT and should be the first step on supporting non-BERT models.

## Requirements


hayatosc:
Thank you for your help! I'll also work on implementing other architectures soon when I have time.

PR #5983, opened 2025-11-10
Support the searchable and exact attributes in the new Settings Indexer

Kerollmops:
This PR aims to improve the current settings by supporting the update of searchable attributes with the new settings indexer. Only the modifications to the searchable attributes and exact attributes are considered in this PR.

The new indexer is designed to reduce disk writes significantly. Its main approach is to read documents directly from LMDB instead of duplicating the dataset into a temporary file and processing it from scratch each time a setting change occurs. Instead, we calculate the difference between the settings and apply only the necessary changes. This second iteration adds support for searchable attributes, following the previous support for embedders.

Global changes like changes to stop words, non-separator tokens, separator tokens, dictionary, exact words, proximity precision, localized attribute rules, prefix search, or disable on numbers are taken into account and will, therefore, use the old settings indexer. A follow-up PR is coming to support them in the new indexer.

## To Do

- [x] Check if default embedder template is correctly impacted.
- [x] Create and call `update_searchable_configs` in settings.rs:L1633.
- [x] Make sure that only changing the searchable or exact attributes enters the new settings indexer.

### DIFF
--- a/crates/meilisearch/tests/settings/parent_seachable_fields.rs
+++ b/crates/meilisearch/tests/settings/parent_seachable_fields.rs
@@ -94,4 +94,21 @@ async fn nested_field_becomes_searchable() {
             "###);
         })
         .await;
+
+    let (response, code) = index
+        .update_settings(json!({
+            "searchableAttributes": ["meta.title"]
+        }))
+        .await;
+    assert_eq!("202", code.as_str(), "{response:?}");
+    server.wait_task(response.uid()).await.succeeded();
+
+    // We expect no documents when searching for
+    // a nested non-searchable field
+    index
+        .search(json!({"q": "many fish"}), |response, code| {
+            snapshot!(code, @"200 OK");
+            snapshot!(json_string!(response["hits"]), @r###"[]"###);
+        })
+        .await;
 }
--- a/crates/milli/src/update/new/extract/searchable/tokenize_document.rs
+++ b/crates/milli/src/update/new/extract/searchable/tokenize_document.rs
@@ -27,16 +27,21 @@ impl DocumentTokenizer<'_> {
         token_fn: &mut impl FnMut(&str, FieldId, u16, &str) -> Result<()>,
     ) -> Result<()> {
         let mut field_position = HashMap::new();
-        let mut tokenize_field = |field_name: &str, _depth, value: &Value| {
-            let (field_id, pattern_match) = should_tokenize(field_name)?;
-            if pattern_match == PatternMatch::Match {
-                self.tokenize_field(field_id, field_name, value, token_fn, &mut field_position)?;
-            }
-            Ok(pattern_match)
-        };
-
         for entry in document.iter_top_level_fields() {
             let (field_name, value) = entry?;
+
+            if let (_, PatternMatch::NoMatch) = should_tokenize(field_name)? {
+                continue;
+            }
+
+            let mut tokenize_field = |field_name: &str, _depth, value: &Value| {
+                let (fid, pattern_match) = should_tokenize(field_name)?;
+                if pattern_match == PatternMatch::Match {
+                    self.tokenize_field(fid, field_name, value, token_fn, &mut field_position)?;
+                }
+                Ok(pattern_match)
+            };
+
             // parse json.
             match serde_json::to_value(value).map_err(InternalError::SerdeJson)? {
                 Value::Object(object) => seek_leaf_values_in_object(


crates/milli/src/update/new/extract/vectors/mod.rs
```diff
@@ -765,7 +789,9 @@ impl<'a, 'b, 'extractor> Chunks<'a, 'b, 'extractor> {
                         }
                         session.request_embedding(metadata, input, unused_vectors_distribution)?;
                     }
-                    ExtractorDiff::Unchanged => { /* do nothing */ }
+                    ExtractorDiff::Unchanged => {
+                        dbg!("coucou unchange"); /* do nothing */
```
ManyTheFish: Coucou @Kerollmops  👋

crates/milli/src/update/new/extract/searchable/extract_word_docids.rs
```diff
@@ -411,3 +464,300 @@ impl WordDocidsExtractors {
...
+        for cache in datastore.into_iter().flat_map(RefCell::into_inner) {
+            merger.push(cache)?;
+        }
+
+        Ok(merger)
+    }
+
+    // TODO find a better name (extract_document_change?)
+    //      and document this method.
+    fn extract_settings_change<SD: SettingsDelta>(
```
ManyTheFish: extract_from_settings_change
extract_document_from_settings_change
settings_change_extract_document

crates/milli/src/update/new/extract/searchable/extract_word_docids.rs
```diff
@@ -411,3 +464,300 @@ impl WordDocidsExtractors {
...
+        // This check doesn't trigger the tokenizer as we never return
+        // PatternMatch::Match.
+        document_tokenizer.tokenize_document(
+            current_document,
+            &mut |field_name| {
+                let fid = new_fields_ids_map.id(field_name).expect("All fields IDs must exist");
+
+                if action == ActionToOperate::ReindexAllFields {
+                    return Ok((fid, PatternMatch::NoMatch));
+                }
```
ManyTheFish, line 614:
```suggestion
                // If the document must be reindexed, early return NoMatch to stop the scanning process. 
                if action == ActionToOperate::ReindexAllFields {
                    return Ok((fid, PatternMatch::NoMatch));
                }
```

crates/milli/src/update/new/extract/searchable/extract_word_docids.rs
```diff
@@ -411,3 +464,300 @@ impl WordDocidsExtractors {
...
+                        ActionToOperate::ReindexAllFields
+                    }
+                    (Metadata { searchable: Some(_), .. }, Metadata { searchable: None, .. }) => {
+                        ActionToOperate::ReindexAllFields
+                    }
+                    (Metadata { searchable: None, .. }, Metadata { searchable: Some(_), .. }) => {
+                        ActionToOperate::IndexAddedFields
+                    }
+                    _ => action,
+                };
```
ManyTheFish, line 636:
```suggestion
                action = match (old_field_metadata, new_field_metadata) {
                    // At least one field is added or removed from the exact fields => ReindexAllFields
                    (Metadata { exact: old_exact, .. }, Metadata { exact: new_exact, .. })
                        if old_exact != new_exact =>
                    {
                        ActionToOperate::ReindexAllFields
                    }
                    // At least one field is removed from the searchable fields => ReindexAllFields
                    (Metadata { searchable: Some(_), .. }, Metadata { searchable: None, .. }) => {
                        ActionToOperate::ReindexAllFields
                    }
                    // At least one field is added in the searchable fields => IndexAddedFields
                    (Metadata { searchable: None, .. }, Metadata { searchable: Some(_), .. }) => {
                        // We can safely overwrite the action, because we early return when action is ReindexAllFields.
                        ActionToOperate::IndexAddedFields
                    }
                    _ => action,
                };
```

crates/milli/src/update/new/extract/searchable/extract_word_docids.rs
```diff
@@ -411,3 +464,300 @@ impl WordDocidsExtractors {
...
+                }
+                ActionToOperate::IndexAddedFields => {
+                    let has_searchable_children =
+                        |field_name: &str, searchable: Option<&Vec<String>>| {
+                            searchable.is_none_or(|fields| {
+                                fields.iter().any(|attr| {
+                                    match_field_legacy(attr, field_name) != PatternMatch::Parent
+                                })
+                            })
+                        };
```
ManyTheFish: What do you think about making it a function and using it at line 655? (inside `ActionToOperate::ReindexAllFields =>`), this would ease the readability 🤔

crates/milli/src/update/new/extract/searchable/extract_word_docids.rs
```diff
@@ -411,3 +464,300 @@ impl WordDocidsExtractors {
...
+                    if !old_field_metadata.is_searchable() && new_field_metadata.is_searchable() {
+                        PatternMatch::Match
+                    // If the field was not a parent of a searchable before and is now
+                    } else if !has_searchable_children(field_name, old_searchable)
+                        && has_searchable_children(field_name, new_searchable)
+                    {
+                        PatternMatch::Parent
+                    } else {
+                        PatternMatch::NoMatch
+                    }
```
ManyTheFish: I know I suggested this optimisation, but this is wrong.
let say:
```
searchableAttributes:
// old
["meta.author"]

// new
["meta.author", "meta.illustrator"]
```

We must index `meta.illustrator`, but when we are on the `meta` field:
- `has_searchable_children(field_name, old_searchable)` => true, `meta` is the parent of `meta.author`
- `has_searchable_children(field_name, new_searchable)` => true, `meta` is the parent of `meta.author` and `meta.illustrator`

giving:
```rust
                    // If the field was not a parent of a searchable before and is now
                    } else if !true && true
                    {
                        PatternMatch::Parent
                    } else {
                        PatternMatch::NoMatch
                    }
```

we return `PatternMatch::NoMatch` and we never extract `meta.illustrator`.

```suggestion
                    // If the field is now a parent of a searchable field
                    } else if has_searchable_children(field_name, new_searchable)
                    {
                        PatternMatch::Parent
                    } else {
                        PatternMatch::NoMatch
                    }
```

crates/milli/src/update/new/indexer/extract.rs
```diff
@@ -391,6 +390,102 @@ where
...
+            merge_and_send_docids(
+                fid_word_count_docids,
+                index.field_id_word_count_docids.remap_types(),
+                index,
+                extractor_sender.docids::<FidWordCountDocids>(),
+                &indexing_context.must_stop_processing,
+            )?;
+        }
+    };
+
```
ManyTheFish, line 519: WordPairProximity must be extracted if a searchable attribute is added or removed as the other searchable databases 😬

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -235,6 +241,12 @@ where
     SD: SettingsDelta + Sync,
 {
     delete_old_embedders_and_fragments(wtxn, index, settings_delta)?;
+    delete_old_fid_based_databases(wtxn, index, settings_delta, must_stop_processing, progress)?;
+
+    // TODO delete useless searchable databases
+    //      - Clear word_pair_proximity if byWord to byAttribute
```
ManyTheFish: can be done in this PR easily

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -235,6 +241,12 @@ where
...
+    //      - Clear fid_prefix_* in the post processing
```
ManyTheFish, line 265: must be done in this PR (related to searchableAttributes)

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -235,6 +241,12 @@ where
...
+    //      - clear the prefix + fid_prefix if setting `PrefixSearch` is enabled
```
ManyTheFish, line 266: can be delayed to another PR

crates/milli/src/update/settings.rs
```diff
@@ -1589,33 +1589,33 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
...
             max_values_per_facet: Setting::NotSet,
             sort_facet_values_by: Setting::NotSet,
             pagination_max_total_hits: Setting::NotSet,
             proximity_precision: Setting::NotSet,
             embedder_settings: _,
             search_cutoff: Setting::NotSet,
-            localized_attributes_rules: Setting::NotSet,
-            prefix_search: Setting::NotSet,
+            localized_attributes_rules: Setting::NotSet, // TODO to start with
+            prefix_search: Setting::NotSet,              // TODO continue with this
```
ManyTheFish, line 1616: Do you plan to support these in this PR?

Kerollmops: Nope, I plan to move that to a following PR. Would you like me to remove these comments?

crates/milli/src/update/new/extract/searchable/extract_word_docids.rs
Kerollmops: 2️⃣ ✅

crates/milli/src/update/settings.rs
ManyTheFish: no I don't care

.github/workflows/test-suite.yml
```diff
@@ -154,8 +154,8 @@ jobs:
           apt-get update && apt-get install -y curl
           apt-get install build-essential -y
       - uses: dtolnay/rust-toolchain@1.89
-      - name: Cache dependencies
-        uses: Swatinem/rust-cache@v2.8.0
+      - name: Run df -h
+        run: df -h
```
ManyTheFish: How is it related to the PR?

crates/milli/src/update/new/extract/searchable/mod.rs
```diff
@@ -27,3 +31,15 @@ pub fn match_searchable_field(
...
+
+fn has_searchable_children<I, A>(field_name: &str, searchable: Option<I>) -> bool
+where
+    I: IntoIterator<Item = A>,
+    A: AsRef<str>,
+{
+    searchable.is_none_or(|fields| {
+        fields
+            .into_iter()
+            .any(|attr| match_field_legacy(attr.as_ref(), field_name) != PatternMatch::Parent)
```
ManyTheFish: We are missing a test for that, specifically one with:
 ```yaml
serachableAttributes:
  old: ["meta.author"]
  new: ["meta.author", "meta.illustrator"]
 ```


```suggestion
            .any(|attr| match_field_legacy(attr.as_ref(), field_name) == PatternMatch::Parent)
```

crates/milli/src/update/new/extract/searchable/mod.rs
```diff
@@ -27,3 +31,15 @@ pub fn match_searchable_field(
 
     selection
 }
+
+fn has_searchable_children<I, A>(field_name: &str, searchable: Option<I>) -> bool
```
ManyTheFish, line 37:
```suggestion
/// return `true` if the provided `field_name` is a parent of at least one of the fields contained in `searchable`,
/// or if `searchable` is `None`.
fn has_searchable_children<I, A>(field_name: &str, searchable: Option<I>) -> bool
```

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -456,6 +492,107 @@ where
...
+                // or if it was never referenced in any document
+                .filter_map(|name| if new.contains(name) { None } else { fields_ids_map.id(name) })
+                .collect()
+        })
+    };
+
+    let fids_to_delete = match fids_to_delete {
+        Some(fids) => fids,
+        None => return Ok(()),
+    };
```
ManyTheFish: personal preference, ignore if you prefer the match. 
```suggestion
    let Some(fids_to_delete) = fids_to_delete else {
        return Ok(());
    };
```

ManyTheFish (changes requested):

.github/workflows/test-suite.yml
Kerollmops: Nice catch, it's not and I'll remove this 👌

PR #5984, opened 2025-11-10
Embedder failure modes

dureuill:
This PR introduces ways to ignore some errors when indexing documents:

1. Allow to ignore errors when document template fails
2. Allow to ignore errors when an embedder request fails

To enable these new error policies, one can set the `MEILI_EXPERIMENTAL_CONFIG_EMBEDDER_FAILURE_MODES` environment variables to comma-separated list of the following values:

1. `ignore_document_template_failures` to ignore document template failures
2. `ignore_embedder_failures` to ignore embedder failures

For example:

- `ignore_document_template_failures,ignore_embedder_failures` ignores both kinds of failures
- `ignore_embedder_failures` only ignores embedder failures
- ` `, `,,,,`, or if the variable is not set: no failures are ignored

---

Furthermore, this PR introduces the `MEILI_EXPERIMENTAL_REST_EMBEDDER_TIMEOUT_SECONDS` environment variable to configure the duration of the REST client timeout.

# Manual testing

The environment variable interface does not lend itself to automated testing. However, manual tests were done with an openAI embedder exhibiting a wrong key:

1. When a document was passed that would not pass the document template, when the variable is not set, the expected errors around document template occurs
2. The same document with the variable set to `ignore_document_template_failures` is indexed successfully (albeit without a vector)
3. A document that passes the document template, when the variable is not set or set to `ignore_document_template_failures`, exposes the expected error around OpenAI authentication
4. The same document when the variable contains `ignore_embedder_failures` passes indexing, albeit the batch contains the authentication-related error message and the document has no embedding
5. Adding a user provided embedder when the variable contains `ignore_embedder_failures` succeeds despite there being documents not containing any `_vectors`

### DIFF
--- a/crates/milli/src/update/new/extract/vectors/mod.rs
+++ b/crates/milli/src/update/new/extract/vectors/mod.rs
@@ -1170,38 +1170,32 @@ struct EmbedderFailureModes {
 
 impl EmbedderFailureModes {
     fn from_env() -> Self {
-        std::env::var_os("MEILI_EXPERIMENTAL_CONFIG_EMBEDDER_FAILURE_MODES")
-            .map(|failure_modes| {
-                Self::parse_from_os_str(
-                    &failure_modes,
-                    "`MEILI_EXPERIMENTAL_CONFIG_EMBEDDER_FAILURE_MODES`",
-                )
-            })
-            .unwrap_or_default()
+        match std::env::var("MEILI_EXPERIMENTAL_CONFIG_EMBEDDER_FAILURE_MODES") {
+            Ok(failure_modes) => Self::parse_from_str(
+                &failure_modes,
+                "`MEILI_EXPERIMENTAL_CONFIG_EMBEDDER_FAILURE_MODES`",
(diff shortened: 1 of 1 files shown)


crates/milli/src/update/new/extract/vectors/mod.rs
```diff
@@ -1078,3 +1161,47 @@ where
...
+            Self { ignore_embedder_failures: true, ignore_document_template_failures: false }
+        } else if failure_modes == " " || failure_modes.is_empty() || failure_modes == "," {
+            Self { ignore_document_template_failures: false, ignore_embedder_failures: false }
+        } else {
+            panic!(
+                "Unexpected value {failure_modes} for {provenance}",
+                failure_modes = failure_modes.to_string_lossy()
+            )
+        }
+    }
```
Kerollmops: Could you please clean that up?

dureuill: done

PR #5985, opened 2025-11-10
Introduce a delta-encoded bitmap

Kerollmops:
This PR introduces a new method for reducing the database size. Our first measures showed that the size of the word-docids database was reduced by half (yes, 50%) and that the word-prefix-docids database was reduced by 30%.

### Changelog

We introduced a more efficient way to store internal data structures on disk. This new technique, called delta-encoding, is very space-efficient when storing many runs of increasing integers, which is the case for Meilisearch. Our measures showed that the database size can be reduced by up to 50% in some cases. Note that the engine will not convert your database to this new format immediately; instead, it will do so gradually during indexing, whenever a data structure is naturally modified. It's like the engine is gradually adopting the new format. This feature can be disabled by using the `--experimental-disable-delta-encoding` or by setting the environment variable `MEILI_EXPERIMENTAL_DISABLE_DELTA_ENCODING=true`.

## Notes for the reviewer(s)

Please check that I only converted `CboRoaringBitmapCodec` to `DeCboRoaringBitmapCodec`, as that is the only valid way to change the codec. `DeCboRoaringBitmapCodec` codecs can read `CboRoaringBitmapCodec` data but not `RoaringBitmapCodec` data, and this can cause many hard-to-debug errors if we don't check for it properly, since it will generate silent errors and bitmaps with invalid values (e.g., 1, 16, ...) will be generated when deserializing them.

This is why you should also double-check that the CLI/env var option is valid, and we must ensure that this new codec is not used in the cloud by default. I added a new workflow test to check that disabling delta-encoding works correctly. Note that disabling delta-encoding doesn't disable deserialization of delta-encoded bitmaps.

TBD: AVX (ARM64) is only. library bitpacking does support large buckets: determine if we want to disable encoding them on ARM.

*Forward-compatible changes are changes to the database such that databases created in an older version of  Meilisearch are still valid in the new version of Meilisearch. They usually represent additive changes, like adding a new optional attribute or setting.*

### Detail the change to the DB format and why it is forward compatible

We introduced a new `DeCboRoaringBitmap` that is capable of deserializing the `CboRoaringBitmaps`, deserializing `DeRoaringBitmap`s (delta-encoded), and encoding the numbers as `DeRoaringBitmaps`. For this, we add a header to the serialization format and check the size of the serialized data; if it is below a threshold, it consists of raw u32s.
		
- [x] Declarative test: add a [declarative test containing a dumpless upgrade](https://github.com/meilisearch/meilisearch/blob/main/TESTING.md#typical-usage)
		A new declarative test named `disabled-delta-encoding` has been introduced and:
		  - Starts from v1.12 without the delta-encoding feature (as it didn't exist),
		  - Indexes some documents,
		  - Switches to the latest version of Meilisearch with delta-encoding **disabled**,
		  - Indexes more documents.
		Another declarative test named `enabled-delta-encoding` has been introduced and:
		  - Starts from v1.12 without the delta-encoding feature (as it didn't exist),
		  - Indexes some documents,
		  - Switches to the latest version of Meilisearch with delta-encoding **enabled**,
		  - Indexes more documents.

## To Do
- [x] Make sure the delta-encoding codec is named `DeRoaringBitmapCodec`.
- [x] Create a `DeCboRoaringBitmap` that manages all encodings.
- [x] Make sure we can disable the delta-encoding via an experimental CLI parameter.
- [ ] ~Implement the conditional compilation with an `fn()` instead of a boolean.~
	   Actually, branching always on the same branch is well optimized by CPUs. We can change this in a future version and [running a pass of PGO](https://en.wikipedia.org/wiki/Profile-guided_optimization) can drastically speed things up.
- [ ] Introduce an alias to simplify the usage in milli's index.
- [x] Add workflow tests to check that disabling delta-encoding works as expected:
	    Disable it after processing tasks, then continue processing more tasks.
- [x] Enable delta encoding by default. Make it possible to disable it by using the `MEILI_EXPERIMENTAL_DISABLE_DELTA_ENCODING=true` env variable.


crates/index-scheduler/src/queue/batches.rs
```diff
@@ -42,11 +42,11 @@ pub struct BatchQueue {
     /// Store the batches associated to an index.
     pub(crate) index_tasks: Database<Str, RoaringBitmapCodec>,
```
Kerollmops, line 43: I think I will switch to `DeCboRoaringBitmap` here, too!

crates/index-scheduler/src/queue/tasks.rs
```diff
@@ -44,11 +44,11 @@ pub struct TaskQueue {
     /// Store the tasks that were canceled by a task uid
     pub(crate) canceled_by: Database<BEU32, RoaringBitmapCodec>,
```
Kerollmops, line 45: I think I will switch to `DeCboRoaringBitmap` here, too!

crates/index-scheduler/src/insta_snapshot.rs
```diff
@@ -4,7 +4,7 @@ use std::fmt::Write;
 use meilisearch_types::batches::{Batch, BatchEnqueuedAt, BatchStats};
 use meilisearch_types::heed::types::{SerdeBincode, SerdeJson, Str};
 use meilisearch_types::heed::{Database, RoTxn};
-use meilisearch_types::milli::{CboRoaringBitmapCodec, RoaringBitmapCodec, BEU32};
+use meilisearch_types::milli::{DeCboRoaringBitmapCodec, RoaringBitmapCodec, BEU32};
```
Kerollmops, line 7: It seems I am using the `RoaringBitmapCodec` and I think I will switch to DeCboRoaringBitmap here, too!

crates/meilisearch/src/option.rs
```diff
@@ -1292,7 +1262,7 @@ where
     T: AsRef<OsStr>,
 {
     if let Err(VarError::NotPresent) = std::env::var(key) {
-        std::env::set_var(key, value);
+        unsafe { std::env::set_var(key, value) }
```
Kerollmops: Not related to this PR.

crates/meilisearch/tests/common/server.rs
Kerollmops, line 1: File changes are mostly not related to this PR.

crates/milli/src/heed_codec/roaring_bitmap/de_cbo_roaring_bitmap_codec.rs
```diff
@@ -0,0 +1,341 @@
...
+        let mut output = Vec::with_capacity(capacity);
+        Self::serialize_into_with_tmp_buffer(item, &mut output, &mut tmp_buffer)?;
+        Ok(Cow::Owned(output))
+    }
+}
+
+/// Manages the global status of the delta encoding.
+///
+/// Whether we must use delta encoding or not when encoding roaring bitmaps.
+pub struct DeltaEncodingStatusLock(OnceLock<DeltaEncodingStatus>);
```
Kerollmops, line 230: I think I should switch to using function pointers to avoid branching instead 🤔

EDIT: We can do that later, and branching is very efficient when the branch is predictable, which it is.

crates/milli/src/heed_codec/roaring_bitmap/de_roaring_bitmap_codec.rs
```diff
@@ -0,0 +1,454 @@
...
+const MAGIC_HEADER: u16 = 36869;
+
+pub struct DeRoaringBitmapCodec;
+
+// TODO reintroduce:
+//  - serialized_size?
+//  - serialize_into_vec
+//  - intersection_with_serialized
+//  - merge_into
+//  - merge_deladd_into
```
Kerollmops: Remove this comment.

crates/milli/src/heed_codec/roaring_bitmap/mod.rs
Kerollmops, line 1: ~~It would be great to make sure that nobody can ever use the `RoaringBitmapCodec` and the `CboRoaringBitmapCodec`. People can still use the `DeCboRoaringBitmapCodec` and, maybe, the `BoRoaringBitmapCodec`.~~

EDIT: `DeCboRoaringBitmapCodec` cannot be used to read data encoded using a `RoaringBitmapCodec`. We must keep the `RoaringBitmapCodec` as is.

crates/milli/src/heed_codec/roaring_bitmap_length/mod.rs
Kerollmops, line 1: Same, we want to restrict the codecs we can use outside of this module.

crates/milli/src/heed_codec/mod.rs
Kerollmops, line 1: We want to restrict the use of all the codecs but the `DeCboRoaringBitmapCodec`.

crates/milli/src/update/index_documents/mod.rs
```diff
@@ -764,8 +764,8 @@ where
 fn execute_word_prefix_docids(
     txn: &mut heed::RwTxn<'_>,
     merger: Merger<CursorClonableMmap, MergeDeladdCboRoaringBitmaps>,
```
Kerollmops: We must delete the `MergeDeladdCboRoaringBitmaps` codec and replace it with a `MergeDeladdDeCboRoaringBitmaps`.

crates/milli/src/update/new/words_prefix_docids.rs
```diff
@@ -87,12 +87,12 @@ impl<'i> WordPrefixDocids<'i> {
...
                         .database
                         .prefix_iter(&rtxn, prefix.as_bytes())?
-                        .remap_types::<Str, CboRoaringBitmapCodec>()
+                        .remap_types::<Str, DeCboRoaringBitmapCodec>()
                         .map(|result| result.map(|(_word, bitmap)| bitmap))
                         .union()?;
 
                     buffer.clear();
-                    CboRoaringBitmapCodec::serialize_into_vec(&output, &mut buffer);
+                    DeCboRoaringBitmapCodec::serialize_into(&output, &mut buffer)?;
```
Kerollmops: Can we optimize this by providing the tmp buffer?

crates/milli/src/lib.rs
```diff
@@ -73,7 +73,7 @@ pub use self::filterable_attributes_rules::{
 };
 pub use self::heed_codec::{
     BEU16StrCodec, BEU32StrCodec, BoRoaringBitmapCodec, BoRoaringBitmapLenCodec,
-    CboRoaringBitmapCodec, CboRoaringBitmapLenCodec, FieldIdWordCountCodec, ObkvCodec,
+    DeCboRoaringBitmapCodec, DeCboRoaringBitmapLenCodec, FieldIdWordCountCodec, ObkvCodec,
     RoaringBitmapCodec, RoaringBitmapLenCodec, StrBEU32Codec, U8StrStrCodec,
     UncheckedU8StrStrCodec,
```
Kerollmops: We want to restrict the use of codecs other than the `DeCboRoaringBitmapCodec`.

crates/meilisearch-types/src/tasks/network.rs
```diff
@@ -461,7 +461,8 @@ impl Serialize for TaskKeys {
     {
         let TaskKeys(task_keys) = self;
         let mut bytes = Vec::new();
-        DeCboRoaringBitmapCodec::serialize_into(task_keys, &mut bytes);
+        // TODO correctly handle this io::Error
+        DeCboRoaringBitmapCodec::serialize_into(task_keys, &mut bytes).unwrap();
```
Kerollmops: Correctly catch this `io::Error`.

Kerollmops:
The code is still not right. I tried it on a very large dataset, and searching for "Renaud" returned no good results. However, using phrase search was working fine.

EDIT: I forgot that Meilisearch has a default search cut-off of 1500ms; it seems that was the issue from the beginning... I checked, and the ranking score details show skipped ranking scores after about 1.5s.

Kerollmops:
Now I trigger this error but I'm not sure why:
> internal: MDB_BAD_VALSIZE: Unsupported size of key/DB name/data, or wrong DUPFIXED size.

EDIT: I have a clue, the failures append when we were processing the word FST, it could be larger than 2/4GiB!? So not related to this PR, probably a settings configuration issue 🤔 

```json
{
  "uid": 43,
  "progress": null,
  "details": {
    "receivedDocuments": 4124647,
    "indexedDocuments": 0
  },
  "stats": {
    "totalNbTasks": 102,
    "status": {
      "failed": 102
    },
    "types": {
      "documentAdditionOrUpdate": 102
    },
    "indexUids": {
      "tracks": 102
    },
    "progressTrace": {
      "processing tasks > retrieving config": "1.36ms",
      "processing tasks > computing document changes > preparing payloads > payload": "3.79s",
      "processing tasks > computing document changes > preparing payloads": "3.79s",
      "processing tasks > computing document changes > assigning documents ids": "92.87s",
      "processing tasks > computing document changes > reordering payload offsets": "288.49ms",
      "processing tasks > computing document changes": "96.95s",
      "processing tasks > reading payload stats": "50.24µs",
      "processing tasks > indexing > extracting documents > document": "503.61s",
      "processing tasks > indexing > extracting documents": "503.61s",
      "processing tasks > indexing > extracting facets > document": "1.53s",
      "processing tasks > indexing > extracting facets": "1.53s",
      "processing tasks > indexing > merging facet caches": "1.46ms",
      "processing tasks > indexing > extracting words > document": "68.61s",
      "processing tasks > indexing > extracting words": "68.61s",
      "processing tasks > indexing > merging word caches > word docids": "427.41s",
      "processing tasks > indexing > merging word caches > word field id docids": "354.13s",
      "processing tasks > indexing > merging word caches > exact word docids": "2.56ms",
      "processing tasks > indexing > merging word caches > word position docids": "400.07s",
      "processing tasks > indexing > merging word caches > field id word count docids": "12.29s",
      "processing tasks > indexing > merging word caches": "1193.91s",
      "processing tasks > indexing > extracting word proximity > document": "51.94s",
      "processing tasks > indexing > extracting word proximity": "51.96s",
      "processing tasks > indexing > merging word proximity": "812.38s",
      "processing tasks > indexing > waiting for database writes": "71.49s",
      "processing tasks > indexing > waiting for extractors": "309.71µs",
      "processing tasks > indexing > writing embeddings to database": "1.56ms",
      "processing tasks > indexing > post processing facets > facet search": "2.95ms",
      "processing tasks > indexing > post processing facets": "4.11ms",
      "processing tasks > indexing > post processing words > word fst": "688.80s",
      "processing tasks > indexing > post processing words": "688.80s",
      "processing tasks > indexing": "3393.01s",
      "processing tasks": "3489.96s",
      "writing tasks to disk > task": "10.62ms",
      "writing tasks to disk": "10.66ms"
    }
  },
  "duration": "PT3489.967929421S",
  "startedAt": "2026-01-16T14:29:03.830053772Z",
  "finishedAt": "2026-01-16T15:27:13.797983193Z",
  "batchStrategy": "batched up to configured batch size limit of 2147483648B with a total of 2160039219B"
}
```

Kerollmops:
The last benchmark failed with an error introduced in this PR:

> An unknown internal document id have been used: `6299712`

This was found by using the `search-movies-subset-hf-embeddings` workload. I'll reproduce and investigate.

crates/index-scheduler/src/queue/batches.rs
Kerollmops: Nope, it is impossible to change a codec from RoaringBitmap to DeCboRoaringBitmap. The latter cannot read the former, and upgrading would break the database. We could implement a conversion during dumpless upgrades, but we can save that for future work.

crates/index-scheduler/src/queue/tasks.rs
Kerollmops: Nope, it is impossible to change a codec from RoaringBitmap to DeCboRoaringBitmap. The latter cannot read the former, and upgrading would break the database. We could implement a conversion during dumpless upgrades, but we can save that for future work.

crates/index-scheduler/src/insta_snapshot.rs
Kerollmops: Nope, it is impossible to change a codec from RoaringBitmap to DeCboRoaringBitmap. The latter cannot read the former, and upgrading would break the database. We could implement a conversion during dumpless upgrades, but we can save that for future work.

crates/milli/src/heed_codec/roaring_bitmap_length/mod.rs
Kerollmops: Actually, we still want to use the RoaringBitmap codec for retro-compatibility.

crates/milli/src/heed_codec/mod.rs
Kerollmops: Actually, we still want to use the RoaringBitmap codec for retro-compatibility.

crates/milli/src/lib.rs
Kerollmops: Actually, we still want to use the RoaringBitmap codec for retro-compatibility.

crates/milli/src/heed_codec/roaring_bitmap/de_cbo_roaring_bitmap_codec.rs
Kerollmops: I'm thinking: why don't we use an atomic? Not necessarily an `AtomicBool` but maybe an `AtomicU8` so that I can encode the uninitialized state too. The disadvantage of using an atomic is that it always requires fetching the data via an atomic load operation, whereas a `OnceLock` removes the need to sync anything; it's just a pointer dereference once the data is initialized.

crates/milli/src/heed_codec/roaring_bitmap/de_cbo_roaring_bitmap_codec.rs
```diff
@@ -0,0 +1,379 @@
...
+    /// storing them in the bitmap we use for the final intersection.
+    pub fn intersection_with_serialized(
+        bytes: &[u8],
+        other: &RoaringBitmap,
+    ) -> io::Result<RoaringBitmap> {
+        if CboRoaringBitmapCodec::bytes_deserialize_as_raw_u32s(bytes) {
+            return CboRoaringBitmapCodec::intersection_with_serialized(bytes, other);
+        }
+
+        // TODO move this tmp buffer outside
```
Kerollmops: Yup, I should do that 👀

Kerollmops:
/bench workloads/hackernews-modify-searchables.json workloads/search/filterable-movies.json workloads/search/hackernews.json workloads/search/movies.json workloads/search/sortable-movies.json

Kerollmops:
/bench workloads/hackernews-modify-searchables.json workloads/search/filterable-movies.json workloads/search/hackernews.json workloads/search/movies.json workloads/search/sortable-movies.json

crates/meilitool/src/main.rs
```diff
@@ -169,7 +192,194 @@ fn main() -> anyhow::Result<()> {
...
+                DatabaseInfo { name: lhs_db_name, database: lhs_database },
+                DatabaseInfo { name: rhs_db_name, database: rhs_database },
+            )| {
+                assert_eq!(lhs_db_name, rhs_db_name);
+
+                let lhs_rtxn = lhs_index.read_txn()?;
+                let rhs_rtxn = rhs_index.read_txn()?;
+
+                for (lhs_result, rhs_result) in
+                    lhs_database.iter(&lhs_rtxn)?.zip(rhs_database.iter(&rhs_rtxn)?)
```
YoEight, line 362: Are we sure `lhs` and `rhs` has the same length? if not `compare_entries` would report success when one database could still have trailing keys.

crates/milli/src/heed_codec/roaring_bitmap/de_cbo_roaring_bitmap_codec.rs
```diff
@@ -0,0 +1,378 @@
...
+    ///
+    /// It tries to decode the input by using the delta-decoded version and
+    /// if it fails, falls back to the CboRoaringBitmap version.
+    pub fn deserialize_from_with_tmp_buffer(
+        input: &[u8],
+        tmp_buffer: &mut Vec<u32>,
+    ) -> io::Result<RoaringBitmap> {
+        // The input is too short to be a valid delta-decoded bitmap.
+        // We fall back to the CboRoaringBitmap version with raw u32s.
+        if CboRoaringBitmapCodec::bytes_deserialize_as_raw_u32s(input) {
```
YoEight, line 96: If the delta-encoded bitmap takes less space, could it be possible that one delta-encoded bitmap could be mistaken for a cbo bitmap considering `bytes_deserialize_as_raw_u32s` uses byte length as a format marker?

crates/meilitool/src/main.rs
Kerollmops: I mean, they should be the same size. But you're probably right, and I should switch to using [the `zip_eq` method](https://docs.rs/itertools/latest/itertools/fn.zip_eq.html) of itertools to check for this case 👍

crates/milli/src/heed_codec/roaring_bitmap/de_cbo_roaring_bitmap_codec.rs
Kerollmops: Actually, that's a bit tricky, but if we don't have enough integers to fit into a simd channel, we will end up using a non-compress chunk, which has the same size as raw integers like the Cbo codec. However, to enter this branch and use the DeCbo codec, we need fewer than THRESHOLD integers. So, it's not possible to encode too few numbers that would cause the Cbo codec to think it must be decoded by it, when it was encoded by the De Cbo one.

Actually... now that I think about it, I remember that delta-encoding can encode a chunk with literally zero bytes. That's the case when the integers are exactly consecutive and fit a chunk size. If we are unlucky enough, we could have exactly one chunk of integers following each other, which would result in the header (2 bytes) + the chunk type identifier (1 byte) + compressed length of the first chunk (usize = 8 bytes) = 11 bytes. Knowing that the THRESHOLD is 7 x u32s (28 bytes), it seems that you are right, and in certain conditions, it is possible that the encoded bytes could be decoded by the Cbo codec.

How can we fix that? We could reencode the integers using the Cbo codec instead of the De Cbo codec if the serialized size is smaller than the THRESHOLD. What do you think about this fix? Do you have a better idea? We could compute the `serialized_len` of the De Cbo codec, but I prefer to avoid calling this method every time, since this situation is very rare.

ManyTheFish: A possible alternative is to pad the header of the delta-encoded data, making sure that the first encoded `u32` can't be mistaken for [a roaring bitmap SERIAL_COOKIE](https://github.com/RoaringBitmap/roaring-rs/blob/83caaca2ec5ea29e27dc19f930b9450b9a246b5b/roaring/src/bitmap/serialization.rs#L82-L100).

ManyTheFish: BTW, reviewing the roaring bitmap serialization, I understood that the roaring bitmap minimal size is not 28bytes (7u32) but 15bytes (3 to 4 u32). I may be wrong on this, but if we have 1 run container with 1 interval:
- 4 bytes: SERIAL_COOKIE + size
- 1 byte: run_container_bitmap for 1 container
- 2 bytes: container key
- 2 bytes: container len
- 2 bytes: intervals
- 2 bytes: interval start
- 2 bytes: interval size

CKC?

Kerollmops: Thanks for the investigation @ManyTheFish. I have implemented padding up to 28 bytes to ensure we don't use delta-encoded bitmaps for flat u32s. However, I haven't implemented it like you describe:
> pad the header of the delta-encoded data, making sure that the first encoded u32 can't be mistaken for [a roaring bitmap SERIAL_COOKIE](https://github.com/RoaringBitmap/roaring-rs/blob/83caaca2ec5ea29e27dc19f930b9450b9a246b5b/roaring/src/bitmap/serialization.rs#L82-L100).

I have simply added zeroes at the end of the delta-encoded integers and used the num_bits to count the number of remaining flat u32s (that can't fit in a bitpacker). What do you think about this solution?

About the minimum size of a roaring bitmap, you are right, if we have a run container, it's smaller than the 28 bytes. However, note that we are using v0.10.12 in Meilisearch because run containers [were introduced in v0.11.0](https://github.com/RoaringBitmap/roaring-rs/releases/tag/v0.11.0) and we cannot use them due to what you describe.

I'm still wondering what could be the best move to move to the latest version of the roaring-rs crate one day. I was thinking about a simple feature flag to disable run-length encoding, an alternative method to serialize bitmaps without run-length encoding, or maybe finally get rid of this ugly length-based codec by rewriting Cbo-encoded bitmaps by classic roaring bitmaps...

ManyTheFish (approved):
LGTM on the serializing part, I didn't reviewed the whole PR.

crates/milli/src/heed_codec/roaring_bitmap/de_cbo_roaring_bitmap_codec.rs
ManyTheFish: > I have simply added zeroes at the end of the delta-encoded integers and used the num_bits to count the number of remaining flat u32s (that can't fit in a bitpacker). What do you think about this solution?

I saw that you've already added a cookie at the start of the serialized string, so LGTM

PR #5990, opened 2025-11-13
Make the limit batched tasks total size defaults to half of the max indexing memory

Kerollmops:
This PR automatically sets the limit for the total size of batched tasks to half of the maximum memory. This value can be overwritten by using the `MEILI_EXPERIMENTAL_LIMIT_BATCHED_TASKS_TOTAL_SIZE` or CLI parameter. This significantly improves the engine's indexing speed by ensuring it doesn't crash with too many tasks (the previous default was to auto-batch all enqueued tasks).

### DIFF
--- a/crates/meilisearch/src/analytics/segment_analytics.rs
+++ b/crates/meilisearch/src/analytics/segment_analytics.rs
@@ -195,7 +195,7 @@ struct Infos {
     experimental_enable_logs_route: bool,
     experimental_reduce_indexing_memory_usage: bool,
     experimental_max_number_of_batched_tasks: usize,
-    experimental_limit_batched_tasks_total_size: u64,
+    experimental_limit_batched_tasks_total_size: Option<u64>,
     experimental_network: bool,
     experimental_multimodal: bool,
     experimental_chat_completions: bool,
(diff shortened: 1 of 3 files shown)


crates/meilisearch/src/lib.rs
```diff
@@ -230,7 +230,9 @@ pub fn setup_meilisearch(
         cleanup_enabled: !opt.experimental_replication_parameters,
         max_number_of_tasks: 1_000_000,
         max_number_of_batched_tasks: opt.experimental_max_number_of_batched_tasks,
-        batched_tasks_size_limit: opt.experimental_limit_batched_tasks_total_size.into(),
+        batched_tasks_size_limit: opt
+            .experimental_limit_batched_tasks_total_size
+            .map_or(u64::MAX, |size| size.as_u64()),
```
ManyTheFish: This is the entry point of the index scheduler; the default value should be computed at this step instead of `export_to_env_if_not_present`. 🤔

crates/meilisearch/src/analytics/segment_analytics.rs
```diff
@@ -359,7 +359,7 @@ impl Infos {
             http_payload_size_limit,
             experimental_max_number_of_batched_tasks,
             experimental_limit_batched_tasks_total_size:
-                experimental_limit_batched_tasks_total_size.into(),
+                experimental_limit_batched_tasks_total_size.map_or(u64::MAX, |size| size.as_u64()),
```
ManyTheFish: We should accept none, or send "default"

crates/meilisearch/src/option.rs
```diff
@@ -701,9 +703,20 @@ impl Opt {
             MEILI_EXPERIMENTAL_MAX_NUMBER_OF_BATCHED_TASKS,
             experimental_max_number_of_batched_tasks.to_string(),
         );
+        // We want to base the maximum number of tasks based on the
+        // size of the tasks on the maximum indexing memory defined.
+        // So we fetch the maximum indexing memory and divide it by 2.
```
ManyTheFish: We may not want to export it if it is not present, but we can set the default value when instantiating the index scheduler.

PR #5991, opened 2025-11-13
Release v1.26.0

dureuill:
- update version to v1.26
- take care of dumpless upgrade
- update snaps


crates/index-scheduler/src/upgrade/mod.rs
```diff
@@ -48,6 +48,8 @@ pub fn upgrade_index_scheduler(
         (1, 22, _) => 0,
         (1, 23, _) => 0,
         (1, 24, _) => 0,
+        (1, 25, _) => 0,
```
Kerollmops, line 51: We forgot last time 😬

PR #5994, opened 2025-11-17
Improve S3 upload by showing errors in the task queue

Kerollmops:
This PR enhances the S3 snapshots by gathering and displaying error messages.

### DIFF
--- a/crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
+++ b/crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
@@ -438,12 +438,15 @@ async fn multipart_stream_to_s3(
     db_name: String,
     reader: std::io::PipeReader,
 ) -> Result<(), Error> {
-    use std::{collections::VecDeque, io, os::fd::OwnedFd, path::PathBuf};
+    use std::collections::VecDeque;
+    use std::io;
+    use std::os::fd::OwnedFd;
+    use std::path::PathBuf;
 
     use bytes::{Bytes, BytesMut};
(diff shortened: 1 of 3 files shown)


crates/index-scheduler/src/scheduler/process_snapshot_creation.rs
```diff
@@ -580,16 +579,19 @@ async fn multipart_stream_to_s3(
...
-                Ok(resp) if resp.status().is_client_error() => {
-                    resp.error_for_status().map_err(backoff::Error::Permanent)
-                }
+                Ok(resp) if resp.status().is_client_error() => match resp.error_for_status_ref() {
+                    Ok(_) => Ok(resp),
+                    Err(_) => Err(backoff::Error::Permanent(Error::S3Error {
+                        status: resp.status(),
+                        body: resp.text().await.unwrap_or_default(),
+                    })),
+                },
```
ManyTheFish:
```suggestion
                Ok(resp) if resp.status().is_client_error() && resp.error_for_status_ref().is_err() => {
                     Err(backoff::Error::Permanent(Error::S3Error {
                           status: resp.status(),
                           body: resp.text().await.unwrap_or_default(),
                    }))
                },
```

Kerollmops: You are right, it's much better with a little `&&` in the match condition.

Kerollmops: I even simplified the code by remove the redondant check on the status, we are checking for a client error already!

PR #6000, opened 2025-11-18
Allow changing network topology

dureuill:
This PR enables to dynamically add or remove remotes from a network, ensuring that documents are properly rebalanced between hosts on network changes.

To do so, it makes the following changes:

1. Breaking change: The `Network` object returned or edited by the `/network` route is modified in the following way:
    - the `sharding` boolean is removed
    - a `leader` field is added as an optional string. When it is not `null`, it has the same effect (and more) than having `sharding: true` in the previous iteration of the `Network` object. The leader is used as a check when receiving task creation requests.
    - a `version` field is added as a UUID, defaulting to the null UUID. The version is used when processing tasks.
2. Breaking change: When a `network.leader` is set, calling one of the following routes will fail with `not_a_leader` error if the target's `network.self` is not the same as its `network.leader`:
    - `POST /indexes`
    - `PATCH/DELETE /indexes/{:indexUid}`
    - `POST/PUT/DELETE /indexes/{:indexUid}/documents`
    -  `POST /indexes/{:indexUid}/documents/delete`
    -  `POST /indexes/{:indexUid}/documents/delete-batch`
    -  `POST /indexes/{:indexUid}/documents/edit`
    - `PATCH/DELETE /indexes/{:indexUid}/settings` and settings sub-routes
    - `PATCH /network` if the target is the **new** leader
    - `POST /swap-indexes`
3. Breaking change: when a `leader` is set, `PATCH /network` no longer returns a `Network` object. Rather, it spawns a new `NetworkTopologyChange` task, and returns the summarized task view. 
4. Breaking change: Tasks are duplicated by the leader to the entire network when calling the following routes:
    - `POST /indexes` (new to this PR)
    - `PATCH/DELETE /indexes/{:indexUid}` (new to this PR)
    - `POST/PUT/DELETE /indexes/{:indexUid}/documents` (was already the case before this PR)
    -  `POST /indexes/{:indexUid}/documents/delete` (was already the case before this PR)
    -  `POST /indexes/{:indexUid}/documents/delete-batch` (was already the case before this PR)
    -  `POST /indexes/{:indexUid}/documents/edit` (was already the case before this PR)
    - `PATCH/DELETE /indexes/{:indexUid}/settings` and settings sub-routes (new to this PR)
    - `PATCH /network` if the target is the **new** leader (new to this PR)
    - `POST /swap-indexes` (new to this PR)
5. New `NetworkTopologyChange` tasks that perform the following:
    1. Execute any remaining task to process with a `network.version` lower than the network task's version
    2. Iterate over all documents in all indexes, determine their new shard, and send the document to the remote that must now have it in the new version , deleting it from the local DB
        - The export route code has been factored and specialized to allow this 
        - Should the export to a remote fail, the corresponding documents are kept locally
        - If there are no documents to send for an index, still call the documents addition with an empty payload and appropriate headers containing the expected metadata
        - If there are no documents to send for an entire remote, call the `network` route with specific headers containing the expected metadata
    3.  Wait for and process tasks from the remotes of the previous version of the network.
6. Breaking change: When importing dumps, we drop the `self` and `leader` from the network
7. Network topology change tasks can be cancelled. In this case the state will be the current one (any moved documents will stay that way). Cancellation needs to happen on all machines.

---

## This PR makes breaking changes

*Breaking changes are changes to the database such that databases created in an older version of Meilisearch need changes to remain valid in the new version of Meilisearch. This typically happens when the way to store the data changed (change of database, new required key, etc). This can also happen due to breaking changes in the API of an experimental feature. ⚠️ This kind of changes are more difficult to achieve safely, so proceed with caution and test dumpless upgrade right before merging the PR.*

- [x] Detail the changes to the DB format,
    - [x] which are compatible, and why
         - The `task.network` object is modified to add a `network_version` with a default value (hence compatible) and an `import_data` new variant (hence compatible)
         - The `network` object is modified with a new defaulted `leader` field (hence compatible)
    - [x] which are not compatible, why, and how they will be fixed up in the upgrade
        - The `network` object is modified with the `sharding` field **removed**. There is an index scheduler migration to fix it.
 - [x] /!\ Ensure all the read operations still work!
    - If the change happened in milli, you may need to check the version of the database before doing any read operation
    - If the change happened in the index-scheduler, make sure the new code can immediately read the old database
    - If the change happened in the meilisearch-auth database, reach out to the team; we don't know yet how to handle these changes
  - [x] Write the code to go from the old database to the new one
    - If the change happened in milli, the upgrade function should be written and called [here](https://github.com/meilisearch/meilisearch/blob/3fd86e8d76d7d468b0095d679adb09211ca3b6c0/crates/milli/src/update/upgrade/mod.rs#L24-L47)
    - If the change happened in the index-scheduler, we've never done it yet, but the right place to do it should be [here](https://github.com/meilisearch/meilisearch/blob/3fd86e8d76d7d468b0095d679adb09211ca3b6c0/crates/index-scheduler/src/scheduler/process_upgrade/mod.rs#L13)
  - [x] Declarative test: add a [declarative test containing a dumpless upgrade](https://github.com/meilisearch/meilisearch/blob/main/TESTING.md#typical-usage)

### DIFF
--- a/crates/index-scheduler/src/scheduler/create_batch.rs
+++ b/crates/index-scheduler/src/scheduler/create_batch.rs
@@ -745,6 +745,7 @@ impl IndexScheduler {
         mut current_batch: ProcessingBatch,
     ) -> Result<Option<(Batch, ProcessingBatch)>> {
         current_batch.processing(Some(&mut task));
+        current_batch.reason(BatchStopReason::NetworkTask { id: task.uid });
 
         let change_version =
             task.network.as_ref().map(|network| network.network_version()).unwrap_or_default();
@@ -777,11 +778,16 @@ impl IndexScheduler {
                         task_version >= change_version
                     });
 
-                let (batch, current_batch) = res?;
+                let (batch, mut current_batch) = res?;
 
                 let batch = match batch {
                     Some(batch) => {
                         let inner_batch = Box::new(batch);
+                        let inner_reason = current_batch.reason.to_string();
+                        current_batch.reason(BatchStopReason::NetworkTaskOlderTasks {
+                            id: task.uid,
+                            inner_reason,
+                        });
 
                         Batch::NetworkIndexBatch { network_task: task, inner_batch }
                     }
@@ -819,10 +825,15 @@ impl IndexScheduler {
                         task_version != change_version
                     });
 
-                let (batch, current_batch) = res?;
+                let (batch, mut current_batch) = res?;
 
                 let batch = batch.map(|batch| {
                     let inner_batch = Box::new(batch);
+                    let inner_reason = current_batch.reason.to_string();
+                    current_batch.reason(BatchStopReason::NetworkTaskImportTasks {
+                        id: task.uid,
+                        inner_reason,
+                    });
 
                     (Batch::NetworkIndexBatch { network_task: task, inner_batch }, current_batch)
                 });
--- a/crates/meilisearch-types/src/tasks/mod.rs
+++ b/crates/meilisearch-types/src/tasks/mod.rs
@@ -899,6 +899,17 @@ pub enum BatchStopReason {
     SettingsWithDocumentOperation {
         id: TaskId,
     },
+    NetworkTask {
+        id: TaskId,
+    },
+    NetworkTaskOlderTasks {
+        id: TaskId,
+        inner_reason: String,
+    },
+    NetworkTaskImportTasks {
+        id: TaskId,
+        inner_reason: String,
+    },
 }
 
 impl BatchStopReason {
@@ -985,6 +996,24 @@ impl Display for BatchStopReason {
                 write!(
                     f,
                     "stopped before task with id {id} because it is a document operation which cannot be batched with settings changes"
+                )
+            }
+            BatchStopReason::NetworkTask { id } => {
+                write!(
+                    f,
+                    "stopped after task with id {id} because it is a network topology change task"
+                )
+            }
+            BatchStopReason::NetworkTaskOlderTasks { id, inner_reason } => {
+                write!(
+                    f,
+                    "stopped after batching network task with id {id} and a batch of older tasks: {inner_reason}"
+                )
+            }
+            BatchStopReason::NetworkTaskImportTasks { id, inner_reason } => {
+                write!(
+                    f,
+                    "stopped after batching network task with id {id} and a batch of import tasks: {inner_reason}"
                 )
             }
         }


Cargo.lock
Kerollmops, line 1: It seems that there are two new versions of some dependencies now

crates/meilisearch-types/src/tasks/network.rs
```diff
@@ -0,0 +1,470 @@
+// Copyright © 2025 Meilisearch Some Rights Reserved
```
Kerollmops: Remove this

crates/meilisearch-types/src/tasks/network.rs
```diff
@@ -0,0 +1,470 @@
...
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct Origin {
+    pub remote_name: String,
+    pub task_uid: u32,
+    pub network_version: Uuid,
```
Kerollmops, line 105: We need a default here

crates/meilisearch-types/src/tasks/network.rs
```diff
@@ -0,0 +1,470 @@
...
+/// 4. Finished.
+///
+/// Furthermore, it maintains some stats
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct NetworkTopologyChange {
+    state: NetworkTopologyState,
+    // in name, `None` if the node is no longer part of the network
+    #[serde(skip_serializing_if = "Option::is_none")]
+    in_name: Option<String>,
```
Kerollmops: Can you create a Linear issue to detect aliasing because it can produce bugs.

crates/meilisearch-types/src/tasks/network.rs
```diff
@@ -0,0 +1,470 @@
...
+                total_documents: right_total_documents,
+                received_documents: right_received_documents,
+                processed_documents: right_processed_documents,
+                task_keys: mut right_task_keys,
+            },
+        ) => {
+            let total_documents = u64::max(left_total_documents, right_total_documents);
+            let received_documents = u64::max(left_received_documents, right_received_documents);
+            let processed_documents = u64::max(left_processed_documents, right_processed_documents);
+            left_task_keys.append(&mut right_task_keys);
```
Kerollmops: Consider RoaringBitmaps

Kerollmops: Consider clearing the keys once the task is finished.

crates/meilisearch-types/src/tasks/network.rs
```diff
@@ -0,0 +1,470 @@
...
+impl InRemote {
+    pub fn new() -> Self {
+        Self { import_state: ImportState::WaitingForInitialTask }
+    }
+}
+
+#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+enum ImportState {
+    /// Initially Meilisearch doesn't know how many documents it should expect from a remote.
```
Kerollmops, line 425: All tasks are doing this

crates/meilisearch-types/src/tasks/network.rs
```diff
@@ -0,0 +1,470 @@
...
+    },
+}
+
+pub enum ReceiveTaskError {
+    UnknownRemote(String),
+    DuplicateTask(DocumentId),
+}
+
+pub mod headers {
+    pub const PROXY_ORIGIN_REMOTE_HEADER: &str = "Meili-Proxy-Origin-Remote";
```
Kerollmops: You should always prefix headers by `X-`

crates/meilisearch-types/src/tasks/network/enterprise_edition.rs
```diff
@@ -0,0 +1,234 @@
...
+        remote_name: &str,
+        index_name: &str,
+        document_count: u64,
+    ) {
+        let remote = self.in_remotes.get_mut(remote_name).unwrap();
+        remote.import_state = match std::mem::take(&mut remote.import_state) {
+            ImportState::WaitingForInitialTask => panic!("no task received yet one processed"),
+            ImportState::Ongoing { mut import_index_state, total_indexes } => {
+                let (index_name, mut index_state) =
+                    import_index_state.remove_entry(index_name).unwrap();
```
Kerollmops, line 184: Consider expect

crates/index-scheduler/src/scheduler/create_batch.rs
```diff
@@ -669,19 +719,109 @@ impl IndexScheduler {
...
+
+        match network_topology_change.state() {
+            NetworkTopologyState::WaitingForOlderTasks => {
+                let res =
+                    self.create_next_batch_unprioritized(rtxn, enqueued, current_batch, |task| {
+                        let has_index = task.index_uid().is_some();
+
+                        if !has_index {
+                            return true;
+                        }
```
Kerollmops: Check boolean 👀

crates/index-scheduler/src/scheduler/create_batch.rs
```diff
@@ -669,19 +719,109 @@ impl IndexScheduler {
...
+
+                        let has_older_network_version = task
+                            .network
+                            .as_ref()
+                            .map(|network| network.network_version() < change_version)
+                            // if there is no version, we never retain the task
+                            .unwrap_or_default();
+
+                        !has_older_network_version
```
Kerollmops: Can you make it more look alike que l'autre

crates/index-scheduler/src/scheduler/create_batch.rs
```diff
@@ -669,19 +719,109 @@ impl IndexScheduler {
...
+                            .network
+                            .as_ref()
+                            .map(|network| {
+                                network.network_version() == change_version
+                                    && network.import_data().is_some()
+                            })
+                            // if there is no version, we never retain the task
+                            .unwrap_or_default();
+
+                        !is_import_task
```
Kerollmops: Review this too

crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        if let Some((remotes, out_name)) = network_topology_change.export_to_process() {
+            network_topology_change.set_moved(self.balance_documents(
+                remotes,
+                out_name,
+                network_topology_change.in_name(),
+                origin,
+                &progress,
+                &self.scheduler.must_stop_processing,
+            )?);
```
Kerollmops: Move the output of the balance_documents into a variable, please :)

crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+                    &agent,
+                    must_stop_processing,
+                );
+
+                match res {
+                    Ok(_) => {}
+                    Err(err) => {
+                        tracing::warn!("Could not signal not to wait documents to `{remote_name}` due to error: {err}");
+                    }
+                }
```
Kerollmops: Can you use a if let Err, please?

crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+            }
+            return Ok(0);
+        }
+
+        let moved_documents: Vec<u64> = self.index_mapper.try_for_each_index(
```
Kerollmops: Would be probably better to use a simple `let mut total` and increment it, instead.

crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+            &scheduler_rtxn,
+            |index_uid, index| -> crate::Result<u64> {
+                indexer_alloc.reset();
+                let err = |err| Error::from_milli(err, Some(index_uid.to_string()));
+                let index_rtxn = index.read_txn()?;
+                let all_docids = index.external_documents_ids();
+                let mut documents_to_move_to: hashbrown::HashMap<String, RoaringBitmap> =
+                    hashbrown::HashMap::new();
```
Kerollmops: Pleas use a poisson 🐟

crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+                    let documents_to_move =
+                        documents_to_move_to.remove(remote_name).unwrap_or_default();
+
+                    let target = TargetInstance {
+                        base_url: &remote.url,
+                        api_key: remote.write_api_key.as_deref(),
+                    };
+                    let options = ExportOptions {
+                        index_uid,
+                        payload_size: None,
```
Kerollmops, line 203: Are you sure you want that?

crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+                let mut mapper_wtxn = self.env.write_txn()?;
+                let stats =
+                    crate::index_mapper::IndexStats::new(index, &index_wtxn).map_err(err)?;
+                self.index_mapper.store_stats_of(&mut mapper_wtxn, index_uid, &stats)?;
+
+                index_wtxn.commit()?;
+                // update stats after committing changes to index
+                mapper_wtxn.commit()?;
+
+                Ok(moved_count)
```
Kerollmops: Can you move this to a function, please?

Cargo.lock
dureuill: I made it so hashbrown uses the same version in `index-scheduler` as in `milli` (0.15.5).

Itertools was already using the same version everywhere, but I suspect some of our dependencies pull different versions

crates/meilisearch-types/src/tasks/network.rs
dureuill: done

crates/meilisearch-types/src/tasks/network.rs
dureuill: done

crates/meilisearch-types/src/tasks/network.rs
dureuill: Done https://linear.app/meilisearch/issue/SCA-297/solve-aliasing-problem

crates/meilisearch-types/src/tasks/network.rs
dureuill: done

crates/meilisearch-types/src/tasks/network/enterprise_edition.rs
dureuill: done

crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
dureuill: done

crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
dureuill: done

crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
dureuill: done

crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
dureuill: done

crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
dureuill: done

crates/index-scheduler/src/scheduler/create_batch.rs
dureuill: done, refactored all lambdas in this file

crates/index-scheduler/src/scheduler/create_batch.rs
dureuill: done

crates/index-scheduler/src/scheduler/create_batch.rs
dureuill: done

crates/index-scheduler/src/scheduler/process_export.rs
```diff
@@ -61,260 +66,435 @@ impl IndexScheduler {
...
+            }))?;
+        }
+
+        let fields_ids_map = ctx.index.fields_ids_map(ctx.index_rtxn)?;
+        let all_fields: Vec<_> = fields_ids_map.iter().map(|(id, _)| id).collect();
+        let total_documents = ctx.universe.len() as u32;
+        let (step, progress_step) = AtomicDocumentStep::new(total_documents);
+        ctx.progress.update_progress(progress_step);
+
+        let limit = options.payload_size.map(|ps| ps.as_u64() as usize).unwrap_or(20 * 1024 * 1024);
```
Kerollmops: defaults to 20 MiB

crates/index-scheduler/src/scheduler/process_export.rs
```diff
@@ -61,260 +66,435 @@ impl IndexScheduler {
...
+            tracing::warn!("remote answered with: {message}");
+            Ok(ControlFlow::Break(()))
+        }
+        // note: there has already been many attempts to get this due to exponential backoff
+        Err(Error::FromRemoteWhenExporting { code, message, .. })
+            if code == Code::ImportTaskWithoutNetworkTask.name() =>
+        {
+            tracing::warn!("remote answered with: {message}");
+            Ok(ControlFlow::Break(()))
+        }
```
Kerollmops, line 532: Can you specify the name of the remote sending the error (the receiver of the document task)

crates/meilisearch/src/proxy/enterprise_edition.rs
```diff
@@ -0,0 +1,799 @@
...
+where
+    T: serde::Serialize,
+    F: FnMut(&str, &Remote, &mut T),
+{
+    if let DbTaskNetwork::Remotes { remote_tasks, network_version } = &mut task_network {
+        let network_version = *network_version;
+        let this = network
+            .local
+            .as_deref()
+            .expect("inconsistent `network.sharding` and `network.self`")
```
Kerollmops: It's not network sharding but leader

crates/meilisearch/src/proxy/enterprise_edition.rs
```diff
@@ -0,0 +1,799 @@
...
+        let content_type = match &body {
+            // for file bodies, force x-ndjson
+            Body::NdJsonPayload(_) => Some(b"application/x-ndjson".as_slice()),
+            // otherwise get content type from request
+            _ => req.headers().get(CONTENT_TYPE).map(|h| h.as_bytes()),
+        };
+
+        let mut in_flight_remote_queries = BTreeMap::new();
+        let client = reqwest::ClientBuilder::new()
+            .connect_timeout(std::time::Duration::from_secs(3))
```
Kerollmops: Can we have a wild env var, please?

crates/meilisearch/src/proxy/enterprise_edition.rs
```diff
@@ -0,0 +1,799 @@
...
+                tokio::spawn({
+                    let url = format!("{}{}", node.url, path_and_query);
+
+                    let url_encoded_this = urlencoding::encode(&this).into_owned();
+                    let url_encoded_task_uid = task.uid.to_string(); // it's url encoded i promize
+
+                    let content_type = content_type.map(|b| b.to_owned());
+
+                    let backoff = backoff::ExponentialBackoffBuilder::new()
+                        .with_max_elapsed_time(Some(std::time::Duration::from_secs(25)))
```
Kerollmops: An env var here, please.

crates/meilisearch/src/proxy/enterprise_edition.rs
```diff
@@ -0,0 +1,799 @@
...
+                    let task_uid = res.task_uid;
+
+                    remote_tasks.insert(node_name, Ok(task_uid).into());
+                }
+                Ok(Err(error)) => {
+                    remote_tasks.insert(node_name, Err(error.as_response_error()).into());
+                }
+                Err(panic) => match panic.try_into_panic() {
+                    Ok(panic) => {
+                        let msg = match panic.downcast_ref::<&'static str>() {
```
Kerollmops, line 294: Consider payload_as_str from the std

crates/meilisearch/src/proxy/enterprise_edition.rs
```diff
@@ -0,0 +1,799 @@
...
+        // for file bodies, force x-ndjson
+        Body::NdJsonPayload(_) => Some("application/x-ndjson".into()),
+        // otherwise get content type from request
+        _ => content_type,
+    };
+
+    let body = body.into_bytes(remote_name, remote).map_err(Box::new)?;
+
+    let client = reqwest::ClientBuilder::new()
+        .connect_timeout(std::time::Duration::from_secs(3))
```
Kerollmops: Env var, please

crates/meilisearch/src/proxy/enterprise_edition.rs
```diff
@@ -0,0 +1,799 @@
...
+        .unwrap();
+
+    let url = format!("{}{}", remote.url, path_and_query);
+
+    // send payload to remote
+    tracing::trace!(remote_name, "sending request to remote");
+    let api_key = remote.write_api_key.clone();
+
+    let backoff = backoff::ExponentialBackoffBuilder::new()
+        .with_max_elapsed_time(Some(std::time::Duration::from_secs(25)))
```
Kerollmops: Env var, please

crates/meilisearch/src/proxy/enterprise_edition.rs
```diff
@@ -0,0 +1,799 @@
...
+        let url = url.clone();
+        let client = client.clone();
+        let content_type = content_type.clone();
+
+        let body = body.clone();
+        let api_key = api_key.clone();
+        let method = method.clone();
+
+        async move {
+            let request = client.request(method, url).timeout(std::time::Duration::from_secs(30));
```
Kerollmops: Env var, please

crates/meilisearch/src/proxy/enterprise_edition.rs
```diff
@@ -0,0 +1,799 @@
...
+                        status_code = status_code.as_u16(),
+                        "remote replied with unexpected status code"
+                    );
+                }
+            }
+
+            let response: U = match parse_response(response).await {
+                Ok(response) => response,
+                Err(response) => {
+                    return Err(backoff::Error::transient(ProxyError::CouldNotParseResponse {
```
Kerollmops: Permanent here?

crates/meilisearch/src/proxy/enterprise_edition.rs
```diff
@@ -0,0 +1,799 @@
...
+            );
+        }
+    }
+
+    let response = match parse_response(response).await {
+        Ok(response) => response,
+        Err(response) => {
+            return Err(backoff::Error::transient(ProxyError::CouldNotParseResponse { response }))
+        }
+    };
```
Kerollmops: Can you factorize with the code in send_request

crates/meilisearch/src/proxy/enterprise_edition.rs
```diff
@@ -0,0 +1,799 @@
...
+                        }
+                    })?)
+                    .map_err(|err| {
+                        MeilisearchHttpError::InvalidHeaderValue {
+                            header_name: PROXY_ORIGIN_NETWORK_VERSION_HEADER,
+                            msg: format!("while URL-decoding network version: {err}"),
+                        }
+                    })?
+                }
+                None => Cow::Borrowed("0"),
```
Kerollmops: Shouldn't it be a UUID here?

crates/index-scheduler/src/scheduler/process_export.rs
```diff
@@ -61,260 +66,435 @@ impl IndexScheduler {
...
 
-            step.store(total_documents, atomic::Ordering::Relaxed);
+fn set_network_ureq_headers(
+    request: ureq::Request,
+    import_data: &ImportData,
+    origin: &Origin,
+    metadata: &ImportMetadata,
+) -> ureq::Request {
+    let request = request
+        .set(headers::PROXY_ORIGIN_REMOTE_HEADER, &origin.remote_name)
```
dureuill: url encode here please

crates/meilisearch/src/proxy/enterprise_edition.rs
```diff
@@ -0,0 +1,799 @@
...
+
+pub fn import_data_from_req(req: &HttpRequest) -> Result<Option<ImportData>, MeilisearchHttpError> {
+    let (remote_name, index_name, documents) = match (
+        req.headers().get(PROXY_IMPORT_REMOTE_HEADER),
+        req.headers().get(PROXY_IMPORT_INDEX_HEADER),
+        req.headers().get(PROXY_IMPORT_DOCS_HEADER),
+    ) {
+        (None, None, None) => return Ok(None),
+        (Some(remote_name), Some(index_name), Some(documents)) => {
+            let remote_name = urlencoding::decode(remote_name.to_str().map_err(|err| {
```
Kerollmops: Can you check that the values are url encoded everywhere

crates/meilisearch/src/routes/network/enterprise_edition.rs
```diff
@@ -0,0 +1,366 @@
...
+    // These checks are by no mean perfect (they are not atomic since the network is involved), but they should
+    // help preventing a bad situation.
+    if merged_network.leader.is_some() {
+        let query = Query {
+            statuses: Some(vec![
+                meilisearch_types::tasks::Status::Enqueued,
+                meilisearch_types::tasks::Status::Processing,
+            ]),
+            types: Some(vec![meilisearch_types::tasks::Kind::NetworkTopologyChange]),
+            ..Default::default()
```
Kerollmops, line 108: Please set a limit to something reasonable

crates/meilisearch/src/routes/network/enterprise_edition.rs
```diff
@@ -0,0 +1,366 @@
...
+                            }
+                        }
+
+                        Ok(())
+                    })
+                }),
+        )
+        .try_buffer_unordered(40)
+        .try_collect::<()>()
+        .await?;
```
Kerollmops: [try_for_each_concurrent](https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_for_each_concurrent)

crates/meilisearch/src/routes/network/enterprise_edition.rs
```diff
@@ -0,0 +1,366 @@
...
+        let mut proxied_network = Network {
+            remotes: Setting::Set(to_settings_remotes(&merged_network.remotes)),
+            local: Setting::NotSet,
+            leader: Setting::some_or_not_set(merged_network.leader.clone()),
+            previous_remotes: Setting::Set(to_settings_remotes(&old_network.remotes)),
+        };
+        let mut deleted_network = old_network;
+
+        let deleted_remotes = &mut deleted_network.remotes;
+        deleted_remotes.retain(|node, _| !merged_network.remotes.contains_key(node));
```
Kerollmops, line 259: Can you document a bit this, please

crates/meilisearch/src/routes/network/enterprise_edition.rs
```diff
@@ -0,0 +1,366 @@
...
+        let remote = remote.try_into_db_node(&name)?;
+        previous_remotes.insert(name, remote);
+    }
+
+    old_network.remotes = previous_remotes;
+
+    let new_network = DbNetwork {
+        local: merged_network.local.set(),
+        remotes,
+        leader: merged_network.leader.set(),
```
Kerollmops: Can you error if the leader is not set

crates/meilisearch/src/routes/network/enterprise_edition.rs
```diff
@@ -0,0 +1,366 @@
...
+            network_size: new_network.remotes.len(),
+            network_has_self: new_network.local.is_some(),
+        },
+        &req,
+    );
+
+    let network_topology_change = NetworkTopologyChange::new(old_network, new_network);
+    let task = KindWithContent::NetworkTopologyChange(network_topology_change);
+    let task = {
+        let index_scheduler = index_scheduler.clone();
```
Kerollmops: Remove this

crates/meilisearch/src/routes/indexes/documents.rs
```diff
@@ -1085,9 +1094,16 @@ async fn document_addition(
         index_uid: index_uid.to_string(),
     };
 
+    // FIXME: not new to #6000, but _any_ error here will cause the payload to unduly persist
```
Kerollmops, line 1097: Good first issue

crates/index-scheduler/src/scheduler/process_export.rs
dureuill: done

crates/meilisearch/src/proxy/enterprise_edition.rs
dureuill: done

crates/meilisearch/src/proxy/enterprise_edition.rs
dureuill: done

crates/meilisearch/src/proxy/enterprise_edition.rs
dureuill: Actually not applicable because it doesn't act on `Box<dyn Any + Send + 'static>` which is what I'm receiving here

crates/meilisearch/src/proxy/enterprise_edition.rs
dureuill: I kept it as an option and unwrap_or_uuid_nil later

crates/meilisearch/src/routes/network/enterprise_edition.rs
dureuill: done

crates/meilisearch/src/proxy/enterprise_edition.rs
dureuill: done

crates/meilisearch/src/routes/network/enterprise_edition.rs
dureuill: set to 1

crates/meilisearch/src/routes/network/enterprise_edition.rs
dureuill: done

crates/meilisearch/src/routes/network/enterprise_edition.rs
dureuill: done

crates/meilisearch/src/routes/network/enterprise_edition.rs
dureuill: useless clone removed

crates/index-scheduler/src/scheduler/process_export.rs
dureuill: done

crates/meilisearch/src/proxy/enterprise_edition.rs
dureuill: done

crates/meilisearch-types/src/tasks/network.rs
dureuill: done

crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
dureuill: done, I didn't change the payload limit here, rather I changed the default to use a value close from the max http payload config from the **sending instance**.

see https://github.com/meilisearch/meilisearch/pull/6000/commits/532684981d23faf57951fac7a78a82cb0e29a3ca

crates/meilisearch/src/proxy/enterprise_edition.rs
dureuill: done

crates/meilisearch/src/proxy/enterprise_edition.rs
dureuill: done

crates/meilisearch/src/proxy/enterprise_edition.rs
dureuill: done

crates/meilisearch/src/proxy/enterprise_edition.rs
dureuill: done

crates/meilisearch/src/proxy/enterprise_edition.rs
dureuill: done

crates/meilisearch-types/src/tasks/network.rs
dureuill: Using roarings as b64-encoded strings.

See ccc54b1d235f7e804a8595df178de83506fe92ba

dureuill: > Consider clearing the keys once the task is finished.

Incidentally, this already happens because the task keys are stored in the index import state `Ongoing` variant, so it is trashed when the variant becomes `Finished`

dureuill:
revisited the headers to support both legacy and new headers

crates/index-scheduler/src/scheduler/process_export.rs
dureuill: done, see https://github.com/meilisearch/meilisearch/pull/6000/commits/2ad094e95dd9486b7190657d1c3e06c827120b38

crates/meilisearch/src/routes/indexes/documents.rs
Kerollmops: Done in #6048

Kerollmops (changes requested):
- [x] The batch reason is still unspecified
- [x] The out payload size is not set correctly and seems fixed at 10MiB

dureuill:
> The out payload size is not set correctly and seems fixed at 10MiB

I checked this, and it is actually because we are using 40 request threads, meaning that each thread sees at most `1/40` of all the documents that need sending. As we send `1/remote_count` docs to each remote on average, that means that each task is at most `docs_db/40*remote_count` bytes.

even with a 50GB docs db and 5 remotes, that's merely 25MB.

dureuill:
> The batch reason is still unspecified

done, see https://github.com/meilisearch/meilisearch/pull/6000/commits/4e4d6dfe2cb6f00e626492e7147723706e3d51f7

PR #6003, opened 2025-11-19
Build x86 and ARM images on Github-hosted runners

paulden:
# Pull Request

## Switch to GitHub-hosted runners to build Docker images

To reduce time needed to build images when releasing a new Meilisearch version, we can use ARM runners instead of QEMU for ARM64 images.

Comparison:
- Using QEMU to build x86 and ARM images: **[~1h39](https://github.com/meilisearch/meilisearch/actions/runs/19519191125)**
- Using `arm64` Github-hosted runners to build ARM image and standard runners to build x86 image: **[~21min](https://github.com/meilisearch/meilisearch/actions/runs/19542511194)**

References:
- https://docs.github.com/en/actions/how-tos/write-workflows/choose-where-workflows-run/choose-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories
- https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners

## Run tests on ARM

To ensure meilisearch runs correctly on ARM, we can run the `Tests on ubuntu-22.04` on both x86 and ARM64 runners.

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -15,7 +15,7 @@ env:
 
 jobs:
   test-linux:
-    name: Tests on ubuntu-22.04
+    name: Tests on ubuntu-24.04
     runs-on: ${{ matrix.runner }}
     strategy:
       matrix:


dureuill:
@Kerollmops : we fixed the compilation of gemm on ARM with @paulden 

however, I think that rather than having ARM CI for `{clippy, rustfmt, with ollama, in debug, on Ubuntu}`, just having the ARM CI for `on Ubuntu` might be sufficient. WDYT?

Kerollmops:
@dureuill Yes, that would be enough I believe!

.github/workflows/test-suite.yml
```diff
@@ -16,7 +16,10 @@ env:
 jobs:
   test-linux:
     name: Tests on ubuntu-22.04
```
Kerollmops:
```suggestion
    name: Tests on ubuntu-24.04
```

Kerollmops (approved):
Nice work 👏

PR #6004, opened 2025-11-19
Make Hannoy the default vector store

Kerollmops:
This PR ensures that the default vector store is Hannoy for newer versions of Meilisearch. The new vector store is stable enough and much more performant than the current arroy version. Drastically reducing the indexing latency and disk use. Relevance is also drastically improved.

- [x] Revert default so that newly created indexes are created with the `experimental` (hannoy) vector store
  - [x] During dumpless upgrade, force `None` settings to `stable` so as not to corrupt existing indexes
  - [x] `None` will always mean `stable`, and new indexes are created with the `experimental` (hannoy) vector store

### DIFF
--- a/crates/meilisearch/tests/settings/get_settings.rs
+++ b/crates/meilisearch/tests/settings/get_settings.rs
@@ -197,7 +197,7 @@ test_setting_routes!(
     {
       setting: vector_store,
       update_verb: patch,
-      default_value: null
+      default_value: "experimental"
     },
 );
 


dureuill:
needs rebase

crates/meilisearch/tests/vector/binary_quantized.rs
```diff
@@ -104,8 +104,8 @@ async fn binary_quantize_before_sending_documents() {
             "manual": {
               "embeddings": [
                 [
-                  -1.0,
```
Kerollmops, line 107: The binary quantization is hamming and no longer cosine so the zeros are zeros and no longer -1s

crates/milli/src/update/upgrade/v1_29.rs
```diff
@@ -0,0 +1,29 @@
...
+
+impl UpgradeIndex for Latest_V1_28_To_V1_29_0 {
+    fn upgrade(
+        &self,
+        wtxn: &mut RwTxn,
+        index: &Index,
+        _original: (u32, u32, u32),
+        _progress: Progress,
+    ) -> Result<bool> {
+        if index.get_vector_store(wtxn)?.is_none() {
```
Kerollmops:
```suggestion
        // The database before v1.29 defaulted to using arroy, so we need
        // to set it explicitly because the new default is hannoy.
        if index.get_vector_store(wtxn)?.is_none() {
```

ManyTheFish (dismissed):
This is simpler than I was expecting. ☺️

ManyTheFish:
db-change or no-db-change?

Kerollmops:
Yes, we have considered this with @dureuill, and in fact, we can only do it this way; otherwise, it will break some search requests during the dumpless upgrade.

dureuill (approved):
perfect

PR #6013, opened 2025-11-27
Fix SDK test to use EE

curquiza:
Fix SDK tests failing because they test Sharding


curquiza:
Super weird you are right....

EDIT: normal, since it was using nightly CE and not nightly EE

PR #6018, opened 2025-11-29
Add support for XLM Roberta models

qdequele:
## Related issue
I conducted numerous tests using the Meilisearch Composite embedder at scale on Meilisearch Cloud, and the results seem promising. However, we are currently limited in the models we can embed. To complete my test on the composite embedder for Meilisearch Cloud, I would need to be able to use not only BERT or ModernBert models but also XLM-Roberta-based models. My target is to use [baai/bge-m3 ](https://huggingface.co/BAAI/bge-m3) model. 

I added the test, tried locally, and it should work. However, the test needs to load the BGE-M3 model, and that seems like an exaggeration to do it every time we need to do a test. Should we remove this from the default tests? 

This involves no changes in the DB. 

## Requirements

    - [] Test that any impacted DB still works as expected after using `--experimental-dumpless-upgrade` on a DB created with the last released Meilisearch

### DIFF
--- a/crates/meilisearch-types/src/network.rs
+++ b/crates/meilisearch-types/src/network.rs
@@ -1,5 +1,6 @@
-use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
+
+use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Default)]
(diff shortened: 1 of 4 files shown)


Kerollmops:
Hey @qdequele 👋  Would you mind rebasing on main, please? We have tweaked the CI slightly to make it faster 🚀

PR #6024, opened 2025-12-03
Limit the number of retrieved task to one in the metrics

Kerollmops:
Thanks, @dureuill, for the quick investigation.


dureuill (approved):
that looks better.

We could maybe consider having these function return iterators rather than vectors, to avoid this kind of eager allocations?

(also, might be worth checking the callers of get_batches and get_tasks for similar issues)

(also, what happens if we run a `GET /batches?limit=1000000`? we shouldn't let that cause an availability issue I guess)

PR #6028, opened 2025-12-03
Add code samples to generated openAPI file

curquiza:
- Closes https://linear.app/meilisearch/issue/ENGPROD-2179/adapt-code-samples-to-be-able-to-automate-api-reference by generating 2 openAPI files to the release assets
  - one that is the same as before, no change
  - one that is the using the newly integrated option `--with-mintlify-code-sample`
- Fix bug in `/compact` route generation
- Fix bug in `/webhook/uuid` route generation
- Add check of OpenAPI file in CI and fixes errors detected

### DIFF
--- a/.github/workflows/check-openapi-file.yml
+++ b/.github/workflows/check-openapi-file.yml
@@ -3,6 +3,9 @@ name: Check OpenAPI file
 on:
   workflow_dispatch:
   pull_request:
+  schedule:
+    # Run daily at 5 AM UTC
+    - cron: '0 5 * * *'
 
 permissions:
   contents: read


.github/workflows/check-openapi-file.yml
```diff
@@ -0,0 +1,50 @@
+name: Check OpenAPI file
+
+on:
+  workflow_dispatch:
+  pull_request:
+  merge_group:
```
Kerollmops: Can you remove it from the merge queue, please?

crates/meilisearch/src/routes/indexes/mod.rs
```diff
@@ -51,7 +51,7 @@ mod similar_analytics;
         (path = "/", api = settings::SettingsApi),
         (path = "/", api = compact::CompactApi),
     ),
-    paths(list_indexes, create_index, get_index, update_index, delete_index, get_index_stats, compact::compact),
+    paths(list_indexes, create_index, get_index, update_index, delete_index, get_index_stats),
```
Kerollmops, line 58: Can we add a test for that, please?

crates/meilisearch-types/src/settings.rs
```diff
@@ -302,7 +302,7 @@ pub struct Settings<T> {
 
     #[serde(default, skip_serializing_if = "Setting::is_not_set")]
     #[deserr(default, error = DeserrJsonError<InvalidSettingsLocalizedAttributes>)]
-    #[schema(value_type = Option<Vec<LocalizedAttributesRuleView>>, example = json!(50))]
+    #[schema(value_type = Option<Vec<LocalizedAttributesRuleView>>, example = json!(null))]
```
Kerollmops: Maybe a better example?

crates/meilisearch/src/routes/indexes/similar.rs
```diff
@@ -79,11 +79,11 @@ pub fn configure(cfg: &mut web::ServiceConfig) {
                   "release_date": 1418256000
                 }
               ],
+              "id": "143",
               "offset": 0,
               "limit": 2,
               "estimatedTotalHits": 976,
-              "processingTimeMs": 35,
-              "query": "american "
+              "processingTimeMs": 35
```
Kerollmops, line 86: Why don't we have the query?

crates/meilisearch/src/routes/batches.rs
```diff
@@ -46,10 +46,7 @@ pub fn configure(cfg: &mut web::ServiceConfig) {
         (status = OK, description = "Return the batch", body = BatchView, content_type = "application/json", example = json!(
             {
                 "uid": 1,
-                "details": {
-                    "receivedDocuments": 1,
-                    "indexedDocuments": 1
-                },
+                "details": {},
```
Kerollmops: Can we do better, here?

crates/meilisearch/src/routes/indexes/similar.rs
curquiza: I checked @Kerollmops 
The similar API uses a document id instead of a text query. Unlike regular search, similar finds documents based on vector similarity to an existing document, so the response returns the id that was used, not a query.

curquiza:
I applied the changes @Kerollmops , from `Remove merged_queue` commit

PR #6029, opened 2025-12-03
Switch to migration-oriented dumpless upgrade

dureuill:
The author recommends that the reviewer reads the PR [commit by commit](https://github.com/meilisearch/meilisearch/pull/6029/commits)

This PR performs the changes described below.

## 1. Switch to a "migration-oriented" dumpless upgrade implementation

Rather than upgrading from version to version, we now only list migrations that require changes to the DB.
A migration knows whether it must be applied depending on the **initial** version of the DB, that is, the version of the DB before performing any migration work.

After applying all applicable migrations, the version is set to the version of the Meilisearch binary.

This change was performed for both the index scheduler upgrade and the index upgrade, although the former doesn't currently have any migration.

## 2. Make snapshot independent from the binary version

Snapshot are doctored a bit to replace any mention of the binary version with the string `"[current version]"`.


ManyTheFish (approved):
LGTM!

dureuill:
> Should we care about [meilitool/src/upgrade.rs](https://github.com/meilisearch/meilisearch/blob/dumpless-upgrade-migrations/crates/meilitool/src/upgrade/mod.rs)?

I would say no.

1. < v1.12.0 is likely a very small portion of users today
2. the code there already doesn't need to change with the current version
3. the code there doesn't need to be maintainable as it is not expected to evolve.

It might be a good idea to remove it at some point, but it doesn't need to be done in this PR :-)

PR #6031, opened 2025-12-04
Url fetching embedder settings

### LINKED ISSUE #6036: URL Fetching for Embeddings
## Overview

Enable Meilisearch to fetch content from URLs during the embedding extraction process automatically. This allows users to store URLs in their documents (e.g., `imageUrl`) and have Meilisearch download the content, convert it to base64, and pass it to the embedder—without persisting the fetched content in the database.

## Problem Statement

Currently, to use image or document embeddings in Meilisearch, users must:
1. Download the media content itself
2. Convert it to base64
3. Include the base64 data in the document

This creates friction for users who already have media hosted on CDNs or external services. They want to simply provide URLs and let Meilisearch handle the fetching.

## Goals

- Allow users to specify a document field containing a URL to fetch
- Fetch content during embedding extraction (not stored in DB)
- Provide security controls (allowed domains, size limits, timeouts)
- Work with any media type (images, PDFs, etc.)
- Simple, minimal configuration for common use cases

## Non-Goals

- Text extraction from PDFs (embedder's responsibility)
- Caching fetched content across indexing sessions
- Storing fetched content in the document database
- Fetching multiple URLs per embedder (keep it simple)

### Embedder Configuration

**Minimal configuration (works out of the box):**
```json
{
  "embedders": {
    "multimodal": {
      "source": "rest",
      "url": "https://embedding-service/embed",
      "fetchUrl": {
        "input": "imageUrl",
        "output": "_image"
      },
      "request": {
        "image": "{{doc._image}}"
      }
    }
  }
}
```

**Full configuration with all options:**
```json
{
  "embedders": {
    "multimodal": {
      "source": "rest",
      "url": "https://embedding-service/embed",
      "fetchUrl": {
        "input": "imageUrl",
        "output": "_image",
        "allowedDomains": ["cdn.example.com", "*.s3.amazonaws.com"],
        "timeout": 15000,
        "maxSize": "20MB",
        "retries": 3,
        "outputFormat": "dataUri"
      },
      "request": {
        "image": "{{doc._image}}"
      }
    }
  }
}
```

#### `fetchUrl` (Object)

Specifies a document field containing a URL and the virtual field name to expose the fetched content.

| Field | Type | Required | Default | Description |
|-------|------|----------|---------|-------------|
| `input` | `string` | Yes | - | Document field containing the URL (supports nested paths like `media.imageUrl`) |
| `output` | `string` | Yes | - | Virtual field name for use in templates |
| `allowedDomains` | `string[]` | No | `["*"]` | Allowed domains. `["*"]` allows any. Supports wildcards: `["*.s3.amazonaws.com"]` |
| `timeout` | `number` | No | `10000` | Request timeout in milliseconds |
| `maxSize` | `string` | No | `"10MB"` | Maximum content size to download |
| `retries` | `number` | No | `2` | Number of retry attempts on failure |
| `outputFormat` | `string` | No | `"dataUri"` | Output format: `"dataUri"` (with MIME prefix) or `"base64"` (raw) |

### Behavior

1. **During document indexing**: When extracting vectors, Meilisearch checks if the embedder has `fetchUrl` configured
2. **URL resolution**: Read the URL from the `input` field (supports nested paths)
3. **Domain validation**: Check URL domain against `allowedDomains`
4. **Content fetching**: Download content with configured timeout/retries
5. **Size validation**: Check content size against `maxSize`
6. **Encoding**: Convert fetched content to base64 or data URI based on `outputFormat`
7. **Template injection**: Make content available as `output` field in template context
8. **No persistence**: Fetched content is discarded after embedding extraction

### Virtual Fields

Virtual fields (e.g., `_image`) are:
- Only available during template rendering for that embedder
- Not stored in the document database
- Not returned in search results
- Prefixed with `_` by convention (not enforced)

### Document Example

**Stored document:**
```json
{
  "id": 1,
  "title": "Product Photo",
  "imageUrl": "https://cdn.example.com/product.jpg"
}
```

**What embedder template sees:**
```json
{
  "id": 1,
  "title": "Product Photo",
  "imageUrl": "https://cdn.example.com/product.jpg",
  "_image": "data:image/jpeg;base64,/9j/4AAQSkZJRg..."
}
```

**What is persisted:** Only original document (URL as string)

### SSRF Protection

- `allowedDomains` defaults to `["*"]` for ease of use
- For production, users should restrict to specific domains
- Wildcard support for subdomains: `["*.example.com"]`
- Private IP ranges should be blocked by default (localhost, 10.x.x.x, 192.168.x.x, etc.)

### Resource Limits

- `maxSize` prevents downloading excessively large files
- `timeout` prevents hanging on slow/unresponsive servers
- Consider global rate limiting across all URL fetches

### Content Validation

- Validate Content-Type header matches expected media types (optional future enhancement)
- Reject responses with suspicious content

## Error Handling

| Scenario | Behavior |
|----------|----------|
| URL field is null/missing | Skip fetch, `output` field is not set |
| URL field is empty string | Skip fetch, `output` field is not set |
| Domain not in allowedDomains | Error logged, document embedding fails |
| Timeout exceeded | Retry up to `retries` times, then fail |
| Content exceeds maxSize | Error logged, document embedding fails |
| HTTP error (4xx, 5xx) | Retry on 5xx, fail on 4xx |
| Network error | Retry up to `retries` times, then fail |

When a fetch fails, the document's embedding for that embedder fails, but other embedders and other documents continue processing.

## Future Enhancements

- **Content-Type validation**: Restrict to specific MIME types
- **Caching**: Cache fetched content during a single indexing batch
- **Async fetching**: Parallel URL fetches for better performance
- **Multiple URLs**: Support an array of fetchUrl mappings if needed
- **Text extraction**: Built-in PDF/DOCX text extraction option

## Open Questions

1. Should we support authentication headers per-URL (e.g., signed URLs)?
2. Should private IP blocking be configurable or always enforced?
3. Should we add metrics/logging for fetch operations?
# TODO

### Are you modifying a database?

- [ ] If not, add the `no db change` label to your PR, and you're good to merge.
- [ ] If yes, add the `db change` label to your PR. You'll receive a message explaining you what to do.

### Reminders when adding features

- [ ] Write unit tests using insta
- [ ] Write declarative integration tests in [workloads/tests](https://github.com/meilisearch/meilisearch/tree/main/workloads/test). Specify the routes to call and then call `cargo xtask test workloads/tests/YOUR_TEST.json --update-responses` so that responses are automatically filled.

- [ ] Update the openAPI file with utoipa:
  - [ ] If a new module has been introduced, create a new structure deriving [the OpenAPI proc-macro](https://docs.rs/utoipa/latest/utoipa/derive.OpenApi.html) and nest it in the main [openAPI structure](https://github.com/meilisearch/meilisearch/blob/f2185438eed60fa32d25b15480c5ee064f6fba4a/crates/meilisearch/src/routes/mod.rs#L64-L78).
  - [ ] If a new route has been introduced, add the [path decorator](https://docs.rs/utoipa/latest/utoipa/attr.path.html) to it and add the route at the top of the file in its openAPI structure.
  - [ ] If a structure which is deserialized or serialized in the API has been introduced or modified, it must derive the [`schema`](https://docs.rs/utoipa/latest/utoipa/macro.schema.html) or the [`IntoParams`](https://docs.rs/utoipa/latest/utoipa/derive.IntoParams.html) proc-macro.
  - [ ] Once everything is done, start Meilisearch with the swagger flag: `cargo run --features swagger`, open `http://localhost:7700/scalar` on your browser, and ensure everything works as expected.

qdequele:
## Related issue

Fixes #6036

## Requirements


qdequele:
The biggest question for me here is that I created a comprehensive structure for DocumentsVirtualFields to manage the added data in the documents. However, I'm unsure if that was the best approach or if it duplicates another logic that already handles this.

qdequele:
Thanks for your comment. My PoV on the different items. 

> By using your current implementation and way to download content, every time a change is performed on any field of a document, it will force a redownload of the image (and possibly a reembedding of the same). It would be great to store a checksum of the image to determine if any indexing is required. We can examine what CDNs are capable of revealing if there have been any recent changes.

100% agree, it's not best what has been done so far. 

> There are too many parameters in the API, and we will reduce the API surface for now. We can add more parameters later on when the API is used.

Parameters are needed for security and making sure people don't do anything stupid with it. If they go with the default options it's easy. 

> We already thought about this feature previously and had the idea to use liquid filters, but we need to think about that.

I found this idea not optimal because it makes this feature limited in terms of options and almost impossible to improve over time. It adds new capabilities to a language that is well known, meaning that the syntax highlighter will not recognize it and the LLM will not understand it.

> We noticed that the mime type is computed based on the file extension in the URL. It's not great to do that based on our experience, and we would prefer to discuss this behavior together. Guessing the mime type based on the actual content would be better.

No in fact it's based on the magic hash on the base64 that can be identified. I could confirm that I did not changed that. 

> It would be great to make sure that an error when fetching an image does not block the indexation and instead generates a template error (ignored when using a specific env var). More generally, how is the engine reacting when the fetches are failing?

Indeed I think I'm missing some test coverage on this part. 

> This type of PR is exciting, and it's really cool to work on, but it is not a priority and creates a lot of unowned code to maintain.

Yes, I agree and fully understand that this feature might have a lot of interconnections that could lead to a future issue. I still believe that the API design is not the best, but I strongly think that creating an extension to Liquid is not a good idea. At least we can close this PR and use it as an ideation of the API. <br><br><br>Thanks for the review and comment!

PR #6035, opened 2025-12-04
added documentation

### LINKED ISSUE #3206: Major lack of documentation in codebase
**Describe the bug**
The codebase significantly lacks documentation. As a newly introduced developer, it makes it hard for me to navigate the codebase and quickly identify which item does what. The documentation is needed as a source of intuition for evaluating whether some code item is valuable for the end-user, and whether it should be modified, and how exactly it should be modified to accomplish the desired behavior. 

Without the said documentation, the codebase can easily become a jungle of functions/structs/enums/traits whose names only suffice for achieving a ballpark of correct understanding rather than the exact understanding required to modify the correct code in the correct way. 

Documenting the codebase may constitute a significant upfront investment, but I believe it will pay off in the long run. 

**To Reproduce**
1. `cargo doc --open --document-private-items` 

**Expected behavior**
Each code item should have documentation that explains how the given code item relates to the greater context from the end-user's point of view.

clairem03:
Related issue

 Fix:

   This PR adds comprehensive documentation to several undocumented functions and modules in the Meilisearch codebase.

Changes Made:

   - Added function documentation to `search_queue.rs` including SearchQueue struct and methods
   - Documented API endpoint handlers in `routes/indexes/documents.rs`
   - Added module-level documentation to `error.rs` explaining error handling
   - Included parameter descriptions, return values, and usage examples
   - Followed Rust documentation conventions with `///` and `//!` comments

   Closes #3206

### DIFF
--- a/crates/meilisearch/src/routes/indexes/documents.rs
+++ b/crates/meilisearch/src/routes/indexes/documents.rs
@@ -233,6 +233,26 @@ impl<Method: AggregateMethod> Aggregate for DocumentsFetchAggregator<Method> {
         )),
     )
 )]
+/// added documentation
+/// Retrieves a single document by ID from the specified index.
+///
+/// This handles GET requests to `/indexes/{index_uid}/documents/{document_id}`
+///
+/// # Arguments
+///
+/// * `index_scheduler` - The index scheduler with read permissions
+/// * `document_param` - Contains the index UID and document ID
+/// * `params` - Optional query parameters for field selection
+///
+/// # Returns
+///
+/// * `200 OK` with the document as JSON
+/// * `404 Not Found` if the document doesn't exist
+///
+///
+/// ```
+/// GET /indexes/movies/documents/123
+/// ```
 pub async fn get_document(
     index_scheduler: GuardedData<ActionPolicy<{ actions::DOCUMENTS_GET }>, Data<IndexScheduler>>,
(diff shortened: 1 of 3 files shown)


vidit-virmani:
I think the PR is adding some obvious stuff as a part of documentation.  We should keep text minimal and let code speak for itself, ensuring any anamolies / special cases are specifically highlighted - in case it gets missed. 

Following the guidelines mentioned by the team.
please see https://github.com/meilisearch/meilisearch/issues/3206#issuecomment-1341132046

Specifically check this one 
> 
> Even when documenting public items, it is important to not systematically document what can be inferred by the convention or idioms of the language and to focus on what cannot be inferred without reading the implementation (returned errors, validity range for parameters, preconditions of a function, etc.).

crates/meilisearch/src/routes/indexes/documents.rs
```diff
@@ -233,6 +233,26 @@ impl<Method: AggregateMethod> Aggregate for DocumentsFetchAggregator<Method> {
         )),
     )
 )]
+/// added documentation
```
vidit-virmani, line 236: this should not be explicitly mentioned.

Kerollmops:
Thank you for your help @clairem03 but this PR is very far behind and we are working on this through a very complete OpenAPI file. Closing this.

PR #6041, opened 2025-12-09
Remove risk of command injection

curquiza:
Reported by Bastion

> Using variable interpolation `${{...}}` with `github` context data in a `run:` step could allow an attacker to inject their own code into the runner. This would allow them to steal secrets and code. `github` context data can have arbitrary user input and should be treated as untrusted. Instead, use an intermediate environment variable with `env:` to store the data and use the environment variable in the `run:` script. Be sure to use double-quotes the environment variable, like this: "$ENVVAR".

This PR fixes this alert

Proof that is running correctly: https://github.com/meilisearch/meilisearch/actions/runs/20070153806

### DIFF
--- a/.github/workflows/sdks-tests.yml
+++ b/.github/workflows/sdks-tests.yml
@@ -25,14 +25,18 @@ jobs:
       - uses: actions/checkout@v5
       - name: Define the Docker image we need to use
         id: define-image
+        env:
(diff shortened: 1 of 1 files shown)


Kerollmops (approved):
Ho! Ok! Nice, we just have to put this in and env block 👌 
Thanks!

PR #6042, opened 2025-12-10
Check LMDB-patch with mutexes works well

Kerollmops:
This PR verifies that LMDB [with mutexes to manage nested read transactions works with Meilisearch](https://bugs.openldap.org/show_bug.cgi?id=10395).


Kerollmops:
Outdated PR.

PR #6045, opened 2025-12-11
Fix MDB_MAP_FULL error handling to allow proper database resizing

### LINKED ISSUE #6038: Engine do not resize the index
**Describe the bug**

```json
"error": {
    "code": "internal",
    "type": "internal",
  }
```

**Expected behavior**

**Meilisearch version:**
v1.27 and previous ones.

AtharvGangwar48:
## What does this PR do?

Fixes #6038 - MDB_MAP_FULL errors are now properly handled to allow database resizing instead of being converted to internal errors.

## Problem

When Meilisearch encounters an `MDB_MAP_FULL` error during indexing, the error was being caught and converted to an `InternalError::StorePut` before it could be properly handled by the existing error conversion logic. This prevented the system from recognizing it as a `MaxDatabaseSizeReached` user error and automatically resizing the database.

## Solution

Modified the error handling in `write_from_bbqueue()` and `write_to_db()` functions to:
- Check if the error is specifically `heed::Error::Mdb(heed::MdbError::MapFull)`
- If it is, allow it to bubble up using `Error::from(error)` for proper conversion
- Otherwise, continue with existing `InternalError::StorePut` wrapping for other errors

## Impact

- `MDB_MAP_FULL` errors will now be properly converted to `MaxDatabaseSizeReached` user errors
- The database resizing mechanism can function as intended
- Other database errors continue to be handled as internal errors
- No breaking changes to the API

## Files Changed
- `crates/milli/src/update/new/indexer/write.rs`


Kerollmops (changes requested):
Hey @AtharvGangwar48 👋

Thank you for the PR. However, I gave more guidance in [this comment](https://github.com/meilisearch/meilisearch/issues/6038#issuecomment-3640760992). Can you take a look at it, please? This way we can take this bug as an opportunity to improve the codebase.

AtharvGangwar48:
I have successfully implemented the fix for the **MDB_MAP_FULL** error handling issue in Meilisearch.  
Here’s a summary of what was accomplished:

---

## Key Changes Made

### 1. Reduced Index Map Size  
- Updated `options.rs` to reduce the index map size **from 2 TiB to 1 GiB**  
- This makes it much easier to reproduce `MDB_MAP_FULL` during testing.

### 2. Added a Factorized Error Handling Function  
- Implemented in `error.rs`  
- This function **ignores `MdbMapFull` errors** and lets them pass through naturally  
- Prevents premature conversion into `StorePut` / `StoreDelete` errors.

### 3. Updated Error Handling in Five Locations  
All updated in `channel.rs` to use the new factorized function:

- `write_key_value()`
- `delete_entry()`
- `WordDocidsSender::write()`
- `FacetDocidsSender::write()`
- `GeoSender::set_geo_faceted()`

---

## How the Fix Works

### **Before**
`MDB_MAP_FULL` errors were caught too early and wrapped into `InternalError::StorePut`,  
which **blocked the system from recognizing database-size issues** and prevented  
automatic resizing.

### **After**
`MDB_MAP_FULL` errors now flow through to the existing logic and are correctly  
converted into:

**`UserError::MaxDatabaseSizeReached`**

This allows Meilisearch to **automatically resize the database** and continue indexing smoothly.

---

##  Validation

The validation script confirms that everything works as intended:

- ✓ `INDEX_SIZE` reduced to **1 GiB**
- ✓ Factorized error-handling function added
- ✓ `MdbMapFull` errors are now properly ignored during early handling
- ✓ All **five** error-handling points updated in `channel.rs`

---

## Result

Meilisearch can now correctly detect when the database hits its size limit,  
trigger automatic resizing, and continue indexing without failure.

This update ensures more robust and predictable behavior during large indexing operations.

Kerollmops:
Thanks for all the work @AtharvGangwar48, but this has been recently implemented in https://github.com/meilisearch/meilisearch/pull/6095.

PR #6046, opened 2025-12-11
Bump hannoy to v0.1.1

Kerollmops:
This PR bumps hannoy to v0.1.1 to fix a bug that improves the recall. This was a regression from v0.1.0 and is fixed now. We don't have to reindex the HNSW disk-based graph, but the recall is impacted when using hannoy v0.1.0.

EDIT: Converted to draft as we [may have spotted a bug](https://github.com/nnethercott/hannoy/blob/e7fda85e366ce8556db30781951a2bda3718bdf3/src/hnsw.rs#L453-L454) on Hannoy v0.1.1 👀 Waiting for @nnethercott's answer.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2698,9 +2698,9 @@ dependencies = [
 
 [[package]]
 name = "hannoy"
-version = "0.1.0-nested-rtxns"
+version = "0.1.1-nested-rtxns"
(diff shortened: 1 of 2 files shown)


Kerollmops:
Superseeded by #6055

PR #6053, opened 2025-12-15
Allow strict document update without creating missing documents

YoEight:
## Summary

This PR adds a `skipCreation` query parameter to the document addition endpoints (POST and PUT on `/indexes/{index}/documents`). When set to `true`, documents that don't already exist in the index are silently ignored rather than created.

## Motivation

Currently, both document addition methods (POST and PUT) create documents that don't exist. There's no way to update a document only if it's already present in the index. This parameter enables workflows where document operations should be applied conditionally, such as systems that need consistent outcomes regardless of operation ordering.

## API changes

Both `POST` and `PUT` on `/indexes/{index}/documents` now accept an optional `skipCreation` boolean query parameter:

```
PUT /indexes/{index_uid}/documents?skipCreation=true
POST /indexes/{index_uid}/documents?skipCreation=true
```

Default is `false`, preserving existing behavior.

### Changelog description

Adds an optional `skipCreation` boolean query parameter to POST and PUT on `/indexes/{index}/documents`. When set to `true`, documents that don't exist in the index are silently ignored rather than created. Default is `false`, preserving existing behavior.


crates/benchmarks/benches/indexing.rs
```diff
@@ -146,7 +146,7 @@ fn indexing_songs_default(c: &mut Criterion) {
 
                 let mut indexer = indexer::DocumentOperation::new();
                 let documents = utils::documents_from(datasets_paths::SMOL_SONGS, "csv");
-                indexer.replace_documents(&documents).unwrap();
+                indexer.replace_documents(&documents, Default::default()).unwrap();
```
Kerollmops: I would prefer that we use the actual type instead of `Default` so that the parameter is more obvious.

crates/meilisearch-types/src/tasks/mod.rs
```diff
@@ -118,6 +118,8 @@ pub enum KindWithContent {
         content_file: Uuid,
         documents_count: u64,
         allow_index_creation: bool,
+        #[serde(default)]
+        on_missing_document: MissingDocumentPolicy,
```
Kerollmops, line 122: Ok, that's what make this change retro compatible 👌

crates/milli/src/update/index_documents/mod.rs
```diff
@@ -67,6 +67,21 @@ pub enum IndexDocumentsMethod {
...
+///
+/// This policy is checked when processing a document whose ID is not found in the index.
+/// It applies to both update and replace operations.
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum MissingDocumentPolicy {
+    /// Create the document if it doesn't exist. This is the default behavior.
+    #[default]
+    Create,
+
+    /// Skip the document silently if it doesn't exist. No error is returned, the document is simply
```
Kerollmops, line 80: Nice! Have you checked that count of indexed documents is correct in the task details? Have you checked that batches of multiple tasks are correctly handled (a mix of skipCreation options, a document just created in a task in the same batch -> not visible through the rtxn)?

crates/milli/src/update/new/indexer/document_operation.rs
```diff
@@ -271,13 +286,21 @@ fn extract_addition_payload_changes<'r, 'pl: 'r>(
 
                             match method {
                                 ReplaceDocuments => {
+                                    if matches!(on_missing_document, MissingDocumentPolicy::Skip) {
+                                        continue;
+                                    }
+
```
Kerollmops: Can we move this just above the `match method` part and even below the `Entry::Vacant` part, please? This way, we don't generate a new internal document ID for a document we don't want.

crates/milli/src/update/new/indexer/document_operation.rs
```diff
@@ -297,13 +320,25 @@ fn extract_addition_payload_changes<'r, 'pl: 'r>(
                 },
                 Entry::Vacant(entry) => match method {
                     ReplaceDocuments => {
+                        if payload_operations.is_new
+                            && matches!(on_missing_document, MissingDocumentPolicy::Skip)
+                        {
+                            continue;
+                        }
```
Kerollmops: Can we also move this part before the `match method`, so that we have it only once in the `Entry::Vacant` branch, please?

Kerollmops (changes requested):
Looks good to me apart from some minor improvements 💪👌

crates/milli/src/update/new/indexer/document_operation.rs
Kerollmops, line 518: I think you should work on this part to ensure Meilisearch ignores updates with the `skipCreation` parameter set to true. As we can merge document deletions, replacements, and updates, we must make sure that if the last change to a document is a deletion and the following changes are replacements or an update with `skipCreation = true`, we must skip them until we get a non-`skipCreation=true` update.

crates/benchmarks/benches/indexing.rs
YoEight: Done!

crates/milli/src/update/new/indexer/document_operation.rs
YoEight: Done

crates/milli/src/update/new/indexer/document_operation.rs
YoEight: Done

crates/milli/src/update/new/indexer/document_operation.rs
YoEight: Nice catch!

I updated the code accordingly, we need to discuss a proper way to cover that use case

crates/milli/src/update/new/indexer/document_operation.rs
```diff
@@ -483,12 +522,38 @@ impl<'pl> PayloadOperations<'pl> {
...
+        on_missing_document: MissingDocumentPolicy,
+    ) {
+        // If the last operation was a deletion, we can safely ignore the update if the policy is
+        // set on skip.
+        if matches!(
+            (self.operations.last(), on_missing_document),
+            (Some(InnerDocOp::Deletion), MissingDocumentPolicy::Skip)
+        ) {
+            return;
+        }
```
Kerollmops, line 555: Looks good to me 👍

crates/milli/src/update/index_documents/mod.rs
YoEight: Yes, I added the appropriate test

YoEight:
@Kerollmops I added the batch related test as discussed.

crates/index-scheduler/src/queue/batches_test.rs
```diff
@@ -487,3 +490,41 @@ fn query_batches_canceled_by() {
...
+            false,
+        )
+        .unwrap();
+    let kind = replace_document_import_task_with_opts(
+        "docs",
+        Some("id"),
+        2,
+        documents_count1,
+        MissingDocumentPolicy::Skip,
+    );
```
Kerollmops, line 524: Can you explain to me what this test does? I thought it was deleting the document with id 2 and updating doc 2 with a `skipCreation=true`. But it seems to me that it is updating document 1 instead. Am I right?

YoEight: It's about 2 documents having the same content:

```json
{
    "id": 1
}
```

They only have different file ids but share the same document id. The test makes sure that when a deletion and a replace document requests are in the same batch, that all update requests that comes after the deletion request are ignored if their `MissingDocumentPolicy` is set to `Skip`

bpolaszek:
Dreamt of this barely a few days ago!! Kudos @YoEight 🫶

PR #6057, opened 2025-12-17
Make the S3-streaming snapshots an Enterprise Edition feature

Kerollmops:
This PR makes the S3-streaming snapshot available only on the enterprise edition. Please note that I would like to merge this into v1.31 next year, rather than into the soon-to-be-released v1.30.1.

### Changelog

The recently introduced S3-streaming snapshots feature is now available as an Enterprise Edition feature. From now on, you'll need a license to use this feature when using Meilisearch as a self-hosted solution. Note that this version introduces a **breaking change** regarding the S3-streaming snapshot feature, which is no longer available in the Community Edition; however, on-disk snapshots remain accessible. You can read our BUSL license for contact information, if you like. Note that if you are using the Community Edition of Meilisearch between [version 1.25](https://github.com/meilisearch/meilisearch/releases/tag/v1.25.0) and this release, you can freely use the S3 Streaming feature without an Enterprise Edition license.

### DIFF
--- a/crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
+++ b/crates/index-scheduler/src/scheduler/enterprise_edition/mod.rs
@@ -3,31 +3,23 @@
 // Use of this source code is governed by the Business Source License 1.1,
 // as found in the LICENSE-EE file or at <https://mariadb.com/bsl11>
 
-use std::env::VarError;
-use std::ffi::OsStr;
-use std::fs;
-use std::path::{Path, PathBuf};
-use std::sync::atomic::Ordering;
 use std::time::Duration;
 
 use bumpalo::Bump;
 use roaring::RoaringBitmap;
 
 use meilisearch_types::milli::documents::PrimaryKey;
-use meilisearch_types::milli::progress::{EmbedderStats, Progress, VariableNameStep};
+use meilisearch_types::milli::progress::{EmbedderStats, Progress};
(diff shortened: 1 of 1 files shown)


dureuill:
Hello, code that is available only in the enterprise edition should:

1. Be moved to an `enterprise_edition` submodule.
2. Include the appropriate BUSL header
3. Have an accompanying `community_edition` submodule that implements stubs of the enterprise functionality.
4. Have a module alias named `current_edition` , set to the appropriate edition depending on compilation options.

See https://github.com/meilisearch/meilisearch/blob/main/crates/meilisearch/src/proxy/mod.rs#L1 for an example.

dureuill:
Nice to have included a changelog entry.

I think it should be clearer about the fact that this is a breaking change:



> ### Changelog
> 
> Breaking change: The recently introduced S3-streaming snapshots feature is now available as an Enterprise Edition feature. From now on, you'll need a license to use this feature when using Meilisearch as a self-hosted solution, and passing the related flags in the Community Edition will cause the binary to exit in error. You can read our BUSL license for contact information, if you like. Note that if you are using the Community Edition of Meilisearch between [version 1.25](https://github.com/meilisearch/meilisearch/releases/tag/v1.25.0) and version v1.30, you can freely use the S3 Streaming feature without an Enterprise Edition license.

dureuill (approved):
code will perform as designed

PR #6059, opened 2025-12-17
Fix task attribution during index swap to prevent cross-index task loss

YoEight:
### Context

We were investigating reports of tasks being incorrectly removed after an index swap followed by an index deletion, we traced the issue back to how tasks were reassigned during swaps.

### Root cause

Given two indexes A and B, the index swap implementation only reassigned tasks in one direction, causing A’s tasks to be moved to B while B’s tasks were left unchanged.

### Fix

This PR corrects the swap behavior so that task attribution is symetrical. A's tasks move to B while B's tasks move to A.

### DIFF
--- a/crates/meilisearch/tests/swap_indexes/mod.rs
+++ b/crates/meilisearch/tests/swap_indexes/mod.rs
@@ -103,7 +103,7 @@ async fn swap_indexes() {
         {
           "uid": 1,
           "batchUid": 1,
-          "indexUid": "b",
+          "indexUid": "a",
           "status": "succeeded",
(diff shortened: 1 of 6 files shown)


crates/meilisearch/tests/swap_indexes/mod.rs
Kerollmops, line 75: Can you update the associated comments, please?

YoEight: Why? it's still true 😄

Kerollmops (approved):

PR #6060, opened 2025-12-18
fix(index-scheduler): align snapshot map_size to page size on Windows

### LINKED ISSUE #6051: Cannot take snapshots on Windows due to map_size and page_size values
**Describe the bug**
Index scheduler fails to initialize during snapshot creation on Windows, leading to an internal error and the inability to create snapshots.

**To Reproduce**
1. Run Meilisearch on Windows (x86) with default options
2. Attempt to create snapshot
3. Snapshot fails with the following error:

```
Batch failed while doing: {"steps":[{"currentStep":"processing tasks","finished":0,"total":2},{"currentStep":"snapshot the index scheduler","finished":1,"total":6}],"percentage":8.333334}

Batch failed map size (1310824018739) must be a multiple of the system page size (4096)
```

**Expected behavior**
The snapshot succeeds

**Screenshots**

**Meilisearch version:**
Reproduced on v1.27.0 and v1.29.0

**Additional context**
We have done analysis of the code where the error originates (found via trace logging) and we believe there is a missing call to `clamp_to_page_size` [here](https://github.com/meilisearch/meilisearch/blob/26e368b1161ff02c3b069239993e61b27bbad03a/crates/index-scheduler/src/lib.rs#L429)

We see that the DEFAULT_BUDGET on Windows is 6TiB:
```
const DEFAULT_BUDGET: usize = 6 * 1024 * 1024 * 1024 * 1024; // 6 TiB, 1 index
```

Naturally, this properly fits into the default page size of 4KiB (4096) for x86-64 machines: 

```
6597069766656÷4096 = 1610612736
```

Given the following trace logs, we can further deduce what happens in the function to make things go awry:
```
memmap budget: 6597069766656B
index budget: 3277060046848B
1 index of 1310824018739B can be opened simultaneously.
```

With the default `task_db_size` of 20GiB [here](https://github.com/meilisearch/meilisearch/blob/main/crates/meilisearch/src/option.rs#L117) and a default `base_map_size` set to the max index size (2TiB) [here](https://github.com/meilisearch/meilisearch/blob/26e368b1161ff02c3b069239993e61b27bbad03a/crates/meilisearch/src/lib.rs#L222) we can trace the calculations:

```
let mut budget = budget / 2;  # 3298534883328
...

budget -= task_db_size;  # 3277060046848 (value from trace log)
...
let mut index_count = budget / base_map_size;   # 1.490234375 < 2 -> enter the conditional
...
let map_size = (budget * 2) / 5;  # 1310824018739.2 floored to the value from the trace
```

This gets returned as the map_size and instantly fails when passed off to heed: `1310824018739 mod 4096 = 819`

If we clamped the map_size to the page size before returning, we believe the issue would be fixed.

Adnanebds:
## Related issue
Fixes #6051

## Description
On Windows, memory mapping requires the file size to be a multiple of the system page size (typically 4096 bytes).
Previously, the `map_size` calculation for snapshots could result in unaligned values (e.g., `1310824018739`), causing `batch failed` errors during snapshot creation.

This PR applies `clamp_to_page_size` to the budget calculation to ensure alignment.

## Requirements

## How Has This Been Tested?
- Reproduced the crash on Windows 11 by triggering a snapshot on a running instance.
- Applied the fix and verified that `POST /snapshots` now returns `202 Accepted` and creates the snapshot successfully without error logs.

### DIFF
--- a/crates/index-scheduler/src/lib.rs
+++ b/crates/index-scheduler/src/lib.rs
@@ -437,7 +437,7 @@ impl IndexScheduler {
         let mut index_count = budget / base_map_size;
         if index_count < 2 {
             // take a bit less than half than the budget to make sure we can always afford to open an index
-            let map_size = (budget * 2) / 5;
+            let map_size = clamp_to_page_size((budget * 2) / 5);
             // single index of max budget
             tracing::debug!("1 index of {map_size}B can be opened simultaneously.");
             return IndexBudget { map_size, index_count: 1, task_db_size };


Adnanebds:
Hey @Kerollmops 

Great thanks!

EclipseAditya:
Hi @Kerollmops ,
Tested this change on my Windows machine, Snapshot creation works without any issues now.

Kerollmops:
Hey @EclipseAditya 👋

That's really nice! Would you mind rebasing on main to fix the tests, please? This way we can merge!

Have a nice and of the year 🌵

EclipseAditya:
Hi @Kerollmops 
should I,
1. Create a new PR with the same fix (since I can't push to @Adnanebds branch), or
2. should we Wait for @Adnanebds to rebase the PR?

Let me know what works best.

Kerollmops:
Hey @EclipseAditya 👋

Thanks for your help and sorry for the (very) late reply. You can create a new PR for your fixes 👍 

Thank you for your help,
Have a good day 🌵

chrisbenincasa:
Any updates here?

PR #6061, opened 2025-12-18
Render auth optional on the metrics endpoint

### LINKED ISSUE #5917: remove auth from metrics endpoint; use different port
**Describe the bug**
For unattended kubernetes deployments its way simpler to create a ServiceMonitor (prometheus scrap target) and point to an (k8s internal; not exposed) container port then adding API tokens to the scraper.

**To Reproduce**
-

**Expected behavior**
the metrics endpoint should not require any auth and the port should be moved away from then api.
at least configurable.

**Meilisearch version:**
v1.22.0

**Additional context**
-

YoEight:
Fixes #5917

### DIFF
--- a/crates/index-scheduler/src/features.rs
+++ b/crates/index-scheduler/src/features.rs
@@ -27,7 +27,7 @@ pub(crate) struct FeatureData {
     network: Arc<RwLock<Network>>,
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Default)]
 pub struct RoFeatures {
     runtime: RuntimeTogglableFeatures,
 }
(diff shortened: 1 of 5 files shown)


dureuill (changes requested):
Hello 👋 

I do not understand what `OptionallyGuardedData` does. I think that, since this is authentication-related, it should be clearly documented.

From reading the code and some testing it appears as if it allows skipping authentication on the `metrics` route when no api key is provided, while still checking api keys when one is provided.

If so, and while I don't believe it is explicitly covered by our stability policy, I think that removing a security property some users might be relying on (the `/metrics` route is authenticated) is a breaking change. I'm adding the associated label.

Now, metrics is an experimental route, so breaking changes are allowed. However I don't find the use case in the original issue to be particularly compelling. I will answer there.

Also, if we are to make the authentication optional, I think we should do like we do for other publicly accessible routes, such as `/health`: remove all authentication altogether.
The mix of "checking if the provided key is valid, but accepting when there are no keys" might be surprising to some users, especially as the behavior is inconsistent with other public routes.

PR #6064, opened 2025-12-18
Hotfix: log a warning when FieldidsWeightsMap is missing entry

ManyTheFish:
- Replace the internal error returning by a warning log and avoid the search to crash due to a database corruption.
- Update Meilisearch to v1.30.1


dureuill (approved):
🏳️‍🌈 looks good 👍

PR #6072, opened 2025-12-22
Log Search performance trace

ManyTheFish:
## Summary

This PR introduces comprehensive progress tracking and logging for search operations in Meilisearch. It adds detailed timing information for each step of the search process, enabling better observability and performance analysis.

related to: https://linear.app/meilisearch/issue/ENGPROD-2184/add-bucket-sort-details-to-debug-search-queries

## Motivation

Tracking search progress helps with:
- **Performance debugging**: Identifying bottlenecks in search operations
- **Observability**: Understanding which parts of the search process take the most time
- **Monitoring**: Providing insights into search performance in production environments

### 1. Search Progress Steps (`crates/milli/src/search/steps.rs`)

Added new progress step enums to track different phases of the search process:

- **`SearchStep`**: High-level search operations
  - `Tokenize`: Query tokenization
  - `Embed`: Embedding generation
  - `Filter`: Filter evaluation
  - `ResolveUniverse`: Universe resolution
  - `KeywordSearch`: Keyword-based search
  - `PlaceholderSearch`: Placeholder search
  - `SemanticSearch`: Semantic/vector search
  - `Format`: Result formatting
  - `FacetDistribution`: Facet computation
  - `Federation`: Federated search operations
  - `Personalization`: Personalization steps

- **`ComputingBucketSortStep`**: Bucket sort algorithm steps
  - `MergeCandidates`, `Distinct`, `Words`, `Typo`, `Proximity`, `AttributePosition`, `WordPosition`, `Exactness`, `Sort`, `GeoSort`, `VectorSort`, `Asc`, `Desc`

- **`RankingRuleStep`**: Ranking rule iteration steps
  - `StartIteration`, `NextBucket`, `NonBlockingNextBucket`

- **`FederatingResultsStep`**: Federated search result merging
  - `WaitForRemoteResults`, `MergeFacets`, `MergeResults`

- **`TotalProcessingTimeStep`**: Overall processing time tracking
  - `WaitForPermit`, `Search`, `Similar`

### 2. Progress Tracking Infrastructure (`crates/milli/src/progress.rs`)

Enhanced the `Progress` struct with:

- **`ScopedProgressStep`**: An helper that automatically finishes a progress step when dropped
  - Prevents manual step management errors
  - Must be assigned to a named variable (`let _step = ...`) to ensure proper scoping

- **`accumulate_durations()` improvements**: 
  - Now sums durations for similar steps instead of keeping only the most recent
  - Essential for iterative processes like bucket sort that may execute the same step multiple times

- **Debug assertions**: Added safety checks to prevent double-starting scoped progress steps

### 3. Logging Integration

Progress traces are logged at the end of search operations in:
- `crates/meilisearch/src/routes/indexes/search.rs`
- `crates/meilisearch/src/routes/indexes/similar.rs`
- `crates/meilisearch/src/routes/multi_search.rs`

The progress information is logged using `tracing::trace!` with accumulated durations for each step.

### Scoped Progress Steps

The `ScopedProgressStep` helper uses Rust's `drop` behavior to ensure progress steps are properly finished:

```rust
let _step = progress.update_progress_scoped(SearchStep::Tokenize);
// ... search operations ...
// Step automatically finished when _step is dropped
```

This prevents common errors like forgetting to finish a step or finishing it multiple times.

> :warning: every call of `update_progress_scoped` must be assigned to a named value (`let value = ..` or `let _value = ..`) in order to be dropped at the end of the scope, if a call is assigned to an anonymous value (`let _ = ..`) or not assigned, it will be dropped directly.

### Duration Accumulation

The `accumulate_durations()` function now properly handles iterative processes:

- Previous behavior: Only kept the most recent duration for each step name
- New behavior: Sums all durations for steps with the same name
- This is critical for algorithms like bucket sort that may execute ranking rules multiple times

### Safety Improvements

- Added debug assertions to prevent double-starting scoped progress steps
- Improved error handling for progress tracking edge cases

## Testing

- Updated all existing search tests to work with the new progress tracking
- Added progress trace in benchmarks

### DIFF
--- a/crates/milli/src/progress.rs
+++ b/crates/milli/src/progress.rs
@@ -50,18 +50,25 @@ struct InnerProgress {
 }
 
 impl Progress {
-    pub fn update_progress<P: Step>(&self, sub_progress: P) {
+    /// Update the progress and return `true` if the step was started, `false` if it was already started.
+    pub fn update_progress<P: Step>(&self, sub_progress: P) -> bool {
         let mut inner = self.steps.write().unwrap();
         let InnerProgress { steps, durations } = &mut *inner;
 
         let now = Instant::now();
         let step_type = TypeId::of::<P>();
         if let Some(idx) = steps.iter().position(|(id, _, _)| *id == step_type) {
+            if steps[idx].1.name() == sub_progress.name() {
+                // The step is already started, so we don't need to start it again.
+                return false;
+            }
+
             push_steps_durations(steps, durations, now, idx);
             steps.truncate(idx);
         }
 
         steps.push((step_type, Box::new(sub_progress), now));
+        true
     }
 
     /// End a step that has been started without having to start a new step.
@@ -71,11 +78,13 @@ impl Progress {
 
         let now = Instant::now();
         let step_type = TypeId::of::<P>();
+
         debug_assert!(
             steps.iter().any(|(id, s, _)| *id == step_type && s.name() == sub_progress.name()),
             "Step `{}` must have been started",
             sub_progress.name()
         );
+
         if let Some(idx) = steps.iter().position(|(id, _, _)| *id == step_type) {
             push_steps_durations(steps, durations, now, idx);
             steps.truncate(idx);
@@ -84,8 +93,15 @@ impl Progress {
 
     /// Update the progress and return a scoped progress step that will end the progress step when dropped.
     pub fn update_progress_scoped<P: Step + Copy>(&self, step: P) -> ScopedProgressStep<'_, P> {
-        self.update_progress(step);
-        ScopedProgressStep { progress: self, step }
+        let started = self.update_progress(step);
+
+        debug_assert!(
+            started,
+            "Step `{}` can't be scoped because it was already started",
+            step.name()
+        );
+
+        ScopedProgressStep { progress: self, step: started.then_some(step) }
     }
 
     // TODO: This code should be in meilisearch_types but cannot because milli can't depend on meilisearch_types
@@ -378,11 +394,13 @@ impl<T: steppe::Step> Step for Compat<T> {
 
 pub struct ScopedProgressStep<'a, P: Step + Copy> {
     progress: &'a Progress,
-    step: P,
+    step: Option<P>,
 }
 
 impl<'a, P: Step + Copy> Drop for ScopedProgressStep<'a, P> {
     fn drop(&mut self) {
-        self.progress.end_progress_step(self.step);
+        if let Some(step) = self.step {
+            self.progress.end_progress_step(step);
+        }
     }
 }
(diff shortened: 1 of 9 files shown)


YoEight:
I conducted a first review pass and I didn't find any showstopper. Good job.

YoEight (approved):

crates/milli/src/progress.rs
```diff
@@ -50,18 +50,58 @@ struct InnerProgress {
...
+        let InnerProgress { steps, durations } = &mut *inner;
+
+        let now = Instant::now();
+        let step_type = TypeId::of::<P>();
+
+        debug_assert!(
+            steps.iter().any(|(id, s, _)| *id == step_type && s.name() == sub_progress.name()),
+            "Step `{}` must have been started",
+            sub_progress.name()
+        );
```
dureuill, line 86: if this debug assert fires, then the lock will be poisoned.

I'm also not sure about having asserts that only fire in debug mode.

If we want to keep it, I say we should have a cfg(dev) to compute the boolean and then explicitly drop the lock before panicking when the assertion doesn't hold

crates/milli/src/progress.rs
```diff
@@ -50,18 +50,58 @@ struct InnerProgress {
...
+
+        if let Some(idx) = steps.iter().position(|(id, _, _)| *id == step_type) {
+            push_steps_durations(steps, durations, now, idx);
+            steps.truncate(idx);
+        }
+    }
+
+    /// Update the progress and return a scoped progress step that will end the progress step when dropped.
```
dureuill, line 94: This function should document its panicking case.

crates/milli/src/progress.rs
```diff
@@ -343,3 +391,16 @@ impl<T: steppe::Step> Step for Compat<T> {
...
+
+pub struct ScopedProgressStep<'a, P: Step + Copy> {
+    progress: &'a Progress,
+    step: Option<P>,
+}
+
+impl<'a, P: Step + Copy> Drop for ScopedProgressStep<'a, P> {
+    fn drop(&mut self) {
+        if let Some(step) = self.step {
+            self.progress.end_progress_step(step);
```
dureuill, line 403: `end_progress_step` may panic (for example, if the coder performed a manual progress.update while there is a ù ScopedProgressStep`), which will cause an abort in some cases as this is a destructor.

(panics in panicking destructors cause an immediate abort).

crates/meilisearch/src/routes/indexes/search.rs
```diff
@@ -394,11 +401,12 @@ pub async fn search_with_url_query(
                 personalize,
                 personalize_query.as_deref(),
                 time_budget,
+                &progress,
             )
             .await?;
     }
 
-    debug!(request_uid = ?request_uid, returns = ?search_result, "Search get");
+    debug!(request_uid = ?request_uid, returns = ?search_result, progress = ?progress.accumulated_durations(), "Search get");
```
dureuill, line 409: This makes logs for all search requests much bigger, might this increase costs?

dureuill (commented):
Hello 👋

I'm coming a bit late with a few comments.

Thank you for your consideration

PR #6079, opened 2025-12-30
Support protocol negociation on plain-text connections

YoEight:
When a client sends a large request body and the server decides to reject it with 413 before the client finishes uploading:

1. Server sends the 413 response and closes (or tries to close) the connection
1. Client is still writing the request body
1. The server's close can cause a TCP RST (reset)
1. The RST can arrive at the client before or during its read of the response

The client might see a "connection reset" error instead of the actual 413 response. HTTP/2 handles this gracefully with stream multiplexing where a request can be rejected without nuking the connection. HTTP/1.1 doesn't have that luxury.

Currently, the server only has protocol negociation when using HTTPS. With this change the server can now do protocol negociation when dealing with plain-text connections.


YoEight:
No longer needed

PR #6080, opened 2025-12-30
Extract document operations from payloads in parallel

Kerollmops:
Closes #5840 and implements it without any additional dependencies or synchronisation primitives. Only using the rayon reduce operation. We ensured that the order of documents was preserved when assigning internal IDs in parallel, consistent with the previous approach before this PR.

### Changelog Description

We accelerate document indexing by processing a large number of tasks in batches or a large number of records in parallel. We expedited the preparation of the payloads by extracting the various changes and assigning internal IDs in parallel. We achieved a 7x speedup on a four-million-document insertion using four CPUs, and the performance scales with the number of CPUs.

The `indexedDocuments` field **in tasks using `skipCreation`** no longer precisely reflects the number of document operations performed, specifically for POST and PUT operations. This count may be higher than the actual number of operations, but it doesn't affect the computation; only the reported count is impacted. We prioritize speed over perfect accuracy here, and the documents are still correctly indexed as before.

### Before

We can clearly see that a large portion of the indexing is spent using a single thread. This portion involves preparing the payload, which extracts external document IDs, assigns an internal document ID, and associates the documents with various operations (deletion, replacement, update with and without `skipCreation`). You may notice that the range on the screenshot is approximately 1 hour and 30 minutes, which corresponds to the maximum size of a batch, equivalent to about 4.1 million documents. As you can see, the mean time the user spends on the CPU is about 47%.

Capture d’écran 2025-12-31 à 12 42 47

### After

Now we can clearly see that the portion dedicated to preparing the payload is not even visible. Indexation now takes 38 minutes to index nearly 4 million documents. As you can see, the mean time the user spends on the CPU is now 76%. In this specific example, we reduced the preparation time from 27 minutes to 4 minutes for the payloads.

Capture d’écran 2025-12-31 à 15 50 29

### DIFF
--- a/crates/milli/src/update/new/indexer/document_operation.rs
+++ b/crates/milli/src/update/new/indexer/document_operation.rs
@@ -1,6 +1,7 @@
 use std::sync::atomic::Ordering;
 use std::{mem, ops, ptr, vec};
 
+use bstr::ByteSlice;
 use bumpalo::collections::vec::Vec as BumpVec;
 use bumpalo::Bump;
 use bumparaw_collections::RawMap;
@@ -93,64 +94,62 @@ impl<'pl> IndexOperations<'pl> {
         progress.update_progress(IndexingStep::PreparingPayloads);
         let Self { operations } = self;
 
-        let documents_ids = index.documents_ids(rtxn)?;
-        let first_document = fetch_first_document(indexer, &operations)?;
-        let primary_key_result = retrieve_or_guess_primary_key(
+        // We first fetch the primary key from the index...
+        let primary_key = retrieve_or_guess_primary_key(
             rtxn,
             index,
             new_fields_ids_map,
             primary_key_from_op,
-            first_document,
-        )?;
+            None,
+        )?
+        .map(|(pk, _)| pk);
 
-        // Whenever retrieving the primary key fails with a user
-        // error we return it as a sub error for the first payload.
-        let primary_key = match primary_key_result.map_err(Some) {
-            Ok((primary_key, _)) => primary_key,
-            Err(mut user_error) => {
-                if operations.iter().all(|p| p.bytes().is_none_or(|b| b.is_empty())) {
-                    // We do not consider errors when all payloads are empty and
-                    // especially when it is an issue guessing the primary key.
-                    user_error = None;
+        // ...and if not present we try to guess it from
+        // the operations and ignore the useless ones.
+        let (primary_key, pre_payload_stats, remaining_operations) = match primary_key {
+            Ok(pk) => (pk, Vec::new(), operations),
+            Err(_user_error) => {
+                let (primary_key, pre_payload_stats, operations) =
+                    fetch_primary_and_ignore_payloads(
+                        indexer,
+                        index,
+                        rtxn,
+                        primary_key_from_op,
+                        new_fields_ids_map,
+                        operations,
+                    )?;
+
+                match primary_key {
+                    Some(pk) => (pk, pre_payload_stats, operations),
+                    None => {
+                        return Ok((
+                            DocumentOperationChanges { docids_version_offsets: &[] },
+                            pre_payload_stats,
+                            None,
+                        ));
+                    }
                 }
-
-                let payload_stats = operations
-                    .iter()
-                    .map(|payload| PayloadStats {
-                        bytes: payload.bytes().map_or(0, |p| p.len() as u64),
-                        document_count: 0,
-                        error: user_error.take(),
-                    })
-                    .collect();
-
-                return Ok((
-                    DocumentOperationChanges { docids_version_offsets: &[] },
-                    payload_stats,
-                    None,
-                ));
             }
         };
 
-        let payload_count: u32 = operations.len().try_into().unwrap();
+        let payload_count: u32 = remaining_operations.len().try_into().unwrap();
         let (step, progress_step) = AtomicPayloadStep::new(payload_count);
         progress.update_progress(progress_step);
 
-        let indexed_document_operations = operations
-            .into_par_iter()
-            .enumerate()
-            .map(|(payload_index, payload)| {
-                if must_stop_processing() {
-                    return Err(InternalError::AbortedIndexation.into());
-                }
-                step.store(payload_index as u32, Ordering::Relaxed);
-                IndexedPayloadOperations::from_payload(payload, &primary_key, shards)
-            })
-            .try_reduce(IndexedPayloadOperations::default, |lhs, rhs| lhs | rhs)?;
+        let IndexedPayloadOperations { document_operations, fields_ids_map, payload_stats } =
+            remaining_operations
+                .into_par_iter()
+                .enumerate()
+                .map(|(payload_index, payload)| {
+                    if must_stop_processing() {
+                        return Err(InternalError::AbortedIndexation.into());
+                    }
+                    step.store(payload_index as u32, Ordering::Relaxed);
+                    IndexedPayloadOperations::from_payload(payload, &primary_key, shards)
+                })
+                .try_reduce(IndexedPayloadOperations::default, |lhs, rhs| lhs | rhs)?;
 
         step.store(payload_count, Ordering::Relaxed);
-
-        let IndexedPayloadOperations { document_operations, fields_ids_map, payload_stats } =
-            indexed_document_operations;
 
         // We must drain the HashMap into a Vec because rayon::hash_map::IntoIter: !Clone
         progress.update_progress(IndexingStep::AssigningDocumentsIds);
@@ -169,6 +168,7 @@ impl<'pl> IndexOperations<'pl> {
             })
             .collect_vec_list();
 
+        let documents_ids = index.documents_ids(rtxn)?;
         let mut available_ids = AvailableIds::new(&documents_ids);
         let number_of_operations = document_operations.len();
         let mut docids_version_offsets = BumpVec::with_capacity_in(number_of_operations, indexer);
@@ -199,29 +199,84 @@ impl<'pl> IndexOperations<'pl> {
         docids_version_offsets
             .par_sort_unstable_by_key(|(_, po)| first_update_pointer(&po.operations).unwrap_or(0));
 
-        let docids_version_offsets = docids_version_offsets.into_bump_slice();
-        Ok((DocumentOperationChanges { docids_version_offsets }, payload_stats, Some(primary_key)))
+        Ok((
+            DocumentOperationChanges {
+                docids_version_offsets: docids_version_offsets.into_bump_slice(),
+            },
+            // Once we got the payload stats for the valid operations
+            // we must prepend the stats from skipped ones.
+            pre_payload_stats.into_iter().chain(payload_stats).collect(),
+            Some(primary_key),
+        ))
     }
 }
 
-/// Fetches the first document from the set of payloads.
-fn fetch_first_document<'b, 'pl: 'b>(
-    bump: &'b Bump,
-    operations: &[Payload<'pl>],
-) -> Result<Option<RawMap<'b, FxBuildHasher>>, InternalError> {
-    operations
-        .iter()
-        .find_map(|payload| {
-            let payload = payload.bytes()?;
-            Deserializer::from_slice(payload)
-                .into_iter::<&RawValue>()
-                .next()
-                .map(|v| v.and_then(|v| RawMap::from_raw_value_and_hasher(v, FxBuildHasher, bump)))
-                .transpose()
-                .map_err(InternalError::SerdeJson)
-                .transpose()
-        })
-        .transpose()
+/// Fetches the primary key from the operations and removes the ignored ones.
+/// Collecting the stats and the useful payloads for future use.
+fn fetch_primary_and_ignore_payloads<'pl>(
+    bump: &'pl Bump,
+    index: &Index,
+    rtxn: &'pl RoTxn<'pl>,
+    primary_key_from_op: Option<&'pl str>,
+    new_fields_ids_map: &mut FieldsIdsMap,
+    operations: Vec<Payload<'pl>>,
+) -> Result<(Option<PrimaryKey<'pl>>, Vec<PayloadStats>, Vec<Payload<'pl>>)> {
+    let mut payload_stats = Vec::new();
+    let mut remaining_operations = Vec::new();
+    let mut primary_key = None;
+
+    for operation in operations {
+        if primary_key.is_some() {
+            remaining_operations.push(operation);
+            continue;
+        }
+
+        let stats = match operation {
+            Payload::Replace { payload: p, .. } | Payload::Update { payload: p, .. } => {
+                // Fetches the first document from payload bytes.
+                let first_document = Deserializer::from_slice(p)
+                    .into_iter::<&RawValue>()
+                    .next()
+                    .map(|v| {
+                        v.and_then(|v| RawMap::from_raw_value_and_hasher(v, FxBuildHasher, bump))
+                    })
+                    .transpose()
+                    .map_err(InternalError::SerdeJson)?;
+
+                let primary_key_result = retrieve_or_guess_primary_key(
+                    rtxn,
+                    index,
+                    new_fields_ids_map,
+                    primary_key_from_op,
+                    first_document,
+                )?;
+
+                match primary_key_result {
+                    Ok((pk, _)) => {
+                        primary_key = Some(pk);
+                        // From now on, we will collect the remaining
+                        // operations in a vector to manage them later on.
+                        remaining_operations.push(operation);
+                        continue;
+                    }
+                    Err(error) => PayloadStats {
+                        bytes: p.len() as u64,
+                        document_count: 0,
+                        // We do not consider errors when payloads are empty.
+                        error: if p.trim().is_empty() { None } else { Some(error) },
+                    },
+                }
+            }
+            Payload::Deletion(_) => {
+                // We reach this when we don't have a primary key so it's impossible to delete documents.
+                PayloadStats { bytes: 0, document_count: 0, error: None }
+            }
+        };
+
+        payload_stats.push(stats);
+    }
+
+    Ok((primary_key, payload_stats, remaining_operations))
 }
 
 /// The correctly ordered operations that were extracted from the payload.
@@ -573,16 +628,6 @@ pub enum Payload<'pl> {
     Replace { payload: &'pl [u8], on_missing_document: MissingDocumentPolicy },
     Update { payload: &'pl [u8], on_missing_document: MissingDocumentPolicy },
     Deletion(&'pl [&'pl str]),
-}
-
-impl<'pl> Payload<'pl> {
-    fn bytes(&self) -> Option<&'pl [u8]> {
-        match self {
-            Payload::Replace { payload, .. } => Some(payload),
-            Payload::Update { payload, .. } => Some(payload),
-            Payload::Deletion(_) => None,
-        }
-    }
 }
 
 pub struct PayloadStats {
--- a/crates/index-scheduler/src/queue/batches_test.rs
+++ b/crates/index-scheduler/src/queue/batches_test.rs
@@ -559,13 +559,12 @@ fn batch_deletion_nothing_and_add_documents_no_guess_pk() {
         )
         .unwrap();
 
-    // non-guessable pk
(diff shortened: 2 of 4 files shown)


crates/index-scheduler/src/queue/snapshots/batches_test.rs/batch_skip_creation_with_deletion/1.snap
```diff
@@ -9,7 +9,7 @@ source: crates/index-scheduler/src/queue/batches_test.rs
...
 2 {uid: 2, batch_uid: 1, status: succeeded, details: { received_document_ids: 1, deleted_documents: Some(1) }, kind: DocumentDeletion { index_uid: "docs", documents_ids: ["1"] }}
-3 {uid: 3, batch_uid: 1, status: succeeded, details: { received_documents: 1, indexed_documents: Some(0) }, kind: DocumentAdditionOrUpdate { index_uid: "docs", primary_key: Some("id"), method: ReplaceDocuments, content_file: 00000000-0000-0000-0000-000000000002, documents_count: 1, allow_index_creation: true, on_missing_document: Skip }}
+3 {uid: 3, batch_uid: 1, status: succeeded, details: { received_documents: 1, indexed_documents: Some(1) }, kind: DocumentAdditionOrUpdate { index_uid: "docs", primary_key: Some("id"), method: ReplaceDocuments, content_file: 00000000-0000-0000-0000-000000000002, documents_count: 1, allow_index_creation: true, on_missing_document: Skip }}
```
dureuill, line 12: why did the number of indexed documents change?

crates/milli/src/update/new/indexer/document_operation.rs
```diff
@@ -83,376 +88,462 @@ impl<'pl> DocumentOperation<'pl> {
...
+        let IndexedPayloadOperations { document_operations, fields_ids_map, payload_stats } =
+            indexed_document_operations;
+
+        // We must drain the HashMap into a Vec because rayon::hash_map::IntoIter: !Clone
+        progress.update_progress(IndexingStep::AssigningDocumentsIds);
+        let external_documents_ids = index.external_documents_ids();
+
+        // We broadcast database reads to retrieve the internal IDs of the existing
+        // documents and mark the ones that needs a new one. To avoid creating a lot of
+        // read transactions we prefer broadcasting the work and creating one by thread.
```
dureuill: we are no longer "broadcasting" since we use `par_keys`

crates/milli/src/update/new/indexer/document_operation.rs
```diff
@@ -83,376 +88,462 @@ impl<'pl> DocumentOperation<'pl> {
...
+        let thread_local_rtxns = ThreadLocal::new();
+        let rtxn_id = rtxn.id();
+        let extracted_docids = document_operations
+            .par_keys()
+            .enumerate()
+            .map(|(_, external_id)| {
+                let local_rtxn = thread_local_rtxns.get_or_try(|| index.read_txn())?;
+                assert_eq!(local_rtxn.id(), rtxn_id);
```
dureuill: I don't understand why we are performing this check? Which error condition are we mitigating, when do we think it might happen?

crates/milli/src/update/new/indexer/document_operation.rs
```diff
@@ -83,376 +88,462 @@ impl<'pl> DocumentOperation<'pl> {
...
+                    payload_stats,
+                    None,
+                ));
+            }
+        };
 
         let payload_count = operations.len();
-        let (step, progress_step) = AtomicPayloadStep::new(payload_count as u32);
+        let (step, progress_step) =
+            AtomicPayloadStep::new(payload_count.min(u32::MAX as usize) as u32);
```
dureuill: should we error explicitly should the payload count exceed `u32::MAX` rather than silently hiding the issue?

I think the `.min` call here is not useful because `payload_count` is already guaranteed to be in range `0..u32::MAX` by the fact that Meilisearch supports at most `u32::MAX` tasks, each with at most 1 payload.

Should this property be disproven at some point, I'd rather have a loud error message (not necessarily at this point) than a silent truncation.

crates/milli/src/update/new/indexer/document_operation.rs
```diff
@@ -83,376 +88,462 @@ impl<'pl> DocumentOperation<'pl> {
...
+        let primary_key_result = retrieve_or_guess_primary_key(
+            rtxn,
+            index,
+            new_fields_ids_map,
+            primary_key_from_op,
+            first_document,
+        )?;
+
+        // Whenever retrieving the primary key fails with a user
+        // error we return it as a sub error for the first payload.
```
dureuill: why the first payload?

This is a bit niche, but I did a sequence of `DELETE /1` and `POST <documents-without-guessable-pk>` on a freshly created index without a pk, and it resulted in:

```jsonc
{
      "uid": 5,
      "batchUid": 4,
      "indexUid": "movies",
      "status": "succeeded", // task marked as succeeded?
      "canceledBy": null,
      "details": {
        "receivedDocuments": 1, // no document was ever indexed
        "indexedDocuments": 0
      },
      "error": null,
      "duration": "PT0.082032S",
      "enqueuedAt": "2026-01-06T13:47:29.51769Z",
      "startedAt": "2026-01-06T13:47:34.07025Z",
      "finishedAt": "2026-01-06T13:47:34.152282Z"
    },
    {
      "uid": 4,
      "batchUid": 4,
      "indexUid": "movies",
      "status": "failed", // the task is marked as failed
      "type": "documentDeletion", // but it is the deletion one
      "canceledBy": null,
      "details": {
        "providedIds": 1,
        "deletedDocuments": 0,
        "originalFilter": null
      },
      "error": {
        "message": "The primary key inference failed as the engine found 2 fields ending with `id` in their names: 'id' and 'noid'. Please specify the primary key manually using the `primaryKey` query parameter.",
        "code": "index_primary_key_multiple_candidates_found",
        "type": "invalid_request",
        "link": "https://docs.meilisearch.com/errors#index_primary_key_multiple_candidates_found"
      },
      "duration": "PT0.082032S",
      "enqueuedAt": "2026-01-06T13:47:10.315417Z",
      "startedAt": "2026-01-06T13:47:34.07025Z",
      "finishedAt": "2026-01-06T13:47:34.152282Z"
    },
```

By contrast, `main` does:

```jsonc
  {
      "uid": 3,
      "batchUid": 2,
      "indexUid": "movies",
      "status": "failed", // document addition fails
      "canceledBy": null,
      "details": {
        "receivedDocuments": 1,
        "indexedDocuments": 0
      },
      "error": {
        "message": "The primary key inference failed as the engine found 2 fields ending with `id` in their names: 'id' and 'noid'. Please specify the primary key manually using the `primaryKey` query parameter.",
        "code": "index_primary_key_multiple_candidates_found",
        "type": "invalid_request",
        "link": "https://docs.meilisearch.com/errors#index_primary_key_multiple_candidates_found"
      },
      "duration": "PT0.122063S",
      "enqueuedAt": "2026-01-06T13:59:07.636106Z",
      "startedAt": "2026-01-06T13:59:23.391651Z",
      "finishedAt": "2026-01-06T13:59:23.513714Z"
    },
    {
      "uid": 2,
      "batchUid": 2,
      "indexUid": "movies",
      "status": "succeeded", // deletion task succeeds
      "type": "documentDeletion",
      "canceledBy": null,
      "details": {
        "providedIds": 1,
        "deletedDocuments": 1,
        "originalFilter": null
      },
      "error": null,
      "duration": "PT0.122063S",
      "enqueuedAt": "2026-01-06T13:59:02.574439Z",
      "startedAt": "2026-01-06T13:59:23.391651Z",
      "finishedAt": "2026-01-06T13:59:23.513714Z"
    },
```

I think we are now marking the wrong task as failed.

crates/milli/src/update/new/indexer/document_operation.rs
```diff
@@ -83,376 +88,462 @@ impl<'pl> DocumentOperation<'pl> {
...
+                        document_count: 0,
+                        error: user_error.take(),
+                    })
+                    .collect();
+
+                return Ok((
+                    DocumentOperationChanges { docids_version_offsets: &[] },
+                    payload_stats,
+                    None,
+                ));
```
dureuill: This skips the indexation of all tasks when the pk is not found in the first payload, this is a change in behavior compared to `main`.

On main, sending:

1. A `POST <documents-without-guessable-pk>`
2. A `POST <documents-with-guessable-pk>`

results in the first task being failed with no indexed documents, **but the second one being succeeded** with the indexed documents (can be fetched via **GET**)

```jsonc
{
      "uid": 7,
      "batchUid": 4,
      "indexUid": "movies",
      "status": "succeeded",
      "canceledBy": null,
      "details": {
        "receivedDocuments": 1,
        "indexedDocuments": 1
      },
      "error": null,
      "duration": "PT0.114398S",
      "enqueuedAt": "2026-01-06T14:01:05.460501Z",
      "startedAt": "2026-01-06T14:01:16.826178Z",
      "finishedAt": "2026-01-06T14:01:16.940576Z"
    },
    {
      "uid": 6,
      "batchUid": 4,
      "indexUid": "movies",
      "status": "failed",
      "canceledBy": null,
      "details": {
        "receivedDocuments": 1,
        "indexedDocuments": 0
      },
      "error": {
        "message": "The primary key inference failed as the engine found 2 fields ending with `id` in their names: 'id' and 'noid'. Please specify the primary key manually using the `primaryKey` query parameter.",
        "code": "index_primary_key_multiple_candidates_found",
        "type": "invalid_request",
        "link": "https://docs.meilisearch.com/errors#index_primary_key_multiple_candidates_found"
      },
      "duration": "PT0.114398S",
      "enqueuedAt": "2026-01-06T14:00:59.99898Z",
      "startedAt": "2026-01-06T14:01:16.826178Z",
      "finishedAt": "2026-01-06T14:01:16.940576Z"
    },
    {
      "uid": 5,
      "batchUid": 4,
      "indexUid": "movies",
      "status": "succeeded",
      "type": "documentDeletion",
      "canceledBy": null,
      "details": {
        "providedIds": 1,
        "deletedDocuments": 1,
        "originalFilter": null
      },
      "error": null,
      "duration": "PT0.114398S",
      "enqueuedAt": "2026-01-06T14:00:55.379227Z",
      "startedAt": "2026-01-06T14:01:16.826178Z",
      "finishedAt": "2026-01-06T14:01:16.940576Z"
    },
```

on this branch the result is that no document is indexed, in spite of the batch and some of the tasks being succeeded:

```jsonc
 {
      "uid": 4,
      "batchUid": 2,
      "indexUid": "movies",
      "status": "succeeded",
      "canceledBy": null,
      "details": {
        "receivedDocuments": 1,
        "indexedDocuments": 0
      },
      "error": null,
      "duration": "PT0.083527S",
      "enqueuedAt": "2026-01-06T14:07:31.093656Z",
      "startedAt": "2026-01-06T14:07:48.134416Z",
      "finishedAt": "2026-01-06T14:07:48.217943Z"
    },
    {
      "uid": 3,
      "batchUid": 2,
      "indexUid": "movies",
      "status": "succeeded",
      "canceledBy": null,
      "details": {
        "receivedDocuments": 1,
        "indexedDocuments": 0
      },
      "error": null,
      "duration": "PT0.083527S",
      "enqueuedAt": "2026-01-06T14:07:27.669203Z",
      "startedAt": "2026-01-06T14:07:48.134416Z",
      "finishedAt": "2026-01-06T14:07:48.217943Z"
    },
    {
      "uid": 2,
      "batchUid": 2,
      "indexUid": "movies",
      "status": "failed",
      "type": "documentDeletion",
      "canceledBy": null,
      "details": {
        "providedIds": 1,
        "deletedDocuments": 0,
        "originalFilter": null
      },
      "error": {
        "message": "The primary key inference failed as the engine found 2 fields ending with `id` in their names: 'id' and 'noid'. Please specify the primary key manually using the `primaryKey` query parameter.",
        "code": "index_primary_key_multiple_candidates_found",
        "type": "invalid_request",
        "link": "https://docs.meilisearch.com/errors#index_primary_key_multiple_candidates_found"
      },
      "duration": "PT0.083527S",
      "enqueuedAt": "2026-01-06T14:07:23.54819Z",
      "startedAt": "2026-01-06T14:07:48.134416Z",
      "finishedAt": "2026-01-06T14:07:48.217943Z"
    },
```

crates/milli/src/update/new/indexer/document_operation.rs
```diff
@@ -83,376 +88,462 @@ impl<'pl> DocumentOperation<'pl> {
...
+                .map_err(InternalError::SerdeJson)
+                .transpose()
+        })
+        .transpose()
+}
 
-    let mut new_docids_version_offsets = hashbrown::HashMap::<&str, PayloadOperations<'pl>>::new();
+/// The correctly ordered operations that were extracted from the payload.
+///
+/// We don't need the payload index as we merge them in order by using the
```
dureuill: truncated comment

crates/milli/src/update/new/indexer/document_operation.rs
```diff
@@ -83,376 +88,462 @@ impl<'pl> DocumentOperation<'pl> {
...
+                ) => continue,
+                (Some(Deletion), Replacement { on_missing_document: Create, .. }) => {
+                    document_operations.clear()
+                }
+                (Some(Replacement { .. } | Update { .. }), Update { .. }) => (),
+                (Some(Deletion), Update { on_missing_document: Create, .. }) => (),
+            }
+
+            document_operations.push(operation);
+        }
```
dureuill: I find the logic difficult to follow, because the output depends on the existence status of the document following the previous operation, but this dependency is not made explicit.

Here is a version that first computes the existence status after the last op, and then uses this explicit status in the match:

```suggestion
            let existence_after_last_op = match document_operations.last() {
                Some(Replacement { .. } | Update { .. }) => DocumentExistence::Exists,
                Some(Deletion) => DocumentExistence::Missing,
                None => document_existence,
            };

            match (existence_after_last_op, operation) {
                // when the document is missing for sure after the last operation,
                // and the next operation requires skipping creation,
                // we skip this operation
                (
                    DocumentExistence::Missing,
                    Replacement { on_missing_document: Skip, .. }
                    | Update { on_missing_document: Skip, .. },
                ) => continue,
                // deletions and replacements delete all previous operations
                (_, op @ (Deletion | Replacement { .. })) => {
                    document_operations.clear();
                    document_operations.push(op);
                }
                // updates executes after the previous operations
                (_, op @ Update { .. }) => {
                    document_operations.push(op);
                }
            }
```

crates/milli/src/update/new/indexer/document_operation.rs
```diff
@@ -83,376 +88,462 @@ impl<'pl> DocumentOperation<'pl> {
...
+
+        if document_operations.is_empty() {
```
dureuill: what should we do when we have (DocumentExistence::missing, vec![Deletion])?

crates/index-scheduler/src/queue/snapshots/batches_test.rs/batch_skip_creation_with_deletion/1.snap
Kerollmops: Hum... that's a mistake on my side. I should have seen this and refuse this snapshot. I'll investigate. Thank you 🤔

crates/milli/src/update/new/indexer/document_operation.rs
Kerollmops: It shouldn't happen, I'm just cautious but, indeed, the write transaction can't be committed as it is borrowed by us. I'll remove this check.

crates/milli/src/update/new/indexer/document_operation.rs
Kerollmops: Hum... indeed. It's niche and untested. I'll add a test and make sure only the addition/update tasks are marked as failed in case of a user error.

crates/milli/src/update/new/indexer/document_operation.rs
Kerollmops: That's also not tested. I'll add a test for this specific case. Thank you. I'll make sure that if we can't guess the pk from a payload we try with the next one (and ignore the ones without pk?).

crates/milli/src/update/new/indexer/document_operation.rs
Kerollmops: You are right, I'll simply unwrap on this part.

crates/milli/src/update/new/indexer/document_operation.rs
Kerollmops: We should ignore the deletion. I just modified the code to do so.

crates/milli/src/update/new/indexer/document_operation.rs
Kerollmops: I fixed it and added a dedicated test with the help of @ManyTheFish 👏

crates/milli/src/update/new/indexer/document_operation.rs
Kerollmops: I added a test (the same as the other issue) that checked this property and more. Thank you for sprotting this bug 👌

crates/index-scheduler/src/queue/snapshots/batches_test.rs/batch_skip_creation_with_deletion/1.snap
Kerollmops: Investogating this one, looks a bit hard to fix correctly 👀

Kerollmops:
We spend a significant amount of time fixing the tasks' stats, and especially the number of indexed documents. However, with @ManyTheFish, we think it's way too much work and time spent to make the stats work. We noticed that the issue is related to the new `skipCreation` feature of @YoEight.

We wanted to know if it was necessary to make it work perfectly, as we don't want to waste this time making it work for a niche feature. @dureuill do you need these stats to be perfectly valid for the sharding or are you using dedicated `RoaringBitmaps` to know which documents were indexed on the machines?

In other words, we want to trash the last commit of the PR and merge it as is.

dureuill:
Hello, these stats are not used by sharding, except as a "visual"  check that everything went well. 

I think we might have to state in the release notes that in some specific situations the count might be off. 

Other than that we can land as-is

crates/milli/src/update/new/indexer/document_operation.rs
```diff
@@ -83,376 +93,557 @@ impl<'pl> DocumentOperation<'pl> {
...
-        };
+    Ok((primary_key, payload_stats, remaining_operations))
+}
 
-        let current_offset = iter.byte_offset();
-        let content = &payload[previous_offset..current_offset];
-        previous_offset = current_offset;
+/// The correctly ordered operations that were extracted from the payload.
+///
+/// We don't need the payload index as we merge them in order by using
```
YoEight, line 284: I could be wrong here but since `PayloadId` was introduced to prevent ordering issue, isn't that comment outdated?

YoEight:
Regarding this line in the PR description:

> Fortunately, we maintained the original order of the documents when assigning internal IDs in parallel.

Shouldn't this say:

> We maintained the original order of the payloads, which allows us to assign payload ids to documents deterministically in parallel.

I'm asking because It's not the "order of documents" that matters. Documents are identified by external IDs and don't have an inherent order, unless I'm wrong here.

What's actually preserved is the payload submission order (0, 1, 2, 3...). The `PayloadId` tagging system ensures that operations from different payloads are processed in the correct order. The determinism comes from the `IndexMap` having a consistent iteration order that results from the ordered payload merging.

crates/milli/src/update/new/indexer/document_operation.rs
Kerollmops: As explained in my last comment, I dropped the last commit in which I introduced the PayloadId. This PayloadId keeps track of which operation was performed on which documents, removing the ability to reduce the number of operations on documents mid-air (before the last reduction). This increases time and memory usage and complicates the codebase. All of this is to ensure that the count of documents is correctly displayed in tasks using the `skipCreation` parameter. So, I prefer dropping it and notifying of the impact on the `indexedDocuments` value in the changelog.

Kerollmops:
Hey @dureuill, Thanks for the answer. I'll update the changelog section to talk about this specific point.

Kerollmops:
Hey @YoEight, you are right. I'll refine the changelog to better explain the behavior (or even not talk about it as it doesn't change) and explicitly talk about the expected regression around the `IndexedDocument` tasks value that is less precise than before, and specifically for operations using the `skipCreation` parameter.

PR #6082, opened 2025-12-31
Add `/fields` route to get all the fields of an index

YoEight:
## Summary

Adds a new `POST /indexes/{indexUid}/fields` endpoint that returns detailed metadata about all fields in an index. This endpoint provides comprehensive information about each field's configuration, including display, search, filtering, and localization settings.

## What's New

- **New endpoint**: `POST /indexes/{indexUid}/fields`

For each field in an index, the endpoint returns:

```json
{
	"name": "cuisine.type",
	"displayed":  { "enabled": true },
	"searchable": { "enabled": true },
	"sortable": { "enabled": false },
	"distinct":   { "enabled": false },
	"rankingRule": {
	  "enabled": true,
	  "order": "asc"
	},
	"filterable": {
	    "enabled": false,
	    "sortBy": "alpha",
	    "facetSearch": true,
	    "equality": true,
	    "comparison": true
	},
	"localized":  { "locales": ["eng", "fra"] }
}
```

### Query Features

All those properties are optional. When multiple filter properties are provided, they work together using AND logic.

```jsonc
{
  // Pagination controls
  "offset": 0,
  "limit": 20,
  
  // Filter configuration - all filters are optional and can be combined
  "filter": {
    // Match fields using attribute patterns (supports wildcards: * for any characters)
    // Examples: "cuisine.*" matches cuisine.type, cuisine.region
    //           "user*" matches user_id, username, user_profile
    //           "*_id" matches all fields ending with _id
    "attributePatterns": ["cuisine.*", "*_timestamp"],
    
    // Filter by whether fields are displayed in search results
    // true = only displayed fields, false = only hidden fields
    "displayed": true,
    
    // Filter by whether fields are searchable (indexed for full-text search)
    // true = only searchable fields, false = only non-searchable fields
    "searchable": true,
    
    // Filter by whether fields can be used for sorting results
    // true = only sortable fields, false = only non-sortable fields
    "sortable": true,
    
    // Filter by whether the field is used as the distinct attribute
    // true = only the distinct field, false = only non-distinct fields
    "distinct": true,
    
    // Filter by whether the field is used in ranking rules
    // true = only fields used in ranking, false = only fields not used in ranking
    "rankingRule": true,
    
    // Filter by whether the field can be used for filtering/faceting
    // true = only filterable fields, false = only non-filterable fields
    "filterable": true
  }
}
```

## Potential Refactoring Opportunity

The enhancements made to `MetadataBuilder::from_index()` in this PR now provide comprehensive access to field metadata and index settings information. This means that in places where the codebase currently calls `MetadataBuilder::from_index()` followed by the `settings()` function (defined in `crates/meilisearch-types/src/settings.rs`) to retrieve index settings, the `settings()` call may no longer be necessary—`MetadataBuilder` alone now provides what's needed.

`MetadataBuilder` currently takes ownership of many strings, which seems unnecessary given that its lifetime could be tied to the LMDB read transaction it's built from, allowing it to use string references instead. The same could be done to some of the collections defined in the `MetadataBuilder` struct.

### DIFF
--- a/crates/meilisearch/src/routes/indexes/mod.rs
+++ b/crates/meilisearch/src/routes/indexes/mod.rs
@@ -1,23 +1,6 @@
 use std::collections::BTreeSet;
 use std::convert::Infallible;
 
-use actix_web::web::Data;
-use actix_web::{web, HttpRequest, HttpResponse};
-use deserr::actix_web::{AwebJson, AwebQueryParameter};
-use deserr::{DeserializeError, Deserr, ValuePointerRef};
-use index_scheduler::IndexScheduler;
-use meilisearch_types::deserr::query_params::Param;
-use meilisearch_types::deserr::{immutable_field_error, DeserrJsonError, DeserrQueryParamError};
-use meilisearch_types::error::deserr_codes::*;
-use meilisearch_types::error::{Code, ResponseError};
-use meilisearch_types::index_uid::IndexUid;
-use meilisearch_types::milli::{self, FieldDistribution, Index};
-use meilisearch_types::tasks::KindWithContent;
-use serde::Serialize;
-use time::OffsetDateTime;
-use tracing::debug;
-use utoipa::{IntoParams, OpenApi, ToSchema};
-
 use super::{
     get_task_id, Pagination, PaginationView, SummarizedTaskView, PAGINATION_DEFAULT_LIMIT,
 };
@@ -28,6 +11,27 @@ use crate::extractors::sequential_extractor::SeqHandler;
 use crate::proxy::{proxy, task_network_and_check_leader_and_version, Body};
 use crate::routes::is_dry_run;
 use crate::Opt;
+use actix_web::web::Data;
+use actix_web::{web, HttpRequest, HttpResponse};
+use deserr::actix_web::{AwebJson, AwebQueryParameter};
+use deserr::{DeserializeError, Deserr, ValuePointerRef};
+use index_scheduler::IndexScheduler;
+use meilisearch_types::deserr::query_params::Param;
+use meilisearch_types::deserr::{immutable_field_error, DeserrJsonError, DeserrQueryParamError};
+use meilisearch_types::error::deserr_codes::*;
+use meilisearch_types::error::{Code, ResponseError};
+use meilisearch_types::facet_values_sort::FacetValuesSort;
+use meilisearch_types::index_uid::IndexUid;
+use meilisearch_types::milli::tokenizer::Language;
+use meilisearch_types::milli::{
+    self, AttributePatterns, FieldDistribution, FieldSortOrder, FilterFeatures,
+    FilterableAttributesFeatures, Index, MetadataBuilder, PatternMatch,
+};
+use meilisearch_types::tasks::KindWithContent;
+use serde::{Serialize, Serializer};
+use time::OffsetDateTime;
+use tracing::debug;
+use utoipa::{IntoParams, OpenApi, ToSchema};
 
 pub mod compact;
 pub mod documents;
@@ -77,6 +81,7 @@ pub fn configure(cfg: &mut web::ServiceConfig) {
                     .route(web::delete().to(SeqHandler(delete_index))),
             )
             .service(web::resource("/stats").route(web::get().to(SeqHandler(get_index_stats))))
+            .service(web::resource("/fields").route(web::post().to(SeqHandler(post_index_fields))))
             .service(web::scope("/documents").configure(documents::configure))
             .service(web::scope("/search").configure(search::configure))
             .service(web::scope("/facet-search").configure(facet_search::configure))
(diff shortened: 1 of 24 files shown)


crates/meilisearch/src/routes/indexes/mod.rs
```diff
@@ -650,3 +655,216 @@ pub async fn get_index_stats(
...
+#[serde(rename_all = "camelCase")]
+pub struct FieldDistinctConfig {
+    pub enabled: bool,
+}
+
+#[derive(Debug, Serialize, Clone, Copy, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct FieldFacetSearchConfig<'a> {
+    pub sort_by: &'a str,
+}
```
ManyTheFish, line 714: The name `facet` is a legacy misleading name refering to multiple things that are stored in the same databases.

the `facetSearch` refers to the filters on which we can search-on using the `/facet-search` route.
However, the `sort_by` field refers to the `facetDistribution`, a feature that can be used in the `/search` route by adding the `facets` fields in the search request and is taken into account when retrieving the facet-search results.

So even if they are related, they are two "independent" things.

crates/meilisearch/src/routes/indexes/mod.rs
```diff
@@ -650,3 +655,216 @@ pub async fn get_index_stats(
...
+    pub contains: Option<String>,
+    #[deserr(default, try_from(&String) = from_string_regex -> DeserrJsonError<InvalidIndexFieldsFilterRegex>, error = DeserrJsonError<InvalidIndexFieldsFilterRegex>)]
+    #[schema(value_type = String)]
+    pub regex: Option<Regex>,
+    #[deserr(default, try_from(&String) = from_string_glob -> DeserrJsonError<InvalidIndexFieldsFilterGlob>, error = DeserrJsonError<InvalidIndexFieldsFilterGlob>)]
+    #[schema(value_type = String)]
+    pub glob: Option<GlobMatcher>,
+    #[deserr(default, error = DeserrJsonError<InvalidIndexFieldsFilterDisplayed>)]
+    pub displayed: Option<bool>,
+}
```
ManyTheFish: Is it inspered from the `TasksFilterQuery`? Could we harmonize the name and the implementation:
- name could be `FieldsFilterQuery`
- using `#[param(...)]` macro

crates/meilisearch/src/routes/indexes/mod.rs
```diff
@@ -650,3 +655,216 @@ pub async fn get_index_stats(
...
+    let fields = builder
+        .fields_metadata()
+        .iter()
+        .filter_map(|(name, metadata)| {
+            let is_filterable = builder.filterable_attributes().iter().any(|rule| match rule {
+                FilterableAttributesRule::Pattern(p) => {
+                    p.match_str(name) == milli::PatternMatch::Match
+                }
+                FilterableAttributesRule::Field(f) => f == name,
+            });
```
ManyTheFish: `Metadata` provides a `is_filterable`

crates/meilisearch/src/routes/indexes/mod.rs
YoEight: It's not inspired from `TasksFilterQuery`, at least I didn't do it knowingly.  I'm confused by your suggestion on using the `param` macro. I went with having the filter in the payload to match a comment you did here:

https://github.com/meilisearch/meilisearch/pull/5718#issuecomment-3023156207

crates/meilisearch/src/routes/indexes/mod.rs
YoEight: I picked the same names from the previous PR and those also match the names on the notion ticket.

crates/meilisearch/src/routes/indexes/mod.rs
YoEight: no it doesn't. it only exposes `filterable_attributes_rule_id` but then you have to do what I did in that snippet to know if the field is filterable.

ManyTheFish:
I would prefer a return value a below:
```json
{
	"name": "cuisine.type",
	"displayed":  { "enabled": true },
	"searchable": { "enabled": true },
	"sortable": { "enabled": false },
	"distinct":   { "enabled": false },
	"rankingRule": {
	  "enabled": true,
	  "order": "asc"
	},
	"filterable": {
	    "enabled": false,
	    "sortBy": "alpha",
	    "facetSearch": true,
	    "equality": true,
	    "comparison": true
	},
	"localized":  { "locales": ["eng", "fra"] }
}
```

In term of query I would prefer removing of `startsWith`, `contains`, `regex`, `glob` and have something like:
```jsonc
{
	// filter by field name
    // related to https://github.com/meilisearch/meilisearch/blob/a034875136b7d531c18934c3c78a1f50431348af/crates/meilisearch/src/routes/indexes/documents.rs#L413-L438
    "attributePatterns": ["cuisine.*", "..."],

	// filter by activated feature
    "displayed":  true,
    "searchable": true ,
    "sortable":  true,
    "distinct":  true,
    "rankingRule":  true,
    "filterable":  true
}
```

YoEight:
Hey @ManyTheFish , I updated both the input and the output JSON. Tell me what do you think about those changes

crates/meilisearch/src/routes/indexes/mod.rs
```diff
@@ -650,3 +654,242 @@ pub async fn get_index_stats(
...
+#[serde(rename_all = "camelCase")]
+pub struct FieldDistinctConfig {
+    pub enabled: bool,
+}
+
+#[derive(Debug, Serialize, Clone, Copy, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct FieldFacetSearchConfig<'a> {
+    pub sort_by: &'a str,
+}
```
ManyTheFish: Is this necessary now?

crates/meilisearch/src/routes/indexes/mod.rs
ManyTheFish:
```suggestion
		let filterable_attributes = builder.filterable_attributes();
		let FilterableAttributesFeatures {
		    facet_search,
		    filter,
		} = metadata.filterable_attributes_features(filterable_attributes);
		let FilterFeatures {
	        equality,
	        comparison,
	    } = filter;
		let is_filterable = equality || comparison || facet_search;
```

crates/meilisearch/src/routes/indexes/mod.rs
```diff
@@ -650,3 +654,242 @@ pub async fn get_index_stats(
...
+#[serde(rename_all = "camelCase")]
+pub struct FieldRankingRuleConfig {
+    pub enabled: bool,
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        serialize_with = "serialize_field_sort_order"
+    )]
+    #[schema(value_type = Vec<String>)]
+    pub order: Option<FieldSortOrder>,
+}
```
ManyTheFish, line 700: A suggestion here, feel free to ignore.
What about factorizing this using a intermediate structure and using [#[serde(flatten)]](https://serde.rs/field-attrs.html#flatten)?

```suggestion
#[derive(Debug, Serialize, Clone, Copy, ToSchema)]
#[serde(rename_all = "camelCase")]
pub struct Field<'a> {
    pub name: &'a str,
    pub displayed: FieldConfig<()>,
    pub searchable: FieldConfig<()>,
    pub sortable: FieldConfig<()>,
    pub distinct: FieldConfig<()>,
    pub ranking_rule: FieldConfig<FieldRankingRuleConfig>,
    pub filterable: FieldConfig<FieldFilterableConfig>,
    pub localized: FieldLocalizedConfig<'a>,
}

#[derive(Debug, Serialize, Clone, Copy, ToSchema)]
#[serde(rename_all = "camelCase")]
pub struct FieldConfig<T> {
    pub enabled: bool,
    #[serde(
        skip_serializing_if = "Option::is_none",
        flatten
    )]
	pub inner: Option<T>,
}

/// impl helpers
impl FieldConfig<T> {
	fn enabled(inner: T) -> Self {
		Self {
			enabled: true,
			inner
		}
	}
	
	fn disabled() -> Self {
		Self {
			enabled: false,
			inner: None
		}
	}
}

/// alternate: implement From
/// Put enabled to `true if inner config is Some(T)
impl From<Option<T>> for FieldConfig {
	fn from(inner: Option<T>) -> Self {
		Self {
			enabled: inner.is_some(),
			inner
		}
	}
}

#[derive(Debug, Serialize, Clone, Copy, ToSchema)]
#[serde(rename_all = "camelCase")]
pub struct FieldRankingRuleConfig {
    #[serde(serialize_with = "serialize_field_sort_order")]
    #[schema(value_type = Vec<String>)]
    pub order: FieldSortOrder,
}
```


I am not sure that `FieldConfig<()>` works, but a workaround could be found.

crates/meilisearch/src/routes/indexes/mod.rs
```diff
@@ -650,3 +654,242 @@ pub async fn get_index_stats(
...
+#[serde(rename_all = "camelCase")]
+pub struct FieldFacetSearchConfig<'a> {
+    pub sort_by: &'a str,
+}
+
+#[derive(Debug, Serialize, Clone, Copy, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct FieldFilterableConfig<'a> {
+    pub enabled: bool,
+    pub sort_by: &'a str,
```
ManyTheFish: could be an enum to avoid lifetime

ManyTheFish (changes requested):

crates/meilisearch/src/routes/indexes/mod.rs
YoEight: True but at the same time, what is the issue of having a lifetime there?

crates/meilisearch/src/routes/indexes/mod.rs
YoEight: While I like the idea of not repeating myself, I'm not necessarily convinced that we should try to be generic at all costs here. Even if your suggestion is easy to understand, it still introduces a layer of indirection. That said, if this were a pattern we were seeing a lot, I would have considered it.

crates/meilisearch/src/routes/indexes/mod.rs
YoEight: it is not indeed

crates/meilisearch/src/routes/indexes/mod.rs
YoEight: how you mean that `Metadata` not the `MetadataBuilder` my bad. I applied your suggestion

crates/meilisearch/src/routes/indexes/mod.rs
```diff
@@ -650,3 +654,231 @@ pub async fn get_index_stats(
...
+                ranking_rule: FieldRankingRuleConfig {
+                    enabled: metadata.is_asc_desc(),
+                    order: metadata.asc_desc,
+                },
+                filterable: FieldFilterableConfig {
+                    enabled: is_filterable,
+                    sort_by: match metadata.sort_by {
+                        OrderBy::Lexicographic => "alpha",
+                        OrderBy::Count => "count",
+                    },
```
ManyTheFish: The logic is duplicated with:
https://github.com/meilisearch/meilisearch/blob/main/crates/meilisearch-types/src/facet_values_sort.rs

crates/meilisearch/tests/index/fields.rs
```diff
@@ -0,0 +1,157 @@
...
+            "cuisine.type", "cuisine.region", "instructions.description"
+        ],
+        "filterableAttributes": [
+            "difficulty", "preparation_time_minutes", "cooking_time_minutes",
+            "total_time_minutes", "servings", "calories_per_serving", "cuisine.type",
+            "cuisine.region", "cuisine.is_authentic", "ingredients.name", "tags",
+            "nutrition.calories", "nutrition.protein", "nutrition.carbs", "nutrition.fat",
+            "nutrition.fiber", "is_vegetarian", "is_vegan", "is_gluten_free", "is_dairy_free",
+            "reviews.rating"
+        ],
```
ManyTheFish, line 141: what about using the other [`filterableAttributes` format](https://www.meilisearch.com/docs/reference/api/settings#example-17), this will allow you to activate ond deactivate some filter and facet search features.

for instance:
```jsonc
    {
      "attributePatterns": ["nutrition.*", "*_time_minutes"],
      "features": {
        "facetSearch": false,
        "filter": {
          "equality": true,
          "comparison": true
        }
      }
    },
    // [ .. ]
```

crates/meilisearch/src/routes/indexes/mod.rs
ManyTheFish: no it's ok for me

crates/meilisearch/src/routes/indexes/mod.rs
YoEight: alright I leveraged the `From<OrderBy>` implementation for `OrderBy` instead.

crates/meilisearch/tests/index/fields.rs
YoEight: done

PR #6083, opened 2025-12-31
Fix PR comments related to #6072

ManyTheFish:
## Related

Fixes #6072 PR comments

## Description

Modify search progress trace to:
- avoid panics when dropping `ScopedProgressStep` by replacing `debug_assert`s by warnings and errors
- avoid calling `Instant::now` when it's is not necessary
- No more tracing inside bucket sort to avoid performance impact (see below benchmarks)

### DIFF
--- a/crates/milli/src/progress.rs
+++ b/crates/milli/src/progress.rs
@@ -57,7 +57,7 @@ impl Progress {
             Ok(inner) => inner,
             Err(error) => {
                 tracing::error!("Failed to start progress step `{}`: {error}", sub_progress.name());
-                return UpdateStepStatus::Failed;
+                return UpdateStepStatus::NotUpdated;
             }
         };
         let InnerProgress { steps, durations } = &mut *inner;
@@ -88,7 +88,7 @@ impl Progress {
             Ok(inner) => inner,
             Err(error) => {
                 tracing::error!("Failed to end progress step `{}`: {error}", sub_progress.name());
-                return UpdateStepStatus::Failed;
+                return UpdateStepStatus::NotUpdated;
             }
         };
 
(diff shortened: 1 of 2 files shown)


ManyTheFish:
/bench workloads/search/*

⬇️ Benchmarks with bucket sort progress trace

ManyTheFish:
/bench workloads/search/*

⬇️ Benchmarks **without** bucket sort progress trace

crates/milli/src/progress.rs
```diff
@@ -400,7 +420,19 @@ pub struct ScopedProgressStep<'a, P: Step + Copy> {
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum UpdateStepStatus {
+    /// The step was updated.
+    Updated,
+    /// The step did not change.
+    NotUpdated,
+    /// The step failed to update because the RWLock failed to lock.
+    Failed,
+}
```
dureuill, line 447: I like having an explicit enum better than a boolean (more legible)

however, I realize I might have been unclear in my comments on the previous PR.

Not being able to lock a mutex is always a code error, and so should result in a panic.
However, we should make our best effort to ensure that the critical sections where the lock is held cannot panic.

So `UpdateStepStatus::Failed` should be dead code.
I won't be blocking the PR on this though.

dureuill (approved):
Thank you for the changes.

I'm a bit sad about the reduced progress granularity, I wonder if:

1. we could introduce a faster way to track time, perhaps using a specialized crate
2. we could introduce a dedicated "profiling" mode where it is understood that the metrics collection is going to take time

Anyway this is work for later, let's reap the benefits of this PR and the preceding one first 😊

dureuill (approved):
Perfect, thank you

crates/milli/src/progress.rs
dureuill: done

PR #6085, opened 2026-01-06
Fixes #6038 Engine do not resize the index

### LINKED ISSUE #6038: Engine do not resize the index

Vipul-045:
## Related issue

Fixes #6038 

- Introduced a centralized helper fun (**handle_store_mdb_error**) in crates/milli/src/error.rs to handle store errors consistently.

- The helper fun :
1. Passes through MDB_MAP_FULL unchanged, allowing the engine to detect it and resize the LMDB map.
2. Wraps all other LMDB errors into InternalError::StorePut / InternalError::StoreDeletion with proper context.
3. Passes through non-LMDB heed::Errors unchanged.
4. Updated all direct database.put / database.delete call sites to use this helper instead of wrapping errors inline.

### DIFF
--- a/crates/milli/src/search/mod.rs
+++ b/crates/milli/src/search/mod.rs
@@ -444,6 +444,7 @@ mod test {
     #[test]
     fn test_kanji_language_detection() {
         use crate::index::tests::TempIndex;
+        let progress = Progress::default();
 
         let index = TempIndex::new();
 
@@ -456,7 +457,7 @@ mod test {
             .unwrap();
 
         let txn = index.write_txn().unwrap();
-        let mut search = Search::new(&txn, &index);
+        let mut search = Search::new(&txn, &index, &progress);
 
         search.query("東京");
         let SearchResult { documents_ids, .. } = search.execute().unwrap();
@@ -468,6 +469,7 @@ mod test {
     #[test]
     fn test_hangul_language_detection() {
         use crate::index::tests::TempIndex;
+        let progress = Progress::default();
 
         let index = TempIndex::new();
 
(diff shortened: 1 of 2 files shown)


Vipul-045:
@Kerollmops check this out sir!

Vipul-045:
Happy New Year, and thank you for reviewing the PR! @Kerollmops 
I’ve already tested this manually with a smaller index size (20 MiB). I wasn’t able to reproduce the internal error that was occurring earlier when adding more documents. Please let me know if you’d like me to run any additional scenarios or share more detailed results.
I’ll take a look at the CI failures shortly and work on fixing them.
Thanks again!

Kerollmops:
Thanks @Vipul-045 😊

Were you able to add a debug print to see the engine react to the `MDBMapFull` error?

Looking forward to merging this PR.

Have a nice day 🌵

Vipul-045:
@Kerollmops I didn’t add a debug print specifically, but while testing with a reduced map size I observed that once the limit is reached, the MDB_MAP_FULL condition is handled and the engine resizes the map instead of surfacing an internal error as before.
If you’d like, I can add a temporary debug log to explicitly confirm the error path and share the output.
Thanks!

Vipul-045:
@Kerollmops I fixed the CI, it should merge now i guess

Kerollmops:
Hey @Vipul-045 👋

Thanks for the update. However, importing main through a commit doesn't work, and the CIs are not starting; I also don't have a button to run them. Can you rebase on main instead?

Have a nice day 🌵

PR #6088, opened 2026-01-08
PROTO: Introduce a Greedy exploration strategy for HNSW

ManyTheFish:


ManyTheFish:
/bench workloads/search/*

⬇️ BASE

ManyTheFish:
/bench workloads/search/*

⬇️ Greedy strategy

PR #6094, opened 2026-01-14
Fix filestore payload store

### LINKED ISSUE #6048: Make sure payloads are effectively deleted on error
We found out that some payloads are not removed from the disk when an error occurs before we are able to put them in the `FileStore`. I would suggest we should create a temporary temporary file and extract it at the last moment.

_Originally posted by @Kerollmops in https://github.com/meilisearch/meilisearch/pull/6000#discussion_r2602149151_

Vipul-045:
Fixes #6048 

Payload files could remain on disk when an error occurs before they are persisted into the FileStore.

This happens because temporary update files were previously created inside the FileStore directory, meaning that early failures could leave orphan files behind.

What this PR does
- Create update payloads using true temporary files (outside the FileStore directory)

### DIFF
--- a/crates/file-store/src/lib.rs
+++ b/crates/file-store/src/lib.rs
@@ -39,7 +39,7 @@ impl FileStore {
     /// Creates a new temporary update file.
     /// A call to `persist` is needed to persist the file in the database.
     pub fn new_update(&self) -> Result<(Uuid, File)> {
-        let file = NamedTempFile::new_in(&self.path)?;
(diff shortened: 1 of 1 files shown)


Vipul-045:
@Kerollmops check this out sir!

Kerollmops:
Thank you, @Vipul-045, but I don't understand why I see commits from your other PR, #6085. Could you please remove them from this PR?

PR #6095, opened 2026-01-14
Fix/engine resize

### LINKED ISSUE #6038: Engine do not resize the index

Vipul-045:
Fixes #6038 

Introduced a centralized helper fun (handle_store_mdb_error) in crates/milli/src/error.rs to handle store errors consistently.

The helper fun :
1. Passes through MDB_MAP_FULL unchanged, allowing the engine to detect it and resize the LMDB map.
2. Wraps all other LMDB errors into InternalError::StorePut / InternalError::StoreDeletion with proper context.
3. Passes through non-LMDB heed::Errors unchanged.
4. Updated all direct database.put / database.delete call sites to use this helper instead of wrapping errors inline.

### DIFF
--- a/crates/milli/src/search/mod.rs
+++ b/crates/milli/src/search/mod.rs
@@ -444,6 +444,7 @@ mod test {
     #[test]
     fn test_kanji_language_detection() {
         use crate::index::tests::TempIndex;
+        let progress = Progress::default();
 
         let index = TempIndex::new();
 
@@ -456,7 +457,7 @@ mod test {
             .unwrap();
 
         let txn = index.write_txn().unwrap();
-        let mut search = Search::new(&txn, &index);
+        let mut search = Search::new(&txn, &index, &progress);
 
         search.query("東京");
         let SearchResult { documents_ids, .. } = search.execute().unwrap();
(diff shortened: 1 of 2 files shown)


Vipul-045:
@Kerollmops Thanks for the clarification! I’ve rebased the branch on top of main and pushed the updated history. CI should now trigger correctly.

Kerollmops:
Actually, no, I can still see commit d19bd5e41e5e3818dbfcadedfbefbfb7ed878ddd, which is part of your previous PR. Could you cherry-pick your two commits: e92597a, and f2e2e6f, on this branch after resetting it, please?

Vipul-045:
@Kerollmops this PR fixes #6038 which requires [d19bd5e](https://github.com/meilisearch/meilisearch/commit/d19bd5e41e5e3818dbfcadedfbefbfb7ed878ddd) and the other two commits were just a addition and removal.

Vipul-045:
I confused you no worries and dont forget to add no-db-change label

PR #6097, opened 2026-01-14
Fixes filestore payload

### LINKED ISSUE #6048: Make sure payloads are effectively deleted on error

Vipul-045:
## Related issue

Fixes #6048

Payload files could remain on disk when an error occurs before they are persisted into the FileStore.

This happens because temporary update files were previously created inside the FileStore directory, meaning that early failures could leave orphan files behind.

What this PR does
- Create update payloads using true temporary files (outside the FileStore directory)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -601,7 +601,7 @@ source = "git+https://github.com/meilisearch/bbqueue#e8af4a4bccc8eb36b2b0442c4a9
 
 [[package]]
 name = "benchmarks"
-version = "1.32.2"
+version = "1.33.0"
 dependencies = [
  "anyhow",
  "bumpalo",
@@ -811,7 +811,7 @@ dependencies = [
 
 [[package]]
 name = "build-info"
-version = "1.32.2"
+version = "1.33.0"
 dependencies = [
  "anyhow",
  "time",
@@ -1835,7 +1835,7 @@ dependencies = [
 
 [[package]]
 name = "dump"
-version = "1.32.2"
+version = "1.33.0"
 dependencies = [
  "anyhow",
  "big_s",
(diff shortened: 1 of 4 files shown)


Vipul-045:
@Kerollmops check this

Vipul-045:
@Kerollmops I have moved the persisting of temporary file after transaction commits, let me know your thoughts now sir!

Kerollmops:
Hey @Vipul-045 👋

We recently thought about this, and we think we should still [persist the file](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html#method.persist) **before** we commit, but we should create a `NamedTempFile` again (from a TempPath and the file) so that if there is something wrong after the commit, dropping it will delete the file. Once we are sure everything is ok and we must keep the file, we should use the [`NamedTempFile::keep` method](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html#method.keep).

Vipul-045:
@Kerollmops Thanks for the detailed explanation sir!, Rewrapping the file in a NamedTempFile and only calling keep() once everything is confirmed sounds like a good approach. I’ll try to implement this by today or tomorrow and commit with an update.

Vipul-045:
Hey @Kerollmops , after checking [persist()](https://github.com/meilisearch/meilisearch/blob/main/crates/file-store/src/lib.rs), it currently returns a StdFile, which drops the TempPath. That makes it hard to re-wrap the file into a NamedTempFile, let me know if Im missing something here.

Kerollmops:
Hey @Vipul-045 👋 Have you looked at the [`NamedTempFile::into_parts` method](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html#method.into_parts), instead?

Vipul-045:
@Kerollmops That makes sense conceptually, but with our current implementation persist() immediately returns a StdFile and drops the TempPath, which makes the operation irreversible. So calling it early doesn’t bring any benefit right now, I think we need to change persist to preserve the deletion guard (e.g. return a TempPath) for this approach to work or we can call the persist later after we get sure everything is ok.

PR #6100, opened 2026-01-14
Implement parallel cleanup of old field IDs

Kerollmops:
This PR enhances [the algorithm to address inconsistencies in field-ID-based databases](https://github.com/meilisearch/meilisearch/pull/6076) by parallelizing it.

## Changelog

We improve the time it takes to perform the dumpless upgrade for instances before v1.32.0 by multi-threading the way we fetch entries in the database. By doing that, we noticed improvements from 2 hours and 50 minutes to a bit less than 7 minutes.

### DIFF
--- a/crates/milli/src/update/upgrade/v1_32.rs
+++ b/crates/milli/src/update/upgrade/v1_32.rs
@@ -196,7 +196,6 @@ fn fetch_keys_to_delete_in_parallel<'txn>(
                     let (_word, fid) =
                         StrBEU16Codec::bytes_decode(key_bytes).map_err(heed::Error::Decoding)?;
 
-                    // TODO optimize allocations
                     if fids_to_delete.contains(&fid) {
                         keys_to_delete.push(key_bytes.to_vec().into_boxed_slice());
                     }


crates/milli/src/update/upgrade/v1_32.rs
```diff
@@ -61,6 +68,171 @@ impl UpgradeIndex for CleanupFidBasedDatabases {
...
+
+        count += 1;
+    }
+
+    // We now push the last bound that
+    // defines the end of the last range
+    bounds.push(None);
+
+    // We create a thread pool and generate enough read transactions for each one of them.
+    let pool = rayon::ThreadPoolBuilder::new().num_threads(keys_by_thread).build()?;
```
ManyTheFish:
```suggestion
    let pool = rayon::ThreadPoolBuilder::new().num_threads(threads_count).build()?;
```

ManyTheFish (dismissed):

crates/milli/src/update/upgrade/v1_32.rs
```diff
@@ -61,6 +68,178 @@ impl UpgradeIndex for CleanupFidBasedDatabases {
...
+fn fetch_keys_to_delete_in_parallel<'txn>(
+    wtxn: &mut RwTxn<'txn>,
+    index: &Index,
+    database: Database<StrBEU16Codec, Unit>,
+    fids_to_delete: &BTreeSet<u16>,
+) -> Result<LinkedList<Vec<Result<Vec<Box<[u8]>>>>>> {
+    let fst = index.words_fst(wtxn)?;
+
+    // TODO get this number from the CLI parameters
+    let threads_count = rayon::current_num_threads() * 4;
```
ManyTheFish: Agree on that, that should be related to max indexing threads. There could be searches during the upgrade process

crates/milli/src/update/upgrade/v1_32.rs
```diff
@@ -61,6 +68,178 @@ impl UpgradeIndex for CleanupFidBasedDatabases {
...
+                let iter = database.remap_types::<Bytes, DecodeIgnore>().range(&rtxn, &bound);
+                for result in iter? {
+                    let (key_bytes, ()) = result?;
+                    let (_word, fid) =
+                        StrBEU16Codec::bytes_decode(key_bytes).map_err(heed::Error::Decoding)?;
+
+                    // TODO optimize allocations
+                    if fids_to_delete.contains(&fid) {
+                        keys_to_delete.push(key_bytes.to_vec().into_boxed_slice());
+                    }
```
ManyTheFish: we could decompose the key and have the following structure (1):
```
{ fid: 1, words: [ <all words to delete for this fid> ] }
{ fid: 2, words: [ <all words to delete for this fid> ] }
```

or the reverse one (2):
```
{ word: "hello", fids: [ <all fids to delete for "hello"> ] }
{ word: "world", fids: [ <all fids to delete for "world"> ] }
```

and recompose the keys during deletion process.

crates/milli/src/update/upgrade/v1_32.rs
Kerollmops: So it means that we will spawn no more than max indexing threads, even though most of the work here is loading pages and IO-based? I am not sure about my own TODO comment especially when we noticed that a large customer can use a lot of CPUs during "only" 7 minutes. What do you think?

crates/milli/src/update/upgrade/v1_32.rs
ManyTheFish: ⚠️  it's a quick calculation, and take into acount that all the words appears in all fields which is not really the case:
```
Base_size = word_count * (fid_count * average_word_size) + word_count * (fid_count * size_of_u16)
expected_size = word_count * average_word_size + word_count * (fid_count * size_of_u16)
maximum_gain = (fid_count - 1) * (word_count * average_word_size)
```

crates/milli/src/update/upgrade/v1_32.rs
```diff
@@ -61,6 +68,178 @@ impl UpgradeIndex for CleanupFidBasedDatabases {
...
+    wtxn: &mut RwTxn<'txn>,
+    index: &Index,
+    database: Database<StrBEU16Codec, Unit>,
+    fids_to_delete: &BTreeSet<u16>,
+) -> Result<LinkedList<Vec<Result<Vec<Box<[u8]>>>>>> {
+    let fst = index.words_fst(wtxn)?;
+
+    // TODO get this number from the CLI parameters
+    let threads_count = rayon::current_num_threads() * 4;
+    let keys_by_thread = (fst.len() / threads_count) + (fst.len() % threads_count);
```
YoEight: I think this formula doesn't distribute "evenly" work and can lead to thread underutilization. 

It think this one works better:

```rust
let keys_by_thread = fst.len().div_ceil(threads_count);
```

Let say the FST has 1,000 keys and we have 8 cores. The threadpool will be comprise of 32 threads. With the wrong formula we allocate 39 keys per threads, which utilizes only 25 threads (1,000 keys / 39).

With the other formula, we allocate 32 keys per threads, which utilizes 31 threads (1,000 keys / 32).

Granted, as the FST grows, the difference narrows, but it’s still worth mentioning.

crates/milli/src/update/upgrade/v1_32.rs
```diff
@@ -61,6 +68,178 @@ impl UpgradeIndex for CleanupFidBasedDatabases {
...
+
+        count += 1;
+    }
+
+    // We now push the last bound that
+    // defines the end of the last range
+    bounds.push(None);
+
+    // We create a thread pool and generate enough read transactions for each one of them.
+    let pool = rayon::ThreadPoolBuilder::new().num_threads(threads_count).build()?;
```
YoEight: We should create that pool only once, maybe in the `delete_old_fid_based_databases_from_fids` so we don't waste time re-creating a threadpool. `delete_old_fid_word_count_docids` is called 3 times, so we create 3 threadpools in total.

crates/milli/src/update/upgrade/v1_32.rs
```diff
@@ -61,6 +68,178 @@ impl UpgradeIndex for CleanupFidBasedDatabases {
...
+                let mut keys_to_delete = Vec::new();
+                let iter = database.remap_types::<Bytes, DecodeIgnore>().range(&rtxn, &bound);
+                for result in iter? {
+                    let (key_bytes, ()) = result?;
+                    let (_word, fid) =
+                        StrBEU16Codec::bytes_decode(key_bytes).map_err(heed::Error::Decoding)?;
+
+                    // TODO optimize allocations
+                    if fids_to_delete.contains(&fid) {
+                        keys_to_delete.push(key_bytes.to_vec().into_boxed_slice());
```
YoEight: Is that allocation optimization TODO comment about `keys_to_delete` doing multiple small allocations over time where we could have used `Vec::with_capacity` instead?

crates/milli/src/update/upgrade/v1_32.rs
```diff
@@ -61,6 +68,178 @@ impl UpgradeIndex for CleanupFidBasedDatabases {
...
+                    }
+                }
+
+                Ok(keys_to_delete) as crate::Result<_>
+            })
+            .collect_vec_list()
```
YoEight: Do we really need to do this? That's a lot of allocations for something that is poised to deletion anyway. why not just sending batch of keys to be deleted in a different thread instead? Like using a channel to communicate with the other thread?

Kerollmops: That's where there is a catch, LMDB doesn't allow doing write operations while **reading the current, uncommitted transaction**. That's what we do when we use the `Env::nested_read_txn` method, passing our write transaction as the parent. It allows us to make sure the FST is up to date with the database we are cleaning.

We could use multiple read transactions on the index to see what was in the database **before** the write transaction began. Still, it would be invalid as we are in the process of a dumpless upgrade, and the dumpless upgrade is processing all of the upgrades on a single index in a single write transaction to allow for rollbacks. If we could commit in between upgrades, we could indeed use a channel to send keys to delete to the main write transaction thread.

crates/milli/src/update/upgrade/v1_32.rs
Kerollmops: I plan to implement a data structure similar to the stacker at some point, as described in [this post](https://fulmicoton.com/posts/tantivy-stacker/). It is basically [an unrolled linked list](https://en.wikipedia.org/wiki/Unrolled_linked_list) with incrementally increasing fixed allocated blocks. The goal would be to store words with a size header to avoid large allocations.

I was also thinking about using [a `SpooledTempFile`](https://docs.rs/tempfile/latest/tempfile/fn.spooled_tempfile.html) for the last block when the memory limit is reached, enabling an append-only mode to send words again, but on disk this time. That would be a cool data structure to work on, but a bit out of scope for this PR. So, I'll remove my TODO comment for now.

crates/milli/src/update/upgrade/v1_32.rs
Kerollmops: You are right, it's much better. I forgot about `usize::div_ceil`, thank you!

crates/milli/src/update/upgrade/v1_32.rs
Kerollmops: I'll remove this comment for now, but I'll probably work on [a specialized data structure](https://github.com/meilisearch/meilisearch/pull/6100#discussion_r2693768509) to store keys on disk. Thank you.

crates/milli/src/update/upgrade/v1_32.rs
ManyTheFish: Sorry I come back late on this. 😞 

> So it means that we will spawn no more than max indexing threads, even though most of the work here is loading pages and IO-based?

Well, I don't have strong opinion on this, however, we have this max indexing threads config and it's wierd to me to not rely on it. 🤔

> I am not sure about my own TODO comment especially when we noticed that a large customer can use a lot of CPUs during "only" 7 minutes. What do you think?

Yes, I understand, but for these customer, the max indexing threads config will be set to an "high" value

PR #6103, opened 2026-01-14
Bump hannoy to 0.1.4-nested-rtxns

Kerollmops:
This PR bumps hannoy to v0.1.4-nested-rtxns.

### Changelog

We updated our internal vector store to speed up the dumpless upgrade by improving the graph rebuilding
and improve the speed and relevance of the search by using the explore factor as a limit to stop document searches rather than the query limit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2698,9 +2698,9 @@ dependencies = [
 
 [[package]]
 name = "hannoy"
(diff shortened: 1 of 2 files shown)


Kerollmops:
We must merge that for v1.34 👀

PR #6106, opened 2026-01-15
Use parallel deletion for field ID cleanup in standard path

### LINKED ISSUE #6105: Copy paste the parallel way of deleting useless field IDs
In #6100, we introduced a new method for deleting old field IDs from the database in parallel, which is a huge win. Now, I would like to see it used [in the standard path](https://github.com/meilisearch/meilisearch/blob/faa238064e5aa5ccfdcd2d7736fc9274f0bbf842/crates/milli/src/update/new/indexer/mod.rs#L549-L574), not just when performing a dumpless upgrade.

Note that we will need to duplicate the code so that we can freeze the dumpless upgrade code and prevent accidental breakage. We must also remove the threadpool creation as we are already running in a rayon threadpool.

calm329:
## Related issue

Fixes #6105

Ports the parallel field ID deletion algorithm from the dumpless upgrade path to the standard indexer path, as introduced in #6100. Duplicated the code to keep dumpless upgrade frozen, and removed ThreadPool creation since we're already in a rayon threadpool.

## Requirements

---

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=148254234

### DIFF
--- a/crates/milli/src/update/new/indexer/mod.rs
+++ b/crates/milli/src/update/new/indexer/mod.rs
@@ -1,4 +1,6 @@
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet, LinkedList};
+use std::iter;
+use std::ops::Bound;
 use std::sync::atomic::AtomicBool;
 use std::sync::{Arc, Once, RwLock};
 use std::thread::{self, Builder};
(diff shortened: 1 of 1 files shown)


calm329:
@Kerollmops Could you please review my PR?

PR #6109, opened 2026-01-16
feat: use parallel deletion for fid-based databases in standard path

### LINKED ISSUE #6105: Copy paste the parallel way of deleting useless field IDs

VedantMadane:
## Summary

Copies the parallel deletion method from the dumpless upgrade path (v1_32.rs) to the standard indexing path in mod.rs. This improves performance when deleting entries for field IDs that are no longer searchable.

## Changes

- Added \compute_fst_bounds()\ to divide work using FST word dictionary
- Added \etch_keys_to_delete_in_parallel()\ for parallel key fetching using nested read transactions
- Added \delete_old_word_fid_docids_parallel()\ to orchestrate parallel deletion
- Removed ThreadPool creation - uses existing rayon threadpool directly

## Notes

As requested in #6105, the code is duplicated (not imported) to freeze the dumpless upgrade code and prevent accidental breakage.

Closes #6105

### DIFF
--- a/crates/meilisearch/src/routes/indexes/mod.rs
+++ b/crates/meilisearch/src/routes/indexes/mod.rs
@@ -1,16 +1,6 @@
 use std::collections::BTreeSet;
 use std::convert::Infallible;
 
-use super::{
-    get_task_id, Pagination, PaginationView, SummarizedTaskView, PAGINATION_DEFAULT_LIMIT,
-};
-use crate::analytics::{Aggregate, Analytics};
-use crate::extractors::authentication::policies::*;
-use crate::extractors::authentication::{AuthenticationError, GuardedData};
-use crate::extractors::sequential_extractor::SeqHandler;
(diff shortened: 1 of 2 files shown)


Kerollmops (dismissed):
Thank you very much for your work @VedantMadane 💪
Merging this right now! Will be available in v1.34.0 👏 

EDIT: I'll merge once the CI (fmt) passes 😅

PR #6116, opened 2026-01-22
Bring back v1.33.1 changes into main

dureuill:
⚠️ This PR disables the `degraded_search_and_score_details_vector` test.

This is because the test is less relevant now that we return `limit` documents in a single iteration with linear scanning in the vector store.

As it is currently unclear how to make the test relevant again, and that merging the fixes from v1.33.1 into main is a priority, we ignore the test.


Kerollmops (dismissed):
Approuveed

Kerollmops (approved):

PR #6117, opened 2026-01-22
Remove openssl from deps

dureuill:
Openssl is not desirable to have as a dep of Meilisearch


dureuill:
> sorry for not checking the Cargo.lock earlier...

some things that could help us rely less on human perfection:

1. have some kind of CI rule forbidding dependencies on some crates
2. unified deps would prevent having different feature sets in different parts of the code maybe? (ideally `reqwest` would not pull `openssl` in its default config)

PR #6118, opened 2026-01-22
Fix geo-types version to avoid multiple rstar deps

dureuill:
geo-types > 0.7.17 is adding multiple `rstar` versions, unduly increasing the number of Meilisearch dependencies


Kerollmops (approved):
Looks good to me! Thank you. Next time bumping `geo-types` we will look closely to the different features 👀

PR #6124, opened 2026-01-26
Support the filterable, sortable and custom attributes in the new Settings Indexer

Kerollmops:
### Changelog

The new settings indexer, a more efficient version of our internal indexer for index settings modifications, now supports filterable, sortable, facet search, and custom (asc/desc) attributes in addition to the already-supported searchable, exact, proximity precision, and embedders. The new settings indexer can still be disabled by setting the `MEILI_EXPERIMENTAL_NO_EDITION_2024_FOR_SETTINGS` to `true`.

### DIFF
--- a/crates/milli/src/update/index_documents/extract/extract_geo_points.rs
+++ b/crates/milli/src/update/index_documents/extract/extract_geo_points.rs
@@ -102,27 +102,19 @@ fn extract_lat_lng(
             match (lat, lng) {
                 (Ok(lat), Ok(lng)) => Ok(Some([lat, lng])),
                 (Ok(_), Err(lng)) => {
-                    Err(Box::new(GeoError::BadLongitude {
-                        document_id: document_id(),
-                        value: lng,
-                    })
-                    .into())
+                    Err(Box::new(GeoError::BadLongitude { document_id: document_id(), value: lng })
+                        .into())
                 }
                 (Err(lat), Ok(_)) => {
-                    Err(Box::new(GeoError::BadLatitude {
-                        document_id: document_id(),
-                        value: lat,
-                    })
-                    .into())
+                    Err(Box::new(GeoError::BadLatitude { document_id: document_id(), value: lat })
+                        .into())
                 }
-                (Err(lat), Err(lng)) => {
-                    Err(Box::new(GeoError::BadLatitudeAndLongitude {
-                        document_id: document_id(),
-                        lat,
-                        lng,
-                    })
-                    .into())
-                }
+                (Err(lat), Err(lng)) => Err(Box::new(GeoError::BadLatitudeAndLongitude {
+                    document_id: document_id(),
+                    lat,
+                    lng,
+                })
+                .into()),
             }
         }
         None => Ok(None),


crates/milli/src/update/settings.rs
```diff
@@ -1600,14 +1586,14 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
...
-            criteria: Setting::NotSet,
+            filterable_fields: _,
+            sortable_fields: _,
+            criteria: _,
             stop_words: Setting::NotSet, // TODO (require force reindexing of searchables)
             non_separator_tokens: Setting::NotSet, // TODO (require force reindexing of searchables)
             separator_tokens: Setting::NotSet, // TODO (require force reindexing of searchables)
             dictionary: Setting::NotSet, // TODO (require force reindexing of searchables)
-            distinct_field: Setting::NotSet,
+            distinct_field: _,
```
Kerollmops, line 1643: Should I do something specific, or is it considered in the filterables?

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -258,6 +259,31 @@ where
...
+    }
+
+    // Clear the geo rtree entry when the geo support is removed
+    // TODO what about PatternMatch::Parent and please test that the geo rtree entry is cleared correctly
+    //      Shouldn't we clear the geo values from the facets databases or is it done above?
+    if settings_delta.old_match_faceted_field(RESERVED_GEO_FIELD_NAME) == PatternMatch::Match
+        && settings_delta.new_match_faceted_field(RESERVED_GEO_FIELD_NAME) == PatternMatch::NoMatch
+    {
+        index.delete_geo_rtree(wtxn)?;
+    }
```
Kerollmops, line 279: Does it make sense to use the newly introduced `old_match_faceted_field` method here?

crates/milli/src/update/settings.rs
ManyTheFish, line 1654: can we check if we handled it properly?

crates/milli/src/update/new/indexer/mod.rs
ManyTheFish: For both geo databases (RESERVED_GEOJSON_FIELD_NAME and RESERVED_GEO_FIELD_NAME), you could only keep the new_match_faceted_field, the cost would be to delete an already empty database.

crates/milli/src/update/settings.rs
ManyTheFish: change comment content
```suggestion
            localized_attributes_rules: Setting::NotSet, // TODO (require force reindexing of searchables)
```

crates/milli/src/update/new/extract/faceted/facet_document.rs
```diff
@@ -1,48 +1,24 @@
...
 pub fn extract_document_facets<'doc>(
     document: impl Document<'doc>,
-    field_id_map: &mut GlobalFieldsIdsMap,
-    filterable_attributes: &[FilterableAttributesRule],
-    sortable_fields: &HashSet<String>,
-    asc_desc_fields: &HashSet<String>,
-    distinct_field: &Option<String>,
+    // return the match result for the given field name.
+    match_field: impl Fn(&str) -> PatternMatch,
+    register_field: &mut impl FnMut(&str) -> Result<(FieldId, Metadata)>,
```
ManyTheFish, line 16: In the searchable extraction, we managed to merge these two closures into one named `should_tokenize` that get the field id from the field id map. and then return a `PatternMatch` with it as `Result<(FieldId, PatternMatch)>`

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -258,6 +259,31 @@ where
...
+    delete_old_fid_from_facet_databases(
+        wtxn,
+        index,
+        settings_delta,
+        must_stop_processing,
+        progress,
+    )?;
+
+    // Clear the cellulite database when the geojson support is removed
+    // TODO what about PatternMatch::Parent and please test that cellulite is cleared correctly
```
ManyTheFish: remaining todo

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -258,6 +259,31 @@ where
...
+    if settings_delta.old_match_faceted_field(RESERVED_GEOJSON_FIELD_NAME) == PatternMatch::Match
+        && settings_delta.new_match_faceted_field(RESERVED_GEOJSON_FIELD_NAME)
+            == PatternMatch::NoMatch
+    {
+        index.cellulite.clear(wtxn)?;
+    }
+
+    // Clear the geo rtree entry when the geo support is removed
+    // TODO what about PatternMatch::Parent and please test that the geo rtree entry is cleared correctly
+    //      Shouldn't we clear the geo values from the facets databases or is it done above?
```
ManyTheFish: remaining todo

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -507,6 +531,213 @@ where
...
+            for id in remove_from_facet_search.iter().copied() {
+                let mut iter = database.prefix_iter_mut(wtxn, &id.to_be_bytes())?;
+                while iter.next().transpose()?.is_some() {
+                    // safety: We don't keep any reference to the database.
+                    unsafe { iter.del_current()? };
+                }
+            }
+
+            db_progress.fetch_add(1, Ordering::Relaxed);
+        }
```
ManyTheFish, line 697: could be factorized with the upper deleting from all filters

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -507,6 +531,213 @@ where
...
+
+            for id in remove_comparison_levels_only.iter().copied() {
+                let mut iter = database.prefix_iter_mut(wtxn, &id.to_be_bytes())?;
+                while let Some((key, _ignored)) = iter.next().transpose()? {
+                    // We want to delete everything that is not from
+                    // the level 0 as it provides equality comparisons.
+                    //
+                    // The first two bytes corresponds to the field ID
+                    // while the third byte corresponds to the level.
+                    if key.get(2 + 1).copied() != Some(0) {
```
ManyTheFish: The first byte is at the index `0` so the third byte is at index `2` and not `3`:
```suggestion
                    if key.get(2).copied() != Some(0) {
```

To showcase this bug, I suggest adding a test that:
1) Add a filterable attribute on string values with the `comparison` and the `equality` features
2) Add documents
3) Remove the `comparison` feature from the filterable attribute
4) try a filter with an equality like `dog = intel`, we MUST find a matching document because `equality` is still available

ManyTheFish: On the implementation side, I suggest factorizing with the upper code only using prefix_iter on bytes

In the DeletingFromAllFilters and DeletingFromFacetsOnly we would prefix_iter on a slice of bytes correspounding to the fieldid only, and in DeletingFromComparisonsOnly we would prefix_iter on a slice of bytes correspounding to the fieldid + level 0.

crates/milli/src/update/new/extract/faceted/extract_facets.rs
```diff
@@ -563,4 +563,221 @@ impl FacetedDocidsExtractor {
...
+                depth,
+                value,
+            )
+        };
+
+        let current_document = document.current(
+            &context.rtxn,
+            context.index,
+            old_fields_ids_map.as_fields_ids_map(),
+        )?;
```
ManyTheFish, line 638: could be moved at line 645 to avoid getting the old_fields_ids_map twice

crates/milli/src/update/new/extract/faceted/extract_facets.rs
```diff
@@ -563,4 +563,221 @@ impl FacetedDocidsExtractor {
...
+                    || new_distinct;
+
+                if !is_old_faceted && is_new_faceted {
+                    PatternMatch::Match
+                } else {
+                    // We force the system to go down the rabbit hole
+                    // and avoid an early break if we return NoMatch
+                    PatternMatch::Parent
+                }
+            },
```
ManyTheFish, line 712: This looks similar to what we did in the facet databases deletion. This is a complex check so I would suggest to factorize it as much as possible to avoid forgetting a part of the code when modifying this strategy.

CF L667:
```
 // TODO do not duplicate this logic and put it in a function
//      for delete_old_fid_from_facet_databases to use it
```

crates/milli/src/update/new/extract/faceted/extract_facets.rs
```diff
@@ -563,4 +563,221 @@ impl FacetedDocidsExtractor {
...
+                        field_name: name.to_string(),
+                        process: "extract_document_facets",
+                    })
+                    .into()
+                })
+            },
+            &mut add,
+        )?;
+
+        // TODO do not duplicate the content of the extract_geo_document function
```
ManyTheFish: remaining TODO

crates/milli/src/update/new/indexer/extract.rs
```diff
@@ -595,6 +595,37 @@ where
         }
     }
 
+    'geo: {
+        let enabled_filterable_geo =
+            !settings_delta.old_filterable_rules().iter().any(|rule| rule.has_geo())
+                && settings_delta.new_filterable_rules().iter().any(|rule| rule.has_geo());
+        let enabled_geo = settings_delta.old_geo_fields_ids().is_none()
+            && settings_delta.new_geo_fields_ids().is_some();
```
ManyTheFish: this condition may be replaced by a check in the sortable instead

https://www.meilisearch.com/docs/learn/filtering_and_sorting/geosearch#sorting-results-with-_geopoint

crates/milli/src/update/new/indexer/extract.rs
```diff
@@ -626,6 +626,31 @@ where
             &indexing_context.must_stop_processing,
         )?;
     }
+
+    'cellulite: {
+        let enabled_filterable_geojson =
+            !settings_delta.old_filterable_rules().iter().any(|rule| rule.has_geojson())
+                && settings_delta.new_filterable_rules().iter().any(|rule| rule.has_geojson());
+        let enabled_geojson = settings_delta.old_geojson_field_id().is_none()
+            && settings_delta.new_geojson_field_id().is_some();
```
ManyTheFish: this condition may be removed because:
> It is not possible to sort documents based on the _geojson attribute.

https://www.meilisearch.com/docs/learn/filtering_and_sorting/geosearch#sorting-results-with-_geopoint

crates/meilisearch/tests/common/server.rs
```diff
@@ -445,7 +445,7 @@ impl<State> Server<State> {
     pub async fn wait_task(&self, update_id: u64) -> Value {
         // try several times to get status, or panic to not wait forever
         let url = format!("/tasks/{update_id}");
-        let max_attempts = 400; // 200 seconds in total, 0.5secs per attempt
```
ManyTheFish, line 490: which test and why?

ManyTheFish (changes requested):

crates/milli/src/update/settings.rs
Kerollmops: Taken into account and the facet trees are generated based on the metadata.

crates/milli/src/update/settings.rs
Kerollmops: Done, it's correctly handled in extract all settings.

crates/milli/src/update/new/indexer/extract.rs
Kerollmops: Thanks, fixed

crates/milli/src/update/new/indexer/extract.rs
Kerollmops: You are so fast!

crates/milli/src/update/new/extract/faceted/extract_facets.rs
Kerollmops: Thanks, I extracted this into [a dedicated issue](https://github.com/meilisearch/meilisearch/issues/6329).

crates/milli/src/update/new/indexer/mod.rs
Kerollmops: Thanks, I fixed the bug, added a test, and created [a dedicated issue](https://github.com/meilisearch/meilisearch/issues/6330) to refactor the code.

crates/milli/src/update/new/extract/faceted/extract_facets.rs
Kerollmops: Done 👍

crates/meilisearch/tests/common/server.rs
Kerollmops: This is no longer relevant, tests are passing now.

crates/milli/src/update/new/indexer/post_processing/mod.rs
```diff
@@ -148,6 +148,7 @@ fn compute_word_fst(
 
     let words_fst = index.words_fst(&rtxn)?;
     let mut word_fst_builder = WordFstBuilder::new(&words_fst)?;
+    // TODO Are you sure you want to use a rtxn, here?
```
Kerollmops: @ManyTheFish, can we discuss this together? I think it's for our future self when we implement more settings. To confirm.

crates/milli/src/update/new/extract/faceted/facet_document.rs
Kerollmops: I prefer keeping that this way for now and we can improve later.

PR #6126, opened 2026-01-27
Return Pagination fields in /fields route

ManyTheFish:
#### ⚠️ **Breaking**: The `/field` route now returns an object containing the `results`, `offset`, `limit`, and `total` fields instead of returning an array of results.

<details>
	<summary>Example: before/after</summary>

Before:
```jsonc
[
  // fields ..
]
```

After:
```jsonc
{
  "results": [
    // [fields .. ]
  ],
  "offset": 0,
  "limit": 20,
  "total": 0
}
```

</details>

#### 🐛 Fix pattern filtering when a parent field was matching a child pattern: `title` was matching the pattern `title.to*`.

<details>
	<summary>Example: before/after</summary>

For the following request:
```jsonc
{
  "filter": {
    "attributePatterns": [
      "title.en"
    ]
  }
}
```

Before Response, `title` field is returned:
```jsonc
{
  "results": [
    {
      "name": "title",
      //metadata...
    },
    
  ],
  "offset": 0,
  "limit": 20,
  "total": 1
}
```

After Response, nothing returned:
```jsonc
{
  "results": [],
  "offset": 0,
  "limit": 20,
  "total": 0
}
```

</details>

## Related issue

Fixes https://linear.app/meilisearch/issue/EXP-856

## Changes

- Move the fields route implementation in a dedicated file (fields.rs)
- Use the `Pagination` helper to build the return value
- Fix a small filter bug when `title` was matching the pattern `title.to*` because we were matching parent fields.

## Documentation

This change breaks the response value of the `/fields` route:
```jsonc
// BEFORE:
[
  // [fields .. ]
]


// AFTER
{
  "results": [
    // [fields .. ]
  ],
  "offset": 0,
  "limit": 20,
  "total": 0
}
```

### DIFF
--- a/crates/meilisearch/src/routes/indexes/fields.rs
+++ b/crates/meilisearch/src/routes/indexes/fields.rs
@@ -159,7 +159,10 @@ impl ListFields {
     fn apply_filter(&self, field: &Field) -> bool {
         if let Some(filter) = &self.filter {
             if let Some(patterns) = &filter.attribute_patterns {
-                if matches!(patterns.match_str(field.name), PatternMatch::NoMatch) {
+                if matches!(
+                    patterns.match_str(field.name),
+                    PatternMatch::NoMatch | PatternMatch::Parent
+                ) {
                     return false;
                 }
             }
--- a/crates/meilisearch/tests/index/fields.rs
+++ b/crates/meilisearch/tests/index/fields.rs
@@ -61,6 +61,22 @@ async fn get_fields_with_documents_and_filters() {
     insta::assert_json_snapshot!((code.as_u16(), response), {
         "[1]" => insta::sorted_redaction(),
     });
+
+    // search parameter should filter
+    // no field should match
+    let (response, code) = index
+        .fields(&ListFieldsPayload {
+            filter: Some(ListFieldsFilterPayload {
+                attribute_patterns: Some(&["title.to*"]),
+                ..Default::default()
+            }),
+            ..Default::default()
+        })
+        .await;
+
+    insta::assert_json_snapshot!((code.as_u16(), response), {
+        "[1]" => insta::sorted_redaction(),
+    });
 }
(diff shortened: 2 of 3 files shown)


ManyTheFish:
@dureuill I've added a collapsible example

YoEight:
Suggestion: This approach collects everything upfront, including items we'll discard. Since this endpoint isn't performance-critical, that's acceptable for now, but it raises the question of whether PaginationView is the right abstraction here.

dureuill:
@YoEight I agree that the collect is unfortunate, however it is not required for `PaginationView` itself. 

The fundamental difference with the previous version of the code is that we are now computing the total number of elements matching the input parameters.

To do this we have to iterate over all elements because we cannot know how many elements will be retained by the filter before applying it.

If we really wanted to avoid collecting elements we then discard, we could iterate twice: once to find out the total, then a second time using `PaginationView::auto_paginate_unsized`, passing the total we computed the first time.

We could also feasibly add yet another `PaginationView` constructor that would accept any iterator and count the total number of elements:

```rust
    /// Given an iterator, returns the selection section, computing the total number of elements at the same time.
    ///
    /// The passed iterator will be entirely consumed to count the total number of elements, so if the iterator
    /// implements `ExactSizeIterator`, prefer [`Self::auto_paginate_sized`].
    ///
    /// If the total is available by any other means, prefer [`Self::auto_paginate_unsized`].
    pub fn auto_paginate_auto_size<T>(
        self,
        content: impl IntoIterator<Item=T>,
    ) -> PaginationView<T> where T:Serialize
    {
        let mut total = 0;
        let mut it = content.into_iter().inspect(|_t| total += 1);
        let content = (&mut it).skip(self.offset).take(self.limit).collect();
        it.last();
        self.format_with(total, content)
    }
```

🌱

PR #6136, opened 2026-01-30
Do not exclude explicitly null fields when sorting

chrisbenincasa:
## Related issue

None yet - something I noticed in my project that deployed a local Meilisearch on user's machines: https://github.com/chrisbenincasa/tunarr

Happy to open one to discuss further if necessary

## Requirements

    - [ ] Test that any impacted DB still works as expected after using `--experimental-dumpless-upgrade` on a DB created with the last released Meilisearch **This seemed to work for me but I can perform something more official if necessary**
- [ ] If necessary, the feature have been tested in the Cloud production environment (with [prototypes](./documentation/prototypes.md)) and the Cloud UI is ready. **Don't know know to do this**

## Open Questions
* Is there an implication that I may not be aware off to adding these null fields to the sortable index if necessary?

### DIFF
--- a/crates/meilisearch/tests/documents/get_documents.rs
+++ b/crates/meilisearch/tests/documents/get_documents.rs
@@ -1602,3 +1602,170 @@ async fn get_document_sort_desc_includes_docs_without_field() {
     }
     "###);
 }
+
+#[actix_rt::test]
+async fn get_document_sort_with_null_values() {
+    let server = Server::new_shared();
+    let index = server.unique_index();
+    let (task, _code) = index.create(None).await;
+    server.wait_task(task.uid()).await.succeeded();
+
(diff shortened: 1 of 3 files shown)


Kerollmops:
Hey @chrisbenincasa 👋

Sorry for the very late reply, but I don't know if we want to merge this PR, as it changes semantics, and we would prefer to make this decision on our end with the Product team.

Thank you anyway for this work and the interest you have in Meilisearch,
Have a nice day 🌵

chrisbenincasa:
Thanks for getting back to me, even if the answer is a bit unfortunate. Just noting that I found the behavior very surprising as it doesn't align with how most storage or search systems treat null (that I know of) and is also effectively a filter operation when merely sorting.

PR #6154, opened 2026-02-12
Introduce the `AttributeRank` and `AttributePosition` criteria

Kerollmops:
This PR introduces two new ranking rules: the `AttributeRank` and the `AttributePosition`.

## Changelog

We released two new ranking rules that Meilisearch had already been using internally for the `attribute` one, which is basically both ranking rules applied one after the other:
- `attributeRank`: A document is considered better if the query words match in a higher searchable attribute. It ignores the position of the query words in this attribute.
- `attributePosition`: A document is considered better if the query words match closer to the beginning of an attribute. The attribute rank is ignored by this rule.

## To Do

- [x] Disallow using the `AttributeRank` or `AttributePosition` along with the `Attribute` rule when changing the settings.
- [x] Add some tests.
- [x] Create a PR for the documentation to introduce those two new rules. https://github.com/meilisearch/documentation/pull/3462
- [x] In a follow-up PR, change the default ranking rules to replace the `Attribute` rule with the `AttributeRank` and `AttributePosition` with the `Sort` one in between. https://github.com/meilisearch/meilisearch/pull/6155

## Generative AI tools

### DIFF
--- a/crates/meilisearch/tests/search/mod.rs
+++ b/crates/meilisearch/tests/search/mod.rs
@@ -831,6 +831,87 @@ async fn test_score_details() {
 }
 
 #[actix_rt::test]
+async fn test_score_details_with_attribute_rank_and_position() {
+    let server = Server::new_shared();
+    let index = server.unique_index();
+
+    index
+        .update_settings(json!({
+        "filterableAttributes": ["id", "title"],
+        "sortableAttributes": ["id", "title"],
+        "rankingRules": [
+            "words",
+            "typo",
+            "proximity",
+            "attributeRank",
+            "attributePosition",
+            "exactness",
+    ] }))
+        .await;
+
+    let documents = DOCUMENTS.clone();
+    let (response, _code) = index.add_documents(documents, None).await;
+    server.wait_task(response.uid()).await.succeeded();
+
+    index
+        .search(
+            json!({
+                "q": "train dragon",
+                "showRankingScoreDetails": true,
+            }),
+            |response, code| {
+                snapshot!(code, @"200 OK");
+                snapshot!(json_string!(response["hits"]), @r###"
+                [
+                  {
+                    "title": "How to Train Your Dragon: The Hidden World",
+                    "id": "166428",
+                    "color": [
+                      "green",
+                      "red"
+                    ],
+                    "_rankingScoreDetails": {
+                      "words": {
+                        "order": 0,
+                        "matchingWords": 2,
+                        "maxMatchingWords": 2,
+                        "score": 1.0
+                      },
+                      "typo": {
+                        "order": 1,
+                        "typoCount": 0,
+                        "maxTypoCount": 2,
+                        "score": 1.0
+                      },
+                      "proximity": {
+                        "order": 2,
+                        "score": 0.75
+                      },
+                      "attributeRank": {
+                        "order": 3,
+                        "score": 1.0
+                      },
+                      "attributePosition": {
+                        "order": 4,
+                        "score": 0.8095238095238095
+                      },
+                      "exactness": {
+                        "order": 5,
+                        "matchType": "noExactMatch",
+                        "matchingWords": 2,
+                        "maxMatchingWords": 2,
+                        "score": 0.3333333333333333
+                      }
+                    }
+                  }
+                ]
+                "###);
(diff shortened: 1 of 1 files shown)


crates/milli/src/criterion.rs
```diff
@@ -27,19 +27,31 @@ pub enum CriterionError {
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
 pub enum Criterion {
     /// Sorted by decreasing number of matched query terms.
-    /// Query words at the front of an attribute is considered better than if it was at the back.
+    /// Query words at the front of an attribute is considered
```
dureuill: drive-by, but since you reformat: isn't this dependent on the matchingStrategy?

crates/milli/src/criterion.rs
```diff
@@ -27,19 +27,31 @@ pub enum CriterionError {
...
-    /// Documents with quey words contained in more important
-    /// attributes are considered better.
+    /// Documents with query words contained in more important
+    /// attributes and at a closer-to-the-front position in it
+    /// are considered better.
     Attribute,
-    /// Dynamically sort at query time the documents. None, one or multiple Asc/Desc sortable
-    /// attributes can be used in place of this criterion at query time.
+    /// Documents with query words contained in more important
+    /// attributes only are considered better. Position of the
```
dureuill: why `only`? I think they are considered better even if they *also* appear in less important attributes. That sentence seems to implies the reverse

crates/meilisearch-types/src/settings.rs
```diff
@@ -1081,9 +1100,18 @@ pub enum RankingRuleView {
...
     /// Sorted by increasing distance between matched query terms.
     Proximity,
-    /// Documents with quey words contained in more important
-    /// attributes are considered better.
+    /// Documents with query words contained in more important
+    /// attributes and at a closer-to-the-front position in it
+    /// are considered better.
     Attribute,
+    /// Documents with query words contained in more important
+    /// attributes only are considered better. Position of the
```
dureuill: same as above; remove only

crates/meilisearch-types/src/settings.rs
dureuill: same as above (also drive-by)

crates/milli/src/search/new/tests/sort.rs
```diff
@@ -344,10 +345,18 @@ fn test_redacted() {
...
     ]);
 
+    let separated = index
+        .criteria(&txn)
+        .unwrap()
+        .iter()
+        .any(|r| matches!(r, Criterion::AttributeRank | Criterion::AttributePosition));
+    let attribute_state =
+        if separated { AttributeState::Separated } else { AttributeState::Unified };
+
```
dureuill: repeated twice (one in test and one in main code) => worth having a `AttributeState::from_criteria(impl Iterator<Item=Criterion>>)` so that this becomes:

```
let attribute_state = AttributeState::from_criteria(index.criteria(&txn).unwrap().iter())
```

?

crates/milli/src/error.rs
```diff
@@ -432,6 +432,9 @@ and can not be more than 511 bytes.", .document_id.to_string()
     InvalidChatSettingsDocumentTemplateMaxBytes,
     #[error("{0}")]
     DocumentEmbeddingError(String),
+    #[error("Mixed usage of the Attribute, AttributeRank, and AttributePosition ranking rules. \
+        You must either use the Attribute ranking rule alone or the AttributeRank and AttributePosition ranking rules.")]
```
dureuill: shouldn't the rule name be camel cased?


```suggestion
    #[error("Mixed usage of the attribute, attributeRank, and attributePosition ranking rules. \
        You must either use the attribute ranking rule alone or the attributeRank and attributePosition ranking rules.")]
```

Kerollmops: Ho! Yeah! Good catch!

dureuill:
> [pascal case](https://github.com/meilisearch/meilisearch/pull/6154/commits/c868ad5c6140a29ef6b5bff2586fe4076b9a1188)

Careful, this commit moves from `PascalCase` to **`camelCase`**, not the reverse like the commit message claims :-p

PR #6128, opened 2026-01-27
Replicated sharding

dureuill:
# Changelog

> [!IMPORTANT]  
> This release contains [**breaking changes**](#replicated-sharding-breaking-changes) for users of the `network` experimental feature.

## <a id="replicated-sharding"></a> Replicated sharding

> [!NOTE]
> Replicated sharding requires Meilisearch Enterprise Edition (EE).
> - Users of Meilisearch Cloud, please contact support if you need replicated sharding.
> - Users of the Community Edition, please contact the sales if you want to use replicated sharding in production.

### <a id="replicated-sharding-breaking-changes"></a> Breaking changes

- `network` objects sent to the `PATCH /network` route must now contain at least one `shard` object containing at least one remote when `leader` is not `null`.

Existing databases will be migrated automatically when upgraded with `--experimental-dumpless-upgrade` when `leader` is not `null`, such that for each remote:
1. A shard with the same name as the remote is created
2. This shard has exactly one remote in its `remotes` list: the remote with the same name as the shard.

This change will not cause any document to be resharded.

To be able to upgrade without resharding, the migration uses the same name for remotes and for shards. However, in new configurations, we recommend using different names for shards and remotes.

<details>

<summary>Example of migration</summary>

For instance, the following network object:

```jsonc
{
  "leader": "ms-00",
  "self": "ms-01",
  "remotes": {
    "ms-00": { /* .. */ },
    "ms-01": { /* .. */ }
  }
}
```

is converted to:

```jsonc
{
  "leader": "ms-00",
  "self": "ms-01",
  "remotes": {
    "ms-00": { /* .. */ },
    "ms-01": { /* .. */ }
  },
  "shards": {  // ✨ NEW
    "ms-00": {  // shard named like the remote
      "remotes": ["ms-00"] // is owned by the remote
    },
    "ms-01": {
      "remotes": ["ms-01"]
    }
  }
}
```

</details>

### Addition of `network.shards`

The `network` object for routes `PATCH /network`  and `GET /network` now contains the new field `shards`, which is an object whose values are `shard` objects, and keys the name of each shard.

Each `shard` object contains a single field `remotes`, which is an array of strings, each string representing the name of an existing remote. 

#### Convenience fields

The `shard` objects in `PATCH /network` contain the additional fields `addRemotes` and `removeRemotes` meant for convenience:

- pass an array of remote names to `shard.addRemotes` to add these remotes to the list of remotes of a shard.
[...]

PR #6156, opened 2026-02-12
Add dynamic search rules persistence module

YoEight:
## Related issue

Fixes https://linear.app/meilisearch/issue/EXP-892/implement-rule-storage
Fixes https://linear.app/meilisearch/issue/EXP-891/define-rule-data-structures

## Generative AI tools

## Requirements

    - [x] Set the `db change` label.

### DIFF
--- a/crates/index-scheduler/src/dynamic_search_rules.rs
+++ b/crates/index-scheduler/src/dynamic_search_rules.rs
@@ -1,0 +1,52 @@
+use std::sync::{Arc, RwLock};
+
+use meilisearch_types::dynamic_search_rules::{DynamicSearchRule, DynamicSearchRules};
+use meilisearch_types::heed::types::{SerdeJson, Str};
+use meilisearch_types::heed::{Database, Env, RwTxn, WithoutTls};
+
+use crate::Result;
+
+const NUMBER_OF_DATABASES: u32 = 1;
+
+mod db_name {
+    pub const DYNAMIC_SEARCH_RULES: &str = "dynamic-search-rules";
+}
+
+#[derive(Clone)]
+pub(crate) struct DynamicSearchRulesStore {
+    pub(crate) persisted: Database<Str, SerdeJson<DynamicSearchRule>>,
+    runtime: Arc<RwLock<DynamicSearchRules>>,
+}
+
+impl DynamicSearchRulesStore {
+    pub(crate) const fn nb_db() -> u32 {
+        NUMBER_OF_DATABASES
+    }
+
+    pub fn new(env: &Env<WithoutTls>, wtxn: &mut RwTxn) -> Result<Self> {
+        let persisted = env.create_database(wtxn, Some(db_name::DYNAMIC_SEARCH_RULES))?;
+        let rules: DynamicSearchRules = persisted
+            .iter(wtxn)?
+            .filter_map(|entry| entry.ok())
+            .map(|(key, rule): (&str, DynamicSearchRule)| (key.to_string(), rule))
+            .collect();
+
+        Ok(Self { persisted, runtime: Arc::new(RwLock::new(rules)) })
+    }
+
+    pub fn put(&self, mut wtxn: RwTxn, value: DynamicSearchRules) -> Result<()> {
+        self.persisted.clear(&mut wtxn)?;
+        for (uid, rule) in &value {
+            self.persisted.put(&mut wtxn, uid, rule)?;
+        }
+        wtxn.commit()?;
+
+        let mut runtime = self.runtime.write().unwrap();
+        *runtime = value;
+        Ok(())
+    }
+
+    pub fn get(&self) -> DynamicSearchRules {
+        DynamicSearchRules::clone(&*self.runtime.read().unwrap())
+    }
+}
--- a/crates/meilisearch-types/src/dynamic_search_rules.rs
+++ b/crates/meilisearch-types/src/dynamic_search_rules.rs
@@ -1,0 +1,289 @@
+use std::collections::BTreeMap;
+
+use serde::{Deserialize, Serialize};
+use time::OffsetDateTime;
+
+pub type DynamicSearchRules = BTreeMap<String, DynamicSearchRule>;
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct DynamicSearchRule {
+    pub uid: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub description: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub priority: Option<u64>,
+    #[serde(default)]
+    pub active: bool,
+    #[serde(default)]
+    pub conditions: Vec<Condition>,
+    pub actions: Vec<RuleAction>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(tag = "scope", rename_all = "camelCase")]
+pub enum Condition {
+    Query(QueryCondition),
+    Time(TimeCondition),
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct QueryCondition {
+    pub is_empty: bool,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct TimeCondition {
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        with = "time::serde::rfc3339::option"
+    )]
+    pub start: Option<OffsetDateTime>,
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        with = "time::serde::rfc3339::option"
+    )]
+    pub end: Option<OffsetDateTime>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct RuleAction {
+    pub selector: Selector,
+    pub action: Action,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct Selector {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub index_uid: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub id: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub filter: Option<serde_json::Value>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(tag = "type", rename_all = "camelCase")]
+pub enum Action {
+    Pin(PinArgs),
+    Boost(BoostArgs),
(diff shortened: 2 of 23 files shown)


crates/meilisearch-types/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,376 @@
...
+use serde::{Deserialize, Serialize};
+
+pub type DynamicSearchRules = BTreeMap<String, DynamicSearchRule>;
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct DynamicSearchRule {
+    /// data format version to support upcasting if the format evolves over time.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub version: Option<String>,
```
dureuill: data that lives in the index-scheduler is already versioned and does not need a separate version.

See e.g.  <https://github.com/meilisearch/meilisearch/blob/main/crates/index-scheduler/src/upgrade/mod.rs#L44> for details

crates/meilisearch-types/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,376 @@
...
+    pub uid: String,
```
dureuill, line 11: More of a note for the route implementation, but we probably want something similar to `IndexUid` here, since we'll be using the `uid` in `GET/UPDATE` routes.

@Kerollmops should know about the rationale for the limits on index uid formats, but I think it was to avoid having to use url encoding in the URL?

If so, same rationale applies here for rule uid

crates/meilisearch-types/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,376 @@
...
+    pub filter: Option<Filter>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct Filter {
+    pub attribute: String,
+    pub op: FilterOp,
+    pub value: String,
+}
```
dureuill: the filter here should be a `serde_json::Value` as in e.g. <https://github.com/meilisearch/meilisearch/blob/main/crates/meilisearch/src/routes/indexes/documents.rs#L490>.

The route will have verified that the filter is syntactically correct.

crates/meilisearch-types/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,376 @@
...
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct Selector {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub index_uid: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub id: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub filter: Option<Filter>,
```
dureuill: We need to have the semantics for filter matching nailed down before inserting it in the selector I think

crates/meilisearch-types/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,376 @@
...
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(tag = "name", content = "args", rename_all = "camelCase")]
+pub enum Action {
+    Pin(PinArgs),
+    Boost(BoostArgs),
+    Bury(BuryArgs),
+    Hide(HideArgs),
+}
```
dureuill, line 78: should we have `filter` injection as an action here?

crates/meilisearch-types/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,376 @@
...
+            ]
+        });
+
+        let rule: DynamicSearchRule = serde_json::from_value(json).unwrap();
+        assert_eq!(rule.version, None);
+        assert_eq!(rule.description, None);
+        assert_eq!(rule.priority, None);
+        assert!(!rule.active);
+        assert!(rule.conditions.is_empty());
+    }
```
dureuill: we usually use `insta` snapshots rather than `asserts` in most tests

crates/meilisearch-types/src/dynamic_search_rules.rs
YoEight: Phase 1 doesn't include query filters for now.

crates/meilisearch-types/src/dynamic_search_rules.rs
YoEight: That was my plan, I didn't think it was needed but I'm willing to make the changej.

YoEight: done

YoEight:
@dureuill I switched the base branch to `dynamic-search-rules`

crates/index-scheduler/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub fn new(env: &Env<WithoutTls>, wtxn: &mut RwTxn) -> Result<Self> {
+        let persisted = env.create_database(wtxn, Some(db_name::DYNAMIC_SEARCH_RULES))?;
+        let rules: DynamicSearchRules =
+            persisted.get(wtxn, db_keys::DYNAMIC_SEARCH_RULES)?.unwrap_or_default();
+
+        Ok(Self { persisted, runtime: Arc::new(RwLock::new(rules)) })
+    }
+
+    pub fn put(&self, mut wtxn: RwTxn, value: DynamicSearchRules) -> Result<()> {
+        self.persisted.put(&mut wtxn, db_keys::DYNAMIC_SEARCH_RULES, &value)?;
```
dureuill: re-reading the specs, I believe on the route side we want to be able to `POST/GET/UPDATE/DELETE` rules individually, is that correct?

In that case, I think that storing each rule with its id as its own key in the `db_name::DYNAMIIC_SEARCH_RULES` db might make the implementation of the update operations much easier. Refer to `chat` for this.

On the other hand, the rules will all be accessed to check if they apply on every search, so it might make sense to load them in a single database read, as iterating on all keys is going to be very costly I/O-wise 🤔 

What are your thoughts? Should we:

1. keep the atomic POST/GET/UPDATE/DELETE operations and a single key for the rules, and paper over the discrepancy in code (e.g. UPDATE'ng a rule would be: (1) fetch all the rules (2) address the modified rule in the list (3) update it (4) rewrite the full list). As we want the maximum performance at search time this might makes sense.
2. keep the atomic POST/GET/UPDATE/DELETE operations, but move to one DB key per rule. So UPDATE'ng would be: (1) put back the updated rule addressed by its id. This might be detrimental to search performance though
3. have the POST/GET/UPDATE operations operate on all the rules (like we do for the `network`) and have a single key for the rules.

crates/meilisearch-types/src/dynamic_search_rules.rs
dureuill: ah OK. It looks like filter injection is a low-hanging fruit though: you iterate on query, and use the `fuse_filters` function on them (the function is introduced in https://github.com/meilisearch/meilisearch/pull/6128/changes/c53c5f406df6de1da87a98efadf05472402ff39e, you may grab its code if we're racing)

crates/meilisearch-types/src/snapshots/meilisearch_types__dynamic_search_rules__tests__boost.snap
```diff
@@ -0,0 +1,10 @@
+---
```
dureuill, line 1: note: you may use inline snapshots for smaller snapshots

crates/meilisearch-types/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,289 @@
...
+pub struct RuleAction {
+    pub selector: Selector,
+    pub action: Action,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct Selector {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub index_uid: Option<String>,
```
dureuill, line 64: should we allow a list of index uid patterns here?

crates/meilisearch-types/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,289 @@
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub id: Option<String>,
```
dureuill, line 66: `docid`?

crates/meilisearch-types/src/dynamic_search_rules.rs
dureuill: done, thank you

crates/meilisearch-types/src/dynamic_search_rules.rs
dureuill: done, thank you

crates/meilisearch-types/src/dynamic_search_rules.rs
dureuill: thank you

crates/index-scheduler/src/dynamic_search_rules.rs
YoEight: Initially, I was in favor of one single LMDB key for all the rules but reading your comment I realize that we can go with one key per rule, combined with an in-memory cache.

We can avoid I/O at search time if we leverage caching:

* We ca iterate all keys once and load the rules into memory at startup
* On any rule mutation, we update LMDB then refresh the cache
* At search time, we read from memory.

That approach still provides simple CRUD operations while keeping search-time access fast. The spec already mentions caching under Rule Resolution, so this aligns with the design.

LMDB becomes just persistence and recovery, not the hot path.

What do you think?

crates/meilisearch-types/src/dynamic_search_rules.rs
YoEight: But how it will correlate with the document ids? `index_uid`s are unique but doc_ids are only unique within an index.

crates/meilisearch-types/src/dynamic_search_rules.rs
YoEight: If you think it's clearer that way, I'm all for it

crates/index-scheduler/src/dynamic_search_rules.rs
dureuill: that looks good.

In general we should be wary because this approach means that the readers (search) will be blocked by the writer (updating rules), however in this case I think it is OK because rewriting a rule is going to be a very fast operation.

crates/meilisearch-types/src/dynamic_search_rules.rs
dureuill: I agree with your observation. I guess what I don't understand is why we want the selector to be unique.

assuming we allow `indexUids` as an array like I was suggesting, we could have:

```jsonc
{
  "selector": {
    "indexUids": ["toto", "titi"],
    "docId": 42
  }
}
```

That would mean that any action defined in this rule would apply to document 42 in the index toto and to document 42 in the index titi.

crates/meilisearch-types/src/dynamic_search_rules.rs
dureuill: for some reason I was sure that the API was using `docId`, but both `/similar` and `/documents/fetch` use `id` so for consistency we should keep `id` here.

PR #6160, opened 2026-02-16
Attempt to fix task DB corruption on snapshots

dureuill:
When the task DB is under heavy writes, the new S3 snapshotting method appears to cause the corruption of that DB.

This PR locks the task DB for writing and uses the LMDB-native way of copying the environment to avoid the concurrent writes leading to the corruption.

🌱 This PR was created without the use of generative AI tooling


Kerollmops (approved):
Looks good to me 💯

PR #6164, opened 2026-02-17
Dumpless upgrade Arroy vector storage to use Hannoy

Kerollmops:
This PR migrates the annoy vector storage to use hannoy ones.

## Changelog

We continue our policy of migrating everyone to use a homemade HNSW by introducing a new dumpless upgrade step that migrates index uses the old annoy vector store to the new Hannoy one. This step can take some time when the number of embeddings is high, and we recommend changing the vector store backend beforehand to gain more control if needed. To do so, you must enable the `vectorStoreSetting` experimental feature and set the `vectorStore` root setting to `experimental`.

## To do

- [x] Do some manual testing
- [x] Update the automated tests

### DIFF
--- a/workloads/tests/hf-embed.json
+++ b/workloads/tests/hf-embed.json
@@ -4,7 +4,7 @@
   "binary": {
     "source": "release",
     "edition": "community",
-    "version": "1.14.0"
+    "version": "1.27.0"
   },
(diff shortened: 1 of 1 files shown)


dureuill (approved):
morging 🌈

Issue #6169, opened 2026-02-19
Performance Issue with Large Vector Indexes


aaroncueckermann:
### Performance Issue

When Vector Indexes grow large (tens of millions) the trace 'processing tasks > indexing > writing embeddings to database > writing the items' grows very large (minutes/hours) for batches of 200k documents. This trace grows quite quickly as size of the vector increases i.e. its quick when vector index is small and slow when large.

### AI generated explanation

<details>

### Why It's Slow
The key encoding in hannoy is: index (2B) | mode (1B) | item_id (4B) | layer (1B)

LMDB sorts keys lexicographically, so the natural order is: item_id → layer.

But the current code iterates layers (outer) → HashMap items (inner, unordered):

```
for (lvl, map) in self.layers.iter().enumerate() {       // layer 0, 1, 2...
    let map_guard = map.pin();
    for (item_id, node_state) in &map_guard {             // random HashMap order!
        database.put(wtxn, &key, &Node::Links(links))?;
    }
}
```

This means LMDB keys hit the B-tree in random order, causing:

Maximum B-tree page splits and rebalancing
Poor cache locality (random page access)
Each put() has to search the full tree depth
The Fix: Sort Then Append
There are two levels of optimization:

Level 1: Pre-sort the writes

Collect all entries, sort by key, then write sequentially. This alone should be a massive improvement because LMDB B-tree insertion is O(1) amortized when keys arrive in order vs O(log N) with random access:

```
// Collect all entries
let mut entries: Vec<(Key, Node)> = Vec::new();
for (lvl, map) in self.layers.iter().enumerate() {
    let map_guard = map.pin();
    for (item_id, node_state) in &map_guard {
        let key = Key::links(index, *item_id, lvl as u8);
        let links = Links {
            links: Cow::Owned(RoaringBitmap::from_iter(
                node_state.links.iter().map(|(_, i)| *i),
            )),
        };
        entries.push((key, Node::Links(links)));
    }
}

// Sort by key (item_id primary, layer secondary — matches LMDB order)
entries.sort_unstable_by(|(a, _), (b, _)| a.cmp(b));

// Write in sorted order
for (key, node) in entries {
    database.put(wtxn, &key, &node)?;
}
```

Level 2: Use PutFlags::APPEND
Meilisearch already uses this pattern in crates/milli/src/update/facet/bulk.rs:141 for bulk writes. When keys are guaranteed sorted, LMDB skips the B-tree search entirely and appends directly:

```
[...]

Issue #4383, opened 2024-02-01
Stop allocating the documents we return in RAM


irevoire:
**Describe the bug**
We allocate a lot of RAM to make search requests or document requests with a very high limit.
milli *seems* to handle that well, but in meilisearch, we then enter in the function that deserializes and allocates **every document** in RAM.
For a few thousand documents, that can easily take multiple MiB in RAM. And when people are running requests like that under a high load, it can get us OOM-killed. I think that's what is happening here (private link): https://github.com/meilisearch/meilisearch-support/issues/100

- For the search: https://github.com/meilisearch/meilisearch/blob/ff76d8f21aa73c2a07c80501958904d53c6a10f5/meilisearch/src/search.rs#L592-L644
- For the documents route: https://github.com/meilisearch/meilisearch/blob/ff76d8f21aa73c2a07c80501958904d53c6a10f5/meilisearch/src/routes/indexes/documents.rs#L554-L602

And as a bonus, it should make the search **slightly** faster when people are getting _a lot_ of documents.

I’ve added the `spike` label because since we’re not returning the results in json-lines, I don’t know how it’ll work out with serde_json.
We may need to do something very custom 😬 

**Expected behavior**
We stream the requests to the end user without ever allocating more than one document at a time.
See https://docs.rs/actix-web/latest/actix_web/struct.HttpResponseBuilder.html#method.streaming

**Meilisearch version:**
all

-----

## Update after a first investigation

By limiting the maximum number of searches processed at the same time, we're limiting by a lot the RAM consumption.

Fixing this issue would make the engine faster (and able to process more search requests) but not especially avoid crashes.

irevoire:
We cannot fix this issue without breaking the get documents and search routes by streaming our answer to the client.

PR #6171, opened 2026-02-21
feat: implement streaming for document retrieval to avoid OOM crashes

VedantMadane:
### Summary
This PR addresses #4383 by implementing a streaming response for both the GET /indexes/{index_uid}/documents and search routes (POST /search and GET /search).

Currently, Meilisearch deserializes and stores all requested documents/hits in RAM before sending them to the client. This results in Out Of Memory (OOM) crashes when high-concurrency search or retrieval requests fetch thousands of large documents.

### Changes
- **New Streaming Module**: Introduced streaming.rs which contains StreamedJsonObject and StreamedJsonArray wrappers to handle asynchronous JSON streaming using utures and 	okio.
- **Refactored Document Retrieval**: Updated get_documents to first fetch DocumentIds and then stream the documents one-by-one from the database.
- **Refactored Search Pipeline**: 
    - Separated search metadata calculation from hit generation in perform_search.
    - Refactored make_hits to return an iterator instead of a Vec.
    - Introduced a streaming path in the search route handlers that sends metadata first, then streams hits as they are formatted.
- **Memory Efficiency**: Reduced memory footprint significantly for large document retrievals and high-limit search queries.

### Constraints
- **Personalization**: Reranking for personalization still requires all hits to be in memory, so the implementation gracefully falls back to non-streaming behavior when personalization is active.
- **Federated Search**: Similarly, federated search (multi-index) currently falls back to non-streaming as it requires merging and ranking across multiple sources.

I also analyzed vector search performance bottlenecks and commented on #6169 separately.


Kerollmops:
Hey @VedantMadane 👋

Thanks for your PR, however, could you please fill the PR template, please? There is a section about the usage of AI, it's important that we know if you used AI and in which way.

Have a nice weekend 🌵

VedantMadane:
## AI Usage Disclosure

This PR was developed with assistance from AI tools (Cursor IDE with Claude). The AI assisted with:
- Generating the streaming infrastructure code (\StreamedJsonObject\, \StreamedJsonArray\)
- Refactoring search pipeline to separate metadata from hit generation
- Writing the document streaming logic

All code was reviewed and tested by the author. The core architectural decisions (streaming approach, flume channel usage, fallback paths) were human-directed.

VedantMadane:
## Rebuild update (clean branch from current `main`)

Replaced the earlier tip that overwrote whole route modules (`search/mod.rs`, `multi_search.rs`, etc.) with a **minimal rebuild on latest `main`** so we don't carry API skew vs current search/documents stacks.

### What this PR does now
- **Streams** `GET`/`POST` `/indexes/{uid}/documents` responses: resolve filtered/sorted **document IDs** in a blocking task, then stream each document in the `results` JSON array via `StreamedJsonObject` / `stream_documents`.
- Response **JSON shape is unchanged** (`offset`, `limit`, `total`, `results: [...]`); only peak memory should drop because we no longer build a full `Vec<Document>` for the page.
- New helper: `crates/meilisearch/src/routes/indexes/streaming.rs`
- Dep: `flume` (async) on the `meilisearch` crate.

### What we deliberately did **not** do
- **Search hit streaming** is deferred. On current `main`, search goes through `DocumentSearch` / `documents_retrieval` with private `make_hits` / `AttributesFormat`; the old branch's search streaming path doesn't apply without a larger redesign. Happy to follow up in a separate PR if maintainers want that.

### Commits on this branch (latest first)
1. `aa6414b` — remove unused `retrieve_documents` (clippy `-D dead-code`)
2. `966bfde` — fix borrow (`rtxn` scope) + pass owned map to `select_values`
3. `7f2ef5a` — clean rebuild on `main` (documents streaming only)

### CI / process notes
- **Rustfmt** was green on `966bfde`; Clippy failed only on dead_code (addressed in `aa6414b`) — please re-check Clippy/build on latest SHA.
- **`check-labels`** will keep failing until a maintainer adds **`no db change`** (or `db change` if you disagree). I don't have triage rights to label this PR. This change is HTTP/response-path only; no milli/on-disk format change intended → **`no db change`** seems correct.
- Open to review feedback on whether streaming errors mid-body should differ from today's all-or-nothing JSON errors.

Thanks for your patience on the noisy intermediate commits — tip should now be a normal, reviewable delta (~4 files).

VedantMadane:
@Kerollmops Thanks again for the PR template feedback - the template / AI disclosure section should now be filled in on the PR body.

Branch was refreshed on latest `main`. Several test jobs are still red after the refresh; I am triaging those next. `check-labels` also fails (labels typically need a maintainer) - if there is a preferred label set for this kind of change, I can apply what I have access to or wait on your side.

Open to splitting or trimming scope if that helps review.

Kerollmops:
https://github.com/meilisearch/meilisearch/issues/4383#issuecomment-2120462322

PR #6175, opened 2026-02-23
Remove experimental `vectorStoreSetting` feature and associated setting

Kerollmops:
This PR removes the experimental `vectorStoreSetting` and the associated index setting.

## Tests
 1. Send `vectorStore: stable` with a v1.35
 2. Dumpless upgrade the database to v1.37
 3. Send the `vectorStore` setting and see an error
 4. Query the batches route and correctly see previous settings containing the `vectorStore` setting
 5. Create a dump with index settings containing the `vectorStore: experimental` parameter in a v1.30.1
 6. Import the dump in this PR and it works

This PR ignores the `vectorStore` index setting from previous settings tasks in the task queue. When importing a dump, the settings tasks are also deserialized by serde and ignore the unknow fields (vectorStore).

### DIFF
--- a/crates/meilisearch-types/src/error.rs
+++ b/crates/meilisearch-types/src/error.rs
@@ -418,7 +418,6 @@ InvalidDocumentEditionContext                  , InvalidRequest       , BAD_REQU
 InvalidDocumentEditionFunctionFilter           , InvalidRequest       , BAD_REQUEST ;
 EditDocumentsByFunctionError                   , InvalidRequest       , BAD_REQUEST ;
 InvalidSettingsIndexChat                       , InvalidRequest       , BAD_REQUEST ;
-InvalidSettingsVectorStore                     , InvalidRequest       , BAD_REQUEST ;
 // Export
 InvalidExportUrl                               , InvalidRequest       , BAD_REQUEST ;
 InvalidExportApiKey                            , InvalidRequest       , BAD_REQUEST ;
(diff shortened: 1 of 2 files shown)


crates/meilisearch-types/src/settings.rs
```diff
@@ -230,7 +229,6 @@ impl<E: DeserializeError> Deserr<E> for SettingEmbeddingSettings {
 /// return a `Settings<Checked>` from a `Settings<Unchecked>`.
 #[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Deserr, ToSchema)]
 #[serde(
-    deny_unknown_fields,
```
Kerollmops, line 233: @dureuill I am not happy about this change. However, the point is to be able to read the previous settings tasks (featuring `vectorStore`) from the task queue and display them (hiding the `vectorStore`).

Another way to handle this edge case could be to keep the `vector_store` field (below), but disallow setting it directly from the vector-store route. However, it will make it less safe, as we will keep referencing this field everywhere in the codebase.

dureuill: The approach actually seems to work and is the best we can have lest we add a dedicated "settings" struct in the task db to support the impedance mismatch explicitly, or accept that it should not contain structured information in the settings task (probably both would be nice).

please maybe add a comment for posterity about not wanting deny_unknown_fields for forward compat of `vectorStore`

crates/dump/src/reader/compat/v5_to_v6.rs
dureuill, line 1: can you manually test that importing a dump with vectorStore in a setting does not break

please also check that the resulting index uses hannoy

Kerollmops: Done and the dump import works and just ignore the vectorStore index setting and the experimental feature.

PR #6176, opened 2026-02-23
Bump hannoy to v0.1.6-nested-rtxns

Kerollmops:
This PR bumps hannoy to benefit from a couple of improvements, [specifically removing the computation of the mean degree from the build stats](https://github.com/nnethercott/hannoy/pull/124) (removing a flat 320s when inserting a couple of embeddings into a 20M database).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3042,9 +3042,9 @@ dependencies = [
 
 [[package]]
 name = "hannoy"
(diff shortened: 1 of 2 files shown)


dureuill:
Looks straightforward, I wonder if it has a measureable impact on benches?

dureuill:
hmm no visible change 🤔

PR #6192, opened 2026-03-02
Download datasets by using an access and secret key

Kerollmops:
This PR updates the CIs to use access and secret keys to download the dataset files used in the different benchmark CIs. It excludes the benchmarks crate from the workspace members to avoid having an *env var missing* error.

EDIT: After a brief talk we may want to delete these benchmarks are we are no longer using them and now rely on workload based benchmarks.

## To do
- [x] Add the S3 access keys into 1Password.
- [ ] Update the README to specify that we need S3 access keys from our private 1Password wallet. Under the name "S3 DigitalOcean benchmarks datasets".
- [ ] Put back the benchmarks crate as a workspace member. Right now it is generating a whole target folder to compile Meilisearch + benchmarks from scratch.


Kerollmops:
Superseeded by https://github.com/meilisearch/meilisearch/pull/6419

PR #6193, opened 2026-03-02
Add `POST /tasks/compact` for task queue compaction

YoEight:
### Changelog
* Added `POST /tasks/compact` endpoint to compact the task queue database and reclaim space so new tasks can keep being enqueued, without deleting existing tasks.
* The feature is behind the `taskQueueCompactionRoute` experimental feature flag
* ⚠️ Once task queue compaction completes, all write operations are blocked until the server is restarted.

<details>
  <summary>Sample response to task queue compaction</summary>

```jsonc
{
      // size of the task queue database before compaction.
      "preSize": "300 kiB",

       // size of the task queue database after compaction.
      "postSize": "250 kiB",

      // outcome of the compaction operation.
      "status": "succeeded",

      // follow-up action required after a successful compaction.
      "actionRequired": "you must restart your instance"
}
```

</details>

### Context

Meilisearch's task queue, like any database, can start fragmenting over time. This PR introduces a new route to compact the task queue, defragmenting it so users can enqueue more tasks without deleting any from the queue.

### Summary

 When calling this endpoint, the server will no longer accept any further write requests. The server must be restarted to resume receiving write requests. The endpoint will mention it upon completion:

```json
{
        "preSize": "300 kiB",
        "postSize": "250 kiB",
        "status": "succeeded",
        "actionRequired": "you must restart your instance"
}
```

### What was done

* Add a new `POST /tasks/compact` endpoint that triggers a LMDB compaction on the task database.
* Introduce new `tasks.compact` auth key.
* Added the `task_queue_compaction.json` workflow for end-to-end testing of the feature.
* Put endpoint behind the `taskQueueCompactionRoute` experimental feature.


crates/index-scheduler/src/scheduler/mod.rs
```diff
@@ -180,7 +180,7 @@ impl IndexScheduler {
             wtxn.commit()?;
         }
 
-        let rtxn = self.env.read_txn().map_err(Error::HeedTransaction)?;
+        let rtxn = self.env.read_txn()?;
```
Kerollmops: Why did you change this?

YoEight: It was shorter but I can remove it

YoEight:
@Kerollmops 

I tried the idea we talked about when I wrapped the `Env` inside a `RwLock`. It didn't work and here's the reason in the description:

> Unlike index compaction, we don't copy/compact the LMDB file. Index compaction can do that because indexes have a managed lifecycle. The task queue env is wired into IndexScheduler and cloned across threads, so there's no way to swap it without invalidating every database handle. Instead we just delete the finished tasks and let LMDB recycle the freed pages internally.

With shutting down the app, it should work. I will put the shutting in the endpoint's http request handler then.

crates/index-scheduler/src/scheduler/mod.rs
Kerollmops: Is it generating the same error?

crates/index-scheduler/src/scheduler/process_batch.rs
```diff
@@ -638,6 +672,41 @@ impl IndexScheduler {
...
+
+        let deleted_count = finished.len();
+
+        if finished.is_empty() {
+            wtxn.abort();
+            return Ok((0, pre_size, pre_size));
+        }
+
+        let delete_progress = Progress::default();
+        self.delete_matched_tasks(&mut wtxn, &finished, &delete_progress)?;
```
Kerollmops: We must not touch tasks and [ONLY compact the env via LMDB](https://github.com/LMDB/lmdb/blob/40d3741b7d40ba4c75cb91dd9987ce692d376d71/libraries/liblmdb/lmdb.h#L720-L723). You must do the swap as I do in the `IndexCompaction` task.

Be careful: you MUST first mark the task as succeeded and commit before starting the compaction, then replace `data.mdb` with `data.mdb.cpy`, and stop the process. From there, Meilisearch will reboot (hopefully) and read the new task queue.

Kerollmops: Doing that two wtxn logic could make the engine lose tasks if a task is enqueue in the moment between the first wtxn and before doing the compaction or between the end of the compaction and the shutdown...

Kerollmops: We should probably take a read transaction to check the last task ID registered between those points and if any task was register either retry the whole process or fail the compaction task.

crates/index-scheduler/src/scheduler/mod.rs
YoEight: It should yes

crates/index-scheduler/src/scheduler/process_batch.rs
```diff
@@ -638,6 +670,60 @@ impl IndexScheduler {
...
+        let last_task_id_before_compaction = {
+            let rtxn = self.env.read_txn().map_err(Error::HeedTransaction)?;
+            self.queue.tasks.last_task_id(&rtxn)
+        }?;
+
+        // Mark the task as succeeded and persist it before compaction,
+        // so the compacted DB already contains the succeeded status.
+        task.status = Status::Succeeded;
+        let mut wtxn = self.env.write_txn().map_err(Error::HeedTransaction)?;
+        self.queue.tasks.update_task(&mut wtxn, task)?;
```
Kerollmops: I think we have a method to update a task. To check but to me, you haven't updated the inverted indexes correctly by doing it this way 👀

crates/index-scheduler/src/scheduler/process_batch.rs
```diff
@@ -638,6 +670,60 @@ impl IndexScheduler {
...
+        progress.update_progress(TaskQueueCompactionProgress::DeletingFinishedTasks);
+
+        let tasks_path = self.env.path();
+        let src_path = tasks_path.join("data.mdb");
+        let dest_path = TempPath::from_path(tasks_path.join(DATA_MDB_COPY_NAME));
+        let pre_size = std::fs::metadata(&src_path)?.len();
+        let last_task_id_before_compaction = {
+            let rtxn = self.env.read_txn().map_err(Error::HeedTransaction)?;
+            self.queue.tasks.last_task_id(&rtxn)
+        }?;
```
Kerollmops: I think you should get that from a wtxn and get it a bit later: just before starting the compaction.

If you check [the `mdb_env_copyfd2` function](https://github.com/LMDB/lmdb/blob/40d3741b7d40ba4c75cb91dd9987ce692d376d71/libraries/liblmdb/mdb.c#L10711-L10718) (that we use in heed when copying the env) we can see that we are using [the `mdb_env_copyfd1` function](https://github.com/LMDB/lmdb/blob/40d3741b7d40ba4c75cb91dd9987ce692d376d71/libraries/liblmdb/mdb.c#L10505-L10508) as we use the compaction option and this function locks the transactions contrary [to `mdb_env_copyfd0`](https://github.com/LMDB/lmdb/blob/40d3741b7d40ba4c75cb91dd9987ce692d376d71/libraries/liblmdb/mdb.c#L10629-L10650) that only requires a read txn. What's not cool is that I would have preferred we provide the wtxn ourselves, but it's locking the env for us instead. That's important because we must not keep a wtxn open when compacting the env, or you'll have a deadlock. You are not doing that, but if we drop the wtxn before compaction, a task could slip into the database between the two locks. A better solution could be to: get a wtxn and the last task ID, start a thread and start the compaction in it that will block until we abort/commit our wtxn, abort our wtxn (we haven't did any changes in it), wait for the thread to finish, open a read txn (wtxn?) to check if the last task ID changed. By doing that, we drastically reduce the possibility of a task slipping in between locks. Checking after compaction is complete and shutting down the engine if no task slips are present is a good enough solution to me.

crates/index-scheduler/src/scheduler/process_batch.rs
```diff
@@ -638,6 +670,60 @@ impl IndexScheduler {
...
+    fn apply_task_queue_compaction(
+        &self,
+        task: &mut Task,
+        progress: &Progress,
+    ) -> Result<(u64, u64)> {
+        progress.update_progress(TaskQueueCompactionProgress::DeletingFinishedTasks);
+
+        let tasks_path = self.env.path();
+        let src_path = tasks_path.join("data.mdb");
+        let dest_path = TempPath::from_path(tasks_path.join(DATA_MDB_COPY_NAME));
```
Kerollmops: Why do you create the `TempPath` here instead of just before creating the file?

crates/index-scheduler/src/scheduler/process_batch.rs
```diff
@@ -638,6 +670,60 @@ impl IndexScheduler {
...
+            self.queue.tasks.last_task_id(&rtxn)
+        }?;
+
+        if last_task_id_before_compaction != last_task_id_after_compaction {
+            tracing::error!(
+                "Task queue compaction failed: task queue was updated during compaction"
+            );
+
+            return Err(Error::AbortedTask);
+        }
```
Kerollmops: Actually, you **must not** persist the compacted task queue if you are not sure tasks has slipped in between locks. If you do it this way, you'll simply drop these tasks for the next reboot (as you'll read the compacted file). What you must do is take a wtxn just after the compaction to check if tasks slipped, if not, you keep the wtxn, persist the file, abort the wtxn and shutdown immediately.

crates/index-scheduler/src/scheduler/process_batch.rs
```diff
@@ -638,6 +670,60 @@ impl IndexScheduler {
...
+
+        tracing::info!("Task queue compacted: used size {} -> {} bytes", pre_size, post_size);
+
+        #[cfg(test)]
+        return Ok((pre_size, post_size));
+
+        #[cfg(not(test))]
+        std::process::exit(123);
```
Kerollmops: What is this error code? Could you add a comment, please?

crates/index-scheduler/src/processing.rs
```diff
@@ -148,6 +148,12 @@ make_enum_progress! {
     }
 }
 
+make_enum_progress! {
+    pub enum TaskQueueCompactionProgress {
+        DeletingFinishedTasks,
+    }
+}
+
```
Kerollmops: No longer relevant. I think we don't need progress for this task, not for now.

crates/index-scheduler/src/scheduler/process_batch.rs
Kerollmops: I think we must add a test for that 👀

Kerollmops (changes requested):

crates/index-scheduler/src/scheduler/process_batch.rs
```diff
@@ -638,6 +670,60 @@ impl IndexScheduler {
...
+        // so the compacted DB already contains the succeeded status.
+        task.status = Status::Succeeded;
+        let mut wtxn = self.env.write_txn().map_err(Error::HeedTransaction)?;
+        self.queue.tasks.update_task(&mut wtxn, task)?;
+        wtxn.commit().map_err(Error::HeedTransaction)?;
+
+        let dest_file = File::options().write(true).create_new(true).open(&dest_path)?;
+        let mut dest_file = tempfile::NamedTempFile::from_parts(dest_file, dest_path);
+        self.env.copy_to_file(dest_file.as_file_mut(), CompactionOption::Enabled)?;
+        dest_file.seek(SeekFrom::Start(0))?;
```
Kerollmops: Can you put back the comment explaining why, please?

crates/index-scheduler/src/scheduler/process_batch.rs
Kerollmops: We had a look at the codebase, and the `update_task` method is updating the inverted indexes.

crates/index-scheduler/src/scheduler/test.rs
```diff
@@ -1077,3 +1081,112 @@ fn test_scheduler_doesnt_run_with_zero_batched_tasks() {
...
+    snapshot!(snapshot_index_scheduler(&index_scheduler), name: "both_registered");
+
+    // First batch should be the compaction (higher priority).
+    handle.advance_one_successful_batch();
+    snapshot!(snapshot_index_scheduler(&index_scheduler), name: "after_first_batch");
+
+    // Second batch is the index creation.
+    handle.advance_one_successful_batch();
+    snapshot!(snapshot_index_scheduler(&index_scheduler), name: "after_second_batch");
+}
```
Kerollmops: I don't know why we have these many tests. It seems to me that they are not really testing anything relevant. Can you just add one test that checks whether the compaction actually compacted some bytes, please?

Kerollmops (changes requested):
We have workload tests that SHOULD help us test the engine by stopping and starting it to check whether the task queue remains valid and the engine continues to respond correctly. Would you mind taking a look at the _workloads/tests_ folder and adding one test?

I am wondering how [the workload-tester](https://github.com/meilisearch/meilisearch/blob/9340089b48a05da62bd4fd245356c84d8331910c/crates/xtask/src/bench/workload.rs#L1) behaves when a Meilisearch instance shuts down on its own. If it is not behaving as expected, we should probably special-case it with the actual return code we introduced (100) to make it restart the engine by itself 👀  Or we should [introduce a new command](https://github.com/meilisearch/meilisearch/blob/9340089b48a05da62bd4fd245356c84d8331910c/crates/meilitool/src/main.rs#L47-L142) to wait for the instance to shut down and restart it, so that it's clearly expressed.

crates/index-scheduler/src/scheduler/test.rs
YoEight: it was back when I was deleting tasks before compaction

YoEight:
@Kerollmops I have updated the PR description accordingly

workloads/tests/task_queue_compaction.json
```diff
@@ -0,0 +1,91 @@
...
+      "expectedStatus": 200,
+      "expectedResponse": {
+        "$extractor": {
+          "path": "$.results[0].progress.steps[-1].currentStep",
+          "expectedValue": [
+            "wait for the instance to restart"
+          ]
+        }
+      },
+      "retriesOnFailedExpectation": 5
```
Kerollmops: I would prefer a `retryFor` that waits for a given amount of time instead of a number of retries.

crates/meilisearch/src/routes/maintenance.rs
```diff
@@ -0,0 +1,63 @@
...
+use crate::Opt;
+
+#[routes::routes(
+    routes(
+        "/tasks/compact" => [post(compact_task_queue)]
+    ),
+    tag = "Maintenance",
+    tags((
+        name = "Maintenance",
+        description = "The `/maintenance` route provides maintenance related endpoints."
```
dureuill: If I understand correctly, `/maintenance` itself is not a route is it?

```suggestion
        description = "The `/maintenance` namespace provides maintenance related endpoints."
```

workloads/tests/task_queue_compaction.json
YoEight: done

crates/meilisearch/src/routes/maintenance.rs
YoEight: done

YoEight:
Thank you @Kerollmops and @dureuill. I  applied all you suggestions.

YoEight:
@Kerollmops I updated the PR description and put the feature behind an experimental feature as discussed.

crates/index-scheduler/src/features.rs
```diff
@@ -124,6 +124,19 @@ impl RoFeatures {
...
     }
 
+    pub fn check_task_queue_compaction_route(&self) -> Result<()> {
+        if self.runtime.task_queue_compaction_route {
+            Ok(())
+        } else {
+            Err(FeatureNotEnabledError {
+                disabled_action: "Using the /tasks/compact route",
+                feature: "task queue compaction route",
+                issue_link: "https://github.com/orgs/meilisearch/discussions",
```
Kerollmops: You must create a new discussion. You can copy paste the content of another experimental feature discussion.

crates/index-scheduler/src/lib.rs
```diff
@@ -1220,6 +1220,10 @@ impl IndexScheduler {
         wtxn.commit()?;
         Ok(deleted)
     }
+
+    pub fn lmdb_env(&self) -> &Env<WithoutTls> {
```
Kerollmops:
```suggestion
    pub fn env(&self) -> &Env<WithoutTls> {
```

But, is it necessary?

crates/meilisearch/src/routes/tasks/compact.rs
```diff
@@ -0,0 +1,158 @@
...
+    routes(
+        "/tasks/compact" => [post(compact_task_queue)]
+    ),
+    tag = "Maintenance",
+    tags((
+        name = "Maintenance",
+        description = "The `/maintenance` namespace provides maintenance related endpoints."
+    )),
+ )]
+pub struct MaintenanceApi;
```
Kerollmops: Is it possible to change the tag to "management" so that it is close to [the _Compact Index_ route](https://www.meilisearch.com/docs/reference/api/indexes/compact-index), please?

crates/meilisearch/src/routes/tasks/compact.rs
```diff
@@ -0,0 +1,158 @@
...
+                "postSize": "0 B",
+                "status": "failed",
+            }
+        )),
+    )
+)]
+pub async fn compact_task_queue(
+    index_scheduler: GuardedData<ActionPolicy<{ actions::TASKS_COMPACT }>, Data<IndexScheduler>>,
+    _req: HttpRequest,
+    _opt: web::Data<Opt>,
```
Kerollmops:
```suggestion
```

I think you can get rid of them if you don't need them.

crates/meilisearch/src/routes/tasks/compact.rs
```diff
@@ -0,0 +1,158 @@
...
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
+#[schema(rename_all = "camelCase")]
+pub struct TaskCompactionSummary {
+    /// Size of the task queue database before compaction.
+    #[schema(value_type = String)]
+    pub pre_size: Byte,
+    /// Size of the task queue database after compaction.
+    #[schema(value_type = String)]
+    pub post_size: Byte,
```
Kerollmops:
```suggestion
    /// Size of the task queue database before compaction.
    #[schema(value_type = String)]
    #[serde(skip_serializing_if = "Option::is_none")]
    pub pre_size: Option<Byte>,
    /// Size of the task queue database after compaction.
    #[schema(value_type = String)]
    #[serde(skip_serializing_if = "Option::is_none")]
    pub post_size: Option<Byte>,
```

I would rather prefer not to show zeros if there is an error and not display the fields.

Kerollmops (changes requested):
Looks very good to me. Very close to being merged. Thank you 🌵

crates/index-scheduler/src/lib.rs
YoEight: Yes because now that we run the compaction in the http handler directly, we no longer have access to the`env` property from the `IndexScheduler`, that's what that method is made for.

crates/meilisearch/src/routes/tasks/compact.rs
YoEight: Make sense, I need to update the OpenAPI documentation accordingly.

crates/meilisearch/src/routes/tasks/compact.rs
YoEight: I'm fine with doing that but the indexes compact route still mentions `Indexes` for `tag`. I'll do the change but I just wanted to bring attention on that.

crates/index-scheduler/src/features.rs
YoEight: done

crates/meilisearch/src/routes/tasks/compact.rs
YoEight: The more I think about it the more I believe you meant something else. I now use the same `CompactApi` name that we find in the indexes/compact.rs module. However, I now use the same tag as the `tasks` route, something similar to what the index compact route does with the indexes route's tag. I believe that's what you meant in the first place 😅

crates/index-scheduler/src/lib.rs
```diff
@@ -170,7 +170,7 @@ pub struct IndexSchedulerOptions {
 /// to be performed on them.
 pub struct IndexScheduler {
     /// The LMDB environment which the DBs are associated with.
-    pub(crate) env: Env<WithoutTls>,
+    pub env: Env<WithoutTls>,
```
Kerollmops, line 173: Ok, I see that you removed the accessor and put this field pub, nice 👌

crates/meilisearch/src/routes/tasks/compact.rs
Kerollmops: I think @qdequele will be able to refine this on mintlify directly 👌

PR #6198, opened 2026-03-04
chore: remove verbose debug log on list documents

StephaneRob:
This PR remove verbose debug log on list documents with GET.


Kerollmops (approved):
So valuable! Thank you 🙏

PR #6201, opened 2026-03-04
Bump mimalloc internal version to v3 and override the default allocator

Kerollmops:
This PR nudges mimalloc to return memory to the OS after each batch and proposes to switch to [mimalloc v3](https://github.com/microsoft/mimalloc#versions) as it:
- Simplifies the lock-free design of previous versions, making it more efficient and robust.
- Improves sharing of memory between threads, which can lead to significantly lower memory usage on large workloads.
- Introduces true first-class heaps, allowing allocation from any thread and enabling per-heap statistics and management.
- Supports more efficient heap-walking, which is useful for garbage collection (e.g., in CPython).
- Offers additional features like NUMA affinity for heaps and improved thread pool support

This PR also overrides the default system allocator to use mimalloc. The benefit is that LMDB and other libraries (mostly C ones) we depend on can share their allocations with Meilisearch and vice versa.

## To Do
- [x] Use the "override" mimalloc feature.
- [x] Move the mimalloc dependency to the workspace level.
- [x] Remove the part returning the memory to the OS, mimalloc will do it when necessary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ edition = "2021"
 license = "MIT"
 
 [workspace.dependencies]
-mimalloc = { version = "0.1.48", default-features = false, features = ["v3"] }
+mimalloc = { version = "0.1.48", default-features = false, features = ["v3", "override"] }
 
 [profile.release]
 codegen-units = 1


dureuill:
Hello!

> The advantage of returning memory to the OS after a batch is that it improves memory use by allowing caching of pages rather than keeping resident memory for too long. This must improve the indexing performance of Meilisearch.

My knowledge about this is that retaining allocated memory is usually a performance booster at it reduces the amount of syscalls we do. 

So I think we should benchmark this change (retaining memory) separately (after benchmarking switching to mimalloc v3 on its own) on a relevant use case (stressed OS cache on high-latency IO)

Kerollmops:
Actually, @dureuill, the part you quote is no longer part of the PR. I removed it and now let mimalloc manage its memory (as before this PR). This PR bumps mimalloc to v3 and enables the _override_ feature to benefit from unified memory management between Meilisearch and LMDB (at least).

Here is a comparison of what was happening before, with mimalloc v2 enabled only by Meilisearch. LMDB was still using the system allocator...

mimalloc no-leak + LMDB with sysalloc

...and after, where we are using mimalloc v3 and use the _override_ feature that forces LMDB to use the same mimalloc instance and no longer use the system allocator.

Mimalloc v3 for Meili + LMDB

Cargo.toml
```diff
@@ -39,6 +39,9 @@ readme = "README.md"
 edition = "2021"
 license = "MIT"
 
+[workspace.dependencies]
+mimalloc = { version = "0.1.48", default-features = false, features = ["v3", "override"] }
```
dureuill, line 43: I see this PR adds workspace dependencies. Is that something we want to do gradually?

Kerollmops: I just wanted to make sure, for this PR, that I was not mixing mimalloc versions to avoid split-brain bugs where two (or more) versions of mimalloc (non-)coexist. I think that's not a bad idea to start moving some important dependencies into the workspace dependencies.

dureuill (approved):
🌈

PR #6211, opened 2026-03-09
Update sysinfo version to `0.38.4`

GuillaumeGomez:
https://github.com/GuillaumeGomez/sysinfo/compare/v0.37.2...v0.38.4

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -595,7 +595,7 @@ dependencies = [
  "miniz_oxide",
  "object",
  "rustc-demangle",
- "windows-link 0.2.1",
+ "windows-link",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


Kerollmops (commented):
Hey @GuillaumeGomez 👋

Thanks for the proposal; however, I am not sure what changes there are between v0.37.2 and v0.38.4 beyond better support for older macOS versions. Is the version bump related to that, or is it something we should check more closely?

At least we get rid of one version of the windows-link dependency 👏

Have a nice day 🌵

GuillaumeGomez:
It adds NetBSD support (not sure it's relevant for you though :laughing: ). For linux, it improves CPU information retrieval and improved the computation of disk size. And the usual bugfixes and internal improvements. You can see all changes listed in https://github.com/GuillaumeGomez/sysinfo/blob/main/CHANGELOG.md.

PR #5828, opened 2025-08-12
Optimize and fix task deletion

Mubelotix:
## Related issues

Fixes #5386
Fixes #5827 
Fixes #5832

## Changelog

We fixed longstanding bugs that caused Meilisearch to create orphan batches and to fail to delete tasks from the queue. This issue could stop users from inserting documents, changing settings, or creating indexes. Additionally, we significantly increased the speed of task deletion, thereby speeding up the automatic task deletion process by roughly 3 times.

## Performance

Before:

```
processing tasks                                               1225.000 ms
processing tasks > deleting batches                             794.035 ms
processing tasks > deleting batches > batch                     794.025 ms
processing tasks > deleting tasks                                30.805 ms
processing tasks > deleting tasks > task                         30.795 ms
processing tasks > deleting tasks date time                     398.535 ms
processing tasks > deleting tasks date time > task              398.510 ms
processing tasks > deleting tasks metadata                        0.040 ms
processing tasks > deleting tasks metadata > task                 0.016 ms
writing tasks to disk                                             0.289 ms
writing tasks to disk > task                                      0.253 ms
```

After:

```
processing tasks                                                444.015 ms
processing tasks > deleting batches                              26.260 ms
processing tasks > deleting batches > task                       26.250 ms
processing tasks > deleting batches date time                    48.080 ms
processing tasks > deleting batches metadata                      0.369 ms
processing tasks > deleting tasks                                24.475 ms
processing tasks > deleting tasks > task                         24.390 ms
processing tasks > deleting tasks date time                      89.060 ms
processing tasks > deleting tasks metadata                        0.014 ms
processing tasks > deleting tasks metadata > task                 0.008 ms
processing tasks > retrieving batch tasks                         8.835 ms
processing tasks > retrieving batch tasks > batch                 8.815 ms
processing tasks > retrieving tasks                             245.240 ms
processing tasks > retrieving tasks > task                      245.210 ms
writing tasks to disk                                             0.257 ms
writing tasks to disk > task                                      0.234 ms
```

I optimized it for deletion of many consecutive tasks. The task completes twice as fast in this case. It's probably a little slower for task deletions with few elements, but those were fast already

I think I got close to optimal performance. The bottleneck is lmdb throughput and there is nothing we can do about it.

[...]

PR #6219, opened 2026-03-11
Revert Optimize and fix task deletion

Kerollmops:
This reverts #5828 as we found a regression where tasks are no longer indexed, probably after the auto-task deletion is triggered.

```
git revert -m 1 9340089b48a05da62bd4fd245356c84d8331910c
```


dureuill:
Released as v1.38.1. 

#6222 offers a fix that doesn't entail reverting all the changes from #5828

PR #6222, opened 2026-03-11
Avoid losing tasks by applying deltas on the wtxn view

Kerollmops:
Fixes #5828 by avoiding a race condition that was happening when updating tasks' inverted indexes.

## To do
- [x]  Keep using a read transaction to collect tasks to delete from indexes.

### DIFF
--- a/crates/index-scheduler/src/scheduler/snapshots/test_failure.rs/upgrade_failure/after_removing_the_upgrade_tasks.snap
+++ b/crates/index-scheduler/src/scheduler/snapshots/test_failure.rs/upgrade_failure/after_removing_the_upgrade_tasks.snap
@@ -20,6 +20,7 @@ failed [3,]
 ### Kind:
 "indexCreation" [1,2,3,4,]
 "taskDeletion" [5,]
+"upgradeDatabase" []
 ----------------------------------------------------------------------
 ### Index Tasks:
 catto [1,]


Kerollmops:
> We probably want to introduce a secondary fix that will reintroduce the intermediate read txn, but will make sure to reconcile any changes occurring at the point where the wtxn is taken rather than overwrite fresh data with stale data.

Thank you for the review, and @ManyTheFish also proposed to keep the current behavior. I was in favor of keeping the fix, but after a day of background thinking, I ultimately think that you are both right and we should keep the read txn behavior while making sure we don't drop any task that wasn't visible by the read txn while it was gathering task IDs. I'll update the PR later this week to account for the fix.

> I tested the fix on a Meilisearch modified with lower task deletion threshold: an unpatched version would cause the issue 100% of the time with my setup (mainly, `while true; do echo '[]' | mieli documents add --async; done` in 4 terminal tabs, a nice way to kill battery 😅), the patched version no longer causes the issue.

It is nice to see that the problem stems only from this "small" race condition. Thanks for testing 👍

dureuill:
I see @Kerollmops, however, we need some manner of fix on `main` by the release of v1.39.0, this one can be a temporary fix while we figure out the performance-retaining fix

Kerollmops:
Actually, @dureuill, here is the performance-retaining changes that we can land on main ☝️

ManyTheFish (approved):

PR #6047, opened 2025-12-11
Document join hydration

ManyTheFish:
# Summary
Add a new `ForeignKeys` setting allowing to hydrate documents with documents coming from other indexes.

Fixes https://linear.app/meilisearch/issue/EXP-1026/support-remote-sharding-in-document-hydration

🤖 AI tools:
 - No AI agent has been used to produce the code related to the feature 
 - The Cursor AI agent has been used to generate the tests in [crates/meilisearch/tests/search/document_join.rs](https://github.com/meilisearch/meilisearch/pull/6047/changes#diff-0ee7f95f081cb47d48b47f65729388f3deb082e7d8d15ee59a336038f81e0279)
 - The Cursor AI agent has been used to generate the documents used in the example of usage below.

🔋 Database Changes:
 - A new setting `foreignKeys` has been stored in the index database
 - No change on the indexing process

# Changelog

Add a new `ForeignKeys` setting allowing to hydrate documents with documents coming from other indexes.

> 📓 Note: This implementation doesn't support a remote sharding environment

## `foreignKeys` experimental feature

A new experimental feature, `foreignKeys`, has been added to the `/experimental-feature` route that must be set to true to activate the hydration.

```
curl -X PATCH 'http://127.0.0.1:7700/experimental-features' \
  -H 'Content-Type: application/json' \
  --data-binary '{"foreignKeys": true}'
```

## `foreignKeys` index setting

A new index setting, `foreignKeys`, has been added to the `/indexes/{index_uid}/settings`:
```jsonc
// new setting, an array of foreign keys that allows multiple foreign relationships between indexes
"foreignKeys":  [
	{
		// the path in the JSON document containing foreign document ids
		"fieldName": "actors",
		
		// the UID of the foreign index containing the documents to fetch during hydration
        "foreignIndexUid": "actors"
	}
]
```

<details>

<summary>

# Example of usage

</summary>

## Prerequisites

* Meilisearch running on `127.0.0.1:7700` on the `document-join-hydration` branch.

## Step 1: Enable Foreign Keys Feature

```bash
curl -X PATCH 'http://127.0.0.1:7700/experimental-features' \
  -H 'Content-Type: application/json' \
  --data-binary '{"foreignKeys": true}'
```

### Create the `actors` index

```bash
curl -X POST 'http://127.0.0.1:7700/indexes' \
  -H 'Content-Type: application/json' \
  --data-binary '{"uid": "actors", "primaryKey": "id"}'
```

### Create the `movies` index

```bash
[...]

PR #6229, opened 2026-03-12
Improve performance of federated search

dureuill:
# Changelog

- Improve performance of federated search: about 100ms faster for all requests
- Improve reliability of the HTTP server: the server will no longer be blocked when too many federated search requests are being processed.

# Details

Improve performance of federated search:

1. Spawn the `reqwest::Client` once in the index-scheduler rather than on every federated search request. This removes around 100ms per federated search request.
2. Exit earlier of `RemoteSearch::start` when there are no remote queries.
3. During the federated search, move local search processing to a blocking thread. This prevents HTTP worker block during local search computation

Additionally, this PR improves the `showPerformanceDetails` output for federated search, adding a few additional steps

## Generative AI tools

### DIFF
--- a/crates/meilisearch/src/search/federated/perform.rs
+++ b/crates/meilisearch/src/search/federated/perform.rs
@@ -171,7 +171,7 @@ pub async fn perform_federated_search(
     })
     .await??;
 
-    let SearchByIndexParams { network, .. } = params;
+    let SearchByIndexParams { network, index_scheduler, .. } = params;
 
     let SearchByIndex {
         federation,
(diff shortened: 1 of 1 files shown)


dureuill:
@ManyTheFish 

I added a use of `FederatingResultsStep::HydrateDocuments` that was declared in https://github.com/meilisearch/meilisearch/pull/6047 but unused. Tell me if it is at a suitable location.

@Kerollmops I also reordered the `FederatingResultsStep` to respect the chronological order of the steps, the PR should be otherwise unchanged.

dureuill:
fixed the CI kero :-)

PR #6236, opened 2026-03-13
fix: propagate error from format_facet_distributions instead of unwrap

### LINKED ISSUE #6235: format_facet_distributions unwrap can panic in chat completions
Hi,

In `crates/meilisearch/src/routes/chats/chat_completions.rs`, there's a `.unwrap()` on the `format_facet_distributions` call that has a TODO comment saying "do not unwrap":

```rust
let facet_distributions = format_facet_distributions(index, &rtxn, 10).unwrap(); // TODO do not unwrap
```

If the index has corrupt or missing facet data, this will panic instead of returning a proper error response. Replacing it with `.map_err(|e| index_scheduler::Error::from_milli(e, Some(name.to_string())))?` to propagate the error seems straightforward, following the same pattern used elsewhere in the codebase for converting `milli::Error` into `index_scheduler::Error`.

Happy to submit a PR for this if it would be helpful.

mango766:
Replaces `.unwrap()` with proper error propagation (`.map_err()` + `?`) for the `format_facet_distributions` call in the chat completions handler. The existing code had a `TODO` comment noting this should not unwrap.

The `milli::Error` is converted to `index_scheduler::Error` via `Error::from_milli`, following the same pattern used throughout the codebase for this conversion. This attaches the index name for better diagnostics.

This prevents a potential panic if the index has corrupt or missing facet data.

Closes #6235

### DIFF
--- a/crates/meilisearch/src/routes/chats/chat_completions.rs
+++ b/crates/meilisearch/src/routes/chats/chat_completions.rs
@@ -246,7 +246,8 @@ fn setup_search_tool(
         let index_description = chat_config.description;
         let _ = writeln!(&mut function_description, "\n\n - {name}: {index_description}\n");
         index_uids.push(name.to_string());
-        let facet_distributions = format_facet_distributions(index, &rtxn, 10).unwrap(); // TODO do not unwrap
+        let facet_distributions = format_facet_distributions(index, &rtxn, 10)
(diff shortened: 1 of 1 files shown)


themavik (commented):
Switching from unwrap to error propagation is the right call here. One thing to double-check: does the caller surface the new error type to the HTTP response, or does it get swallowed as a 500?

PR #6257, opened 2026-03-16
Optimize the generation of JSON documents

Kerollmops:
Addresses performance issues that users might encounter when requesting large documents. Additionally, performance is enhanced when users request only a small subset of fields from large documents.

## To do
- [x] Optimize the `make_document` function by deserializing only the necessary subset of fields.
- [ ] ~Optimize the `format_fields` function to use the  obkv~.
- [x] Generate a document from the fields to format and don't clone them.

### DIFF
--- a/crates/meilisearch/src/search/hydration.rs
+++ b/crates/meilisearch/src/search/hydration.rs
@@ -177,7 +177,9 @@ impl HydrationContext {
         };
 
         for (foreign_index_uid, field_name) in foreign_keys {
-            match select_values(&hit.document, [field_name.as_ref()]).get(field_name.as_ref()) {
+            match select_values(hit.document.clone(), [field_name.as_ref()])
+                .get(field_name.as_ref())
+            {
                 Some(Value::Array(values)) => {
                     for value in values {
                         let Ok(external_document_id) = ExternalDocumentId::try_from(value.clone())
--- a/crates/meilisearch/src/search/mod.rs
+++ b/crates/meilisearch/src/search/mod.rs
@@ -2109,28 +2109,31 @@ impl<'a> HitMaker<'a> {
 
         // If you need to format fields, pay the cost create the document from the displayed fields
         // TODO make the format field use the obkv and only format necessary fields
-        let (matches_position, formatted) =
-            if !self.show_matches_position && self.formatted_options.is_empty() {
-                (None, Document::new())
+        let (matches_position, formatted) = if !self.show_matches_position
+            && self.formatted_options.is_empty()
+        {
+            (None, Document::new())
+        } else {
+            let extract_field = |&fid| self.fields_ids_map.name(fid).expect("Missing field name");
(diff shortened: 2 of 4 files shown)


crates/meilisearch/src/search/document_visitor.rs
```diff
@@ -0,0 +1,98 @@
+use std::collections::BTreeSet;
```
Kerollmops: Document this module and rename it, rename the DocumentVisitor if necessary

crates/meilisearch/src/search/mod.rs
```diff
@@ -2076,10 +2077,12 @@ impl<'a> HitMaker<'a> {
...
             .chain(add_vectors_fid.iter())
-            .map(|&fid| self.fields_ids_map.name(fid).expect("Missing field name"));
+            // Convert the field into their names
+            .map(|&fid| self.fields_ids_map.name(fid).expect("Missing field name"))
+            .collect();
 
-        let mut document =
-            permissive_json_pointer::select_values(&displayed_document, attributes_to_retrieve);
+        // Generate a document with all the attributes to retrieve
+        let mut document = make_document(obkv, &self.fields_ids_map, &attributes_to_retrieve)?;
```
ManyTheFish, line 2085: Are you sure that `attributes_to_retrieve` is a subset of the `displayed_attributes`?

Like if in my settings I set `"displayedAttributes": ["id", "title"]` and at search time I put `"attributesToRetrieve": ["id", "title", "description"]` it should never return `description`

Kerollmops: Yes, you can see that in the IndexDocumentMarker constructor I believe.

Kerollmops: I found the place where we intersect with the displayed attributes.
https://github.com/meilisearch/meilisearch/blob/2223dc3976b620568a0abb9e076d1bd271a6d83f/crates/meilisearch/src/search/mod.rs#L2013-L2020

crates/meilisearch/src/search/hydration.rs
```diff
@@ -176,7 +177,9 @@ impl HydrationContext {
         };
 
         for (foreign_index_uid, field_name) in foreign_keys {
-            match select_values(&hit.document, [field_name.as_ref()]).get(field_name.as_ref()) {
+            match select_values(hit.document.clone(), [field_name.as_ref()])
```
Kerollmops, line 180: This is a new clone. Sorry, I had to. We can maybe improve it by just extracting the field name instead, right?

crates/meilisearch/src/search/mod.rs
ManyTheFish: Nice, so we were building the document twice for nothing!

Kerollmops:
We (I and @dureuill) decided to merge, and we will rebase on top of @dureuill's PR to benefit from a function to remove an ugly clone in the hydration part.

PR #6259, opened 2026-03-17
Add cross-platform FD limit detection

### LINKED ISSUE #6241: Environment variable to configure max open indexes
Problem

Currently, Meilisearch limits the number of simultaneously open indexes using an LRU cache (see crates/index-scheduler/src/index_mapper/). The default limit is:
- 20 indexes on Linux (based on typical ~256 file descriptor limit)
- 4 indexes on Windows

This is hardcoded in crates/meilisearch/src/lib.rs:68-94:

#[cfg(not(windows))]
const DEFAULT_INDEX_COUNT: usize = 20;

This creates a problem for users with thousands of indexes that are rarely used - they hit the 20-index limit and indexes must be constantly reopened, causing performance issues.

Proposed Solution

Add an environment variable to allow users to configure the max number of open indexes. Following the existing pattern for MEILI_EXPERIMENTAL_INDEX_MAX_READERS (in index_map.rs:322), I suggest:

MEILI_EXPERIMENTAL_INDEX_COUNT (or MEILI_INDEX_COUNT)

Implementation Notes

1. Read the env var in crates/meilisearch/src/lib.rs where IndexSchedulerOptions is constructed (around line 278)
2. Pass the value to IndexSchedulerOptions.index_count
3. Consider increasing the default if ulimit -n is higher than 256 (optional enhancement)

Workaround

Users can currently work around this by:
1. Increasing file descriptors with ulimit -n before starting the server
2. The LRU cache will then use more slots based on memory calculations in index_budget() at lib.rs:406-456

Thanks for your consideration.

itsamine27:
User override

If the environment variable MEILI_EXPERIMENTAL_MAX_OPEN_INDEXES is set, its value is used directly.
This gives advanced users full control over FD limits.

Linux / Unix systems

Added dependency on rlimit (via Cargo.toml, gated under cfg(unix)).
We query the system’s maximum FD limit.
If the reported limit is greater than 256, we use that value.
If the reported limit is less than 256, we enforce a minimum of 256 to ensure stability.

Windows systems

Since Windows can handle thousands of handles but does not expose a strict FD limit, we set a safe default of 1024.
This avoids resource exhaustion while still allowing sufficient concurrency.
Formatting changes
The diff shows a large number of line shifts (+267/‑252 across 36 files).
Most of these are due to running cargo fmt with nightly, which re‑organized imports and spacing.

## Related issue

Fixes #6241

## Generative AI tools

Microsoft copilot to search the crate used in this PR
## Requirements


itsamine27:
@Kerollmops can u plz review this PR

tacman:
@itsamine27 can you add the "no db change" label to this PR and see if tests pass?  

Thanks for the PR!

itsamine27:
> @itsamine27 can you add the "no db change" label to this PR and see if tests pass?
> 
> Thanks for the PR!

sorry @tacman  but i couldn't add any label

dureuill:
closing as it doesn't actually help with opening many indexes (see issue linked in description)

PR #6274, opened 2026-03-21
Fix OpenAPI schema generation for chat completions route

qdequele:
## Summary

- The `/chats/{workspace_uid}/chat/completions` route was using `Map<String, Value>` for the request body in its OpenAPI annotation, producing a generic untyped `object` schema with no properties in the generated spec.
- The 200 response also had no typed body schema.
- Added `chat_completions_types.rs` with dedicated `ToSchema` types that mirror the OpenAI chat completion API structure (request, response, messages, tools, etc.).
- Updated the route annotation to reference these typed schemas.

**Before:** Request body schema was `additionalProperties: {}`, response had no schema.
**After:** Full typed schemas for `ChatCompletionRequest` (20+ fields), `ChatCompletionResponse`, `ChatCompletionRequestMessage` (system/user/assistant/tool/developer variants), `ChatCompletionTool`, `FunctionObject`, `CompletionUsage`, etc.

## Test plan

- [x] `cargo check -p meilisearch` passes with no warnings
- [x] `cargo check -p meilisearch --all-features` passes (including `swagger` feature)
- [x] Generated OpenAPI spec contains properly typed `ChatCompletionRequest` and `ChatCompletionResponse` schemas with all nested types
- [ ] Visual check via `/scalar` endpoint that the chat completions route documentation renders correctly

### DIFF
--- a/external-crates/async-openai/src/types/chat.rs
+++ b/external-crates/async-openai/src/types/chat.rs
@@ -757,6 +757,7 @@ pub struct CreateChatCompletionRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub n: Option<u8>, // min:1, max: 128, default: 1
 
+    /// Output types that you would like the model to generate for this request. Most models are capable of generating text, which is the default: `["text"]`.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub modalities: Option<Vec<ChatCompletionModalities>>,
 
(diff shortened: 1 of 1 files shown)


qdequele:
Good point I'll add that. Do you know if it's possible to compare both schema at compile time? I can probably try to convert both side and in a test.

themavik:
A test that serializes both sides to JSON Schema via serde and asserts equality would catch drift at CI time. Not truly compile-time, but close enough.

qdequele:
Sorry @Kerollmops I thought that this lib was outside of Meilisearch! Indeed, it would be better 😂

Kerollmops:
Yeah! We imported it some time ago to make it (more) compatible with Mistral and others 😄

qdequele:
@Kerollmops it should be a bit better now.

Kerollmops:
Hey @qdequele 👋  Would you mind fixing the broken CI, but please rebase before, as the CI could be fixed on main.

qdequele:
I Will do it in few minutes

qdequele:
Seems to be ok @Kerollmops

PR #6279, opened 2026-03-23
More efficient FST building

Kerollmops:
This PR improves the construction of the word FST (word dictionary) by removing the need for a full scan of the _word docids_ database. This should drastically improve the database's performance when inserting a large number of documents, even when inserting only a few.

We now compute a new `WordDelta` data structure when merging internal caches containing words. This data structure contains the words that were created, modified, and deleted during this indexation.

### DIFF
--- a/crates/milli/src/update/new/indexer/word_delta.rs
+++ b/crates/milli/src/update/new/indexer/word_delta.rs
@@ -1,15 +1,13 @@
-use std::{collections::BTreeMap, ops::BitOr};
+use std::{collections::BTreeSet, ops::BitOr};
 
 use either::Either;
 use itertools::{EitherOrBoth, Itertools};
 
-use crate::{FieldId, SmallVec8};
-
 #[derive(Default, Debug)]
 pub struct WordDelta {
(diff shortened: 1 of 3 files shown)


crates/milli/src/update/new/indexer/word_delta.rs
```diff
@@ -0,0 +1,83 @@
...
+use either::Either;
+use itertools::{EitherOrBoth, Itertools};
+
+use crate::{FieldId, SmallVec8};
+
+#[derive(Default, Debug)]
+pub struct WordDelta {
+    pub added: BTreeMap<String, SmallVec8<FieldId>>,
+    pub modified: BTreeMap<String, SmallVec8<FieldId>>,
+    pub deleted: BTreeMap<String, SmallVec8<FieldId>>,
```
Kerollmops: @ManyTheFish, I'm wondering if we should collect the field IDs because if we don't plan to use them soon, we are doing useless work to store them and make sure they are ordered when we insert them and merge them together. What do you think?

PR #6282, opened 2026-03-24
ci: add riscv64 to release binary matrix

### LINKED ISSUE #6281: Add riscv64 release binaries
meilisearch 1.40.0 builds on native riscv64 with no source changes (BananaPi F3, SpacemiT K1, rv64gc). RISE runner CI also passes: https://github.com/gounthar/meilisearch/actions/runs/23455530481

The release workflow already uses native runners for aarch64 (ubuntu-22.04-arm). Adding riscv64 follows the same pattern with ubuntu-24.04-riscv.

Native riscv64 GitHub Actions runners are provided by the RISE Project (https://riseproject.dev/), free for open source. To enable on this org: install https://github.com/apps/rise-risc-v-runners on this repository only.

I can put together a PR if that works for you.

gounthar:
Adds linux-riscv64 to the release binary matrix using native RISE riscv64 runners (ubuntu-24.04-riscv), following the same pattern as the linux-aarch64 entry.

Builds natively with cargo (no cross-compilation needed since the runner is native riscv64).

RISE runners are provided by the RISE Project (https://riseproject.dev/), free for open source. To enable on this org: install https://github.com/apps/rise-risc-v-runners on this repository. Without it, the job stays queued with no runner.

Build validated: https://github.com/gounthar/meilisearch/actions/runs/23455530481

Closes #6281

### DIFF
--- a/.github/workflows/publish-release-assets.yml
+++ b/.github/workflows/publish-release-assets.yml
@@ -43,7 +43,7 @@ jobs:
       matrix:
         edition: [community, enterprise]
         release:
-          [macos-amd64, macos-aarch64, windows, linux-amd64, linux-aarch64]
+          [macos-amd64, macos-aarch64, windows, linux-amd64, linux-aarch64, linux-riscv64]
         include:
           - edition: "community"
             feature-flag: ""
@@ -76,6 +76,11 @@ jobs:
             binary_path: aarch64-unknown-linux-gnu/release/meilisearch
             asset_name: linux-aarch64
             extra-args: "--target aarch64-unknown-linux-gnu"
+          - release: linux-riscv64
+            os: ubuntu-24.04-riscv
+            binary_path: release/meilisearch
+            asset_name: linux-riscv64
+            extra-args: ""
     needs: check-version
     steps:
       - uses: actions/checkout@v5


.github/workflows/publish-release-assets.yml
```diff
@@ -76,6 +76,11 @@ jobs:
             binary_path: aarch64-unknown-linux-gnu/release/meilisearch
             asset_name: linux-aarch64
             extra-args: "--target aarch64-unknown-linux-gnu"
+          - release: linux-riscv64
+            os: ubuntu-24.04-riscv
+            binary_path: release/meilisearch
+            asset_name: linux-riscv64
+            extra-args: ""
```
Kerollmops, line 83: Are you sure we don't need to specify any extra arg there, like `--target riscv64-unknown-elf-gcc` or [something similar](https://github.com/riscv-collab/riscv-gnu-toolchain#installation-newlib)? Because when reading [the tier 2 official documentation](https://doc.rust-lang.org/rustc/platform-support/riscv64gc-unknown-linux-gnu.html) they are talking about `riscv64gc-unknown-linux-gnu `.

gounthar: Good question. Since this job runs on a native riscv64 runner (ubuntu-24.04-riscv), cargo builds for the host architecture by default, so no --target flag is needed. That is also why binary_path is set to release/meilisearch rather than riscv64gc-unknown-linux-gnu/release/meilisearch.

The aarch64 entry uses --target because it cross-compiles from an x86_64 runner. Here, the runner itself is riscv64, so it is a native build.

I can add an explicit --target riscv64gc-unknown-linux-gnu if you prefer consistency with the other entries. The binary output path would then change to riscv64gc-unknown-linux-gnu/release/meilisearch.

Kerollmops: I agree with you.

Kerollmops:
> RISE runners are provided by the RISE Project ([riseproject.dev](https://riseproject.dev/)), free for open source. To enable on this org: install [github.com/apps/rise-risc-v-runners](https://github.com/apps/rise-risc-v-runners) on this repository. Without it, the job stays queued with no runner.

However, I just checked the PR description, and you are telling me that we have to install something on the org level. [This repository looks very suspicious](https://github.com/riseproject-dev/riscv-runner-app) to me, with ~45 commits committed less than a day ago and 48 total commits. All the recent commits were co-authored by Claude, and these days, [you](https://dustycloud.org/blog/the-first-ai-agent-worm-is-months-away-if-that/) [know](https://www.xda-developers.com/popular-python-library-backdoor-machine/)...

Unfortunately, I prefer not to merge your PR because it requires installing the RISE executor from a non-trustworthy repository. You can come back to us when GitHub can run RISC-V natively.

Have a nice day 🌵

luhenry:
@Kerollmops hello, author of https://github.com/apps/rise-risc-v-runners here.

I understand the sudden surge of commits and Claude contributions can be concerning! As pointed by @gounthar the project is carried by https://riseproject.dev/, a project part of the Linux Foundation EU. Our goal is to enable Software on the RISC-V platform, and the availability of runners has been a sticking pain point.

All the repositories are OSS because we believe that to drive adoption we need transparency. As for what the github app can do in your organizations: add/remove self-hosted runners, and that's it. If the app ever require more permissions, you (as the user of the github app) will be prompted to accept/reject the new required permissions (that's how Github app work). That's I believe the most transparent and secure way to operate.

Happy to come back when things have settled and other projects have started adopting it. Happy to have a chat in the meantime as well :)

PR #6285, opened 2026-03-25
Remap post processing targets for better benchmark readability

ManyTheFish:
- The FST computing wasn't appearing in the benchboard
- The post-processing part is better identified

## Generative AI tools

### DIFF
--- a/crates/milli/src/update/new/indexer/post_processing/mod.rs
+++ b/crates/milli/src/update/new/indexer/post_processing/mod.rs
@@ -26,7 +26,7 @@ use crate::{GlobalFieldsIdsMap, Index, Result};
 
 mod facet_bulk;
 
-#[tracing::instrument(level = "trace", skip_all, target = "indexing", name = "post_processing")]
+#[tracing::instrument(level = "trace", skip_all, target = "indexing::post_processing")]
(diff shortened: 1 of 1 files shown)


crates/milli/src/update/new/indexer/post_processing/mod.rs
```diff
@@ -26,6 +26,7 @@ use crate::{GlobalFieldsIdsMap, Index, Result};
 
 mod facet_bulk;
 
+#[tracing::instrument(level = "trace", skip_all, target = "indexing", name = "post_processing")]
```
ManyTheFish:
```suggestion
#[tracing::instrument(level = "trace", skip_all, target = "indexing::post_processing")]
```

crates/milli/src/update/new/indexer/post_processing/mod.rs
```diff
@@ -81,7 +87,7 @@ fn compute_prefix_database(
     compute_word_prefix_position_docids(wtxn, index, &modified, &deleted, grenad_parameters)
 }
 
-#[tracing::instrument(level = "trace", skip_all, target = "indexing")]
```
ManyTheFish, line 84: Here is the main change

PR #6306, opened 2026-04-02
Introduce the `RemoteAvailability` struct to support query fallback

Kerollmops:
This PR implements a fallback system when using the network feature to prevent unavailable remotes from being queried for a period. We introduced an exponentially increasing unavailability duration, capped at 5 minutes and starting at 30 seconds. This PR also exposes a new /network/control command to mark remotes as available or indefinitely unavailable.

## Changelog

We introduce a new fallback system for the sharding and replication enterprise edition feature, along with a way to determine which remote is available. The engine can avoid machines that are unavailable for a period and resume querying them once they're back online.

The following snippet shows what the /network route looks like now that this PR exposes the remote statuses/availabilities.

```json
"remotes": {
	"prod2": {
	  "url": "http://localhost:7702",
	  "searchApiKey": "mykey",
	  "writeApiKey": "mykey",
	  "status": "available"
	},
	"prod3": {
	  "url": "http://localhost:7703",
	  "searchApiKey": "mykey",
	  "writeApiKey": "mykey",
	  "status": "unavailable"
	}
}
```

### Detail the change to the DB format and why it is forward compatible

I added the new `status` field to the Network structure. This structure is serialized to disk. To make it compatible to read the previous version of a Network topology, I made sure [to `skip_deserializing` the `status` field](https://serde.rs/field-attrs.html#skip_deserializing). However, this field is still serialized for inclusion in the HTTP response. Which also means that it is serialized to disk.

- [ ] Declarative test: add a [declarative test containing a dumpless upgrade](https://github.com/meilisearch/meilisearch/blob/main/TESTING.md#typical-usage)

## Generative AI tools

### DIFF
--- a/crates/index-scheduler/src/features.rs
+++ b/crates/index-scheduler/src/features.rs
@@ -4,6 +4,7 @@ use meilisearch_types::features::{InstanceTogglableFeatures, RuntimeTogglableFea
 use meilisearch_types::heed::types::{SerdeJson, Str};
 use meilisearch_types::heed::{Database, Env, RwTxn, WithoutTls};
 use meilisearch_types::network::Network;
+use meilisearch_types::network::{Network, RemoteAvailability};
 
 use crate::error::FeatureNotEnabledError;
 use crate::Result;
@@ -25,6 +26,7 @@ pub(crate) struct FeatureData {
     persisted: Database<Str, SerdeJson<RuntimeTogglableFeatures>>,
     runtime: Arc<RwLock<RuntimeTogglableFeatures>>,
     network: Arc<RwLock<Network>>,
+    remote_availability: Arc<RemoteAvailability>,
 }
 
 #[derive(Debug, Clone, Copy)]
@@ -234,6 +236,7 @@ impl FeatureData {
             persisted: runtime_features_db,
             runtime,
             network: Arc::new(RwLock::new(network)),
+            remote_availability: Arc::new(RemoteAvailability::new()),
         })
     }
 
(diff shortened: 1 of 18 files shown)


crates/meilisearch-types/src/network.rs
```diff
@@ -26,6 +31,8 @@ pub struct Remote {
     pub search_api_key: Option<String>,
     #[serde(default)]
     pub write_api_key: Option<String>,
+    #[serde(skip_deserializing)]
+    pub status: route::Status,
```
Kerollmops, line 35: Note that this struct is stored in the database. I am adding a field to it, which could lead to incompatibilities, so I made sure not to deserialize this new field. I kept the serialization because we are serializing it for the HTTP response. I tested it, and it is possible to operate a database that was created from the main branch (14efbaeef).

crates/index-scheduler/src/features.rs
```diff
@@ -278,6 +281,18 @@ impl FeatureData {
...
-        Network::clone(&*self.network.read().unwrap())
+        let mut network = Network::clone(&*self.network.read().unwrap());
+        for (remote_name, remote) in network.remotes.iter_mut() {
+            remote.status = if self.remote_availability.is_available(remote_name) {
+                Status::Available
+            } else {
+                Status::Unavailable
+            };
+        }
+        network
```
Kerollmops, line 292: We are hydrating the network availability here, and the network structure needs to be serialized (including the status field) in an HTTP response later.

ManyTheFish (dismissed):
LGTM

ManyTheFish (approved):

PR #6308, opened 2026-04-02
Prevent shell injection in benchmark workflows

curquiza:
## Related issue

Fixes https://linear.app/meilisearch/issue/ENGPROD-2411/security-report-github-actions-expression-injection-in-benchmark

GitHub’s guidance for untrusted / user-controlled input is to put it in an environment variable and reference "$VAR" in the shell. That keeps the value out of inline script substitution (where hostile quotes or shell syntax can still bite) and ensures the shell treats the whole string as a single argument.

For step outputs that are already produced inside the workflow (e.g. basename), wrapping the substituted value in double quotes in the shell ("${{ … }}") is enough to avoid re-tokenization on spaces/; in the path; using env there would also be fine and is mostly a style choice.

## Generative AI tools

- [X] This PR does not use generative AI tooling

## Requirements

### DIFF
--- a/.github/workflows/bench-manual.yml
+++ b/.github/workflows/bench-manual.yml
@@ -24,4 +24,4 @@ jobs:
 
       - name: Run benchmarks - workload ${WORKLOAD_NAME} - branch ${{ github.ref }} - commit ${{ github.sha }}
         run: |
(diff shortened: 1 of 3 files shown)


Kerollmops (approved):
LGTM, thanks 👏

Kerollmops:
Hey @curquiza 👋

I'm sorry, but this PR broke the ability to expand files (e.g., using the \* character), which is really impactful when we want to run many benchmarks. I'll ask my best bud Claude to improve this, but I doubt we can do anything but revert this PR to benefit from globing.

The [above check](https://github.com/meilisearch/meilisearch/pull/6308#issuecomment-4183558240) was meaningless because it used the CI configuration for main.

PR #6310, opened 2026-04-03
Try the latest version of LMDB for nested read txn

Kerollmops:
This PR is mostly to run the CI tests on the new LMDB version with support for nested read transactions.

This version is a bit better than the LMDB fork we are currently using because it has been cleaned up by Howard Chu (maintainer of LMDB). It is much more reliable than what I did previously (we had only one memory leak). I plan to make this version of LMDB the version heed is provided with in the future.


Kerollmops:
So, now that we've validated (using the CI) that the new patch works great, I'll rebase and [squash this PR](https://github.com/meilisearch/heed/pull/307) and release a new version of heed.

Issue #6293, opened 2026-03-17
API reference: code samples input and output don't match


Strift:
I noticed the API reference code samples’ input and output don’t match, see here:
- request has `limit = 3`
- response has `limit = 1`

<img width="998" height="1002" alt="Image" src="https://github.com/user-attachments/assets/5eaa0423-f766-4a17-9df5-d341502d5acc" />

It's acceptable for this example, but it can become confusing for a less self-explanatory API example.

curquiza:
The change of the answer should be done in the engine code base: https://github.com/meilisearch/meilisearch, which is the simpler to avoid changing all the SDKs examples

0RBalaji:
Hi @curquiza, I'd like to pick this up as a good first issue!

My plan:
- Find the `getIndexes` API reference example in the documentation repo
- Fix the response block to correctly reflect the request parameters (e.g., `limit: 3` in the request should show `"limit": 3` in the response, not `1`)
- Ensure the request and response samples are consistent across all SDK examples

Let me know if this is still open and I'll get started. Thanks!

curquiza:
Hey @0RBalaji 
Thanks for your interest in this project 🔥 You are definitely more than welcome to open a PR for this!

For your information, we prefer not assigning people to our issues because sometimes people ask to be assigned and never come back, which discourages the volunteer contributors from opening a PR to fix this issue.
We will accept and merge the first PR that fixes correctly and well implements the issue following our [contributing guidelines](../blob/main/CONTRIBUTING.md).   

We are looking forward to reviewing your PR 😊

0RBalaji:
I’ve opened a PR to fix this: [Fix mismatch between request and response in list indexes example](https://github.com/meilisearch/meilisearch/pull/6315)

PR #6313, opened 2026-04-06
docs(openapi): align list API keys limit example

Hartnakig:
## Summary
- update the `List API keys` OpenAPI response example so its `limit` matches the request sample shown in the API reference

## Why
- addresses #6293 by removing a confusing mismatch between the request and response code samples on the keys API reference page

## Validation
- confirmed the current mismatch on https://www.meilisearch.com/docs/reference/api/keys
- could not run `cargo run -p openapi-generator` locally because this environment does not currently have the Rust toolchain installed

### DIFF
--- a/crates/meilisearch/src/routes/api_key.rs
+++ b/crates/meilisearch/src/routes/api_key.rs
@@ -140,7 +140,7 @@ impl ListApiKeys {
                         "updatedAt": "2021-11-12T10:00:00Z"
                     }
                 ],
-                "limit": 20,
+                "limit": 3,
                 "offset": 0,
                 "total": 1
             }


crates/meilisearch/src/routes/api_key.rs
```diff
@@ -140,7 +140,7 @@ impl ListApiKeys {
                         "updatedAt": "2021-11-12T10:00:00Z"
                     }
                 ],
-                "limit": 20,
+                "limit": 3,
```
Kerollmops, line 143: Thanks but I think you must also update the total and the results to include more entries.

Kerollmops (changes requested):

PR #6314, opened 2026-04-06
Document join: filtering

ManyTheFish:
## Related issue

Fixes [EXP-973](https://linear.app/meilisearch/issue/EXP-973)

## Generative AI tools

    - generate tests in `crates/meilisearch/tests/search/document_join.rs`
    
# Changelog

This enhancement extends the `Cross-index document hydration introduced` in [v1.39.0](https://github.com/meilisearch/meilisearch/releases/tag/v1.39.0) by allowing the user to filter on the foreign indexes to retrieve the documents.

> 📓 Note: This implementation doesn't support a remote sharding environment

## `foreignKeys` experimental feature

The`foreignKeys` experimental feature must be activated to use the foreign filters:

```
curl -X PATCH 'http://127.0.0.1:7700/experimental-features' \
  --data-binary '{"foreignKeys": true}'
```

## `foreignKeys` + `filter` index setting

To be able to use the foreign filters, the related field must be set as a foreignKey and as a filterableAttribute in `/indexes/{index_uid}/settings`:
```jsonc
{
	// new setting, an array of foreign keys that allows multiple foreign relationships between indexes
	"foreignKeys":  [
		{
			// the path in the JSON document containing foreign document ids
			"fieldName": "actors",
			
			// the UID of the foreign index containing the documents to fetch during hydration
	        "foreignIndexUid": "actors"
		}
	],
	// the actors field must be filterable on equality
	"filterableAttributes": [
		{
	    	"attributePatterns": ["actors"],
	     	"features": {
				"facetSearch": false,
				"filter": {
					"equality": true,
					"comparison": false
				}
			}
		}
	]
}
```

## filtering using the `_foreign` filter

On the search route, a new `_foreign` verb has been introduced and should be used as follows:

```jsonc
{
	"q": "<query>",

	// filters on the movie index:
	// genres = action
    // AND
    // the foreign documents from the actor index match:  birthday STARTS WITH \"1958-\" AND popularity >= 3.5
	"filter": "genres = action AND _foreign(actors, birthday STARTS WITH \"1958-\" AND popularity >= 3.5)"
}
```
 
> Note: nesting foreign filters is not supported and will return an error

<details>

<summary>

## Example of usage

</summary>

### Prerequisites

* Meilisearch running on `127.0.0.1:7700` on the `document-join-hydration` branch.

### Step 1: Enable Foreign Keys Feature

```bash
curl -X PATCH 'http://127.0.0.1:7700/experimental-features' \
  --data-binary '{"foreignKeys": true}'
```

#### Create the `actors` index

```bash
curl -X POST 'http://127.0.0.1:7700/indexes' \
  --data-binary '{"uid": "actors", "primaryKey": "id"}'
```

#### Create the `movies` index

```bash
curl -X POST 'http://127.0.0.1:7700/indexes' \
  --data-binary '{"uid": "movies", "primaryKey": "id"}'
```

### Step 3: Add Documents to the `actors` Index

```bash
curl -X POST 'http://127.0.0.1:7700/indexes/actors/documents' \
  --data-binary '[
  {"id": 1, "name": "Tom", "familyName": "Hanks", "birthDate": "1956-07-09"},
  {"id": 2, "name": "Meryl", "familyName": "Streep", "birthDate": "1949-06-22"},
  {"id": 3, "name": "Leonardo", "familyName": "DiCaprio", "birthDate": "1974-11-11"},
  {"id": 4, "name": "Emma", "familyName": "Watson", "birthDate": "1990-04-15"}
]'
```

### Step 4: Add Documents to the `movies` Index

```bash
curl -X POST 'http://127.0.0.1:7700/indexes/movies/documents' \
  --data-binary '[
  {"id": 1, "title": "Forrest Gump", "description": "The presidencies of Kennedy and Johnson, the Vietnam War, the Watergate scandal and other historical events unfold from the perspective of an Alabama man with an IQ of 75.", "actors": [1]},
  {"id": 2, "title": "The Devil Wears Prada", "description": "A smart but sensible new graduate lands a job as an assistant to Miranda Priestly, the demanding editor-in-chief of a high fashion magazine.", "actors": [2, 4]},
  {"id": 3, "title": "Inception", "description": "A thief who steals corporate secrets through the use of dream-sharing technology is given the inverse task of planting an idea into the mind of a C.E.O.", "actors": [3]},
  {"id": 4, "title": "Cast Away", "description": "A FedEx executive undergoes a physical and emotional transformation after crash landing on a deserted island.", "actors": [1]}
]'
```

### Step 5: Configure Foreign Keys on the `movies` Index

```bash
curl -X PATCH 'http://127.0.0.1:7700/indexes/movies/settings' \
  --data-binary '{"foreignKeys": [{"fieldName": "actors", "foreignIndexUid": "actors"}], "filterableAttributes": [{"attributePatterns": ["actors"],"features": {"facetSearch": false,"filter": {"equality": true,"comparison": false}}}]}'
```

### Step 6: Configure filterable on the `actors` Index

```bash
curl -X PATCH 'http://127.0.0.1:7700/indexes/actors/settings' \
  --data-binary '{"filterableAttributes": [{"attributePatterns": ["birthDate"],"features": {"facetSearch": false,"filter": {"equality": true,"comparison": false}}}]}'
```

### Step 7: Perform a Federated Search

```bash
curl -X POST 'http://127.0.0.1:7700/multi-search' \
  --data-binary '{
  "queries": [
    {
      "indexUid": "movies",
      "q": "Forrest",
      "filter": "_foreign(actors, birthDate = \"1956-07-09\")"
    }
  ],
  "federation": {
    "limit": 20,
    "offset": 0
  }
}'
```

### Expected Result

The federated search should return movie documents with the `actors` array automatically hydrated with full actor objects instead of just IDs:

```json
{
  "hits": [
    {
      "id": 1,
      "title": "Forrest Gump",
      "description": "...",
      "actors": [
        {
          "id": 1,
          "name": "Tom",
          "familyName": "Hanks",
          "birthDate": "1956-07-09"
        }
      ],
      "_federation": {
        "indexUid": "movies",
        "queriesPosition": 0,
        "weightedRankingScore": 0.9848484848484849
      }
    }
  ],
  "processingTimeMs": 208,
  "limit": 20,
  "offset": 0,
  "estimatedTotalHits": 1
}
```

</details>


crates/filter-parser/src/lib.rs
ManyTheFish, line 727: create maintainance issue to factorize all the `<function_call>(<parameter....>)`

crates/index-scheduler/src/filter.rs
```diff
@@ -0,0 +1,306 @@
...
+                    unreachable!()
+                };
+                Ok(IndexFilterCondition::In { fid, els })
+            })
+            .map(|condition| Some(IndexFilter { condition }))
+        })
+        .collect()
+}
+
+pub fn filters_into_index_filters_unchecked<'a>(
```
ManyTheFish, line 244: add some documentation

crates/meilisearch/src/search/mod.rs
```diff
@@ -1622,15 +1623,20 @@ pub fn fuse_filters(left: Option<Value>, right: Option<Value>) -> Option<Value>
     }
 }
 
-pub fn prepare_search<'t>(
+#[allow(clippy::too_many_arguments)]
+pub fn prepare_search<'t, 'q>(
```
ManyTheFish: lifetime 'q may not be needed anymore?

crates/index-scheduler/src/filter.rs
```diff
@@ -0,0 +1,306 @@
...
+                    fid.fragment()
+                ))),
+                index_uid: Some(source_index_uid.as_ref().to_string()),
+            })?;
+
+        filters_per_foreign_index.entry(foreign_index_uid.clone()).or_default().push(i);
+    }
+
+    // open each foreign index once and process the filters
+    // TODO: do remote document filtering here
```
ManyTheFish:
```suggestion
    // TODO: do remote document filtering here (linear: EXP-1027)
```

crates/index-scheduler/src/filter.rs
```diff
@@ -0,0 +1,306 @@
...
+    foreign_keys_per_index: &ForeignKeysPerIndex,
+    index_scheduler: &IndexScheduler,
+    progress: &Progress,
+) -> Result<Vec<Option<IndexFilter<'a>>>> {
+    // list all the foreign filters
+    let mut foreign_filters = Vec::new();
+    for (index_uid, filter) in filters.iter() {
+        let Some(filter) = filter else { continue };
+        for foreign_filter in filter.condition.list_foreign_filters() {
+            foreign_filters.push((index_uid.clone(), Some(foreign_filter.clone())));
```
ManyTheFish: parse foreign filters here instead of L155

crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -1035,6 +1074,12 @@ impl PartitionedQueries {
                                         meilisearch_types::error::Code::InvalidMultiSearchRemote,
                                     ));
                                 };
+
+                                // Insert back the filter into the query as a string before sending it to the remote
+                                query.filter = precomputed_filter
+                                    .as_ref()
+                                    .map(|f| serde_json::Value::String(f.condition.to_string()));
```
YoEight: Unfortunately, commit 4ae71410bc5ad9494cea203c0bf55f4b4367bd96 didn't fix the issue as it still persists with compounded filter.

I have a repro when setting up two remotes using this payload for a search:

```json
{
    "q": "Batman Returns",
    "filter": "kind = keep AND tenant = b",
    "useNetwork": true
}
```
`f.condition.to_string()` produces `AND[kind = keep, tenant = b, ]` which the remote parser rejects. I think your last commit only addressed braces issues with  the token's `Display` implementation.

crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -103,10 +107,51 @@ pub async fn perform_federated_search(
...
-        .runtime_features()
-        .foreign_keys
-        .then(|| HydrationContext::new(queries.iter().map(|q| q.index_uid.to_string())));
+    let filter_values = queries.iter_mut().map(|q| q.filter.take()).collect::<Vec<_>>();
+    let (mut hydration_cache, precomputed_filters) = if features.runtime_features().foreign_keys
+        && !is_proxy
+    {
+        let index_uids: Vec<_> =
+            queries.iter().map(|q| SourceIndexUid(Rc::from(q.index_uid.as_str()))).collect();
+        let foreign_keys_settings = retrieve_foreign_keys_settings(&index_scheduler, &index_uids)?;
```
YoEight, line 116: This loads foreign key settings too early on the first node that receives the request. If a query uses `federationOptions.remote`, the index might exist only on the remote node. In that case, `retrieve_foreign_keys_settings` tries to find the index locally and fails with `index_not_found` before the query is sent to the remote. The `!is_proxy` check only skips this on the remote node after proxying, not on the first node handling the request.

YoEight (changes requested):
Nice work!

I only found two major issues so far.  I'm looking for lower priority ones now.

crates/index-scheduler/src/scheduler/process_export.rs
```diff
@@ -84,6 +85,13 @@ impl IndexScheduler {
             let index = self.index(uid)?;
             let index_rtxn = index.read_txn()?;
             let filter = filter.as_ref().map(Filter::from_json).transpose().map_err(err)?.flatten();
+            let filter = filter
+                .map(|f| {
+                    // evaluate foreign key filter
+                    filter_into_index_filter(f, &index, &index_rtxn, self, &progress, uid)
```
YoEight, line 91: because we didn't get `filter` through the `parse_filter` function, it's possible for a user to pass a filter with `_foreign` even if the feature is disabled.

YoEight:
About tests, happy-path coverage for search is fine but not much for negative paths:
 * using `_foreign` on a field that is not declared in `foreignKeys`
 * using `_foreign` on an index with no `foreignKeys` configured
 * when using nested foreign filters
 * when using `_foreign` if the inner filter validation propagates correctly. For example with `_foreign(author, birthday = "1958-06-15")`, check if `birthday` is not filterable if the error propagates correctly to the user.
 * when using both a regular and foreign filter, when either one fails if we report the right one

crates/meilisearch/src/search/federated/perform.rs
ManyTheFish: I discussed this with Louis.
Even if that's a possibility, we consider that all the instances of the network have all the index information.

crates/meilisearch/src/search/federated/perform.rs
ManyTheFish: Fixed in [b88e12e](https://github.com/meilisearch/meilisearch/pull/6314/changes/b88e12e05c4b442f29da2868ede062e4a74b5c9f):
- Made a different serialization
- Added tests

crates/index-scheduler/src/scheduler/process_export.rs
ManyTheFish: wow! But it means it's the same issue for the CONTAINS operator and all the other features checked in parse_filter 😱

ManyTheFish: done in a different PR at this commit:
https://github.com/meilisearch/meilisearch/pull/6319/commits/9de76d104883cffbc1344ae33dd0796b99ddf4b1

crates/meilisearch/src/search/federated/perform.rs
YoEight: I will resolve this then. It might be worthwhile documenting that assumption though but it's not a blocking issue.

crates/benchmarks/benches/utils.rs
```diff
@@ -189,7 +192,7 @@ pub fn run_benches(c: &mut criterion::Criterion, confs: &[Conf]) {
                                 .terms_matching_strategy(TermsMatchingStrategy::default());
                             if let Some(filter) = conf.filter {
                                 let filter = Filter::from_str(filter).unwrap().unwrap();
-                                search.filter(filter);
+                                search.filter(filter_to_index_filter(filter));
```
YoEight, line 195: Do you think your filter patch also addresses this one here as well?

context: https://github.com/meilisearch/meilisearch/pull/6314#discussion_r3048496796

crates/benchmarks/benches/utils.rs
```diff
@@ -411,3 +414,36 @@ impl<R: Read> Iterator for CSVDocumentDeserializer<R> {
...
+        }
+        FilterCondition::GeoBoundingBox { top_right_point, bottom_left_point } => {
+            IndexFilterCondition::GeoBoundingBox { top_right_point, bottom_left_point }
+        }
+        FilterCondition::GeoPolygon { points } => IndexFilterCondition::GeoPolygon { points },
+        FilterCondition::Foreign { .. } => {
+            unreachable!("Foreign filters are not supported in index conditions")
+        }
+    }
+}
```
YoEight, line 449: suggestion: do you think having a `From<Filtercondition>` implementation for `IndexFilterCondition` be more idiomatic?

Same can be said with a potential `From<Filter>` implementation for `IndexFilter`

crates/filter-parser/src/lib.rs
```diff
@@ -149,6 +232,100 @@ pub enum VectorFilter<'a> {
...
+                filter.fids(depth)
+            }
+            Self::And(subfilters) | Self::Or(subfilters) => {
+                let depth = depth.saturating_sub(1);
+                Box::new(subfilters.iter().flat_map(move |f| f.fids(depth)))
+            }
+            _ => Box::new(std::iter::empty()),
+        }
+    }
+}
```
YoEight, line 264: suggestion: I understand that this is not a hot path at all but the version I propose does less allocations. Also it does no longer use recursion, is stack safe while remaining a depth-first traversal. It renders the `depth` parameter useless even if it's still there to keep the original intent. Again, it's just a suggestion. Deal with it as you please 😄 
```suggestion
impl<'a> IndexFilterCondition<'a> {
    pub fn fids(&self, depth: usize) -> impl Iterator<Item = &Token<'a>> {
        FidIter { stack: vec![(depth, self)] }
    }
}

struct FidIter<'a, 'b> {
    stack: Vec<(usize, &'a IndexFilterCondition<'b>)>,
}

impl<'a, 'b> Iterator for FidIter<'a, 'b> {
    type Item = &'a Token<'b>;

    fn next(&mut self) -> Option<Self::Item> {
        loop {
            let (depth, current) = self.stack.pop()?;

            match current {
                IndexFilterCondition::Condition { fid, .. }
                | IndexFilterCondition::In { fid, .. } => return Some(fid),

                IndexFilterCondition::Not(next) if depth > 0 => {
                    self.stack.push((depth - 1, &next));
                }

                IndexFilterCondition::And(others) | IndexFilterCondition::Or(others)
                    if depth > 0 =>
                {
                    self.stack.extend(std::iter::repeat(depth - 1).zip(others.iter()));
                }

                _ => {}
            }
        }
    }
}
```

YoEight (commented):
Nice improvements!

However, I don't see the test coverage I suggested here: https://github.com/meilisearch/meilisearch/pull/6314#issuecomment-4203178730

I only added a couple of comments

crates/benchmarks/benches/utils.rs
ManyTheFish: Yes, but I don't want to allow an easy public path to convert a filter to an index filter.
Because you can't safely convert a filter to an index filter safely using from.

crates/filter-parser/src/lib.rs
ManyTheFish: created here:
https://github.com/meilisearch/meilisearch/issues/6322

crates/filter-parser/src/lib.rs
ManyTheFish: applied, thanks

crates/benchmarks/benches/utils.rs
ManyTheFish: We are in the benchmarks, so checking the experimental feature doesn't make sense to me. 🤔

YoEight: It does if you want to benchmark exactly what happens in production

YoEight (approved):
Well done!

PR #6316, opened 2026-04-07
Use the latest version of heed with nested rtxns support

Kerollmops:
This PR bumps the versions of crates that use heed to [the latest version, v0.22.1](https://github.com/meilisearch/heed/releases/tag/v0.22.1). This version finally stabilized a long-standing piece of work we were doing with Howard Chu: nested read transactions. We no longer have to rely on unstable pre-releases, but rather on a clean, stable version of LMDB (still a fork, but a better one).

## Generative AI tools


ManyTheFish (approved):
Stabilization! 🥲 🎉

PR #6318, opened 2026-04-08
Restore globbing functionality in the bench pr workflow

Kerollmops:
Fixes a regression from #6308, where the globbing functionality was broken and made it hard to benchmark multiple workflow files simultaneously.

## Generative AI tools

    - I asked Claude to [investigate this PR](https://github.com/meilisearch/meilisearch/pull/6308) to restore the globbing behavior.

### DIFF
--- a/.github/workflows/bench-pr.yml
+++ b/.github/workflows/bench-pr.yml
@@ -75,7 +75,7 @@ jobs:
           cargo xtask bench --api-key "${{ secrets.BENCHMARK_API_KEY }}" \
              --dashboard-url "${{ vars.BENCHMARK_DASHBOARD_URL }}" \
              --reason "[Comment](${{ github.event.comment.html_url }}) on [#${{ github.event.issue.number }}](${{ github.event.issue.html_url }})" \
-             -- "$BENCH_ARGS" > benchlinks.txt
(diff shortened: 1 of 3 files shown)


.github/workflows/bench-pr.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           cargo xtask bench --api-key "${{ secrets.BENCHMARK_API_KEY }}" \
              --dashboard-url "${{ vars.BENCHMARK_DASHBOARD_URL }}" \
              --reason "[Comment](${{ github.event.comment.html_url }}) on [#${{ github.event.issue.number }}](${{ github.event.issue.html_url }})" \
-             -- "$BENCH_ARGS" > benchlinks.txt
+             -- $BENCH_ARGS > benchlinks.txt
```
Kerollmops, line 78: Claude wording:
> The fix is safe because **environment variables assigned via `env:` are set directly by GitHub Actions without shell interpretation**. This prevents injection attacks—malicious input like `$(command)` becomes a literal string, not executable code. When we later use the unquoted variable, the shell only performs word-splitting and glob expansion on the already-sanitized value. Combined with the permission checks requiring push access, this maintains security while restoring glob functionality.

curquiza (approved):

PR #6323, opened 2026-04-09
Rename some of the search performance traces

ManyTheFish:
Rename some of the search performance traces to be more self-explanatory:

```
// changed
WaitForPermit -> WaitInQueue
Tokenize -> TokenizeQuery
Embed -> EmbedQuery
Filter -> EvaluateFilter
ResolveUniverse -> EvaluateQuery
KeywordSearch -> KeywordRanking
PlaceholderSearch -> PlaceholderRanking
SemanticSearch -> SemanticRanking

// not changed
Format
FacetDistribution
Personalization
```

### `WaitForPermit -> WaitInQueue`

`WaitForPermit` is related to how the code handles the search queue, but we don't really talk about permits to users,  only the search queue. `WaitInQueue` makes more sense from the user POV; however, I was thinking about `WaitForAvailableThread` or `WaitForAnAvailableThread`, longer, but we directly understand the source of the bottleneck.

### `Tokenize -> TokenizeQuery` & `Embed -> EmbedQuery`

Use the same pattern for both terms because they refer to an equivalent process for the different search methods.

### `Filter -> EvaluateFilter` & `ResolveUniverse -> EvaluateQuery`

Use the same pattern for both terms because they refer to an equivalent process, retrieving a set of documents related to a query or a filter.
Moreover, `ResolveUniverse` can only be understood by a Meilisearch developer. `EvaluateQuery` makes more sense from the user POV.

### `KeywordSearch -> KeywordRanking` & `PlaceholderSearch -> PlaceholderRanking` & `SemanticSearch -> SemanticRanking`

Replacing `xxxxSearch` with `xxxxRanking` is more accurate about what the code is doing. Here, we already retrieved the candidates; we rank them now.

### DIFF
--- a/crates/meilisearch/src/routes/chats/chat_completions.rs
+++ b/crates/meilisearch/src/routes/chats/chat_completions.rs
@@ -350,7 +350,7 @@ async fn process_search_request(
     let search_kind =
         search_kind(&query, index_scheduler.get_ref(), index_uid.to_string(), &index)?;
 
-    progress.update_progress(TotalProcessingTimeStep::WaitForPermit);
+    progress.update_progress(TotalProcessingTimeStep::WaitInQueue);
     let permit = search_queue.try_get_search_permit().await?;
     progress.update_progress(TotalProcessingTimeStep::Search);
     let features = index_scheduler.features();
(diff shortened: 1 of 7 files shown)


Kerollmops (approved):
Looks good to me 👌

mayasssh:
Looks good to me too, much more self-explanatory! Thanks, @ManyTheFish!

PR #6325, opened 2026-04-09
Speedup facet search generation

Kerollmops:
This PR improves facet search indexing performance by avoiding full scans of too many database entries. In a follow-up PR, I plan to parallelize the generation of the facet search FSTs so that the more facets, the more parallelization.

## Generative AI tools


crates/milli/src/fields_ids_map/global.rs
```diff
@@ -117,6 +124,21 @@ impl<'indexing> GlobalFieldsIdsMap<'indexing> {
...
     }
+
+    /// Iterate over all fields with their metadata.
+    ///
+    /// This first iterates over the global fields, then over
+    /// the local fields so you should take care of possible duplicates.
+    pub fn for_each_metadata(&self, mut f: impl FnMut(FieldId, &str, Metadata)) {
+        let global = self.global.read().unwrap();
+        for (field_id, field_name, metadata) in global.iter() {
+            f(field_id, field_name, metadata);
```
dureuill, line 125: wrap in catch_unwind to prevent poisoning, then in case of panic, propagate it after dropping the lock.

dureuill: not actually necessary for a read lock

crates/milli/src/fields_ids_map/global.rs
```diff
@@ -117,6 +124,21 @@ impl<'indexing> GlobalFieldsIdsMap<'indexing> {
...
+        }
+        drop(global);
+        for (field_id, field_name, metadata) in self.local.iter() {
+            f(field_id, field_name, metadata);
+        }
```
dureuill: this will always return duplicates after iterating on global, as local acts as a local cache.

Please remove

crates/milli/src/update/new/indexer/post_processing/mod.rs
```diff
@@ -218,30 +221,56 @@ fn compute_facet_search_database(
...
+                .prefix_iter(txn, field_id)
+                .map_or_else(
+                    |e| Either::Left(iter::once(Err(e))),
+                    |it| {
+                        Either::Right(
+                            it.remap_types::<FacetGroupKeyCodec<StrRefCodec>, DecodeIgnore>()
+                                .filter(|r| r.as_ref().map_or(true, |(k, _)| k.level == 0)),
+                        )
+                    },
+                )
```
dureuill, line 250: Consider prefixing by the level 0 directly rather than filtering it:

```
pub struct FacetGroupKey<T> {
    pub field_id: u16,
    pub level: u8,
    pub left_bound: T,
}
```

crates/milli/src/update/new/indexer/post_processing/mod.rs
```diff
@@ -218,30 +221,56 @@ fn compute_facet_search_database(
 
     let localized_attributes_rules = index.localized_attributes_rules(wtxn)?;
     let filterable_attributes_rules = index.filterable_attributes_rules(wtxn)?;
+
+    // Get the list of ranges corresponding to the facets that are marked as facet searchable.
+    // This way we can avoid iterating over all the facet values.
+    let mut facet_searchable_fids = BTreeSet::new();
+    global_fields_ids_map.for_each_metadata(|field_id, _field_name, metadata| {
```
dureuill, line 228: is this function used in the settings reindexer? If so, we should take care of the fact that the metadata may be different in the previous vs current settings, meaning some fields could become facet-searchable or become NOT facet-searchable

crates/milli/src/update/new/indexer/post_processing/mod.rs
Kerollmops: Good catch! [The level is encoded just after the fieldID](https://github.com/meilisearch/Meilisearch/blob/adcdedf95919cabaef8f60985f3f962c1a1165a1/crates/milli/src/heed_codec/facet/mod.rs#L68-L75) ✅

crates/milli/src/update/new/indexer/post_processing/mod.rs
ManyTheFish: > some fields could become facet-searchable

If we ensure that it is the new field ID map, then it will index everything; however, we may reindex fields that were already indexed.
We may want to rely on the field delta given in the parameters of the post-processing 🤔 

> become NOT facet-searchable

No issue there, there is a pre-processing step deleting all the data we have to delete during the setting re-indexing

Kerollmops: > No issue there, there is a pre-processing step deleting all the data we have to delete during the setting re-indexing

Ok, so I can keep the code I have, right? It works as before and I can use the settings delta in a followup PR.

ManyTheFish: Yep, I just checked, the global field id map is built from the new field id map in the settings indexer :)

PR #6331, opened 2026-04-13
Fix legacy indexer issues

dureuill:
## Changelog

- Fix issues when using the legacy settings indexer:
  - When using the experimental feature "multimodal", removing a fragment would cause an internal error in subsequent search requests
  - When using the experimental feature "multimodal", modifying fragments would sometimes cause an internal error at indexing time
  - `regenerate: false` would be ignored when modifying embedder settings
  - Index a nested field declared as searchable even if its parents are not declared as searchable
- Add new CI run that tests the stable settings indexer

## Generative AI tools

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -158,6 +158,33 @@ jobs:
           command: test
           args: --locked -p meilisearch --features test-ollama ollama
 
+  test-legacy-indexer:
+    name: Test legacy settings indexer
+    runs-on: ubuntu-22.04
+    env:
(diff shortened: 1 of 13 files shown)


dureuill:
As we can see, the new CI check works as expected, as the remaining test that doesn't pass when using the legacy indexer fails as expected: https://github.com/meilisearch/meilisearch/actions/runs/24393052669/job/71243823906?pr=6331

ManyTheFish (approved):

dureuill (commented):
✅ Approving many's commit

PR #6334, opened 2026-04-14
Reduce the memory usage of the post processing

Kerollmops:
This PR removes the use of spooled temporary files to reduce the memory usage of the post-processing file, reduces the size of the data structures we write to disk, and introduces a new `MiniString` data structure to store up to 12-byte strings, improving memory efficiency.

### Changelog

Reduces allocated memory when computing prefixes and speeds up some operations to avoid unnecessary deserialization. If you still see high memory usage while the engine is post-processing, we recommend using the `--experimental-reduce-indexing-memory-usage` option, which makes the engine write directly to the memory-mapped database file.


crates/milli/src/update/new/indexer/post_processing/mod.rs
```diff
@@ -118,9 +112,12 @@ where
 
     let mut output = BTreeSet::new();
     loop {
+        // Current prefixes are only inserted once and each prefix is only inserted once.
         if current_word.as_bytes().starts_with(current_prefix) {
             let current_prefix = std::str::from_utf8(current_prefix)?;
-            output.insert(current_prefix.into());
+            // safety: Prefixes are 3 bytes or less
+            let current_prefix = MiniString::new(current_prefix).unwrap();
```
ManyTheFish, line 119: `MiniString::new` reconverts the string into bytes internally, but this string has just been converted from bytes.

Should we allow to instantiate MiniString from bytes?

ManyTheFish (approved):
Small mini micro comment.

Question! I see that we replaced `spooled_tempfile`; we introduced it to avoid writing on disk if it's not necessary.
Why do we remove it?

crates/milli/src/update/new/indexer/post_processing/mod.rs
Kerollmops: I would rather prefer that we keep the constructor safe. In this case we chack that the current_prefix is a valid string and we later expose the `as_str(&self) -> &str` method that does an `unsafe` call to `from_utf8_unchecked` to output the `&str`.

Kerollmops:
> Question! I see that we replaced `spooled_tempfile`; we introduced it to avoid writing on disk if it's not necessary.
> Why do we remove it?

I removed it because the spooled temp file is consuming a lot of memory (max memory / number of threads) and is not working as I originally intended. Once the spooled in-memory file is full, it writes the entire in-memory data to disk and then continues writing the remaining to disk. I thought it was keeping the in-memory data and writing the remaining bytes to disk without dumping everything. By removing the spooled file, we have better memory usage and a simpler engine.

PR #6336, opened 2026-04-15
Promote dynamic search rules to index

YoEight:
## Description
This PR reworks dynamic search rules so they are stored and queried through a dedicated internal Meilisearch index instead of a scheduler-owned key-value database plus an in-memory runtime map.

## Changelog
This is mostly an internal architecture change. The goal is to make dynamic search rules more scalable and to prepare the feature for future activation conditions, especially filter-based conditions, by relying on the same primitives Meilisearch already uses for indexed lookup rather than repeatedly scanning all rules.

At a high level, this change:

* Promotes dynamic search rules to documents in a hidden internal index managed by the index scheduler.
* Keeps the public DSR API unchanged while introducing an index-friendly internal representation.
* Configures the internal index with searchable, filterable, and sortable fields needed to list rules and resolve search-time candidates.
* Moves rule candidate resolution into the index scheduler so search only receives rules that are relevant for the current index/query context.
* Preserves rule behavior for priority, pin deduplication, time windows, query-empty/query-contains conditions, and global vs index-scoped actions.
* Simplifies the search-side dynamic rules code so it focuses on applying already-selected pinning actions.
* Updates dumps and route handling to read DSRs from the new store.
* Support filter-based activation conditions.

## Breaking change
This PR doesn't import previously created DSRs. They need to be created again.

## Generative AI tools

## Requirements


crates/index-scheduler/src/dynamic_search_rules.rs
```diff
@@ -1,81 +1,959 @@
...
+                return this.intersect(other);
+            }
+
+            return false;
+        }
+
+        false
+    }
+
+    fn collect_facets(&self) -> Vec<String> {
```
dureuill, line 463: I tested a filter condition of:

```json
{"scope":"filter","filter":"'price' > '40'"}
```

I got no error message while sending this condition. However, if I pass a filter of `price = 42` or even `price > 40` in the query, then the rule with that filter condition does not activate, which is contrary to expectations.

This is because the facet value collector is ignoring comparison operators. I think it is smart to reuse the filter syntax and collect the facet values, however:

1. It is a lot of additional code that will need to follow the filter code somewhat closely
2. I'm worried of the impedance mismatch between the general filter syntax of Meilisearch and what is actually supported here.

I think it would be clearer from a user's standpoint to eschew the filter syntax completely, and instead accept a list of facets and values in the filter condition.

crates/index-scheduler/src/dynamic_search_rules.rs
```diff
@@ -1,81 +1,959 @@
...
+        fn go(ast: &FilterAst) -> Option<HashSet<String>> {
+            match ast {
+                FilterAst::Binary { lhs, op, rhs } => match (lhs.as_ref(), op, rhs.as_ref()) {
+                    (FilterAst::Ident(attr), FilterOp::Equal, FilterAst::Ident(value)) => {
+                        Some(HashSet::from_iter([format!("{attr}\\u0000{value}")]))
```
dureuill, line 468: I feel uneasy that we are relying on this `\u0000` trick to store facet values.

1. Sentinel values are dangerous, for example we can put a `\u0000` inside of JSON key, this may interfere with the use of the sentinel, if not now, perhaps in the future
2. My understanding is that we are storing the `"{attr}\\u0000{value}"` as a facet value. Facet values are stored as keys in LMDB to provide inverted indexes of values. Keys in LMDB are limited to 512B in size, which means that a facet value has a max size of around 468 bytes. Longer facet values are truncated. This may cause an unfortunate interaction because we are storing "key + value" as a facet value (https://github.com/meilisearch/meilisearch/blob/main/crates/milli/src/lib.rs#L138)

I'd rather like for us to add each facet as a filterable attribute (perhaps by having `facet_values.*` as a filterable attribute pattern), and then normally index the DSR as a document containing the `facet_values` subobject.

crates/index-scheduler/src/dynamic_search_rules.rs
```diff
@@ -1,81 +1,959 @@
...
+            settings.set_facet_search(true);
+
+            settings
+                .execute(&|| false, &Progress::default(), &options.ip_policy, Arc::default())
+                .map_err(dsr_milli_error)?;
+
+            wtxn.commit()?;
+        }
+
+        // TODO - should I check the index version? I presume yes
```
dureuill, line 584: we need to handle index migration.

When processing the database upgrade task, we need to call the function that upgrades an index on the special DSR index.

crates/index-scheduler/src/dynamic_search_rules.rs
```diff
@@ -1,81 +1,959 @@
...
+                });
+
+                if rule.actions.is_empty() {
+                    continue;
+                }
+
+                let normalized_query = ctx.query.map(milli::normalize_facet);
+
+                // all conditions were already checked individually, but if we have multiple conditions,
+                // we need to tell that they are all true as a whole for the rule to be activated
```
dureuill, line 885: is there a way to spare us that double check?

1. we want to avoid doing work on an unbounded number of rules
2. this means we have two implementations of the checks, which could be distinct and cause discrepancies.

dureuill (changes requested):
Hello 👋 

Thank you for the PR 

I did a first pass focussing on the new filter condition.

crates/index-scheduler/src/dynamic_search_rules.rs
```diff
@@ -1,81 +1,959 @@
...
+}
 
-mod db_name {
-    pub const DYNAMIC_SEARCH_RULES: &str = "dynamic-search-rules";
+fn matches_attribute_patterns(
+    rule: &DynamicSearchRule,
+    attribute_patterns: Option<&AttributePatterns>,
+) -> bool {
+    attribute_patterns.is_none_or(|patterns| {
+        !matches!(patterns.match_str(&rule.uid), PatternMatch::NoMatch | PatternMatch::Parent)
```
ManyTheFish, line 228:
```suggestion
        patterns.match_str(&rule.uid) == PatternMatch::Match
```

crates/index-scheduler/src/dynamic_search_rules.rs
```diff
@@ -1,81 +1,959 @@
...
+                            op,
+                            rhs: Box::new(go(child)?),
+                        };
+                    }
+
+                    Some(binary)
+                }
+
+                // unsupported
+                _ => None,
```
ManyTheFish, line 444: What's left?

We may want to enumerate them in case we introduce a new index filter (This will not compile, and we will be able to choose if we want to support DSR or not).

crates/index-scheduler/src/dynamic_search_rules.rs
```diff
@@ -1,81 +1,959 @@
...
+                    }
+
+                    milli::Condition::Contains { word, .. } => Some(FilterAst::Binary {
+                        lhs: Box::new(FilterAst::Ident(fid.fragment().to_string())),
+                        op: FilterOp::Contains,
+                        rhs: Box::new(FilterAst::Ident(milli::normalize_facet(word.fragment()))),
+                    }),
+
+                    // unsupported for now
+                    _ => None,
```
ManyTheFish, line 403: What's left?

We may want to enumerate them in case we introduce a new index filter (This will not compile, and we will be able to choose if we want to support DSR or not).

ManyTheFish (changes requested):

crates/index-scheduler/src/dynamic_search_rules.rs
ManyTheFish: A possible alternative could be to parse and store [Conditions](https://github.com/meilisearch/meilisearch/blob/475ed56e5612df0dbb826748add5f93e0e7d5500/crates/filter-parser/src/condition.rs#L19-L33), then compare them by implementing a method on `Condition`.

`collect_facets` would be a `collect_conditions`.

We would have a special case to handle if we want to support the `IN` and `NOT` operators. 🤔

crates/index-scheduler/src/dynamic_search_rules.rs
YoEight: My first implementation resolved activation fully at search time.

The issue is that we allow multiple conditions, and we treat them like they are all joined with AND. That is why we still need the second check.

At search time, the logic is mostly just a pre-filter. We could move all activation logic there only if each rule had a single activation condition.

Another option would be to express the activation condition only using the filter syntax. That should address this too.

crates/index-scheduler/src/dynamic_search_rules.rs
YoEight: At that level:
```
VectorExists
GeoLowerThan
GeoBoundingBox
GeoPolygon
```

I will enumerate them then.

crates/index-scheduler/src/dynamic_search_rules.rs
dureuill: we want to avoid checking all rules individually though, so we need to have an inverted index of "something" -> roaring bitmap of rule ids

I suggest we pick an easy and user-legible road of supporting explicit facet declarations

crates/index-scheduler/src/dynamic_search_rules.rs
YoEight: At this level:

```
    GreaterThan
    GreaterThanOrEqual
    Null
    Empty
    Exists
    LowerThan
    LowerThanOrEqual
    Between
    StartsWith
```

I will list all of them in the next commits

crates/index-scheduler/src/dynamic_search_rules.rs
dureuill: I'm sorry I must be missing something 😳 can we not compute the roaring for each condition and then take the intersection of that?

crates/index-scheduler/src/dynamic_search_rules.rs
```diff
@@ -1,81 +1,959 @@
...
+
+        let docids_with_query_scope = if query.is_some() {
+            let docids_with_contains = self.run_filter(
+                &rtxn,
+                &format!(r#"{base_filter} AND conditions.kind = "queryContains""#),
+            )?;
+            let docids_not_empty_query = self.run_filter(
+                &rtxn,
+                &format!(r#"{base_filter} AND conditions.kind = "queryIsNotEmpty""#),
+            )?;
```
ManyTheFish, line 796:
```suggestion
            let docids_with_contains = self.run_filter(
                &rtxn,
                &format!(r#"{base_filter} AND conditions.kind IN ["queryContains", "queryIsNotEmpty"]"#),
            )?;
```

dureuill:
Superseded by #6484

PR #6340, opened 2026-04-15
Normalize before filter

### LINKED ISSUE #6335: String comparisons with `>` yielding false positives
**Describe the bug**
Create an index with a filterable field, insert strings into it and then search with `>` on that field. The search results produce unexpected values.

**To Reproduce**
Start a Meilisearch and run the following script; observe that two records are returned even though the filter should only return one.

```
MEILI_HOST="http://localhost:7700"
INDEX_NAME="movies"

echo "Creating index..."
curl -X POST "$MEILI_HOST/indexes" \
  --json '{ "uid": "movies", "primaryKey": "id"}'

echo -e "\nSetting filterable attribute..."
curl -X PUT "$MEILI_HOST/indexes/movies/settings/filterable-attributes" \
  --json '[ "released" ]'

echo -e "\nAdding documents..."
curl -X POST "$MEILI_HOST/indexes/movies/documents" \
  --json '[
    {
      "id": 1,
      "title": "Inception",
      "released": "2026-01-01T00:00:00"
    },
    {
      "id": 2,
      "title": "The Dark Knight",
      "released": "2026-01-01T01:02:03"
    }
  ]'

echo -e "\nDone."

sleep 2

echo -e "This query should only return \"The Dark Knight\":"


curl $MEILI_HOST/indexes/movies/search --json '{"filter": "released > \"2026-01-01T01:00:00\""}'

```

**Expected behavior**
Only "The Dark Knight" should be a search result.

**Meilisearch version:**
1.42.1 (but also in 1.23.0)

**Additional context**
Doesn't seem to be architecture dependent. Negating the search negates the search results.

dureuill:
## Related issue

Fixes #6335 

# Changelog

Fixes a bug where string facet values appearing in `<, <=, >, >=` and `IN` filters where not normalized before comparison to facet values.

This would cause some values in document (e.g. `2026-01-01T00:00:00`) to appear to be higher than their filter counterpart, due to being normalized (e.g. like `2026-01-01t00:00:00`). 

Thanks to @njaard for [reporting the issue](https://github.com/meilisearch/meilisearch/issues/6335) :heart: 

# Implementation

- Normalize before passing the value to facet range search
- Add test that fails before the fix and passes after the fix.

Maintainability note: Such bugs could be prevented in the future by reifying "(hyper)normalized" values as dedicated types rather than using Strings and manually maintaining their invariants. Had the bounds accepted a `NormalizedStr<'a>` instead of a `&'a str`, the bug would not have been written in the first place.

## Generative AI tools

### DIFF
--- a/crates/meilisearch/tests/search/filters.rs
+++ b/crates/meilisearch/tests/search/filters.rs
@@ -1170,3 +1170,53 @@ async fn vector_filter_regenerate() {
     }
     "###);
 }
+
+#[actix_rt::test]
+async fn issue_6335() {
+    test_settings_documents_indexing_swapping_and_search(
+        &NESTED_DOCUMENTS,
+        &json!({"filterableAttributes": [{"attributePatterns": ["cattos"],
+        "features": {
+                  "facetSearch": false,
+                  "filter": {"equality": true, "comparison": true}
+        }}]}),
+        &json!({
+            "filter": "cattos < pestiféré"
+        }),
+        |response, code| {
+            snapshot!(code, @"200 OK");
+            snapshot!(json_string!(response["hits"]), @r#"
+            [
+              {
+                "id": 750,
+                "father": "romain",
+                "mother": "michelle",
+                "cattos": [
+                  "enigma"
+                ]
+              },
+              {
+                "id": 951,
+                "father": "jean-baptiste",
+                "mother": "sophie",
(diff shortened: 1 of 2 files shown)


ManyTheFish (approved):
Nice catch! 👍 

LGTM

PR #6344, opened 2026-04-20
Replace boolean by PatternMatch for searchable

ManyTheFish:
## Related issue

Change the Metadata structure to allow identifying if a field:
- match a feature
- doesn't match a feature
- is a parent field of a field matching the feature

This allows us to better explore the nested document without having to match the fields agaisnt the settings everywhere.

Diff:
```rust
pub struct Metadata {
    // old
    pub searchable: Option<Weight>,
    // new
    pub searchable: (PatternMatch, Option<Weight>),

    // old
    pub exact: bool,
    // new
    pub exact: PatternMatch,

    // old
    pub sortable: bool,
    // new
    pub sortable: PatternMatch,

    // old
    pub distinct: bool,
    // new
    pub distinct: PatternMatch,

    // old
    pub asc_desc: Option<FieldSortOrder>,
    // new
    pub asc_desc: (PatternMatch, Option<FieldSortOrder>),

    // old
    pub geo: bool,
    // new
    pub geo: PatternMatch,

    // old
    pub geo_json: bool,
    // new
    pub geo_json: PatternMatch,

    // old
    pub displayed: bool,
    // new
    pub displayed: PatternMatch,

    pub localized_attributes_rule_id: Option<NonZeroU16>,

    // old
    pub filterable_attributes_rule_id: Option<NonZeroU16>,
    // new
    pub filterable_attributes_rule_id: (PatternMatch, Option<NonZeroU16>),

    pub sort_by: OrderBy,
}
```

## Generative AI tools

### DIFF
--- a/crates/milli/src/update/new/indexer/mod.rs
+++ b/crates/milli/src/update/new/indexer/mod.rs
@@ -756,10 +756,10 @@ where
         old_fields_ids_map
             .iter_id_metadata()
             .filter_map(|(id, metadata)| {
-                if metadata.is_searchable()
+                if metadata.is_searchable() == PatternMatch::Match
                     && new_fields_ids_map
                         .metadata(id)
-                        .is_none_or(|metadata| !metadata.is_searchable())
+                        .is_none_or(|metadata| metadata.is_searchable() != PatternMatch::Match)
                 {
                     Some(id)
                 } else {
--- a/crates/milli/src/update/new/indexer/mod.rs
+++ b/crates/milli/src/update/new/indexer/mod.rs
@@ -266,17 +266,7 @@ where
         must_stop_processing,
         progress,
     )?;
-
-    // Clear the cellulite database when the geojson support is removed
-    if settings_delta.new_match_faceted_field(RESERVED_GEOJSON_FIELD_NAME) == PatternMatch::NoMatch
-    {
-        index.cellulite.clear(wtxn)?;
-    }
-
-    // Clear the geo rtree entry when the geo support is removed
-    if settings_delta.new_match_faceted_field(RESERVED_GEO_FIELD_NAME) == PatternMatch::NoMatch {
-        index.delete_geo_rtree(wtxn)?;
-    }
+    delete_old_geo_databases(wtxn, index, settings_delta, must_stop_processing, progress)?;
 
     // Clear word_pair_proximity if byWord to byAttribute
     let old_proximity_precision = settings_delta.old_proximity_precision();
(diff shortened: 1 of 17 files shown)


crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -526,6 +516,47 @@ where
...
+    if must_stop_processing() {
+        return Err(Error::InternalError(InternalError::AbortedIndexation));
+    }
+
+    let new_geo_is_faceted =
+        new_fields_ids_map.id_with_metadata(RESERVED_GEO_FIELD_NAME).is_some_and(
+            |(_, metadata)| metadata.is_faceted(new_filterable_rules) == PatternMatch::Match,
+        );
+    // Clear the geo rtree entry when the geo support is removed
+    if new_geo_is_faceted {
```
YoEight: It seems you are doing the opposite of what the comment is saying. You clear the geo rtree entry when the geojson support is still there.

crates/milli/src/update/new/indexer/mod.rs
```diff
@@ -526,6 +516,47 @@ where
...
+
+    let new_fields_ids_map = settings_delta.new_fields_ids_map();
+    let new_filterable_rules = settings_delta.new_filterable_rules();
+
+    let new_geojson_is_faceted =
+        new_fields_ids_map.id_with_metadata(RESERVED_GEOJSON_FIELD_NAME).is_some_and(
+            |(_, metadata)| metadata.is_faceted(new_filterable_rules) == PatternMatch::Match,
+        );
+    // Clear the cellulite database when the geojson support is removed
+    if new_geojson_is_faceted {
```
YoEight: It seems you are doing the opposite of what the comment is saying. You clear the cellulite database when the geojson support is still there.

ManyTheFish:
```suggestion
    if !new_geojson_is_faceted {
```

YoEight:
I have concerns about the new metadata helper that could be inconsistent with another check that we have in `milli`.  Say we have this configuration:

```json
{
  "filterableAttributes": [
    {
      "attributePatterns": ["_geo"],
      "features": {
        "filter": {
          "equality": false,
          "comparison": false
        }
      }
    }
  ]
}
```

If I understand correctly, the new metadata helper `is_geo_enabled` would return `true` in this case  while this version in `milli` should return `false`: https://github.com/YoEight/meilisearch/blob/02ff9a4a1cae79ce405638b6323fbd2f939f7f67/crates/milli/src/index.rs#L1085

**edit**: As discussed on Discord, it turns out is not a problem for `_geo` and `_geojson`.

YoEight (approved):
Nice work

PR #6346, opened 2026-04-22
Make the health route return a 500 when the task compaction succeeded

Kerollmops:
This PR changes the `/health` status to a `mustRestart` status when the server must restart, and specifically after a successful compaction has finished. The main benefit is that the Cloud will automatically restart the user instance after a successful task queue compaction.

## Generative AI tools


dureuill (approved):
nice and sweet 🌈

PR #6352, opened 2026-04-27
Introduce a feature to offload indexes to an S3 storage

Kerollmops:
This PR introduces a feature that keeps a small number of indexes on disk and offloads those that exceed the limit. This is not correlated to the current number of loaded (opened) indexes. There is a maximum of 18 open indexes simultaneously, and we plan to have a higher limit for the number of local indexes (like 50-100).

EDIT: following comment may be a bit outdated. We now have the available, unavailable, deleting and transfered fields in the `IndexMap`. The unavailable become an Lru and deleting is a simple HashSet.

### Notes About the Current Implementation

I have modified the `IndexMapper` structure to accept an async tokio runtime handle. This handle is used to fire asynchronous operations for uploading and downloading indexes. It seems the runtime is optional when creating the `IndexScheduler` (from which I obtain the runtime handle). Adding a runtime to the `IndexMapper` is probably only beneficial to the offloading system, which is an enterprise edition feature and shouldn't contaminate community-edition code. I suppose it's not very complex to make the `IndexMapper` accept a runtime handle, though...

I have introduced a new `closed` field in the `IndexMap` struct to track the indexes I need to upload when there are too many on disk, especially those I need to offload when the time comes. It's basically an LRU, but for indexes that aren't in my opened/available list. I don't yet have a good idea of how to behave when the engine starts, and we see that there are more indexes on disk than the `closed` capacity. The easiest solution would be to scan the on-disk indexes and upload extra indexes that must not be kept on disk (bonus: delete any potential unknown ones). Scanning disk indexes would scale well, since the number of indexes is small enough and defaults to 100. As to how we define the extra indexes to upload, it's hard to know. I can insert them into my LRU, which will cause the earliest-inserted UUIDs to be uploaded, depending on [the order they are returned by the directory scan](https://doc.rust-lang.org/stable/std/fs/fn.read_dir.html).

Another issue I have with the `closed` structure is that it doesn't account for open indexes and is essentially a mirror of the unavailable index list. A better approach would be to have two LRUs: one to track opened indexes and another to track unavailable ones. Both already exist, but the latter is not an LRU. By converting it into one, we can track which one must be uploaded. However, it also means that once we insert a fresh index into the unavailable ones and pop one, we need to track the upload status somewhere, and it cannot be tracked from the unavailable list, since it is already full at that point. We can introduce a new `uploading` list of indexes to track the status of their upload, and ensure _we are not downloading_ an index that is being uploaded, instead keeping it inaccessible (?) while it is uploading. By keeping the index in the "indexes" folder, we keep it in a valid state for a future startup: if the engine restarts, the indexes remain available, and we do not download a potentially outdated version from the S3 bucket. Once the upload is successful, we immediately delete the index from disk. Next time the upload task is pulled, and the task is successful, we remove the entry from the `uploading` set.

A better approach for the **index downloader and uploader is to have a single async task** dedicated to downloading that accepts tasks via an async channel. Network tasks would notify of completion via a one-shot channel, i.e., [`Shared<oneshot::Receiver>`](https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.shared). On the other end, the unavailable set of indexes could wait for the upload/download completion by using the other end of the one-shot channel. By doing that, we factorize a lot and drastically reduce the download and upload weights of indexes. Upload is done [asynchronously and in parallel](https://github.com/meilisearch/meilisearch/blob/bb968aaf0d2981a615d8e748d13c930b3cbc76c7/crates/index-scheduler/src/scheduler/enterprise_edition/s3.rs#L374-L566) (with a maximum in-flight parts), and could even be prioritized depending on the main network tasks channel. Another advantage of this approach is the control we have over the memory used for uploading and downloading indexes, as we can limit the number of tasks processing simultaneously to one upload and one download. I see one limitation to this approach: a `Shared<oneshot::Receiver>` requires `F::Output: Clone` and `io::Error` is not, I could [try using an `Arc<io::Error>`, maybe](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#impl-Error-for-Arc%3CT%3E).

About the function signature of the different **IndexMapper methods. I would like them to be async**. It changes many calls to this method, but it's definitely a better approach. First, because opening an index can take time ([like a lot](https://github.com/meilisearch/meilisearch/blob/bb968aaf0d2981a615d8e748d13c930b3cbc76c7/crates/index-scheduler/src/index_mapper/mod.rs#L413-L417)), we mostly call `IndexMapper::index` in async contexts. That's really bad for an async HTTP runtime. Right now, we are already doing a lot of work using loops, mutexes, and locks. We may have to switch to using async mutexes. I just have some doubts about the way we will use these async methods in our non-async tests. It's feasible, but a lot of changes.

Note that to be able to make those methods async, the `RoTxn` and `RwTxn` must be able to jump over `.await` point. This applies only to `Send` types. [`RoTxn<WithoutTls>` is `Send`](https://docs.rs/heed/latest/heed/struct.RoTxn.html#impl-Send-for-RoTxn%3C'_,+WithoutTls%3E), and `RwTxn` wraps an `RoTxn<WithoutTls>`, [so it is also `Send`](https://docs.rs/heed/latest/heed/struct.RwTxn.html#impl-Send-for-RwTxn%3C'p%3E).

```rust
impl IndexMapper {
  // This method tries to block_on the task that downloads the index.
  // Once the download is complete, the index is marked as missing (among the opened ones),
  // the loop continues, opens the index, and returns it.
  pub async fn index(&self, rtxn: &RoTxn, name: &str) -> Result<Index>;

  // Calling `index` internally so it's inherently async.
  pub async fn create_index(
      &self,
      wtxn: RwTxn,
      name: &str,
      date: Option<(OffsetDateTime, OffsetDateTime)>,
      shards: Option<Shards>,
  ) -> Result<Index>;

  // Doing a loop and sleeping for 6s a couple of times to wait for the index to close.
  // Would be better if it were spawning a blocking task to wait for closing.
  pub async fn delete_index(&self, wtxn: RwTxn, name: &str) -> Result<()>;

  // Calls close_for_resize, which calls IndexMap::close, which should probably be async
  pub async fn resize_index(&self, rtxn: &RoTxn, name: &str) -> Result<()>;

  // Calls close_for_resize, which calls IndexMap::close, which should probably be async
  pub async fn close_index(&self, rtxn: &RoTxn, name: &str) -> Result<()>;

  // Calls close_for_resize, which calls IndexMap::close, which should probably be async
  pub async fn rollback_index(
      &self,
      rtxn: &RoTxn,
      name: &str,
      to: (u32, u32, u32),
  ) -> Result<RollbackOutcome>;

  // What should this method do?
  // A simple loop in an async context with basic .await on the different indexes?
  //  I probably need to switch the FnMut into an AsyncFnMut.
  // Right now, the loop tries to download and offload all indexes.
  pub async fn try_for_each_index<U, V>(
        &self,
        rtxn: &RoTxn,
        f: impl FnMut(&str, &Index) -> Result<U>,
    ) -> Result<V>
    where
        V: FromIterator<U>;
}

impl IndexMap {
  // The method currently returns an Option<Index> to indicate that the index is being downloaded.
  // I would rather make the user .await for the index to be ready.
  pub async fn create(
        &mut self,
        uuid: &Uuid,
        path: &Path,
        date: Option<(OffsetDateTime, OffsetDateTime)>,
        enable_writemap: bool,
        map_size: usize,
        create_or_open: CreateOrOpen,
    ) -> Result<Index>;

  pub async fn close_for_resize(
      &mut self,
      uuid: &Uuid,
      enable_mdb_writemap: bool,
      map_size_growth: usize,
  );

  // This method will probably .await a download to be ready.
  fn async close(
      &mut self,
      uuid: Uuid,
      index: Index,
      enable_mdb_writemap: bool,
      map_size_growth: usize,
  );

  // This method should probably send a delete request for the index to the S3 bucket and .await for it.
  pub async fn start_deletion(
      &mut self,
      uuid: &Uuid,
  ) -> std::result::Result<Option<EnvClosingEvent>, Option<ClosingIndex>>;
}
```

Note that it is invalid [to `block_on` a future in an async context](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.block_on), and guess what? We are always opening indexes in an async context, directly in the HTTP route handler. As I designed the download system via an async task, I had to do the following to make it work, and it's ugly. It's ugly because I am spawning a blocking task (using a dedicated thread) to avoid blocking the async task, and inside... I am using a block_on to wait for an async task to complete...

```rust
// before, it was simple
let index = index_scheduler.index(&index_uid)?;

// right now, it's a bit ugly and not efficient
let index = tokio::task::spawn_blocking({
    let index_scheduler = index_scheduler.clone();
    let index_uid = index_uid.clone();
    move || index_scheduler.index(&index_uid)
})
.await
.unwrap()?;

// in the future, I want something as clean as that
let index = index_scheduler.index(&index_uid).await?;
```

To **avoid TOCTOU and such**, I would prefer opening an index and handling the `NotFound` io error rather than checking whether the path exists. To ensure we do not open a partial index (from a download task), we can download indexes to a dedicated downloads subfolder and persist them after the download completes. I [heard that renaming files is atomic](https://rcrowley.org/2010/01/06/things-unix-can-do-atomically.html) and can only be done when files are on the same device.

I would like to **open an index while it is being uploaded** to avoid having the user wait for something that is already available, but I think that's an optimization we could do later. Enabling this behavior would complicate some specific cases, such as when index A is uploaded and opened simultaneously, and index B is opened, causing index A to be kicked out and uploaded again. I would prefer to keep the code simple for now and avoid this optimization.

### To Do
- [ ] Make this an experimental feature.
- [ ] Make this an enterprise-edition feature.
- [ ] Document how it works on the user-end.
	   We keep an open HTTP connection until the index is loaded and return the time it took to download the offloaded index.
- [ ] Make sure `showPerformanceDetails` accounts for the time it takes to open an index.
- [ ] Use the unavailable list of indexes to make sure not to upload/download indexes multiple times and concurrently.
- [ ] Actually offload to S3. No LMDB compaction, but maybe gzip compression?
	   I am currently faking offloading by moving the index to another directory and sleeping for a while (equivalent to a 10MiB/s network speed).
- [ ] Use an async runtime in a dedicated thread to download and upload indexes.
	    Currently, we are spawning large, complex async tasks to download/upload indexes. We must factorize the logic and keep a single system that pulls downloads/uploads.
- [ ] Think and document how to process snapshots.
	   To check, but right now the snapshot system will load each individual offloaded index, open it, and copy it into the tarball stream to S3. Is it something we want to do?
- [ ] What about sharding? Shouldn't we keep a subset of (full) indexes on each instance instead?

## Generative AI tools

### DIFF
--- a/crates/index-scheduler/src/index_mapper/index_map.rs
+++ b/crates/index-scheduler/src/index_mapper/index_map.rs
@@ -18,6 +18,9 @@ use uuid::Uuid;
 use super::IndexStatus::{self, Available, BeingDeleted, Closing, Missing};
 use crate::clamp_to_page_size;
 use crate::lru::{InsertionOutcome, LruMap};
+
+type IndexTransferCompletion = Shared<oneshot::Receiver<Result<(), Arc<io::Error>>>>;
+
 /// Keep an internally consistent view of the open indexes in memory.
 ///
 /// This view is made of an LRU cache that will evict the least frequently used indexes when new indexes are opened.
@@ -46,7 +49,7 @@ pub struct IndexMap {
     /// Note that opening it after the upload has completed will block until the **download** completes.
     ///
     /// One must wait for the download to complete before opening an index that is being downloaded.
-    transferring: HashMap<Uuid, Shared<oneshot::Receiver<Result<(), Arc<io::Error>>>>>,
+    transferring: HashMap<Uuid, IndexTransferCompletion>,
 
     /// A monotonically increasing generation number, used to differentiate between multiple successive index closing requests.
     ///
@@ -65,8 +68,8 @@ pub struct IndexMap {
 
 #[derive(Clone)]
 pub enum TransferState {
-    Downloading(Shared<oneshot::Receiver<Result<(), Arc<io::Error>>>>),
-    Uploading(Shared<oneshot::Receiver<Result<(), Arc<io::Error>>>>),
+    Downloading(IndexTransferCompletion),
+    Uploading(IndexTransferCompletion),
 }
 
 #[derive(Clone)]
@@ -208,6 +211,8 @@ impl IndexMap {
 
     /// Attempts to create a new index that wasn't existing before.
     ///
+    /// None is returned if the index is being transferred.
+    ///
     /// # Status table
     ///
     /// | Previous Status | New Status |
@@ -225,12 +230,35 @@ impl IndexMap {
         enable_mdb_writemap: bool,
         map_size: usize,
         create_or_open: CreateOrOpen,
-    ) -> Result<Index> {
+    ) -> Result<Option<Index>> {
         if !matches!(self.get_unavailable(uuid).await, Missing) {
+            // The index could be in the transferring state, right?
             panic!("Attempt to open an index that was unavailable");
         }
-        let index =
-            create_or_open_index(path, date, enable_mdb_writemap, map_size, create_or_open)?;
+
+        let index = match create_or_open {
+            CreateOrOpen::Open => {
+                // 1. check if the index exists
+                // 2. if it does, open it;
+                // 3. if not, register a download task and wait for it to complete
+                //    Warning: we must NOT wait for the download to complete here
+                //    or we will block the entire index management system.
+                //    We *MUST* rather return immediately an async task that the
+                //    caller can await on. We can return an Either<Index, Shared<OneShot>>.
+                // 4. Once the download is complete, open the index
+                // 5. insert it in the available map
+                let (sender, receiver) = oneshot::channel();
+                todo!("Do the download");
+                let receiver = receiver.shared();
+                todo!("do not insert another transferring job if one is already in progress");
+                self.transferring.insert(*uuid, receiver.clone());
+                return Ok(None);
+            }
+            create_or_open @ CreateOrOpen::Create { .. } => {
+                create_or_open_index(path, date, enable_mdb_writemap, map_size, create_or_open)?
+            }
+        };
+
         match self.available.insert(*uuid, index.clone()) {
             InsertionOutcome::InsertedNew => (),
             InsertionOutcome::Evicted(evicted_uuid, evicted_index) => {
@@ -240,7 +268,8 @@ impl IndexMap {
                 panic!("Attempt to open an index that was already opened")
             }
         }
-        Ok(index)
+
+        Ok(Some(index))
     }
 
     /// Increases the current generation. See documentation for this field.
--- a/crates/index-scheduler/src/index_mapper/index_map.rs
+++ b/crates/index-scheduler/src/index_mapper/index_map.rs
@@ -39,6 +39,7 @@ pub struct IndexMap {
     unavailable: LruMap<Uuid, ClosingIndex>,
     /// A set of indexes that have been deleted.
     deleting: HashSet<Uuid>,
+
     /// The set of indexes that are currently being uploaded and downloaded.
     ///
     /// One must wait for the upload to complete before opening an index that is being uploaded.
@@ -60,6 +61,12 @@ pub struct IndexMap {
     /// closing request was made, so the reader that "lost the race" has the old generation and will need to wait again for the index
     /// to close.
     generation: usize,
+}
+
+#[derive(Clone)]
+pub enum TransferState {
+    Downloading(Shared<oneshot::Receiver<Result<(), Arc<io::Error>>>>),
+    Uploading(Shared<oneshot::Receiver<Result<(), Arc<io::Error>>>>),
 }
 
 #[derive(Clone)]
@@ -113,8 +120,8 @@ impl ReopenableIndex {
     /// | Closing         | Available or Closing depending on generation |
     /// | Available       | Available                                    |
     ///
-    pub fn reopen(self, map: &mut IndexMap, path: &Path) -> Result<()> {
-        if let Closing(reopen) = map.get(&self.uuid) {
+    pub async fn reopen(self, map: &mut IndexMap, path: &Path) -> Result<()> {
+        if let Closing(reopen) = map.get(&self.uuid).await {
             if reopen.generation != self.generation {
                 return Ok(());
             }
@@ -126,7 +133,8 @@ impl ReopenableIndex {
                 self.enable_mdb_writemap,
                 self.map_size,
                 CreateOrOpen::Open,
-            )?;
+            )
+            .await?;
         }
         Ok(())
     }
@@ -144,8 +152,8 @@ impl ReopenableIndex {
     /// | BeingDeleted    | BeingDeleted                               |
     /// | Closing         | Missing or Closing depending on generation |
     /// | Available       | Available                                  |
-    pub fn close(self, map: &mut IndexMap) {
-        if let Closing(reopen) = map.get(&self.uuid) {
+    pub async fn close(self, map: &mut IndexMap) {
+        if let Closing(reopen) = map.get(&self.uuid).await {
             if reopen.generation != self.generation {
                 return;
             }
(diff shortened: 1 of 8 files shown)


crates/index-scheduler/src/index_mapper/index_map.rs
```diff
@@ -30,7 +39,17 @@ pub struct IndexMap {
     /// or because they are being closed.
     ///
     /// If they are being deleted, the UUID points to `None`.
-    unavailable: BTreeMap<Uuid, Option<ClosingIndex>>,
+    unavailable: LruMap<Uuid, ClosingIndex>,
```
Kerollmops, line 48: Could be preferable to LRU based on disk space rather than index count.

crates/index-scheduler/src/index_mapper/mod.rs
```diff
@@ -419,9 +436,24 @@ impl IndexMapper {
                     // take the lock to reopen the environment.
                     reopen
                         .reopen(&mut self.index_map.write().unwrap(), &index_path)
+                        .await
                         .map_err(|e| Error::from_milli(e, Some(uuid.to_string())))?;
                     continue;
```
Kerollmops, line 450: We must reduce the tries by one when we successfully reopen it.

crates/index-scheduler/src/index_mapper/index_map.rs
```diff
@@ -295,12 +378,54 @@ impl IndexMap {
...
+        );
+        match self.transferring.remove(uuid) {
+            Some(task) => {
+                // safety: we must be the last ones to get this error
+                task.await.unwrap().map_err(|err| Arc::into_inner(err).unwrap())
+            }
+            None => {
+                panic!("Attempt to finish the download of an index that is not being transferred")
+            }
+        }
```
Kerollmops: This method must check that the transferring is None, and if so just skip. Also remove the asserts.

crates/index-scheduler/src/index_mapper/mod.rs
```diff
@@ -519,21 +557,26 @@ impl IndexMapper {
     ///
     /// Since `f` is allowed to return a result, and `Index` is cloneable, it is still possible to wrongly build e.g. a vector of
     /// all the indexes, but this function makes it harder and so less likely to do accidentally.
-    pub fn try_for_each_index<U, V>(
+    pub async fn try_for_each_index<U, V>(
```
Kerollmops, line 568: It was deterministic because we were using a BTreeMap, we switched to an HashMap.

crates/index-scheduler/src/index_mapper/mod.rs
```diff
@@ -519,21 +557,26 @@ impl IndexMapper {
...
-                    .and_then(|(name, _)| self.index(rtxn, name).and_then(|index| f(name, &index)))
-            })
-            .collect()
+        let mut tmp = Vec::new();
+        for result in self.index_mapping.iter(rtxn)? {
+            match result {
+                Ok((name, _)) => {
+                    let index = self.index(rtxn, name).await?;
+                    // TBD not a big fan as `f` will block
+                    tmp.push(f(name, &index)?);
```
Kerollmops, line 582: It would be preferable to have a lambda that generates an async function (spawn_blocking).

crates/index-scheduler/src/index_mapper/mod.rs
```diff
@@ -519,21 +557,26 @@ impl IndexMapper {
...
+                }
+                Err(err) => return Err(err.into()),
+            }
+        }
```
Kerollmops, line 586: Use [`StreamExt::collect`](https://docs.rs/futures/latest/futures/prelude/stream/trait.StreamExt.html#method.collect)

PR #6296, opened 2026-03-31
ci: add riscv64 to release binary matrix via cross-rs

gounthar:
## What this PR does

Adds `linux-riscv64` to the `publish-binaries` matrix in `publish-release-assets.yml`.

The build uses [cross-rs](https://github.com/cross-rs/cross) for cross-compilation from a standard `ubuntu-22.04` runner; no external runner app installation required on the org.

## Changes

- `linux-riscv64` added to the release matrix (`riscv64gc-unknown-linux-gnu` target)
- `cross` is installed via `cargo install cross --locked` for the riscv64 job only
- Build step conditionally uses `cross` vs `cargo` based on `matrix.use_cross`
- All existing targets are unchanged

## Why riscv64?

RISC-V is a growing platform for embedded and server workloads. Several major Go/Rust projects (ripgrep, fd, bat, zellij, zoxide, just) have recently added riscv64 release binaries. Having an official meilisearch binary makes it usable on devices like the BananaPi F3 (SpacemiT K1) and upcoming riscv64 server hardware.

## Validation

The `cross` tool supports `riscv64gc-unknown-linux-gnu` out of the box via its pre-built Docker image. No changes to the test suite or merge queue checks — this only touches the release publishing workflow.

Replaces #6282 (which required installing the RISE runner app).
Closing #6286

PR #6359, opened 2026-05-01
fix: force bash shell for Build step to restore Windows CI

gounthar:
## Summary

My earlier PR #6296 changed the Build step from a hardcoded `cargo build` to a variable-driven `$BUILD_TOOL build` to support both `cargo` and `cross`. This works on Linux/macOS (bash) but broke the Windows publish jobs because PowerShell cannot invoke a command from a bare variable reference (`$BUILD_TOOL build` is a parser error in PowerShell).

The fix is one line: `shell: bash` on the Build step, which forces Git Bash on Windows runners and keeps the variable syntax working across all platforms.

## Changes

- Add `shell: bash` to the Build step in `publish-release-assets.yml`

## Verification

- Windows publish jobs were failing with `ParserError: Unexpected token 'build'`
- Adding `shell: bash` explicitly selects Git Bash on Windows, making `$BUILD_TOOL build` valid
- Linux and macOS jobs are unaffected (they already use bash)

Fixes the regression introduced in #6296. Sorry for the breakage.

### DIFF
--- a/.github/workflows/publish-release-assets.yml
+++ b/.github/workflows/publish-release-assets.yml
@@ -90,6 +90,7 @@ jobs:
         if: matrix.use_cross
         run: cargo install cross --locked
       - name: Build
+        shell: bash
         env:
           BUILD_TOOL: ${{ matrix.use_cross && 'cross' || 'cargo' }}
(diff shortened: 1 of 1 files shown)


Kerollmops (approved):
Thank you for the fix 👏

PR #6370, opened 2026-05-04
Forbid changing `self` when a `leader` is defined

dureuill:
## Breaking change

This PR contains the following breaking change:

- When the `network` experimental feature is enabled, sending a `PATCH /network` to an existing remote of a network now responds with a HTTP 400 `network_invalid_self` error if a `leader` has been defined.

## Rationale

- Renaming a remote is not supported
- However, it might happen if a user sends a network payload meant for the leader to a different remote.
- This PR protects against this specific error by making it forbidden to change `self` once a leader has been defined.
- This PR does not fix the general issue of remote aliasing, which is tracked [here](https://linear.app/meilisearch/issue/SCA-297/solve-aliasing-problem) (internal discussion)

# Implementation

- Detect and reject the condition where a remote is renamed when a leader exists (already, or as a result of the new payload)
- Add integration test cases for this case and other error cases related to misconfiguration when a leader is present

## Generative AI tools

### DIFF
--- a/crates/meilisearch/src/error.rs
+++ b/crates/meilisearch/src/error.rs
@@ -142,6 +142,8 @@ pub enum MeilisearchHttpError {
     InvalidHeaderValue { header_name: &'static str, msg: String },
     #[error("This remote is not the leader of the network.\n  - Note: only the leader `{leader}` can receive new tasks.")]
     NotLeader { leader: String },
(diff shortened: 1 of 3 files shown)


Kerollmops (approved):
Looks good to me, thanks!

PR #6371, opened 2026-05-06
feat: add query to federated search metadata

StephaneRob:
## Description

This pull request add query to federated search metadata and doesn't impact search and multi search due to query presence in response.

## Generative AI tools

### DIFF
--- a/crates/meilisearch/src/search/mod.rs
+++ b/crates/meilisearch/src/search/mod.rs
@@ -1378,6 +1378,9 @@ impl FacetValue {
 #[serde(rename_all = "camelCase")]
 #[schema(rename_all = "camelCase")]
 pub struct SearchMetadata {
+    /// Query string set for the query. (federated search only).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub query: Option<String>,
     /// Unique identifier for the query.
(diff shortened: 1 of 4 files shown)


ManyTheFish (dismissed):
LGTM

crates/meilisearch/src/search/mod.rs
```diff
@@ -1378,6 +1378,9 @@ impl FacetValue {
 #[serde(rename_all = "camelCase")]
 #[schema(rename_all = "camelCase")]
 pub struct SearchMetadata {
+    /// Query string set for the query. (federated search only).
```
dureuill:
```suggestion
    /// Query string set for the query, if not already present in the response.
    ///
    /// Only set if:
    ///
    /// 1. the metadata corresponds to a query in a federated search request, **and**
    /// 2. that federated search request query is not a placeholder search (empty or missing `q`)
    /// 
    /// Never set for regular search and non-federated multi-search because 
    /// the query(ies) are already present in the response.
```

dureuill (changes requested):
Minor clarification in the comment

dureuill (approved):
Thank you 🌈

PR #6375, opened 2026-05-11
Remote facet search

dureuill:
# Changelog

## Remote facet search

Meilisearch now has the ability to search across all shards of a network during facet search via the existing dedicated facet search route.

- If you are using the `network` experimental feature and have a `leader` defined for your network, **remote calls are now the default** for calls to `POST /indexes/{indexUid}/facet-search`
- The behavior can be controlled explicitly via the new `useNetwork` parameter of the [facet search object](https://www.meilisearch.com/docs/reference/api/facet-search/search-in-facets#body-facet-name).

# Implementation

The tests require to pass https://github.com/meilisearch/meilisearch/pull/6380

## Generative AI tools

### DIFF
--- a/crates/meilisearch/src/routes/indexes/facet_search.rs
+++ b/crates/meilisearch/src/routes/indexes/facet_search.rs
@@ -105,6 +105,35 @@ pub struct FacetSearchQuery {
     #[schema(required = false)]
     #[deserr(default, error = DeserrJsonError<InvalidFacetSearchExhaustiveFacetCount>, default)]
     pub exhaustive_facet_count: Option<bool>,
+    /// When `true`, runs the query on the whole network (all shards covered exactly once).
+    ///
+    /// When `false`, the query runs locally.
+    ///
+    /// When omitted or `null`, the default value depends on whether the sharding is enabled for the instance:
+    ///
+    /// - If the instance has sharding enabled (has a leader), defaults to `true`.
+    /// - Otherwise defaults to `false`.
+    ///
+    /// **Enterprise Edition only.** This feature is available in the Enterprise Edition.
+    ///
+    /// It also requires the `network` [experimental feature](http://localhost:3000/reference/api/experimental-features/configure-experimental-features).
+    ///
+    /// Values: `true` = use the whole network; `false` = local, default = see above.
+    ///
+    /// When using the network, the index must exist with compatible settings on all remotes.
+    #[schema(required = false)]
+    #[deserr(default, error = DeserrJsonError<InvalidSearchUseNetwork>)]
+    pub use_network: Option<bool>,
+}
+
+impl NetworkableQuery for FacetSearchQuery {
+    fn use_network_field(&mut self) -> &mut Option<bool> {
+        &mut self.use_network
+    }
+
+    fn has_remote(&self) -> bool {
+        false
+    }
 }
 
 #[derive(Default)]
@@ -137,6 +166,7 @@ impl FacetSearchAggregator {
             ranking_score_threshold,
             locales,
             exhaustive_facet_count,
+            use_network,
         } = query;
 
         Self {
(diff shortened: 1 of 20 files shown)


crates/meilisearch/src/routes/indexes/facet_search.rs
```diff
@@ -275,24 +311,275 @@ pub async fn search(
...
+            // true is left, false is right
+            |(remote, _)| Some(remote) == network.local.as_ref(),
+        );
+
+    let timeout = std::env::var("MEILI_EXPERIMENTAL_REMOTE_SEARCH_TIMEOUT_SECONDS")
+        .ok()
+        .map(|p| p.parse().unwrap())
+        .unwrap_or(25);
+
+    let deadline = before_search + std::time::Duration::from_secs(timeout);
```
ManyTheFish: I don't understand why we instantiate a deadline here in addition to the one that will be instantiated in `json_proxy` (L411). Shouldn't we use this deadline in `json_proxy` (L411) or rely on `json_proxy` to handle deadlines?

crates/meilisearch/src/routes/indexes/facet_search.rs
```diff
@@ -275,24 +311,275 @@ pub async fn search(
...
+                                    &index_uid,
+                                )
+                            })
+                            .transpose()?
+                    }
+                    None => None,
+                })
+            })
+            .collect();
+        let filters = filters?;
```
ManyTheFish: Using `filter_into_index_filter` will open indexes with the index scheduler depending on the foreign filters' needs.
We may open the same index multiple times if some queries have foreign filters referencing the same index.

[filters_into_index_filters](https://github.com/meilisearch/meilisearch/blob/44fffeca6806b4f3b9b667fa26719b8e0cae01b3/crates/index-scheduler/src/filter.rs#L103-L106) may be more suitable.

ManyTheFish (changes requested):
Some comments.

I have one concern regarding foreign filters.

In the PR implementation, the order is:
1) partitioning
2) send remote queries
3) resolve local queries
  3.1) parse and resolve filters (including foreign filters)
  3.2) resolve queries index by index
4) merge remote and local


This means that you are delaying the foreign filters resolution to each remote for their queries, and so, if we want to support remote foreign filters, each remote would have to query their friends to resolve the foreign filters, which is not what we want.

This is not a big deal for this PR, but I may need to move the foreign filter resolution earlier in the process in future PRs.

dureuill:
> This is not a big deal for this PR, but I may need to move the foreign filter resolution earlier in the process in future PRs.

Wondering if it would make sense to have at partitioning time?

crates/meilisearch/src/routes/indexes/facet_search.rs
dureuill: here there is a single "original filter", that we repeatedly clone and to which we simply append `_shard = some_shard_name`.

So IMO it would make sense to resolve the foreign filters even before the partitioning, on the original request, and then clone this resolved filter

crates/meilisearch/src/routes/indexes/facet_search.rs
dureuill: yes it looks redundant. Looking at the git history I think it might have been an issue when merging changes back to main

dureuill: It makes more sense to use a single global deadline set a the start than to reinitialize it for each request

dureuill: See https://github.com/meilisearch/meilisearch/pull/6375/changes/4f7ba148e48dfb1f4c7dbedf1d5e6ab67750ab62

ManyTheFish:
> Wondering if it would make sense to have at partitioning time?

What do you mean?

Something like, during the partitioning, we start by processing the foreign filter?

dureuill:
yes, but I went with a different approach for facet search: I'll be computing a single filter with a conjunction to a disjunction of all the shard filters, and turn that single filter into an index filter, once.

dureuill:
PR update incoming, I'm trying to get my tests to pass :-p

ManyTheFish (approved):
LGTM

PR #6382, opened 2026-05-14
fix: duplicated words in milli/indexer/write.rs and meilisearch option.rs

quyentonndbs:
Two one-line typo fixes for duplicated words in source comments:
- `crates/milli/src/update/new/indexer/write.rs` — "// Used by by the HannoySetVector to copy the embedding into an" → "// Used by the HannoySetVector to copy the embedding into an"
- `crates/meilisearch/src/option.rs` — "/// The size of the the multipart parts." → "/// The size of the multipart parts."

No code/behavior change.


Kerollmops (approved):
Thanks!

PR #6383, opened 2026-05-15
fix: return task documents as valid ndjson

### LINKED ISSUE #6379: Get tasks documents, confusing `content-type`
**Describe the bug**

The `/tasks/${uid}/documents` route returns an unexpected format, akin to NDJSON. 

The API returns `content-type: application/x-ndjson`, but the HTTP body actually contains `{"id":1,...}{"id":2,...}{"id":3,...}`. It’s neither standard JSON, nor truly “newline-delimited” as NDJSON should be. It’s readable as a stream, but it’s [not streamed](https://github.com/meilisearch/meilisearch/blob/475ed56e5612df0dbb826748add5f93e0e7d5500/crates/meilisearch/src/routes/tasks.rs#L711-L712) by Meilisearch.

The response is formatted as [concatenated JSON](https://en.wikipedia.org/wiki/JSON_streaming#Concatenated_JSON).

**Expected behavior**
Aligned response format with content type

**Meilisearch version:**
v1.42.0

atharvasingh7007:
## Summary
- write converted JSON and CSV document payloads as newline-delimited JSON records
- keep `/tasks/{taskUid}/documents` aligned with its `application/x-ndjson` response type
- add regression coverage for the document formatter and the task documents endpoint

## Root cause
JSON and CSV uploads were converted into the task update file by writing each JSON object back-to-back without a separator. The task documents endpoint serves that file as `application/x-ndjson`, so JSON-array uploads could return bodies like `{...}{...}`, which are not valid NDJSON.

## Testing
- `cargo test -p meilisearch-types document_formats`
- `cargo test -p meilisearch --test integration get_task_documents_returns_valid_ndjson_for_json_payload`
- `cargo test -p index-scheduler scheduler::test_document_addition::document_addition`
- `rustfmt --edition 2021 --check crates\meilisearch-types\src\document_formats.rs crates\meilisearch\tests\tasks\mod.rs`

Closes #6379

### DIFF
--- a/crates/meilisearch-types/src/document_formats.rs
+++ b/crates/meilisearch-types/src/document_formats.rs
@@ -10,7 +10,7 @@ use milli::documents::Error;
 use milli::Object;
 use rustc_hash::FxBuildHasher;
 use serde::de::{SeqAccess, Visitor};
-use serde::{Deserialize, Deserializer};
+use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::error::Category;
 use serde_json::value::RawValue;
 use serde_json::{to_writer, Map, Value};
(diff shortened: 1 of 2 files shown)


ManyTheFish:
Hello @atharvasingh7007,

For now, we are fine with the [concatenated JSON](https://en.wikipedia.org/wiki/JSON_streaming#Concatenated_JSON) format. This route is only meant to be used for debug purposes.

Thank you for your contribution.

PR #6387, opened 2026-05-18
docker: fix signal handling and cleanup unused variable

RinZ27:
## Related issue

None. This is a proactive improvement to ensure data integrity and clean up the container build process.

## Generative AI tools

## Requirements

- [x] Automated tests have been added. (Verified via existing CI suite)
- [ ] ⚠️ If there is any change in the DB: (No DB schema changes)

This PR addresses a subtle but important issue with how Meilisearch handles termination signals in Docker. By using the shell form for `CMD`, the process was being wrapped by `/bin/sh`, which often doesn't forward `SIGTERM` to the underlying application. Switching to the exec form ensures `tini` can correctly manage the lifecycle of the binary, preventing potential LMDB corruption during unexpected container stops.

Additionally, an unused `apkArch` variable in the compiler stage has been removed to keep the Dockerfile clean.

These changes should be completely transparent to existing users while improving overall reliability.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,6 @@ ENV     RUSTFLAGS="-C target-feature=-crt-static"
 
 COPY    . .
 RUN     set -eux; \
-        apkArch="$(apk --print-arch)"; \
         cargo build --release -p meilisearch -p meilitool ${EXTRA_ARGS}
 
 # Run
@@ -43,4 +42,4 @@ WORKDIR /meili_data
 EXPOSE  7700/tcp
 
 ENTRYPOINT ["tini", "--"]
-CMD     /bin/meilisearch
+CMD     ["/bin/meilisearch"]


Kerollmops:
Hello @RinZ27 👋 

> This PR addresses a subtle but important issue with how Meilisearch handles termination signals in Docker. By using the shell form for `CMD`, the process was being wrapped by `/bin/sh`, which often doesn't forward `SIGTERM` to the underlying application. Switching to the exec form ensures `tini` can correctly manage the lifecycle of the binary, preventing potential LMDB corruption during unexpected container stops.

Do you have any official documentation about this behavior? 🤔 Because, it seems an important issue indeed.

> Additionally, an unused `apkArch` variable in the compiler stage has been removed to keep the Dockerfile clean. The `rand` crate was also bumped to `0.8.6` in `meilisearch-auth` to address the unsoundness issue reported in RUSTSEC-2026-0097, ensuring the key generation remains robust.

Thank but this has already been fixed on main. It's probably because I'm reviewing your PR a bit late 😅

Have a nice day 🌵

RinZ27:
Here is the [official Docker documentation](https://docs.docker.com/reference/dockerfile/#cmd) regarding the `CMD` shell form behavior, @Kerollmops. It explicitly states that using the shell form wraps the executable in `/bin/sh -c`, which doesn't pass Unix signals downstream, meaning the application will completely miss the `SIGTERM` from a docker stop.

Just force-pushed a rebase against `main` to clear the conflicts. Since the `rand` crate vulnerability was already addressed upstream, I went ahead and dropped that change from the branch, leaving only the Dockerfile fix.

PR #6389, opened 2026-05-18
Autobatch deletions by filter with additions

Kerollmops:
This PR improves the situation when deletions by filter interleave document additions. It is now possible to autobatch deletions by filter with additions if they appear before the document addition.

Since we resolve filters before deleting the resulting documents, it is correct to delete them before any additions, since the inverted indexes for the filtering contain the deleted documents. It is still invalid to autobatch delete by filter after document additions, because the filter indexes are not yet populated with document information and thus cannot be found for deletion.

## Generative AI tools

### DIFF
--- a/crates/index-scheduler/src/scheduler/autobatcher.rs
+++ b/crates/index-scheduler/src/scheduler/autobatcher.rs
@@ -376,11 +376,14 @@ impl BatchKind {
                 deletion_ids.push(id);
                 Continue(BatchKind::DocumentClear { ids: deletion_ids })
             }
-            // we can't autobatch the deletion and import if the document deletion contained a filter
+            // we can autobatch the deletion by filter and import but only if the document addition is *after* the deletion
             (
-                this @ BatchKind::DocumentDeletion { deletion_ids: _, includes_by_filter: true },
-                K::DocumentImport { .. }
-            ) => Break((this, BatchStopReason::DeletionByFilterWithDocumentOperation { id })),
+                BatchKind::DocumentDeletion { mut deletion_ids, includes_by_filter: true },
+                K::DocumentImport { allow_index_creation, primary_key }
+            ) => {
+                deletion_ids.push(id);
+                Continue(BatchKind::DocumentOperation { allow_index_creation, primary_key, operation_ids: deletion_ids })
+            },
             // we can autobatch the deletion and import if the index already exists
             (
                 BatchKind::DocumentDeletion { mut deletion_ids, includes_by_filter: false },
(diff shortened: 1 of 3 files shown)


crates/index-scheduler/src/scheduler/autobatcher.rs
```diff
@@ -376,11 +376,14 @@ impl BatchKind {
...
+            // we can autobatch the deletion by filter and import but only if the document addition is *after* the deletion
             (
-                this @ BatchKind::DocumentDeletion { deletion_ids: _, includes_by_filter: true },
-                K::DocumentImport { .. }
-            ) => Break((this, BatchStopReason::DeletionByFilterWithDocumentOperation { id })),
+                BatchKind::DocumentDeletion { mut deletion_ids, includes_by_filter: true },
+                K::DocumentImport { allow_index_creation, primary_key }
+            ) => {
+                deletion_ids.push(id);
+                Continue(BatchKind::DocumentOperation { allow_index_creation, primary_key, operation_ids: deletion_ids })
```
dureuill, line 385: can we remove `BatchStopReason::DeletionByFilterWithDocumentOperation` without breaking the deserialization of tasks?

Kerollmops: I'm not sure how to follow. We are still using this variant when a deletion by filter is after a document addition, if your question was: "Can we remove the variant?" If the question is: "Will it break people that are expecting deletion by filter to never batch with additions," I suppose we don't care; that's internal behavior.

Kerollmops: Ok, I think I understand your point. The question is: "Can we deserialize batches with this reason in the list of batches?" and the answer is: "I have not removed the variant from the enum, so we can still deserialize it".

dureuill: the question is more "can we remove this variant without breaking anything, as it is no longer used?" 😄 

sorry for being unclear

Kerollmops: Actually, it is still used so the question can be asked later 😅

dureuill: It is not, though?

I checked, I think we're safe removing it, BatchStopReason are never deserialized from DB, they are stored as strings: https://github.com/meilisearch/meilisearch/blob/37cb8fb412c064e2eeeee5cdbb59e07bc3b9b958/crates/meilisearch-types/src/batches.rs#L34

this was the intentional design

Kerollmops: > It is not, though?

Crap! You're right, there are two of them: `DeletionByFilterWithDocumentOperation` and `DocumentOperationWithDeletionByFilter`. I thought they were the same, but they actually expose the order in which tasks are tried to batch.

Let me remove the variant then, and thanks for the review 👌

dureuill:
adding the spooky label, because this changes the behavior of the autobatcher.

However the change is restricted to delete by filter, and I understand it _should not_ cause issues

dureuill (approved):
let's go

PR #6396, opened 2026-05-19
Fix an internal error when changing the binary quantization

Kerollmops:
This PR fixes an internal error preventing users to change the binary quantization parameters of embedders and introduces a test to catch this internal error.

## Generative AI tools


dureuill (approved):
nice we found it 👌

PR #6399, opened 2026-05-20
Remove the vector store backend from the settings

Kerollmops:
This PR completely removes the possibility of changing the vector store backend.

## Generative AI tools


ManyTheFish:
I am not sure we want to introduce an error message forbidding the usage of this setting.

I would rather remove the setting completely, wdyt?

workloads/tests/hf-embed-hannoy.json
```diff
@@ -14,38 +14,14 @@
     }
   },
   "commands": [
-    {
-      "route": "experimental-features",
```
dureuill, line 18: I don't think we should be removing these commands, why are we?

workloads/tests/hf-embed.json
```diff
@@ -155,18 +155,6 @@
         ]
       }
     },
-    {
```
dureuill, line 158: I don't think we should be removing this

workloads/tests/hf-embed-hannoy.json
Kerollmops: Hum... that's my bad on that. I thought we were using the current version of Meilisearch, but we are not. I restored those lines. And the following others.

workloads/tests/hf-embed.json
Kerollmops: Same, my bad. I'll rollback that.

Kerollmops:
I really don't get why this change is making the CI break:

```diff
- "VectorStoreBackend":
- {
-     "type": "string",
-     "enum":
-     [
-         "stable",
-         "experimental"
-     ]
- },
```

EDIT: doesn't seem related to this PR actually.

dureuill (approved):
LGTM

ManyTheFish (approved):
lgtm

PR #6405, opened 2026-05-26
Better limit read bytes when creating the S3 multipart part

### LINKED ISSUE #6400: Error when uploading snapshots to S3
**Describe the bug**
When uploading a snapshot to S3, we received the following error. The error is most likely [in the enterprise edition's S3 Processing of snapshots](https://github.com/meilisearch/meilisearch/blob/e9c00b48e73f6fce1d81c4d7b49538ee7a9964e4/crates/index-scheduler/src/scheduler/enterprise_edition/s3.rs).

```
S3 error: status: 400 Bad Request, body: <Error><Code>EntityTooSmall</Code><Message>Your proposed upload is smaller than the minimum allowed size</Message><ProposedSize>5016510</ProposedSize><MinSizeAllowed>5242880</MinSizeAllowed><PartNumber>1</PartNumber><ETag>19671454289b673399a593e05d8c7181</ETag><RequestId>MR1CC63CNMYPEVRC</RequestId><HostId>REDACTED</HostId></Error>
```

**To Reproduce**
1. Setup Meilisearch to send snapshots to AWS S3 with temporary tokens
1. Send a snapshot request to Meilisearch
2. Wait for it to be processed
3. See the error in the tasks

**Expected behavior**
The snapshot should be correctly processed without error.

**Meilisearch version:**
???

**Additional context**
On the cloud, the database was small.

Kerollmops:
This PR fixes an issue we had with the multipart part size by ensuring we never construct a part larger than the defined multipart part size, rather than doing an ugly division by two. The division by two caused a bug when the multipart part size was set slightly below 10MiB, resulting in a multipart part of 4.99MiB, which is slightly below the AWS S3 multipart part size limit of 5MiB. With this fix, we always create a multipart with the provided multipart part size, except for the last part. Thanks, @vidit-virmani, for the help investigating the issue.

Fixes #6400 and Closes #6404.

## Generative AI tools

### DIFF
--- a/crates/index-scheduler/src/scheduler/enterprise_edition/s3.rs
+++ b/crates/index-scheduler/src/scheduler/enterprise_edition/s3.rs
@@ -395,7 +395,7 @@ async fn multipart_stream_to_s3(
     use std::os::fd::OwnedFd;
     use std::path::PathBuf;
 
-    use bytes::{Bytes, BytesMut};
+    use bytes::{BufMut as _, Bytes, BytesMut};
     use http_client::reqwest::{Client, Response};
     use rusty_s3::actions::CreateMultipartUpload;
     use rusty_s3::{Bucket, BucketError, Credentials, S3Action as _, UrlStyle};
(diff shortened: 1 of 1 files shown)


ManyTheFish (approved):
LGTM ! :D

PR #6407, opened 2026-05-26
chore(config): add commented-out chat completions experimental feature entry

### LINKED ISSUE #5669: update config.toml file to reflect chatCompletion experimental feature

liaoyl830:
This adds `experimental_chat_completions` as a commented-out entry in config.toml, matching the format of other experimental features.

This makes users aware of the option so they can enable chat completion via config file at startup, rather than having to dispatch an API request.

Closes #5669.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -132,3 +132,7 @@ experimental_reduce_indexing_memory_usage = false
 
 # Experimentally reduces the maximum number of tasks that will be processed at once, see: <https://github.com/orgs/meilisearch/discussions/713>
 # experimental_max_number_of_batched_tasks = 100
+
+# Experimental chat completions feature. For more information, see: <https://www.meilisearch.com/docs/learn/experimental/chat_completions>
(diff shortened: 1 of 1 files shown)


Kerollmops (changes requested):
Nice catch @liaoyl830 and thanks for the work 👌 

However, it seems that the link points to a 404. Would you mind linking to the write page, [like this one](https://www.meilisearch.com/docs/capabilities/conversational_search/getting_started/chat) or [even better](https://github.com/orgs/meilisearch/discussions/835), please?

Have you checked that Meilisearch can read the config file with this option enabled?

ManyTheFish:
Closing this PR for no activity.

Don't hesitate to reopen the PR when you are ready,

See you 👋

PR #6409, opened 2026-05-26
Support settings that require two tokenizers in the new settings indexer

Kerollmops:
This PR makes the new settings indexer support all the settings that require two indexers to reindex correctly. It means that the new indexer correctly supports changing the localized attribute rules, stop words, non-separator tokens, separator tokens, and the dictionary.

## Generative AI tools


ManyTheFish (dismissed):
GOOD GOOD!

dureuill (approved):

PR #6411, opened 2026-05-28
Fix internal vector stores quantization config desync

Kerollmops:
This PR fixes an issue we noticed when using the dumpless upgrade on a database from v1.33.1 to v1.45.0, but it is most likely affecting earlier versions as well. I initially thought that the bug was related to the configurations of the multiple internal vector stores we use to support multiple embedding by document, but it seems it's a bit more complex than that, and sometimes internal stores are not configured the same as another internal store (tried to open it as binary quantized, but it shouldn't, and vice versa).

We fix this bug by ensuring the internal vector stores are uniformly configured, clearing those that are not, and potentially fixing the internal millis' knowledge of the vector store quantization.

## Generative AI tools

### DIFF
--- a/crates/milli/src/update/upgrade/v1_37.rs
+++ b/crates/milli/src/update/upgrade/v1_37.rs
@@ -73,7 +73,7 @@ impl UpgradeIndex for ConvertArroyToHannoy {
                     config.config.quantized = match vector_store.clean_stores(wtxn)? {
                         Some(QuantizationStatus::Quantized) => Some(true),
                         Some(QuantizationStatus::NonQuantized) => Some(false),
-                        None => None,
+                        None => config.config.quantized,
                     };
                 }
             }
--- a/crates/milli/src/update/upgrade/v1_37.rs
+++ b/crates/milli/src/update/upgrade/v1_37.rs
@@ -1,11 +1,11 @@
+use arroy::Error::UnmatchingDistance as ArroyUnmatchingDistance;
+use hannoy::Error::UnmatchingDistance as HannoyUnmatchingDistance;
 use heed::RwTxn;
 
 use super::UpgradeIndex;
 use crate::update::upgrade::UpgradeParams;
-use crate::vector::{VectorStore, VectorStoreBackend};
+use crate::vector::{QuantizationStatus, VectorStore, VectorStoreBackend};
 use crate::{Error, Index, InternalError, Result};
-use arroy::Error::UnmatchingDistance as ArroyUnmatchingDistance;
-use hannoy::Error::UnmatchingDistance as HannoyUnmatchingDistance;
 
 /// Convert old Annoy vector stores to Hannoy ones
 pub(super) struct ConvertArroyToHannoy();
@@ -51,18 +51,31 @@ impl UpgradeIndex for ConvertArroyToHannoy {
                 Ok(_) => vector_store,
             };
 
-            // We make sure to only do the backend conversion when using arroy
-            // TODO what happen for unmatching distances when using hannoy
-            if backend == VectorStoreBackend::Arroy {
-                // Continue the hannoy conversion with the right quantization. Note that when changing
-                // the backend the miss-configured quantization stores are simply deleted.
-                vector_store.change_backend(
-                    &rtxn,
-                    wtxn,
-                    progress.clone(),
-                    &|| must_stop_processing.get(),
-                    None,
-                )?;
+            match backend {
+                VectorStoreBackend::Arroy => {
+                    // We make sure to only do the backend conversion when using arroy.
+                    //
+                    // Continue the hannoy conversion with the right quantization. Note that
+                    // when changing the backend the misconfigured quantization stores are
+                    // simply deleted.
+                    vector_store.change_backend(
+                        &rtxn,
+                        wtxn,
+                        progress.clone(),
+                        &|| must_stop_processing.get(),
+                        None,
+                    )?;
+                }
+                VectorStoreBackend::Hannoy => {
+                    // If the store is hannoy we clean the stores in case some were misconfigured.
+                    // Note that we never experienced an issue with hannoy stores but we are not sure
+                    // they are affected too.
+                    config.config.quantized = match vector_store.clean_stores(wtxn)? {
+                        Some(QuantizationStatus::Quantized) => Some(true),
+                        Some(QuantizationStatus::NonQuantized) => Some(false),
+                        None => None,
+                    };
+                }
             }
         }
 
--- a/crates/milli/src/update/upgrade/v1_37.rs
+++ b/crates/milli/src/update/upgrade/v1_37.rs
@@ -70,16 +70,9 @@ impl UpgradeIndex for ConvertArroyToHannoy {
                     // If the store is hannoy we clean the stores in case some were misconfigured.
                     // Note that we never experienced an issue with hannoy stores but we are not sure
                     // they are affected too.
+                    let detected = vector_store.clean_stores(wtxn)?;
                     config.config.quantized =
-                        match (config.config.quantized, vector_store.clean_stores(wtxn)?) {
-                            (None, None) => None,
-                            (None, Some(QuantizationStatus::NonQuantized)) => None, // None defaults to false
-                            (None, Some(QuantizationStatus::Quantized)) => Some(true),
-                            (Some(value), None) => Some(value),
-                            (Some(true), Some(QuantizationStatus::NonQuantized)) => Some(false),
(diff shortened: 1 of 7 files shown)


crates/milli/src/update/upgrade/v1_37.rs
```diff
@@ -17,30 +19,54 @@ impl UpgradeIndex for ConvertArroyToHannoy {
...
+                // Continue the hannoy conversion with the right quantization. Note that when changing
+                // the backend the miss-configured quantization stores are simply deleted.
+                vector_store.change_backend(
+                    &rtxn,
+                    wtxn,
+                    progress.clone(),
+                    &|| must_stop_processing.get(),
+                    None,
+                )?;
+            }
```
Kerollmops: I must make sure this upgrade task runs on all databases and especially new ones > 1.37 so that it fixes the broken DBs. It will have no impact on databases that are already using hannoy.

Kerollmops (commented):
We must introduce another dumpless upgrade migration for indexes above or equal to 1.37.0 and below v1.46.0 to delete/clear internal stores with unmatched distances. Maybe call the must_upgrade of the ChangeBackend struct and do it only if it says that it must not upgrade.

Kerollmops (commented):
Base this on v1.45.1.

crates/milli/src/update/upgrade/v1_37.rs
Kerollmops: Done, in the last commit where we take care of hannoy databases.

crates/milli/src/update/upgrade/v1_37.rs
```diff
@@ -17,30 +19,54 @@ impl UpgradeIndex for ConvertArroyToHannoy {
...
+                        )
+                    }
+                    otherwise => return Err(otherwise.into()),
+                },
+                Err(e) => return Err(e),
+                Ok(_) => vector_store,
+            };
+
+            // We make sure to only do the backend conversion when using arroy
+            // TODO what happen for unmatching distances when using hannoy
```
dureuill: remove this TODO?

Should we `else { store.clean_stores }`?

crates/milli/src/update/upgrade/v1_37.rs
```diff
@@ -17,30 +19,54 @@ impl UpgradeIndex for ConvertArroyToHannoy {
...
+            if backend == VectorStoreBackend::Arroy {
+                // Continue the hannoy conversion with the right quantization. Note that when changing
+                // the backend the miss-configured quantization stores are simply deleted.
```
dureuill:
```suggestion
                // the backend the misconfigured quantization stores are simply deleted.
```

crates/milli/src/update/upgrade/v1_45.rs
```diff
@@ -0,0 +1,53 @@
...
+                embedder_info.embedder_id,
+                config.config.quantized(),
+            );
+
+            // Read the dimensions to be able to know the real quantization
+            // parameter, it corresponds to the quantization of the first store.
+            config.config.quantized = match vector_store.clean_stores(wtxn)? {
+                Some(QuantizationStatus::Quantized) => Some(true),
+                Some(QuantizationStatus::NonQuantized) => Some(false),
+                None => None,
```
dureuill: should we leave the value unchanged in this case? For instance, the config could be quantized on an empty store


```suggestion
                None => config.config.quantized,
```

crates/milli/src/vector/store.rs
```diff
@@ -149,6 +149,9 @@ impl VectorStore {
         }
     }
 
+    /// Converts the vector store from arroy to hannoy and the other way around.
+    ///
+    /// Note that when changing the backend the miss-configured quantization stores are simply deleted.
```
dureuill:
```suggestion
    /// Note that when changing the backend the misconfigured quantization stores are simply deleted.
```

crates/milli/src/vector/store.rs
```diff
@@ -1133,6 +1137,13 @@ impl VectorStore {
                 match arroy::Reader::open(arroy_rtxn, index, self.database.remap_types()) {
                     Ok(reader) => reader,
                     Err(arroy::Error::MissingMetadata(_)) => continue,
+                    Err(arroy::Error::UnmatchingDistance { .. }) => {
+                        // TODO it's ugly
+                        // TODO explain why AD while it must not
```
dureuill: remove the TODOs?

crates/milli/src/vector/store.rs
```diff
@@ -1205,6 +1216,159 @@ impl VectorStore {
         }
         Ok(())
     }
+
+    /// Returns the `true` if the stores are quantized, `false` otherwise.
```
dureuill:
```suggestion
    /// Returns the quantization status as known to the store
```

crates/milli/src/vector/store.rs
```diff
@@ -149,6 +149,9 @@ impl VectorStore {
         }
     }
 
+    /// Converts the vector store from arroy to hannoy and the other way around.
+    ///
```
dureuill, line 153: to the best of my understanding, we are only cleaning the stores when switching from arroy to hannoy bq. Shouldn't we do that in all cases?

dureuill: or at least also when switching from arroy to hannoy nonbq

crates/milli/src/update/upgrade/v1_45.rs
Kerollmops: Uninitialized and therefore empty stores, return the `MissingMetadata` error, we are deleting them. Is a store is empty but initialized ones have metadata and were therefore checked to be aligned with the distance/quantization of the first store and cleared if not. I am not sure to get what "the config could be quantized on an empty store" means. Can you clarify, please?

crates/milli/src/vector/store.rs
Kerollmops: I jus implemented the clean when doing the non-bq conversion from arroy to hannoy in d6d7103. I never saw this kind of misconfigured quantization on non-bq stores, though.

crates/milli/src/update/upgrade/v1_37.rs
Kerollmops: Yup! You are right, thanks for the review, doing it.

crates/milli/src/update/upgrade/v1_45.rs
dureuill: Sure.

I create an embedder with binaryQuantized set  to `true` on an empty index. All stores are empty, so `clean_stores` returns `None`. We reset the config.quantized to `None`, which sets it back to `binaryQuantized: false`. I feel this is wrong.

crates/milli/src/vector/store.rs
dureuill: maybe we should specify in the comment in that case?

crates/milli/src/update/upgrade/v1_45.rs
Kerollmops: Agreed, just push a commit for this. Thanks for the help.

crates/milli/src/vector/store.rs
Kerollmops: Done, thanks!

crates/milli/src/update/upgrade/v1_45.rs
```diff
@@ -0,0 +1,56 @@
...
+            // Read the dimensions to be able to know the real quantization
+            // parameter, it corresponds to the quantization of the first store.
+            config.config.quantized =
+                match (config.config.quantized, vector_store.clean_stores(wtxn)?) {
+                    (None, None) => None,
+                    (None, Some(QuantizationStatus::NonQuantized)) => None, // None defaults to false
+                    (None, Some(QuantizationStatus::Quantized)) => Some(true),
+                    (Some(false), None) => Some(false), // Keep it explicit
+                    (Some(true), None) => Some(true),   // Keep it explicit
+                    (Some(expected), Some(_)) => Some(expected),
```
dureuill: I would have expected `(Some(true), Some(QuantizationStatus::NonQuantized)) => Some(false)` and `(Some(false), Some(QuantizationStatus::Quantized)) => Some(true)` instead of the catch-all.

Kerollmops: Ho yeah! I'm dumb...

dureuill (approved):
Looks good and tested in situ! Can I mark it as spooky, though, given it attempts to fix a DB and possibly destroys user vectors?

dureuill (approved):

PR #6414, opened 2026-05-28
Search refactor

ManyTheFish:
## Related issue

Fixes https://linear.app/meilisearch/issue/ENGPROD-2514/factorize-search-avoiding-duplicated-implementation

## Generative AI tools

## Changelog

Refactor the search pipeline to mutualize the code.
The new implementation will always do a federated search internally, and then the output will be transformed into the expected route's output.

Noticeable changes from the user perspective:
- Some error messages have been modified
- Few error code changes `MultiSearch<Error>` <--> `Search<Error>`

### Enhancement

Personalization will now work with Federated search using the `federation` search attribute.

Like `page`/`hitPerPage` or `limit`/`offset`, the `personalization` must be put in the `federation` attribute to work properly.
Otherwise, an error will be returned asking to move the attribute in `federation`.

### Experimental feature to switch between the new and old implementation

This PR introduces an experimental feature allowing switching back to the legacy implementation if some regressions are found.

In the experimental feature routes, a new feature has been added:
PATCH `/experimental-features`

```json
{
  "legacySearch": true
}
```

available values:
- `true`: force old implementation
- `false`: force new implementation	
- `null`: use the instance default value

#### Set the instance default value for legacy search

In addition to the `/experimental-features` route, a config has been added to choose the default value when the value is set to `default` on the experimental feature route.

environment variable: MEILI_EXPERIMENTAL_LEGACY_SEARCH_DEFAULT
config: experimental-legacy-search-default


crates/meilisearch-types/src/features.rs
```diff
@@ -24,6 +24,7 @@ pub struct RuntimeTogglableFeatures {
     pub multimodal: bool,
     pub foreign_keys: bool,
     pub queue_documents_fetch: bool,
+    pub legacy_search: bool,
```
Kerollmops: Can we enable the legacy search automatically at boot time?

crates/meilisearch/src/routes/indexes/search.rs
```diff
@@ -537,7 +539,90 @@ pub fn fix_sort_query_parameters(sort_query: &str) -> Vec<String> {
...
+        let permit = search_queue.try_get_search_permit().await?;
+        progress.update_progress(TotalProcessingTimeStep::Search);
+        let index_uid = IndexUid::try_from(index_uid.into_inner())?;
+
+        let query: SearchQuery = params.into_inner().try_into()?;
+
+        let mut aggregate = SearchAggregator::<SearchGET>::from_query(&query);
+
+        let include_metadata = parse_include_metadata_header(&req);
+        let is_proxy = false;
```
Kerollmops, line 563: Are you sure it's always false for a GET search/ ?

crates/meilisearch/src/routes/indexes/search.rs
```diff
@@ -745,6 +840,90 @@ pub async fn search_with_post(
...
+        let permit = search_queue.try_get_search_permit().await?;
+        progress.update_progress(TotalProcessingTimeStep::Search);
+
+        let query = params.into_inner();
+        debug!(request_uid = ?request_uid, parameters = ?query, "Search post");
+
+        let mut aggregate = SearchAggregator::<SearchPOST>::from_query(&query);
+
+        let include_metadata = parse_include_metadata_header(&req);
+        let is_proxy = false;
```
Kerollmops, line 853: Shouldn't you take the proxy value from the header?

crates/meilisearch/src/routes/multi_search.rs
```diff
@@ -163,20 +245,26 @@ pub async fn multi_search_with_post(
...
-    let features = index_scheduler.features();
+    // check remote header
+    let is_proxy = req
+        .headers()
+        .get(PROXY_SEARCH_HEADER)
+        .is_some_and(|value| value.as_bytes() == PROXY_SEARCH_HEADER_VALUE.as_bytes());
+
+    let include_metadata = parse_include_metadata_header(&req);
 
+    let auth_filter = index_scheduler.filters();
```
Kerollmops: Can you simplify the diff by moving back the lines from where they come from.

crates/meilisearch/tests/search/metadata.rs
```diff
@@ -111,6 +112,7 @@ async fn search_with_metadata_header_and_primary_key() {
       "estimatedTotalHits": 1,
       "requestUid": "[uuid]",
       "metadata": {
+        "query": "glass",
```
Kerollmops, line 115: What do you think about always exposing the query in the metadata, @dureuill?

Kerollmops (changes requested):

crates/meilisearch/src/routes/indexes/search.rs
ManyTheFish: Yes it is expected

crates/meilisearch/src/routes/indexes/search.rs
ManyTheFish: Yes it is expected

Kerollmops (approved):
Thanks for the huge work 💪

ManyTheFish:
This PR is considered spooky because it refactors the /multi-search and /indexes/:uid/search, which are important routes for us.

If any bug is encountered, please report it to us, and try using the `legacy_search` experimental feature to use the old implementation.

dureuill:
Question: I see this is added as an experimental feature. 

Can the Cloud easily enable the feature by default for users on upgrade?

dureuill:
nvm, I didn't read far enough, disregard previous comment

PR #6417, opened 2026-05-29
OpenAPI CI: disable rule to avoid crash (workaround)

curquiza:
## Related issue

Fixes OpenAPI CI (https://linear.app/meilisearch/issue/ENGPROD-2513/fix-the-openapi-broken-ci)
This is the least worst workaround I found by testing and investigating with Claude: it disables a rule to avoid nimma (dependency of spectral) to crash... this only crashes in the CI btw. I was not able to reproduce it locally

What we lose with this rule: nothing really impacting because this rule triggers warnings, so this would not have stopped our CI anyway since we do not fail for warnings

## Generative AI tools

- [X] This PR uses generative AI tooling and respect the [related policies](https://github.com/meilisearch/meilisearch/blob/main/CONTRIBUTING.md#use-of-generative-ai-tools)
    - for investigating


Kerollmops (approved):
LGTM, thanks @curquiza 👏

PR #6421, opened 2026-06-02
Fix a panic with incomplete filters

Kerollmops:
This PR fixes an internal panic when a filter is icomplete by returning an error.

## Generative AI tools


dureuill (approved):
replaced the finish with its inner match + matched on the incomplete

PR #6430, opened 2026-06-03
Expose metrics to compute the indexed documents throughput

Kerollmops:
This PR exposes more Prometheus metrics so that we can expose the documents throughput of Meilisearch. We decided to hide the metrics after 5 minutes after the last document addition, update, deletion, or edition (with function) so that no points are drawn on the graph instead of misleading information. This may allow customers or the support team to detect when a machine is stuck with an index setting or similar. They can also differentiate slow indexing from no indexing at all.

### Example Metrics Output

```
# HELP meilisearch_last_indexed_documents_count The last number of indexed documents in a batch
# TYPE meilisearch_last_indexed_documents_count gauge
meilisearch_last_indexed_documents_count{batch_uid="0",index="movies"} 31944
# HELP meilisearch_last_indexed_documents_duration_ms The duration in ms of the last batch that indexed documents
# TYPE meilisearch_last_indexed_documents_duration_ms gauge
meilisearch_last_indexed_documents_duration_ms{batch_uid="0",index="movies"} 17904
```

## Generative AI tools

### DIFF
--- a/crates/meilisearch/src/routes/metrics.rs
+++ b/crates/meilisearch/src/routes/metrics.rs
@@ -6,9 +6,10 @@ use meilisearch_types::deserr::query_params::Param;
 use meilisearch_types::error::ResponseError;
 use meilisearch_types::keys::actions;
 use meilisearch_types::milli::progress::ProgressStepView;
-use meilisearch_types::tasks::Status;
+use meilisearch_types::task_view::DetailsView;
+use meilisearch_types::tasks::{Kind, Status};
 use prometheus::{Encoder, TextEncoder};
-use time::OffsetDateTime;
+use time::{Duration, OffsetDateTime};
 
 use crate::extractors::authentication::policies::ActionPolicy;
 use crate::extractors::authentication::{AuthenticationError, GuardedData};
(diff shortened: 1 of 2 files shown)


crates/meilisearch/src/routes/metrics.rs
```diff
@@ -220,6 +227,51 @@ pub async fn get_metrics(
...
+        auth_filters,
+    )?;
+    // ...and get the last batch only.
+    if let Some(batch) = batches.into_iter().next() {
+        // If this last batch is too old to be kept, just keep the metrics reset (unspecified)
+        if let Some(finished_at) = batch.finished_at {
+            let batch_duration = finished_at - batch.started_at;
+            let now = OffsetDateTime::now_utc();
+            // We add a small offset of 5min so that Prometheus can catch the value.
+            if now >= finished_at && now - finished_at <= batch_duration + Duration::minutes(5) {
```
ManyTheFish, line 254: Instead of adding 5 minutes to any duration, I'd prefer putting a minimal threshold and taking the max between duration and threshold.

ManyTheFish (approved):
I put a small change request, but it's not blocking me, so we can merge if you don't think my request is relevant.

PR #6442, opened 2026-06-09
Replace hf-hub git dependency by the official one

Kerollmops:
This PR ensures we use the official hf-hub dependency instead of a fork pulled from GitHub, now that the maintainers have correctly fixed the issue we had with ureq.

## Generative AI tools

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1315,8 +1315,19 @@ dependencies = [
  "encode_unicode",
  "libc",
  "once_cell",
- "unicode-width",
  "windows-sys 0.59.0",
+]
+
+[[package]]
+name = "console"
+version = "0.16.3"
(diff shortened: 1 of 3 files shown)


dureuill:
can you check how the agent is built in the hf-hub, to check if they call [`http_status_as_error`](https://docs.rs/ureq/latest/ureq/config/struct.ConfigBuilder.html#method.http_status_as_error)?

Kerollmops:
Sorry for not mentioning it, but I already checked: they are not calling it, and the default value for [`http_status_as_error`](https://docs.rs/ureq/latest/ureq/config/struct.ConfigBuilder.html#method.http_status_as_error) is true, which is what we expect.

dureuill (approved):
let's go to the future 🚀

Issue #5392, opened 2025-03-05
Create test for the dumpless upgrade


Kerollmops:
We must create at least one integration test that pulls the last version of Meilisearch (based on the version in the TOML), indexes a set of documents, changes some settings, and, if possible, keeps some enqueued tasks. Once the database state is set, we must be able to read this database with the `--experimental-dumpless-upgrade` flag.

msantoshdora:
Hey @Kerollmops , 
I would like to work on this ticket to add the asked integration test.

dureuill:
Fixed by https://github.com/meilisearch/meilisearch/pull/5861

PR #6448, opened 2026-06-11
Add dumpless-upgrade orchestration test on a user-populated database

AymanHamdoun:
## Related issue

Addresses #5392.

This is a partial step toward the issue — see [Scope](#scope) below for what is and isn't covered.

## What this PR adds

One new integration test in `crates/meilisearch/tests/upgrade/mod.rs`:

```
dumpless_upgrade_from_v1_12_with_local_db
```

It exercises the dumpless-upgrade orchestration end-to-end without depending on a checked-in `.ms` fixture:

1. Boots a server with the current binary, creates an index, updates settings (`searchableAttributes`, `filterableAttributes`, `sortableAttributes`) and adds two documents — waits for every task to succeed.
2. Copies the populated database into a fresh tempdir and overwrites the `VERSION` file with `1.12.0`. The copy step sidesteps re-opening the same LMDB env twice in the same process.
3. Boots a second server on the new tempdir with `experimental_dumpless_upgrade: true` and asserts that:
   - the `VERSION` file is rewritten to the current binary version,
   - an `upgradeDatabase` task is auto-enqueued with `details.upgradeFrom = "v1.12.0"` and succeeds,
   - the user's index, settings (all three attribute lists) and documents survive,
   - a sort+filter search returns the expected hit (`age >= 5` → `kefir`),
   - the original two task uids (`0 = settingsUpdate`, `1 = documentAdditionOrUpdate`) remain queryable.

## Why `from = 1.12.0`

While picking a `from` version I noticed anything other than 1.12.0 silently no-ops. `update_version_file_for_dumpless_upgrade` (in `crates/meilisearch/src/lib.rs`) only syncs the index-scheduler's internal version DB on the `from_minor == 12` branch — that's the entry `upgrade_index_scheduler` reads to decide whether to run migrations. For any later `from`, the internal version is left at the current binary, so `upgrade_index_scheduler` early-returns at `initial_version == target_version` and no upgrade task is ever enqueued. Picking 1.12.0 is what actually fires the migration pipeline. There's also a doc comment on the test calling this out for future readers.

## Scope

This is an **orchestration** test on a current-format database — it verifies the `check_version → update_version_file_for_dumpless_upgrade → Versioning::new → upgrade_index_scheduler → UpgradeDatabase task` chain works end-to-end and preserves user data. It is **not** a real format-migration test: the underlying LMDB bytes were written by the current binary, so the per-version migration functions run but have no meaningful format work to do.

Real format migration is already covered by the existing `crates/meilisearch/tests/upgrade/v1_12/v1_12_0.rs` test, which loads a checked-in v1.12.0 database fixture. The new test complements that one — it adds coverage for the orchestration on a user-populated DB, which the static-fixture test can't (its DB is a fixed reference).

A follow-up PR could add a real fixture for a more recent starting version (e.g. v1.45.0) once someone with the previous-version binary available can generate and commit the `.ms` directory.

## Test plan

Local (Rust 1.91.1, macOS):

- `cargo test -p meilisearch --test integration upgrade` — all 6 upgrade tests pass (4 pre-existing in `mod.rs` + `import_v1_12_0` + the new test).
- `cargo clippy -p meilisearch --tests -- -D warnings` — clean.
- `cargo fmt -- --check crates/meilisearch/tests/upgrade/mod.rs` — clean. (Pre-existing fmt diff in `crates/meilisearch/tests/upgrade/v1_12/v1_12_0.rs:7` exists on `main` too — not touched.)

## Generative AI tools

    - Claude Code (Anthropic, Opus 4.7): used to research the dumpless-upgrade code path and write the test.

## Requirements

- [ ] If some tests cannot be automated, manual rigorous tests should be applied. — N/A
- [ ] ⚠️ If there is any change in the DB: — N/A (no DB-format change; this PR only adds a test)
- [ ] If necessary, the feature have been tested in the Cloud production environment — N/A
- [ ] If necessary, the [documentation](https://github.com/meilisearch/documentation) related to the implemented feature in the PR is ready. — N/A
- [ ] If necessary, the [integrations](https://github.com/meilisearch/integration-guides) related to the implemented feature in the PR are ready. — N/A

### DIFF
--- a/crates/meilisearch/tests/upgrade/mod.rs
+++ b/crates/meilisearch/tests/upgrade/mod.rs
@@ -6,7 +6,8 @@ use std::{fs, io};
 use meili_snap::snapshot;
 use meilisearch::Opt;
 
-use crate::common::{default_settings, Server};
+use crate::common::{Server, Value, default_settings};
+use crate::json;
 
 fn copy_dir_all(src: impl AsRef<Path>, dst: impl AsRef<Path>) -> io::Result<()> {
     fs::create_dir_all(&dst)?;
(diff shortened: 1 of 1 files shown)


dureuill:
Thank you for your contribution,

However, the linked issue was actually already handled satisfactorily by https://github.com/meilisearch/meilisearch/pull/5861.

PR #6455, opened 2026-06-16
Truncate rendered document templates to documentTemplateMaxBytes

### LINKED ISSUE #6150: The documentTemplateMaxBytes setting is completely ignored when passing documents to the chat
**Describe the bug**

I'm seeing an issue where the chat completions endpoint completely ignores the index setting `chat.documentTemplateMaxBytes`. No matter what I set this limit to, the documents are passed to the LLM in full without truncation.

**To Reproduce**

1. Create an index with some documents in it that have a good amount of content.
2. Set the index setting `chat.documentTemplateMaxBytes` to `100` (or whatever, as long as it's shorter than the indexed documents after they're put through `documentTemplate`).
3. Create a chat workspace that has access to this index. When calling chat completions, make sure to include the [`_meiliAppendConversationMessage` tool](https://www.meilisearch.com/docs/learn/chat/chat_tooling_reference#meiliappendconversationmessage).
4. Call chat completions and inspect the appended messages. The documents are returned in full, not truncated to the specified limit.

**Expected behavior**

The specified limit should be enforced, not ignored.

**Meilisearch version:**
1.29.0

**Additional context**

I'm not familiar with the codebase (or Rust), but I've asked Claude about this issue, and it claims to have found the source of the bug. I can't verify if that is correct, but including it in case it's useful:

> The `render_document` method in `crates/milli/src/prompt/mod.rs` (lines 108-132) only uses `max_bytes` to set the initial Vec capacity but never truncates the rendered output. 

> In contrast, `render_kvdeladd` (lines 134-150) in the same file correctly calls `truncate(&mut rendered, max_bytes.get())` when `max_bytes` is set.

> The fix is to add truncation logic before returning from `render_document`:

```rust
if let Some(max_bytes) = self.max_bytes {
    let max_bytes = max_bytes.get();
    if rendered.len() > max_bytes {
        for i in (0..=max_bytes).rev() {
            if std::str::from_utf8(&rendered[..i]).is_ok() {
                rendered.truncate(i);
                break;
            }
        }
    }
}
```

### LINKED ISSUE #6434: `documentTemplateMaxBytes` has no effect during document embedding extraction
**Describe the bug**
documentTemplateMaxBytes does not truncate rendered document templates during embedding extraction in the primary indexing path. The setting only affects the initial memory allocation size, not the final output length.

**To Reproduce**
1. Create an index with an embedder and set "documentTemplateMaxBytes": 1000
2. Index documents whose rendered template exceeds 1000 bytes
3. Observe that the rendered text sent to the embedder is not truncated at 1000 bytes

**Expected behavior**
Rendered document templates should be truncated to documentTemplateMaxBytes bytes.

**Screenshots**
N/A

**Meilisearch version:**
[v1.45.2](https://github.com/meilisearch/meilisearch/releases/tag/v1.45.2)

**Additional context**
N/A

genisis0x:
## Related issue

Fixes #6434.

Also fixes #6150: chat completions render documents through the same `Prompt::render_document` (`crates/meilisearch/src/routes/chats/utils.rs` builds `Prompt::new(template, max_bytes)` from `chat_config` and calls `render_document` per result), so `chat.documentTemplateMaxBytes` was equally unenforced and documents appended to the LLM via `_meiliAppendConversationMessage` were sent untruncated. Bounding the output in `render_document` fixes both the embedding-extraction path and the chat path.

## What does this PR do?

`documentTemplateMaxBytes` was only sizing the initial allocation, not bounding the output. In `Prompt::render_document` (`crates/milli/src/prompt/mod.rs`) the setting is passed to `with_capacity_in`:

```rust
let mut rendered = bumpalo::collections::Vec::with_capacity_in(
    self.max_bytes.unwrap_or_else(default_max_bytes).get(),
    doc_alloc,
);
self.template.render_to(&mut rendered, &context)?;
```

but `render_to` can write past that capacity, and the result was returned untruncated. So on the primary (new-indexer) embedding-extraction path, the rendered template sent to the embedder could exceed `documentTemplateMaxBytes`. The chat completions path shares this function, so `chat.documentTemplateMaxBytes` was unenforced there too. The sibling path `render_kvdeladd` already truncates:

```rust
if let Some(max_bytes) = self.max_bytes {
    truncate(&mut rendered, max_bytes.get());
}
```

This PR brings `render_document` to parity. Since that path renders into a bump-allocated `Vec<u8>` (not a `String`), it adds `truncate_bytes`, the byte-buffer counterpart of the existing `truncate`, which cuts on a UTF-8 char boundary so the output stays valid UTF-8.

## Requirements

  - `template_truncation_bytes` in `prompt::test` mirrors the existing `template_truncation` test on the same multibyte input, asserting char-boundary cuts and the no-op case.
- [x] `cargo test -p milli --lib prompt::`, `cargo clippy -p milli`, and `cargo fmt -- --check` pass locally.

### DIFF
--- a/crates/milli/src/prompt/mod.rs
+++ b/crates/milli/src/prompt/mod.rs
@@ -129,6 +129,12 @@ impl Prompt {
                 liquid_error,
             )
         })?;
+        // `with_capacity_in` above only sizes the initial allocation; `render_to`
+        // can still write past it, so enforce the byte budget here the same way
+        // `render_kvdeladd` does for the other indexing path.
+        if let Some(max_bytes) = self.max_bytes {
+            truncate_bytes(&mut rendered, max_bytes.get());
+        }
         Ok(std::str::from_utf8(rendered.into_bump_slice())
             .expect("render can only write UTF-8 because all inputs and processing preserve utf-8"))
     }
@@ -162,6 +168,21 @@ fn truncate(s: &mut String, max_bytes: usize) {
             break;
         }
     }
+}
+
+/// Truncate a UTF-8 byte buffer to at most `max_bytes`, cutting on a `char`
+/// boundary so the result stays valid UTF-8. This is the bump-allocated
+/// counterpart of [`truncate`] used by [`Prompt::render_document`].
+fn truncate_bytes(bytes: &mut bumpalo::collections::Vec<'_, u8>, max_bytes: usize) {
+    if max_bytes >= bytes.len() {
+        return;
+    }
+    // Walk back off any UTF-8 continuation byte (0b10xx_xxxx) to a char boundary.
+    let mut end = max_bytes;
+    while end > 0 && (bytes[end] & 0b1100_0000) == 0b1000_0000 {
+        end -= 1;
+    }
+    bytes.truncate(end);
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


genisis0x:
Hi @ManyTheFish — gentle first ping on this one, no rush. It fixes #6434 (and #6150): documentTemplateMaxBytes was never actually enforced, so rendered templates — including the documents appended to the LLM on chat completions — could exceed the configured cap. The real test gates are green; the red check-labels is just the triage-label gate on your side. Happy to rebase or adjust if anything's needed.

Kerollmops (changes requested):
Hello @genisis0x 👋

Would you mind fill in the PR template and especially the part around AI usage. Note that your code seems pretty dangerous toward masking UTF-8 characters. We have already implemented the fix in [this commit](https://github.com/meilisearch/meilisearch/pull/6368/changes/72015dec6835580ab2e4fafbeac96aaca082bb4a) if you want to open a new PR about it.

Have a nice day 🌵

PR #6457, opened 2026-06-16
move the performance trace computation after the personalization

ManyTheFish:
## Related issue

Fixes https://linear.app/meilisearch/issue/ENGPROD-2502/personalization-is-missing-in-search-performance-details

## Generative AI tools

Fix the search performance details of the route `/indexes/:uid/search` that was missing the personalization trace.


Kerollmops (approved):
LGTM 🌵

PR #6458, opened 2026-06-17
fix(auth): re-expand wildcard key actions at lookup for post-creation actions

### LINKED ISSUE #5691: Wildcard `Action`s lack new `Action`s that get added in updates
**Describe the bug**

Wildcard API key `Action`s lack new `Action`s for new features that get added in updates.

**To Reproduce (allegedly)**

1. Create a meilisearch database with an old version of meilisearch, before it supported chat completions.
2. Update meilisearch to the latest version.
3. Setup chat completions.
4. Attempt to use them with the wildcard admin key that was created before the update.
5. The wildcard key will be rejected.

**Expected behavior**

When a new key `Action` is introduced, it should be available to wildcard keys created before.

**Meilisearch version:**
[e.g. v0.20.0]

**Additional context**

It seems that the only place where actions get stored is when the key is created: https://github.com/meilisearch/meilisearch/blob/7c1a9113f9e38fb079426460c2f038443b7e62c6/crates/meilisearch-auth/src/store.rs#L77-L160

However, the action check logic doesn't handle wildcards on its own. Wildcards only work because they are expanded at creation.
https://github.com/meilisearch/meilisearch/blob/7c1a9113f9e38fb079426460c2f038443b7e62c6/crates/meilisearch-auth/src/store.rs#L221-L247

genisis0x:
## Summary

Fixes #5691.

Keys created with `Action::All` (or any group wildcard like `Action::ChatsAll`) have their actions expanded at creation time via `enum_iterator::all::<Action>()` and stored in an inverted DB. After a dumpless upgrade that introduces new leaf actions (e.g. `ChatsGet`, `ChatsDelete` added after a key was created), those keys never receive the new actions in the inverted DB and silently cannot use the new features.

### Root cause

`put_api_key` snapshots `enum_iterator::all::<Action>()` at write time. The inverted DB entry for a new action is never created retroactively, so `get_expiration_date` / `prefix_first_expiration_date` return `None` → `is_key_authorized` returns `false`.

### Fix

**Extract `expand_actions(&[Action]) -> HashSet<Action>`** from `put_api_key` (no behavioural change at creation time) and call it as a runtime fallback in both lookup functions:

1. `get_expiration_date` — when the inverted DB returns `None`, look up the stored `Key` and re-expand its `actions` slice against the *current* `Action` enum. If the requested action is now covered and the key's index patterns permit the requested index, return `key.expires_at`.
2. `prefix_first_expiration_date` — same fallback when the prefix scan returns nothing.

Index restrictions are fully honoured in the fallback path (`IndexUidPattern::matches_all` / `matches_str`). Expiration dates from the stored key are respected.

## Tests

Two new unit tests in `store::tests`:
- `test_expand_actions_all_covers_new_leaf_actions` — asserts `Action::All` expands to every non-wildcard action in the current enum (regression guard for future action additions).
- `test_expand_actions_all_get_covers_only_read_actions` — asserts `Action::AllGet` covers exactly the read-only subset.

All 6 tests pass; `cargo clippy` clean.

### DIFF
--- a/crates/meilisearch-auth/src/store.rs
+++ b/crates/meilisearch-auth/src/store.rs
@@ -85,85 +85,7 @@ impl HeedAuthStore {
         // create inverted database.
         let db = self.action_keyid_index_expiration;
 
-        let mut actions = HashSet::new();
-        for action in &key.actions {
-            match action {
-                Action::All => {
-                    actions.extend(enum_iterator::all::<Action>());
-                    actions.remove(&Action::AllGet);
-                }
-                Action::AllGet => {
-                    actions.extend(enum_iterator::all::<Action>().filter(|a| a.is_read()))
-                }
-                Action::DocumentsAll => {
-                    actions.extend(
-                        [Action::DocumentsGet, Action::DocumentsDelete, Action::DocumentsAdd]
-                            .iter(),
-                    );
-                }
-                Action::IndexesAll => {
-                    actions.extend(
-                        [
-                            Action::IndexesAdd,
-                            Action::IndexesDelete,
-                            Action::IndexesGet,
-                            Action::IndexesUpdate,
-                            Action::IndexesSwap,
-                            Action::IndexesCompact,
-                        ]
-                        .iter(),
-                    );
-                }
-                Action::SettingsAll => {
-                    actions.extend([Action::SettingsGet, Action::SettingsUpdate].iter());
-                }
-                Action::DumpsAll => {
-                    actions.insert(Action::DumpsCreate);
-                }
-                Action::SnapshotsAll => {
-                    actions.insert(Action::SnapshotsCreate);
-                }
-                Action::TasksAll => {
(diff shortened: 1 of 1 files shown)


genisis0x:
Hi @ManyTheFish, thanks again for merging #6454. Whenever you get a moment, would really appreciate a look at this one, no rush. It closes #5691: a key created with a wildcard (`*`) action wasn't picking up actions introduced after the key was created, so newer endpoints stayed inaccessible until the key was recreated. I just rebased it on latest main; the earlier red embedder tests were a flaky run unrelated to this auth-only change (verified green again locally and on the fresh CI). Happy to adjust anything.

ManyTheFish:
Hello @genisis0x,

We prefer modifying the auth part of the code ourselves.
Moreover, this part needs a real refactor, switching to bitmask logic instead.

Sorry for the delay,

Have a good day!

PR #6459, opened 2026-06-17
fix(chats): avoid internal error on chat completions against empty index

### LINKED ISSUE #5808: internal: missing primary-key in the main database - chat completions feature
**Describe the bug**
When doing a /chat/completions request with an apikey tied to an index with no data, I get the following error first:
```
{
  "event_id": "24fe74e6-eacc-45f2-9f5a-3d029cee5a21",
  "type": "error",
  "error": {
    "type": "response_error",
    "code": "500",
    "message": "internal: missing primary-key in the main database.",
    "param": null,
    "event_id": null
  }
}
```

Then a chat stream as expected.

**To Reproduce**
- Create an index with no data
- Create an apikey for this index with chat completions access
- Trigger a `${[url]}/chats/${[ chatWorkspace ]}/chat/completions` request
- See the error

**Expected behavior**
No internal error

**Screenshots**
Image

**Meilisearch version:**
v1.16.0

genisis0x:
## Summary

Fixes #5808.

When a chat completions request hits an index that has never had documents indexed, `format_documents` calls `index.external_id_of(...)`, which unconditionally reads the primary key from the main database before iterating over document IDs. On an empty index, no primary key has been set, so `ok_or(InternalError::DatabaseMissingEntry { ... })` fires and the client receives:

```json
{
  "type": "error",
  "error": {
    "code": "500",
    "message": "internal: missing primary-key in the main database."
  }
}
```

This error is emitted before the chat stream continues, which is both confusing and incorrect — an empty result set is perfectly valid.

## Fix

Early-return `Ok(Vec::new())` in `format_documents` when `internal_docids` is empty. No documents to render means no document IDs to look up, and no primary key is needed. This matches the intent of the `external_id_of` precondition comment ("never called on an empty index") while making the call site safe for the zero-results case.

The change is 5 lines in `crates/meilisearch/src/routes/chats/utils.rs`.

### DIFF
--- a/crates/meilisearch-auth/src/store.rs
+++ b/crates/meilisearch-auth/src/store.rs
@@ -85,85 +85,7 @@ impl HeedAuthStore {
         // create inverted database.
         let db = self.action_keyid_index_expiration;
 
-        let mut actions = HashSet::new();
-        for action in &key.actions {
-            match action {
-                Action::All => {
-                    actions.extend(enum_iterator::all::<Action>());
-                    actions.remove(&Action::AllGet);
-                }
-                Action::AllGet => {
-                    actions.extend(enum_iterator::all::<Action>().filter(|a| a.is_read()))
-                }
-                Action::DocumentsAll => {
-                    actions.extend(
-                        [Action::DocumentsGet, Action::DocumentsDelete, Action::DocumentsAdd]
-                            .iter(),
-                    );
-                }
-                Action::IndexesAll => {
-                    actions.extend(
-                        [
-                            Action::IndexesAdd,
-                            Action::IndexesDelete,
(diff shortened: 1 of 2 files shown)


genisis0x:
Hi @ManyTheFish, gentle ping on this one when you have bandwidth, no urgency at all. It closes #5808: chat completions against an empty index surfaced an internal error instead of a clean empty result. Small and self-contained, tests passing. Thanks for taking the time, glad to tweak if you'd prefer a different shape.

Kerollmops:
Hey @genisis0x 👋

It's too dangerous to merge this PR. We will handle that ourselves.

Thank you 🌵

PR #6444, opened 2026-06-09
Bump and removes unused dependencies

Kerollmops:
This PR upgrades most of our dependencies and removes unnecessary features and dependencies to reduce indexing time a bit. I marked the PR spooky because bumping incompatible dependencies could impact serialization or similar operations.

## Generative AI tools

- [x] This PR does not use generative AI tooling
- [ ] This PR uses generative AI tooling and respects the [related policies](https://github.com/meilisearch/meilisearch/blob/main/CONTRIBUTING.md#use-of-generative-ai-tools)

PR #6470, opened 2026-06-22
Fix the workflow after broken ranking scores

Kerollmops:
This PR is a follow-up to fix [the breaking change](https://github.com/meilisearch/meilisearch/pull/6444/changes/367613a8489a690ab949bf6abc7cdad4ee163b54) that caused the hf-embed workflow to break ranking scores.

## Generative AI tools

### DIFF
--- a/workloads/tests/hf-embed.json
+++ b/workloads/tests/hf-embed.json
@@ -20,17 +20,9 @@
       "method": "PATCH",
       "body": {
         "inline": {
-          "filterableAttributes": [
-            "genres",
-            "release_date"
-          ],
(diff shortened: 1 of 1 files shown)


dureuill (changes requested):
ah, sorry if it wasn't clear. I'd like for us to also keep the v1.47.0 checkpoint, so that we have a lasting proof that values don't change before v1.48.0.

I suggest keeping both the initial test of v1.47 and your change to v1.48

dureuill (approved):
thank you 👌

PR #6484, opened 2026-06-30
DSR as index, mk II

dureuill:
## Breaking changes

This PR introduces several breaking changes

### Request type changes

1. `priority` has been replaced with `precedence`, which better reflects the behavior (lower precedence means the rule is applied first)
2. `conditions` has been modified from an array to an object with two fields: "query" of type `QueryCondition` and "time" of time `TimeCondition`
3. New type `QueryCondition` that contains the fields `isEmpty` (as previously) and `words` instead of `contains` (same type)
4. It is now possible to pass `isEmpty: false` with `words` in a `QueryCondition`. Passing `isEmpty:true` with `words` still results in a synchronous error.
5. New type `TimeCondition` with fields `start` and `end` (unchanged semantics from previous type).
6. When specifying the `selector` of an `Action`, it is now mandatory to specify an `id`. Previously, it was optional, but the action would never trigger.
7. When listing rules with `POST /dynamic-search-rules`, `filter.attributePatterns` has been replaced with `filter.query`, an optional string that searches in rule `description` and `conditions.query.words`.
8. When calling `DELETE /dynamic-search-rules/{:ruleUid}` or `PATCH /dynamic-search-rules/{:ruleUid}` in a sharded configuration, endpoint will not return a HTTP 400 error if called on a follower remote rather than on the leader.

### Response changes

1. `PATCH /dynamic-search-rules/{:ruleUid}` and `DELETE /dynamic-search-rules/{:ruleUid}` now register an asynchronous task..
2. The response is modified to return the registered task instead of the modified dynamic search rule.
3. HTTP 404 is no longer returned if the `{:ruleUid}` portion of the URL refers to a rule that doesn't exist. This is because rules are processed asynchronously, and is consistent with the behavior of `DELETE /indexes/{:indexUid}/documents/{:docId}` for `{:docId}` 

## Behavior changes

- The `conditions.query.words` behaves differently from `query.contains`: previously, a rule would match if its conditions `query.contains` would be substrings of `q` in the search query in the sense of [`str::contains`](https://doc.rust-lang.org/std/primitive.str.html#method.contains). Now, a rule matches if all the words in `conditions.query.words` appear in `q` (after normalization). For `q = hero super`, `query.contains = super hero` would **not** match, whereas `conditions.query.words = super hero` does now match. This behavior is more in line with regular search, and allows improving performance.
- Dynamic search rules are now replicated from the leader to its follower, when in a sharded configuration

## Additions

- Add a new `DELETE /dynamic-search-rule` route that deletes all the DSRs
- Add the concept of "DSR fuel" that determines how much energy is spent resolving DSR during a search. The fuel is initialized with some default variables that can be overriden using environment variables:
    - `MEILI_EXPERIMENTAL_DSR_FUEL_MAX_COUNTED_WORDS`: max number of words considered inside of a search query for the purpose of finding `conditions.query.words` constraints. Defaults to 10, max value is 255
    - `MEILI_EXPERIMENTAL_DSR_FUEL_MAX_ACTIVE_RULES`: max number of active rules whose actions are evaluated. Defaults to 1000, max value is 4294967295
    - `MEILI_EXPERIMENTAL_DSR_FUEL_MAX_PIN_ACTIONS`: max number of pin actions that are applied. Defaults to 100, max value is 4294967295
    - `MEILI_EXPERIMENTAL_DSR_FUEL_WORD_FUEL`: max number of constraint combinations that are evaluated for the purpose of finding `conditions.query.words` constraints. Defaults to 4096, max value is 4294967295

## Implementation changes

- Introduce a special dsr index. It is created if missing when creating a DSR
    - It is snapshotted and saved/restored in dumps 
    - It is migrated along with other indexes during dumpless upgrade
    - It is replicated during network rebalancing, from the leader's version of the index
- Introduce several `IndexUid` types, to represent validated user indexes or the special dsr index.
- Make index mapper implementation generic over the IndexUid type
- Add DSR related tasks
- Search changes:
    -  Build located query term before calling execute_search, resolve pins there
    -  Filtered universe optionally accepts a set of base candidates
    - Use time OffsetDateTime instead of std::Instant to measure elpased time
    - Pass uid and OffsetDateTime to lower levels in search so they can be used to resolve DSRs

## Generative AI tools


crates/meilisearch/src/lib.rs
Kerollmops, line 1: Would you mind testing if the dumps work, please?

crates/index-scheduler/src/index_mapper/mod.rs
```diff
@@ -647,3 +647,67 @@ impl IndexMapper {
...
+    {
+        UserIndex::new(uid).ok_or_else(|| Error::InvalidIndexUid { index_uid: uid.to_string() })
+    }
+
+    fn uid(&self) -> &'a str {
+        UserIndex::uid(self)
+    }
+
+    fn count(db: Database<Str, UuidCodec>, rtxn: &RoTxn) -> Result<u64> {
+        Ok(AnyIndex::count(db, rtxn)?.saturating_sub(DsrIndex::count(db, rtxn)?))
```
Kerollmops: Maybe add a test to make sure introducing a new `.meili` index doesn't break the count.

crates/index-scheduler/src/index_mapper/mod.rs
```diff
@@ -483,16 +487,31 @@ impl IndexMapper {
...
     /// the lock.mdb and an optional data.mdb.cpy file.
-    pub fn index_path(&self, uuid: Uuid) -> PathBuf {
+    fn index_path_from_uuid(&self, uuid: Uuid) -> PathBuf {
         self.base_path.join(uuid.to_string())
     }
 
-    pub fn rollback_index(
+    pub fn index_path<'a>(&self, rtxn: &RoTxn, name: impl IndexUid<'a>) -> Result<Option<PathBuf>> {
+        let name = name.uid();
+        let index_path = match self.index_mapping.get(rtxn, name)? {
```
Kerollmops: Use the `index_uuid` method instead.

crates/index-scheduler/src/scheduler/process_batch.rs
```diff
@@ -227,48 +229,57 @@ impl IndexScheduler {
 
                 Ok((tasks, info))
             }
-            Batch::IndexCreation { index_uid, primary_key, task } => {
+            Batch::IndexCreation { index_uid: index_name, primary_key, task } => {
                 progress.update_progress(CreateIndexProgress::CreatingTheIndex);
+                let index_uid = AnyIndex::new(&index_name);
```
Kerollmops: It may be a UserIndex here.

crates/index-scheduler/src/scheduler/process_batch.rs
```diff
@@ -439,6 +436,8 @@ impl IndexScheduler {
                     unreachable!()
                 };
 
+                let index_uid = AnyIndex::new(index_uid);
```
Kerollmops: UserIndex

crates/meilisearch-types/src/dynamic_search_rules.rs
```diff
@@ -46,6 +46,94 @@ pub struct DynamicSearchRule {
...
+            actions: vec![],
+        }
+    }
+
+    pub fn try_from_meili_doc<'a>(
+        doc: impl Document<'a>,
+        fault_source: FaultSource,
+    ) -> Result<Self, milli::Error> {
+        let to_milli_error = |err| {
+            if let FaultSource::User = fault_source {
```
Kerollmops: I would prefer a match here, please.

crates/meilisearch-types/src/dynamic_search_rules.rs
```diff
@@ -46,6 +46,94 @@ pub struct DynamicSearchRule {
...
+                        milli::Error::UserError(milli::UserError::MissingDocumentId {
+                            primary_key: "uid".to_string(),
+                            document: Default::default(),
+                        })
+                    } else {
+                        milli::Error::InternalError(milli::InternalError::DatabaseMissingEntry {
+                            db_name: "dsr index",
+                            key: None,
+                        })
+                    }
```
Kerollmops: Same here, please

crates/index-scheduler/src/scheduler/process_index_operation.rs
```diff
@@ -571,4 +571,198 @@ impl IndexScheduler {
...
+            filterable_attributes: Setting::Set(vec![
+                // filter by active or inactive rules
+                FilterableAttributesRule::Field("active".into()),
+                // used to find time constraints
+                FilterableAttributesRule::Field("conditions.time.start".into()),
+                // used to find time constraints
+                FilterableAttributesRule::Field("conditions.time.end".into()),
+                // used to find query isEmpty constraints
+                FilterableAttributesRule::Field("conditions.query.isEmpty".into()),
+            ]),
```
Kerollmops, line 606: Would be better if we limit this further to only the equal condition.

crates/index-scheduler/src/scheduler/process_index_operation.rs
```diff
@@ -571,4 +571,198 @@ impl IndexScheduler {
...
+        let (document_changes, operation_stats, primary_key) = indexer
+            .into_changes(
+                &indexer_alloc,
+                index,
+                &rtxn,
+                Some("uid"),
+                &mut new_fields_ids_map,
+                &must_stop_processing,
+                progress.clone(),
+                None,
```
Kerollmops, line 720: Comment that no sharding is used, please

crates/milli/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,541 @@
...
+        &self,
+        query_terms: &[LocatedQueryTerm],
+        universe: &mut RoaringBitmap,
+        search_context: &SearchContext,
+        fuel: DsrFuel,
+    ) -> Result<Vec<PinDoc>> {
+        let active_rules = self.active_rules_for_query(query_terms, search_context, fuel)?;
+
+        self.find_pins(self.rule_ids_sorted_by_precedence(active_rules)?, search_context)
+            .take(fuel.max_active_rules())
```
Kerollmops: To put inside the find_pins method, please

crates/milli/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,541 @@
...
+    }
+
+    pub fn all_rule_ids(&self) -> Result<RoaringBitmap> {
+        Ok(self.index.documents_ids(self.rtxn)?)
+    }
+
+    pub fn search_in_description_and_words(
+        &self,
+        query: Option<String>,
+        universe: RoaringBitmap,
```
Kerollmops: &RoaringBitmap

crates/milli/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,541 @@
...
+
+        if let Some(query) = query {
+            search.query(query);
+        }
+
+        search.candidates(&universe);
+
+        search.exhaustive_number_hits(true);
+        search.max_total_hits(Some(
+            self.index.pagination_max_total_hits(self.rtxn)?.unwrap_or(1000) as usize,
```
Kerollmops: Are you sure we can see more than 1000 docs or will they be hidden?

crates/milli/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,541 @@
...
+                }
+            } else {
+                let is_query_empty_key =
+                    FacetGroupKey { field_id: is_query_empty_fid, level: 0, left_bound: "true" };
+                if let Some(FacetGroupValue { size: _, bitmap: is_query_empty_rules }) =
+                    self.index.facet_id_string_docids.get(self.rtxn, &is_query_empty_key)?
+                {
+                    active_rules -= is_query_empty_rules;
+                }
+            }
```
Kerollmops: Can we factorize this, please? Only the left_bound changes

crates/milli/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,541 @@
...
+        let words_count =
+            query_terms.len().min(MAX_COUNTED_WORDS).min(fuel.max_counted_words()) as u8;
+        if let Some(query_words_fid) = self.db_fields_ids_map.id("conditions.query.words") {
+            let word_count_db = &self.index.field_id_word_count_docids;
+
+            // 4. exclude words with more word constraints than present in the query
+            if let Some(words_count_plus_one) = words_count.checked_add(1) {
+                for res in word_count_db.range(
+                    self.rtxn,
+                    &((query_words_fid, words_count_plus_one)..=(query_words_fid, u8::MAX)),
```
Kerollmops, line 291: May be easier to read: `&((query_words_fid, words_count_plus_one)..`

crates/milli/src/dynamic_search_rules.rs
```diff
@@ -0,0 +1,541 @@
...
+                    continue;
+                };
+
+                word_rules &= &active_rules;
+
+                if word_rules.is_empty() {
+                    continue;
+                }
+
+                words_rules.push(word_rules);
```
Kerollmops, line 312: What happens if we have duplicate words? Rules could be applied multiple times.

Kerollmops (changes requested):

crates/meilisearch/src/lib.rs
```diff
@@ -243,6 +243,33 @@ pub fn setup_meilisearch(
         }
     };
 
+    let dsr_fuel = {
+        let max_counted_words = match std::env::var("MEILI_EXPERIMENTAL_DSR_FUEL_MAX_COUNTED_WORDS")
```
Kerollmops, line 247: Can you document a bit those env vars in the PR description, please?

crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -993,8 +998,8 @@ fn merge_metadata(
         performance_details: _,
     } in remote_results
     {
-        let this_remote_duration = Duration::from_millis(*processing_time_ms as u64);
-        max_remote_duration = Duration::max(this_remote_duration, max_remote_duration);
+        let this_remote_duration = time::Duration::nanoseconds(*processing_time_ms as i64);
```
Kerollmops: Should probably be millis here ?

crates/index-scheduler/src/upgrade/v1_49.rs
```diff
@@ -0,0 +1,210 @@
...
+                        let new_end = match (existing_end.as_ref().copied(), end) {
+                            (None, None) => None,
+                            (None, Some(end)) | (Some(end), None) => Some(end),
+                            (Some(existing), Some(new)) => Some(existing.min(new)),
+                        };
+
+                        if let (Some(start), Some(end)) = (new_start, new_end) {
+                            if start > end {
+                                have_impossible_conditions = true;
+                                break;
```
Kerollmops: Remove this boolean and early return instead

Kerollmops (changes requested):

crates/meilisearch/src/lib.rs
dureuill: the dumps work after the latest changes: f67aa8044615ff0303e6138387f8f103d25b2361

crates/milli/src/dynamic_search_rules.rs
dureuill: After adding a dedicated test, it seems to be working. We are probably capping the number of hits before calling the search in the regular endpoints

crates/milli/src/dynamic_search_rules.rs
dureuill: Fixed by 67afe82f00d5e89f9734580a9de1ed6586b5e79f and f66d2df5001d849e846cd22a0021665d4f858505 adds a test

crates/milli/src/dynamic_search_rules.rs
dureuill: done in c233a53e199f44a4d7280de95216476b62bce0d3

crates/milli/src/dynamic_search_rules.rs
dureuill: done in c233a53e199f44a4d7280de95216476b62bce0d3

crates/milli/src/dynamic_search_rules.rs
dureuill: done in c233a53e199f44a4d7280de95216476b62bce0d3

crates/milli/src/dynamic_search_rules.rs
dureuill: done in c233a53e199f44a4d7280de95216476b62bce0d3

Kerollmops: My bad on that one, you don't want to read other FIDs 😬

dureuill: actually, reverted because we don't want to iterate on higher fids... bfa7d645d79e3490264a1b4aa773d20d1d314e6a

crates/index-scheduler/src/upgrade/v1_49.rs
dureuill: done in 41d8fba59dcdea054a05b47e87417c2227c9fcac

crates/index-scheduler/src/index_mapper/mod.rs
dureuill: done in  12a194b31c5044049f5cd521dd44e38877aaaa13

crates/index-scheduler/src/scheduler/process_index_operation.rs
dureuill: done in d3ac496f70d048cbbe6c3539cf6398198a7f36a1

crates/index-scheduler/src/scheduler/process_index_operation.rs
dureuill: done here d3ac496f70d048cbbe6c3539cf6398198a7f36a1

crates/index-scheduler/src/scheduler/process_batch.rs
dureuill: done in d3ac496f70d048cbbe6c3539cf6398198a7f36a1

crates/index-scheduler/src/scheduler/process_batch.rs
dureuill: done in d3ac496f70d048cbbe6c3539cf6398198a7f36a1

crates/meilisearch/src/search/federated/perform.rs
dureuill: done in 106fc58369c66e64b4f3d935c3b28f7e2b48ebc8

crates/meilisearch-types/src/dynamic_search_rules.rs
dureuill: done in 106fc58369c66e64b4f3d935c3b28f7e2b48ebc8

crates/meilisearch-types/src/dynamic_search_rules.rs
dureuill: done in 106fc58369c66e64b4f3d935c3b28f7e2b48ebc8

crates/index-scheduler/src/index_mapper/mod.rs
dureuill: done in ed32270a0ed0a00dfcf222b7e75259ce298ef439

also changed the way we do the count to iter on .meili indexes to be more future-proof

crates/meilisearch/src/lib.rs
dureuill: done, please see the description

Kerollmops (approved):
LGTM, thank you!

# 🥧

PR #6486, opened 2026-07-01
Stabilize dumpless upgrade

curquiza:
## Related issue

Fixes https://linear.app/meilisearch/issue/ENGPROD-2628/stabilize-dumpless-upgrade

## Generative AI tools

- [X] This PR uses generative AI tooling and respect the [related policies](https://github.com/meilisearch/meilisearch/blob/main/CONTRIBUTING.md#use-of-generative-ai-tools)
    - to point out places to changes

## Requirements

- [ ] If necessary, the [documentation](https://github.com/meilisearch/documentation) related to the implemented feature in the PR is ready -> I will update the documentation


dureuill:
do we want to fully remove the switch?

I was thinking we would still like to keep an `--upgrade` or `--upgrade-db` flag, so that we cannot accidentally upgrade a DB by e.g. starting the wrong binary.

We'd remove the `experimental`, and IMO the `dumpless` part is also not bearing its load

what do you think @Kerollmops

Kerollmops:
> do we want to fully remove the switch?
> 
> I was thinking we would still like to keep an `--upgrade` or `--upgrade-db` flag, so that we cannot accidentally upgrade a DB by e.g. starting the wrong binary.
> 
> We'd remove the `experimental`, and IMO the `dumpless` part is also not bearing its load
> 
> what do you think @Kerollmops

Hey @curquiza and @dureuill 👋 

Indeed, I don't think we want to remove the switch entirely and would prefer to keep it explicit that we want to upgrade the database to avoid locking someone into a newer version (rollback is not supported). I would prefer we keep the `--dumpless-upgrade` naming, but we can discuss this.

curquiza:
Hey,
I agree with you that keeping the flag is important to let the user upgrade consciously
 
My favorite naming is `--upgrade-db` because
- `--upgrade` looks like we are going to perform the full upgrade, including the change of version, which is not the case. User has to download the new binary and then, use the flag
- `--dumpless-upgrade` -> the user is not aware they should do a dump to upgrade, except if they read the docs carefully, which is not the case for most people. For me "dumpless upgrade" is an internal wording

What do you think?

crates/index-scheduler/src/upgrade/mod.rs
```diff
@@ -62,7 +62,7 @@ pub fn upgrade_index_scheduler(
 
     if initial_version < (1, 12, 0) {
         bail!(
-                "Database version {initial_major}.{initial_minor}.{initial_patch} is too old for the experimental dumpless upgrade feature. Please generate a dump using the v{initial_major}.{initial_minor}.{initial_patch} and import it in the v{target_major}.{target_minor}.{target_patch}",
+                "Database version {initial_major}.{initial_minor}.{initial_patch} is too old to be upgraded automatically. Please generate a dump using the v{initial_major}.{initial_minor}.{initial_patch} and import it in the v{target_major}.{target_minor}.{target_patch}",
```
dureuill:
```suggestion
                "Database version {initial_major}.{initial_minor}.{initial_patch} is too old to be upgraded with `--db-upgrade`. Please generate a dump using the v{initial_major}.{initial_minor}.{initial_patch} and import it in the v{target_major}.{target_minor}.{target_patch}",
```

crates/index-scheduler/src/lib.rs
```diff
@@ -158,8 +158,6 @@ pub struct IndexSchedulerOptions {
     pub export_default_payload_size_bytes: Byte,
     /// The experimental features enabled for this instance.
     pub instance_features: InstanceTogglableFeatures,
-    /// Whether the index scheduler is able to auto upgrade or not.
-    pub auto_upgrade: bool,
```
dureuill, line 164: was indeed unused, apparently 👍

crates/meilisearch-types/src/versioning.rs
```diff
@@ -81,13 +81,14 @@ pub enum VersionFileError {
     MalformedVersionFile { context: String },
     #[error(
         "Your database version ({major}.{minor}.{patch}) is incompatible with your current engine version ({}).\n\
-        To migrate data between Meilisearch versions, please follow our guide on https://www.meilisearch.com/docs/learn/update_and_migration/updating.",
+        To migrate data between Meilisearch versions, please follow our guide on https://www.meilisearch.com/docs/learn/update_and_migration/updating.\n\
+        Alternatively, you can set the `--upgrade-db` flag (or the `MEILI_UPGRADE_DB` environment variable) to upgrade the database automatically on startup.",
```
dureuill:
```suggestion
        Alternatively, you can set the `--upgrade-db` flag (or the `MEILI_UPGRADE_DB` environment variable) to upgrade the database on startup.",
```

crates/meilisearch-types/src/versioning.rs
```diff
@@ -81,13 +81,14 @@ pub enum VersionFileError {
...
         env!("CARGO_PKG_VERSION").to_string()
     )]
     VersionMismatch { major: u32, minor: u32, patch: u32 },
     #[error("Database version {major}.{minor}.{patch} is higher than the Meilisearch version {VERSION_MAJOR}.{VERSION_MINOR}.{VERSION_PATCH}. Downgrade is not supported")]
     DowngradeNotSupported { major: u32, minor: u32, patch: u32 },
-    #[error("Database version {major}.{minor}.{patch} is too old for the experimental dumpless upgrade feature. Please generate a dump using the v{major}.{minor}.{patch} and import it in the v{VERSION_MAJOR}.{VERSION_MINOR}.{VERSION_PATCH}")]
+    #[error("Database version {major}.{minor}.{patch} is too old to be upgraded automatically. Please generate a dump using the v{major}.{minor}.{patch} and import it in the v{VERSION_MAJOR}.{VERSION_MINOR}.{VERSION_PATCH}")]
```
dureuill:
```suggestion
    #[error("Database version {major}.{minor}.{patch} is too old to be upgraded via `--db-upgrade`. Please generate a dump using the v{major}.{minor}.{patch} and import it in the v{VERSION_MAJOR}.{VERSION_MINOR}.{VERSION_PATCH}")]
```

crates/meilisearch/tests/upgrade/mod.rs
```diff
@@ -51,7 +51,7 @@ async fn version_too_old() {
     let current_version = format!("{major}.{minor}.{patch}");
     let err = err.replace(&current_version, "[current version]");
 
-    snapshot!(err, @"Database version 1.11.9999 is too old for the experimental dumpless upgrade feature. Please generate a dump using the v1.11.9999 and import it in the v[current version]");
+    snapshot!(err, @"Database version 1.11.9999 is too old to be upgraded automatically. Please generate a dump using the v1.11.9999 and import it in the v[current version]");
```
dureuill:
```suggestion
    snapshot!(err, @"Database version 1.11.9999 is too old to be upgraded via `--upgrade-db`. Please generate a dump using the v1.11.9999 and import it in the v[current version]");
```

dureuill (commented):
I find the "upgrade automatically" wording a bit confusing, so i suggested its removal.

Otherwise I'm all good with this 👍 

Just wondering if we should allow both flags for a few versions to make switching from one flag to another easier 🤔

curquiza:
> Just wondering if we should allow both flags for a few versions to make switching from one flag to another easier 🤔

I think it's more work for a switch that the users will have to do at some point when upgrading, and it's experimental so ok to do this as long as the communication is clear on the changelogs

dureuill (approved):
thank you,

waiting for the changes on Cloud, otherwise ✈️

curquiza:
https://github.com/meilisearch/meilisearch-cloud/pull/7735 is merged @dureuill , so I merge

emilpriver:
Hey!

Is `MEILI_EXPERIMENTAL_DUMPLESS_UPGRADE` changed as well? :D

dureuill:
@emilpriver 

yes!

`MEILI_UPGRADE_DB`

https://github.com/meilisearch/meilisearch/blob/main/crates/meilisearch/src/option.rs#L448

emilpriver:
> MEILI_UPGRADE_DB

Perfect! merci :D I didn't find it in the docs so thats why I asked

dureuill:
No problem,

the value should be in the docs now: https://www.meilisearch.com/docs/resources/self_hosting/configuration/reference#dumpless-upgrade

emilpriver:
Thank you :D 

I don't know if you know but if you click documentation on meilisearch page are you sent to https://www.meilisearch.com/docs/getting_started/overview.md and not https://www.meilisearch.com/docs/getting_started
image

just a FYI :D

dureuill:
ahaha thanks for letting me know. I wasn't aware, I always type `docs.meilisearch.com` directly :-p

will forward to the team

PR #6494, opened 2026-07-06
Retry requests to create a multipart upload for the S3 snapshots

Kerollmops:
This _potentially_ fixes an issue we have when sending a request to AWS S3 to create a new multipart upload, ensuring we resend the request if it fails.

### DIFF
--- a/crates/index-scheduler/src/scheduler/enterprise_edition/s3.rs
+++ b/crates/index-scheduler/src/scheduler/enterprise_edition/s3.rs
@@ -396,7 +396,7 @@ async fn multipart_stream_to_s3(
     use std::path::PathBuf;
 
     use bytes::{BufMut as _, Bytes, BytesMut};
-    use http_client::reqwest::{Client, Response};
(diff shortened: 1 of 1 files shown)


crates/index-scheduler/src/scheduler/enterprise_edition/s3.rs
```diff
@@ -423,19 +423,31 @@ async fn multipart_stream_to_s3(
...
+                    .await
+                    .map_err(http_client::reqwest::Error::from)
+                    .map_err(Error::S3HttpError)
+                    .map_err(backoff::Error::Permanent)
+            } else {
+                Err(backoff::Error::transient(Error::S3Error {
+                    status,
+                    body: resp.text().await.unwrap_or_default(),
+                }))
+            }
```
dureuill: I'm not sure I get in which case we retrieve a successful response body, can you clarify for me please?

dureuill (changes requested):

dureuill (approved):
thank you, I hope this will fix the issue. Retry logic looks good

PR #6495, opened 2026-07-06
Implement Document route sharding

ManyTheFish:
# summary

Support sharding and network on **GET** `indexes/:uid/documents`, **GET** `indexes/:uid/documents/:document_id` and on **POST** `indexes/:uid/documents/fetch`

# Changelog

**GET** `indexes/:uid/documents`, **GET** `indexes/:uid/documents/:document_id` and **POST** `indexes/:uid/documents/fetch` will now fetch the documents from all the shards in the configured network.

Moreover, a new `useNetwork` parameter is available to activate or deactivate the usage of the network.

## Breaking change

The default behavior for users using the network experimental feature will change on the following routes:
- **GET** `indexes/:uid/documents`
- **GET** `indexes/:uid/documents/:document_id`
- **POST** `indexes/:uid/documents/fetch`

Meilisearch will now fetch the documents from all the shards and not only on the local machine when processing the request.
To keep the same behavior as before, users will have to set `useNetwork` to `false` when making their request.


crates/meilisearch/src/routes/indexes/documents.rs
```diff
@@ -478,6 +482,49 @@ pub struct BrowseQuery {
...
+        &mut self.use_network
+    }
+
+    fn has_remote(&self) -> bool {
+        false
+    }
+}
+
+impl ProxyQuery for &BrowseQuery {
+    /// The only things that can change are the filter on shard and the remote, so recover this
```
dureuill: I'm not sure that this comment is still current since we now return the entire BrowseQuery.

crates/meilisearch/src/routes/indexes/documents.rs
```diff
@@ -478,6 +482,49 @@ pub struct BrowseQuery {
...
+    type ProxiedQuery = (String, BrowseQuery);
+
+    fn proxy_with_remote(&self, remote: String) -> Self::ProxiedQuery {
+        let mut query = (*self).clone();
+        query.limit += self.offset;
+        query.offset = 0;
```
dureuill, line 605: perhaps comment that we're fixing the limit and offset to prepare for federation?

crates/meilisearch/src/routes/indexes/documents.rs
```diff
@@ -666,6 +719,7 @@ pub async fn get_documents(
         filter,
         ids: ids.map(|ids| ids.into_iter().map(Into::into).collect()),
         sort: sort.map(|attr| fix_sort_query_parameters(&attr)),
+        use_network: None,
```
dureuill: should we allow `useNetwork` in get queries?

crates/meilisearch/src/routes/indexes/documents.rs
```diff
@@ -686,7 +740,7 @@ pub async fn get_documents(
         &req,
     );
 
-    let ret = documents_by_query(index_scheduler.clone(), index_uid, query).await;
+    let ret = documents_by_query(index_scheduler.clone(), index_uid, query, false).await;
```
dureuill, line 900: Is this route never called by proxy? if so, can we document that?

crates/meilisearch/src/routes/indexes/documents.rs
```diff
@@ -752,15 +824,200 @@ pub async fn get_documents(
...
+                }
+            }
+        }
+        in_flight_requests
+            .push_back(tokio::spawn(async move { (request.await, remote_name, query_id) }));
+    }
+
+    // Perform local search
+    for (query_id, (_, query)) in local_queries {
+        // TODO: clone index_scheduler and index_uid maybe not needed
```
dureuill: do we want to remove the clones?

dureuill (commented):
Thank you, I have a few comments

crates/meilisearch/src/routes/indexes/documents.rs
ManyTheFish: done (use network is now allowed on **GET** `indexes/:uid/documents` and **GET** `indexes/:uid/documents/:document_id`)

crates/meilisearch/src/routes/indexes/documents.rs
ManyTheFish: done (removed)

crates/meilisearch/src/routes/indexes/documents.rs
ManyTheFish: done (added a comment)

crates/meilisearch/src/routes/indexes/documents.rs
ManyTheFish: never, it's always the POST route

crates/meilisearch/src/routes/indexes/documents.rs
ManyTheFish: I removed the comment, we do a spawn_blocking just after, so we must clone

dureuill (approved):
Thank you, looks great!

Can you specify in the PR description (+label) that this is a **breaking change** for users of sharding (with network experimental feature), as document fetch will now default to federated?

Then you can merge

PR #6497, opened 2026-07-06
Support attribute patterns when asking for facet distribution

Kerollmops:
## Changelog

The `facets` parameter in search and federated search now supports more wildcards. Previously, only the single wildcard `"*"` was supported, requesting all filterable fields.

Now, patterns *containing* `*`  are supported with the same matching rules as in `filterableAttributes.attributePatterns` and `localizedAttributes.attributePatterns`, such as `dogs.*`, which will add to the facet distribution all filterable fields that match the pattern (such as `dogs.intel`, `dogs.kefir`, etc.).

---

This PR enables providing patterns when requesting the facet distribution of certain fields. Before, it was only possible to provide either a whole attribute name, i.e., `title`, `dogs.intel`, or everything through the `*` special pattern. Now, a user can provide patterns, i.e., `title`, `dogs.*`, `*` and retrieve all of the facets that are filterable. An error is still thrown if a pattern is found to match a non-filterable pattern or matches nothing in the filterable attributes list.

## To Do

- [x] Update the documentation
- [x] Support attribute patterns on the search route
- [x] Support attribute patterns on the multi-search/federation route
- [x] Make sure it works with sharding

### DIFF
--- a/crates/meilisearch/Cargo.toml
+++ b/crates/meilisearch/Cargo.toml
@@ -138,7 +138,7 @@ tempfile = { version = "3.27.0", optional = true }
 zip = { version = "6.0.0", optional = true, default-features = false, features = ["deflate-flate2"] }
 
 [features]
-default = ["meilisearch-types/all-tokenizations", "mini-dashboard"]
+default = ["enterprise", "meilisearch-types/all-tokenizations", "mini-dashboard"]
 swagger = ["utoipa-scalar"]
 test-ollama = []
 mini-dashboard = [
--- a/crates/meilisearch/Cargo.toml
+++ b/crates/meilisearch/Cargo.toml
@@ -138,7 +138,7 @@ tempfile = { version = "3.27.0", optional = true }
 zip = { version = "6.0.0", optional = true, default-features = false, features = ["deflate-flate2"] }
 
 [features]
-default = ["enterprise", "meilisearch-types/all-tokenizations", "mini-dashboard"]
+default = ["meilisearch-types/all-tokenizations", "mini-dashboard"]
 swagger = ["utoipa-scalar"]
 test-ollama = []
 mini-dashboard = [
--- a/crates/meilisearch/src/search/mod.rs
+++ b/crates/meilisearch/src/search/mod.rs
@@ -1,13 +1,13 @@
 use core::fmt;
 use std::cmp::min;
+use std::collections::HashMap;
 use std::collections::{BTreeMap, BTreeSet, HashSet};
+use std::iter;
 use std::ops::Not as _;
 use std::rc::Rc;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Instant;
-use std::collections::HashMap;
-use std::iter;
 
 use deserr::Deserr;
 pub use federated::ProxyQuery;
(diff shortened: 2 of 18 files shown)


crates/meilisearch/Cargo.toml
```diff
@@ -138,7 +138,7 @@ tempfile = { version = "3.27.0", optional = true }
 zip = { version = "6.0.0", optional = true, default-features = false, features = ["deflate-flate2"] }
 
 [features]
-default = ["meilisearch-types/all-tokenizations", "mini-dashboard"]
+default = ["enterprise", "meilisearch-types/all-tokenizations", "mini-dashboard"]
```
dureuill: I don't think this change should stay, I'll be removing it, we can add it as a separate PR if needs be, after discussion

crates/milli/src/error.rs
```diff
@@ -147,33 +147,24 @@ and can not be more than 511 bytes.", .document_id.to_string()
...
-            match .matching_rule_indices.get(field) {
+        {
+            let field = .invalid_facet_pattern;
+            match .matching_rule_index {
                 Some(rule_index) => format!("Attribute `{}` matched rule #{} in filterableAttributes, but this rule does not enable filtering.\nHint: enable filtering in rule #{} by modifying the features.filter object\nHint: prepend another rule matching `{}` with appropriate filter features before rule #{}",
                     field, rule_index, rule_index, field, rule_index),
                 None => match .valid_patterns.is_empty() {
-                    true => format!("Attribute `{}` is not filterable. This index does not have configured filterable attributes.", field),
-                    false => format!("Attribute `{}` is not filterable. Available filterable attributes patterns are: `{}`.",
+                    true => format!("Pattern `{}` is not filterable. This index does not have configured filterable attributes.", field),
```
dureuill, line 156: The "pattern" terminology feels a bit strange when the passed facet is not a pattern (does not contain `*`). I wonder if we should reuse the `attribute` terminology in this case

crates/meilisearch/src/search/mod.rs
```diff
@@ -2049,20 +2051,64 @@ fn compute_facet_distribution_stats<S: AsRef<str>>(
...
+        return Err(Error::UserError(UserError::InvalidFacetsDistribution {
+            invalid_facet_pattern: facet_pattern.clone(),
+            valid_patterns: fetch_valid_patterns(),
+            // Not a single pattern matched our facet pattern
+            matching_rule_index: None,
+        })
+        .into());
     }
 
+    let fidmap = index.fields_ids_map_with_metadata(rtxn)?;
```
dureuill: I think we've been trying very hard in the past to avoid spawning a fidmap with metadata during a search due to the possibly involved cost as the number of fields increases. 

The issue is that we'll be computing the entire set of metadata for all fields even if we only have a single facet that has been specified (and not even a pattern). Doing so involves checking all fields against the patterns in filterable attributes until we find the matching pattern.

I wonder if we shouldn't do this only for  fields that match at least a pattern in the requested...

dureuill: Fixed:

1. Add facilities to compute partial metadata in MetadataBuilder https://github.com/meilisearch/meilisearch/pull/6497/commits/7044bb49afca11ff0027b3f5d296e6971e8b1f6f
2. Use them in https://github.com/meilisearch/meilisearch/pull/6497/commits/288993c69d84be744e12c76cc55ef56c005e096e
3. Reuse fidmap rather than deserializing it twice

crates/milli/src/error.rs
dureuill: can be fixed later

crates/meilisearch/Cargo.toml
dureuill: done

crates/milli/src/fields_ids_map/metadata.rs
```diff
@@ -296,7 +296,6 @@ pub struct MetadataBuilder {
     distinct_attribute: Option<String>,
     asc_desc_attributes: HashMap<String, FieldSortOrder>,
     displayed_attributes: Option<HashSet<String>>,
-    fields_metadata: BTreeMap<String, Metadata>,
```
ManyTheFish, line 292: 👋

ManyTheFish (dismissed):

dureuill (approved):
Tested, looks nice 🌈

PR #6499, opened 2026-07-07
Fix yet another filter round-trip serialization error

### LINKED ISSUE #6493: Escaped quotes/backslashes in filters return `invalid_search_filter` error
## Describe the bug
A filter expression containing escaped backslashes and double quotes in both the attribute name and value is rejected by Meilisearch with `invalid_search_filter`. 

## Expected behavior
The request should be accepted and parsed as a valid filter, as it was in earlier versions. The server should not reject properly escaped quoted attribute/value pairs.

## Actual behavior
Meilisearch returns HTTP `400` with `invalid_search_filter`, e.g. `Was expecting an operation ... at "\"crazy_\\_\\\"field\\\"\" = \"real \\\\\"crazy\\\"\""`. The same payload previously worked.

## Steps to reproduce
1. Create an index, make the unusual field filterable, and add one document:
```bash
curl -X POST 'http://localhost:7700/indexes' \
  -H 'Authorization: Bearer masterKey' \
  --data-binary '{"uid":"movies","primaryKey":"id"}'

curl -X PATCH 'http://localhost:7700/indexes/movies/settings/filterable-attributes' \
  -H 'Authorization: Bearer masterKey' \
  --data-binary '["crazy_\\_\"field\""]'

curl -X POST 'http://localhost:7700/indexes/movies/documents' \
  -H 'Authorization: Bearer masterKey' \
  --data-binary '[{"id":1,"crazy_\\_\"field\"":"real \\\"crazy\\\""}]'
```

2. Run a `multi-search` using the escaped filter:
```bash
curl -X POST 'http://localhost:7700/multi-search' \
  -H 'Authorization: Bearer masterKey' \
  --data-binary '{
    "queries": [
      {
        "indexUid": "movies",
        "q": "",
        "filter": ["\"crazy_\\\\_\\\\\\\"field\\\\\\\"\"=\"real \\\\\\\"crazy\\\\\\\"\""]
      }
    ]
  }'
```

3. Observe that the server returns `400 invalid_search_filter` instead of accepting the query.

## Version
Reproduced on `v1.48.3`

## Additional context (optional)
This bug is visible in `meilisearch-js-plugins` [failing SDK tests](https://github.com/meilisearch/meilisearch/actions/runs/28733007711/job/85202058112)

dureuill:
## Changelog

Fix a bug where some filters containing escaped characters (such as `\\`) would cause search requests to fail with `invalid_search_filter`

## Related issue

Fixes #6493 

## Implementation

The bug stems from the fact that we're manually escaping `"` to `\"`, which leaves out other escaped characters such as `\`.

To fix this issue I added  `serde-json` and use it to escape the filter fragments as a serde-json string, which makes sense when serializing them and should take care of all escaped characters.

## Generative AI tools

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2274,6 +2274,7 @@ dependencies = [
  "levenshtein_automata",
  "nom",
  "nom_locate",
+ "serde_json",
  "unescaper",
 ]
 
--- a/crates/filter-parser/Cargo.toml
+++ b/crates/filter-parser/Cargo.toml
@@ -16,6 +16,7 @@ nom = "7.1.3"
 nom_locate = "4.2.0"
 unescaper = "0.1.6"
 levenshtein_automata = { version = "0.2.1", features = ["fst_automaton"] }
+serde_json = { version = "1.0.150", features = ["preserve_order"] }
 
 [dev-dependencies]
 # fixed version due to format breakages in v1.40
--- a/crates/filter-parser/src/lib.rs
+++ b/crates/filter-parser/src/lib.rs
@@ -175,7 +175,7 @@ impl<'a> Token<'a> {
 
     /// The fragment with escaped double quotes.
     pub fn escaped_fragment(&self) -> String {
-        self.fragment().replace('"', r#"\""#)
+        serde_json::to_string(self.fragment()).unwrap()
     }
 
     /// Returns the extra fragment of the token.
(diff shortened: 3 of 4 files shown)


ManyTheFish (approved):
OUI!

LGTM, so it was indeed the `\` that was not escaped.

PR #6501, opened 2026-07-07
Make sure not to store empty or long synonym keys

Kerollmops:
This PR makes sure we do not store empty synonym keys in LMDB and simply ignore them. Note that, as before, we keep the original user synonyms unchanged.


dureuill (approved):
👍

PR #6505, opened 2026-07-09
Add filter condition to DSRs

dureuill:
### New `filter` rule activation condition

Dynamic search rules (DSR) can now declare a new `filter` condition: it contains a single `values` key, whose value is a JSON object.
The keys of this `values` JSON object are the  facet names (e.g.  `color`, `genres`, ...), while their values are the values that a filter must resolve to for these facets, so that the rule is active.

<details>

<summary>Example of sending a DSR with a filter condition</summary>

```jsonc
// PATCH /dynamic-search-rules/test-filter

{
  "conditions": {
    "filter": {
      "values": {
        "color": "red",
        "category": "shirt"
      }
    }
  },
  "actions": [
    {
      "selector": { "id": "red shirt on sales" },
      "action": { "type": "pin", "position": 0 }
    }
  ]
}
```

</details>

<details>
<summary>Example of search query that enables the DSR above</summary>

```jsonc
{
  "filter": "(category = shirt AND color = red) OR (category = jeans AND color = blue)"
}
```

</details>

Note that the filter only needs to have one branch resolving to all the facet values declared in the rule for the rule to be active.

### New `lastUpdatedAt` field

Dynamic search rules returned by `GET /dynamic-search-rules/{:ruleUid}` and `POST /dynamic-search-rules` contain an additional `lastUpdatedAt` field. 

This field is automatically updated with the `enqueuedAt` value of the last task that modified the rule.

### Rules are listed with the most recently updated first

`POST /dynamic-search-rules` now lists rules in descending `lastUpdatedAt` order, meaning that the most recently updated rules will be listed first.

### New environment variables

New environment variables are defined to control the behavior of the DSR fuel:

- `MEILI_EXPERIMENTAL_DSR_FUEL_FILTER_FUEL` (in range 0..4294967296): controls how many filter constraint combinations Meilisearch will attempt to resolve at search time before giving up and applying partial rules.
- `MEILI_EXPERIMENTAL_DSR_FUEL_FILTER_OR_FUEL` (in range 0..65536): controls how many filter disjunctions Meilisearch will build when turning a filter to its canonical shape
- `MEILI_EXPERIMENTAL_DSR_FUEL_FILTER_AND_FUEL` (in range 0..65536): controls how many filter conjunctions Meilisearch will build in total when turning a filter to its canonical shape
- `MEILI_EXPERIMENTAL_DSR_FUEL_FILTER_DEPTH_FUEL` (in range 0..256): controls the maximum recursive depth that Meilisearch allows when turning a filter to its canonical shape

## Implementation

- Modify the way updates to DSR are handled: updated rules are now computed by merging the previous and new versions of the rule, and the computed payload is now a replacement rather than an update.
- Add DSR index upgrade logic that reapplies the dsr settings if in a given version range.

### Generative AI tools


crates/filter-parser/src/constraint.rs
```diff
@@ -0,0 +1,254 @@
...
+        if self.remaining_and.0 == 0 {
+            ControlFlow::Break(())
+        } else {
+            ControlFlow::Continue(())
+        }
+    }
+
+    fn consume_depth_fuel(&mut self) -> ControlFlow<(), ()> {
+        self.remaining_depth -= 1;
+        if self.remaining_depth.0 == 0 {
```
Kerollmops, line 238: I would prefer you use `self.remaining_xxx.saturating_sub(1)` here because nothing prevents you from creating a `FilterConstraintFuel` from all zeroes, and you will end up with a panic from overflow.

crates/filter-parser/Cargo.toml
```diff
@@ -16,7 +16,9 @@ nom = "7.1.3"
 nom_locate = "4.2.0"
 unescaper = "0.1.6"
 levenshtein_automata = { version = "0.2.1", features = ["fst_automaton"] }
+itertools = "0.14.0"
 serde_json = { version = "1.0.150", features = ["preserve_order"] }
+wip = "0.2.0"
```
Kerollmops: Don't you want to remove it from the dependencies?
Can't we make sure this dep never ends up in the repo?
Via a CI check or a config.toml stuff, maybe?

crates/index-scheduler/Cargo.toml
```diff
@@ -47,6 +47,7 @@ uuid = { version = "1.18.1", features = ["serde", "v4"] }
 backoff = { version = "0.4.0", features = ["futures", "tokio"] }
 rusty-s3 = "0.8.1"
 tokio = { version = "1.52.3", features = ["full"] }
+wip = "0.2.0"
```
Kerollmops, line 50: Don't you want to remove it?

crates/meilisearch-types/Cargo.toml
```diff
@@ -53,6 +53,7 @@ utoipa = { version = "5.4.0", features = [
     "openapi_extensions",
 ] }
 uuid = { version = "1.18.1", features = ["serde", "v4"] }
+wip = "0.2.0"
```
Kerollmops: Same?

crates/milli/src/dynamic_search_rules.rs
```diff
@@ -208,65 +216,86 @@ impl<'a> DynamicSearchRulesView<'a> {
...
 
-        // 3. exclude rules that have the a different query emptiness condition
+    fn apply_query_conditions(
+        &self,
+        active_rules: &mut RoaringBitmap,
+        query_terms: &[LocatedQueryTerm],
+        search_context: &SearchContext<'_>,
+        mut fuel: DsrFuel,
+    ) -> Result<(), crate::Error> {
+        // 1. exclude rules that have the a different query emptiness condition
```
Kerollmops:
```suggestion
        // 1. exclude rules that have a different query emptiness condition
```

crates/milli/src/dynamic_search_rules.rs
```diff
@@ -515,20 +752,27 @@ pub struct DsrFuel {
     max_counted_words: u8,
     max_active_rules: u32,
     max_pin_actions: u32,
-    remaining_word_fuel: std::num::Saturating<u32>,
+    remaining_word_fuel: Saturating<u32>,
+    remaining_filter_fuel: Saturating<u32>,
```
Kerollmops, line 759: I thought they were `u16` somewhere else? Is it because this is a "sum" of them all?

crates/milli/src/dynamic_search_rules.rs
```diff
@@ -545,6 +789,15 @@ impl DsrFuel {
         }
     }
 
+    pub fn consume_filter_combination(&mut self) -> ControlFlow<(), ()> {
+        self.remaining_filter_fuel -= 1;
```
Kerollmops, line 796: Same, I would rather prefer you use `.saturating_sub(1)` here.

crates/milli/Cargo.toml
```diff
@@ -122,6 +122,7 @@ twox-hash = { version = "2.1.2", default-features = false, features = [
 ] }
 geo-types = "=0.7.17" # fixed version: see <https://github.com/meilisearch/meilisearch/pull/6118>
 zerometry = "0.3.0"
+wip = "0.2.0"
```
Kerollmops: Same?

Kerollmops (changes requested):
> This field is automatically updated with the `enqueuedAt` value of the last task that modified the rule.

Why not use the `processedAt` instead? Do you want to "fake" the actual value and not show the actual moment the DSR was lastly modified and those modifications applied?

> Note that the filter only needs to have one branch resolving to all the facet values declared in the rule for the rule to be active.

If I understand correctly you are not matching a DSR filter rule on the syntax but on the matching documents, and you do that by testing the eqaulity between the rule and every filter components?

crates/filter-parser/src/constraint.rs
dureuill: no, because `remaining_xxx` is a [Saturating](https://doc.rust-lang.org/std/num/struct.Saturating.html), which already encapsulates this semantics

crates/filter-parser/Cargo.toml
dureuill: yeah will remove the corresponding commit with a rebase

crates/milli/src/dynamic_search_rules.rs
dureuill: u16 is for the fuel that builds the constraints out of the filter, not for resolving constraint combinations.

So they're two different things

crates/milli/src/dynamic_search_rules.rs
dureuill: same answer as above

dureuill:
> Why not use the processedAt instead? Do you want to "fake" the actual value and not show the actual moment the DSR was lastly modified and those modifications applied?

`enqueuedAt` is going to be more stable and easier to know when a change originate with. Autobatching means that many DSR could share very close (or identical) processing time.

Reading your comment, I'm no longer sure of what we want here. Perhaps the `latestTaskUid` that touched the rule would be more suitable than the date? That way we can still roughly sort by "update time" since tasks with higher uids tend to be processed later than tasks with lower uids (especially when they are of the same kind and target the same special index), and users get to find the processedAt and enqueuedAt by looking for the task. wdyt @Kerollmops ?

dureuill:
> If I understand correctly you are not matching a DSR filter rule on the syntax but on the matching documents, and you do that by testing the eqaulity between the rule and every filter components?

You are correct that I'm not using the syntax to match a DSR rule to a filter.
We have to be careful with what we mean with "documents" here.

What I do is a bit subtle:

1. I canonicalize the filter to a shape where they can be expressed as a list of disjunctions (ORs), where each member is a list of fields, and for each field a list of conjunctions (AND) on the permissible values for that field.
2. I index the DSRs with their `filter.values.*` as filterable.
3. At resolution time, I use the canonical filter shape to find all rules that satisfy at least one disjunction by having a permissible value for all fields in that disjunction.

crates/milli/Cargo.toml
dureuill: rebased without the offending commit

crates/meilisearch-types/Cargo.toml
dureuill: rebased without the offeending commit

crates/index-scheduler/Cargo.toml
dureuill: rebased without the offending commit

crates/filter-parser/Cargo.toml
dureuill: done

Kerollmops:
> Perhaps the `latestTaskUid` that touched the rule would be more suitable than the date? That way we can still roughly sort by "update time" since tasks with higher uids tend to be processed later than tasks with lower uids (especially when they are of the same kind and target the same special index), and users get to find the processedAt and enqueuedAt by looking for the task.

At first, I thought it was a great idea, but the fact that Meilisearch deletes tasks is not ideal, as users could end up with no data to show. Maybe that is enough, but thinking about UI stuff, I believe your idea of keeping the `enqueuedAt` date is probably the best in fact 👌

crates/milli/src/dynamic_search_rules.rs
dureuill: I updated the description to document all newly introduced env variables

crates/filter-parser/src/constraint.rs
Kerollmops: my bad, I miss that info. Perfect then 👌

Kerollmops (approved):
Amazing _demande de tirage_, loves it ❤️

PR #6506, opened 2026-07-09
Don't omit rules without precedence when there exist rules with precedence

dureuill:
- Add test that fails on main and succeeds on this branch

## Generative AI tools


ManyTheFish (approved):
OKEY!


What is the behavior of ascending_facet_sort for null facets?
I thought it was returning them after the documents having the facet 🤔

PR #6509, opened 2026-07-13
Own parsed filters structure

ManyTheFish:
## Related

Fixes: https://linear.app/meilisearch/issue/ENGPROD-2617

This work is related to the document join filtering.

The goal here is to parse the filter in a preprocessing step, which implies storing the parsed value in an intermediate struct.

This PR removes the reference to the filter string in the parsed filter structure in order to own it completely.

### DIFF
--- a/crates/filter-parser/src/error.rs
+++ b/crates/filter-parser/src/error.rs
@@ -3,7 +3,7 @@ use std::fmt::Display;
 use nom::error::{self, ParseError};
 use nom::Parser;
 
-use crate::{CowSpan, IResult, Span};
+use crate::{IResult, OwnedSpan, Span};
 
 pub trait NomErrorExt<E> {
     fn is_failure(&self) -> bool;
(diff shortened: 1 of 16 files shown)


crates/filter-parser/src/error.rs
```diff
@@ -303,11 +303,8 @@ impl Display for Error<'_> {
...
         }
-        if let Some(base_column) = self.context.get_utf8_column() {
-            let size = self.context.fragment().chars().count();
-            write!(f, "{}:{} {}", base_column, base_column + size, self.context.extra())
-        } else {
-            write!(f, "{}", self.context.extra())
-        }
+        let base_column = self.context.get_utf8_column();
+        let size = self.context.fragment().chars().count();
+        write!(f, "{}:{} {}", base_column, base_column + size, self.context.extra())
```
ManyTheFish, line 308: @dureuill

dureuill (approved):
j'aime bien 🐰

PR #6511, opened 2026-07-13
simplify partitioning step

ManyTheFish:
This PR simplifies the partitioning step in the federated search.

Before, the federated query was destructured into a tuple containing:
- simple query
- the extracted federation option
- the extracted index uid

Then this tuple was directly used to reconstruct the federated query before calling `to_partition`.

This PR removes that intermediate transformation.

I may be missing something, so that's why I created a dedicated PR for it.

@dureuill, let me know if this intermediate transformation must be kept. 🤔

### DIFF
--- a/crates/meilisearch/src/search/federated/network.rs
+++ b/crates/meilisearch/src/search/federated/network.rs
@@ -109,17 +109,6 @@ impl Partition {
         })
     }
 
-    pub fn to_query_partition(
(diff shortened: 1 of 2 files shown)


dureuill (approved):
This removes the call to fixup the federation for federated queries that `useNetwork`, but I believe it cannot have an effect on already federated queries

PR #6520, opened 2026-07-15
Fix frequently failing tests on Windows

dureuill:
## Changelog

Fix two tests that tended to fail on Windows because they would not await some tasks.

## Generative AI tools


ManyTheFish (approved):
Approved

PR #6528, opened 2026-07-20
Speed up search requests by reducing the number of deserializations

Kerollmops:
This PR improves Meilisearch search speed by ensuring we avoid doing unnecessary work internally. We drastically reduced the number of times we retrieve data from disk to a single time across the whole search pipeline.

- [x] Fix the performance difference between `q: ""` and `q: null`, i.e., -55% on placeholder search requests.
- [x] Drastically reduce the number of fields ids map deserializations, i.e., -62% on placeholder search requests.
- [x] Log the time it takes to deserialize the fields ids map in the progress, i.e., `showPerformanceDetails`.

## Generative AI tools

### DIFF
--- a/crates/meilisearch/src/search/federated/perform.rs
+++ b/crates/meilisearch/src/search/federated/perform.rs
@@ -1707,7 +1707,13 @@ impl SearchByIndex {
         let estimated_total_hits = candidates.len() as usize;
         let facets = facet_patterns_by_index
             .map(|facets_by_index| {
-                compute_facet_distribution_stats(&facets_by_index, &index, &rtxn, candidates)
+                compute_facet_distribution_stats(
+                    &facets_by_index,
+                    &index,
+                    &rtxn,
+                    &fidmap,
+                    candidates,
+                )
             })
             .transpose()
             .map_err(|mut error| {
@@ -1776,9 +1782,13 @@ impl SearchByIndex {
             }
 
             if let Some(facets) = facets {
-                if let Err(mut error) =
-                    compute_facet_distribution_stats(&facets, &index, &rtxn, Default::default())
-                {
+                if let Err(mut error) = compute_facet_distribution_stats(
+                    &facets,
+                    &index,
+                    &rtxn,
+                    &fidmap,
+                    Default::default(),
+                ) {
                     if self.show_federation_info == ShowFederationInfo::Always {
                         error.message = format!(
                             "Inside `.federation.facetsByIndex.{index_uid}`: {}\n - Note: index `{index_uid}` is not used in queries",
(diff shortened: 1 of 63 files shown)


crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -1429,7 +1432,9 @@ impl SearchByIndex {
                     (SearchKind::SemanticOnly { .. }, _) => {
                         ranking_rules::CanonicalizationKind::Vector
                     }
-                    (_, Some(q)) if !q.is_empty() => ranking_rules::CanonicalizationKind::Keyword,
+                    (_, Some(q)) if !q.trim().is_empty() => {
```
dureuill, line 1435: 🥲

Kerollmops: It could be faster to just check if both ends characters are whitespaces, but... 🤷

crates/milli/src/test_index.rs
```diff
@@ -245,7 +246,14 @@ impl TempIndex {
     }
 
     pub fn search<'a>(&'a self, rtxn: &'a heed::RoTxn<'a>) -> Search<'a> {
-        self.inner.search(rtxn, "test", time::OffsetDateTime::now_utc(), &self.progress)
+        let fields_ids_map = self.inner.fields_ids_map(rtxn).unwrap();
```
dureuill: I think we should require the fidmap in parameter of this function

crates/milli/src/search/mod.rs
```diff
@@ -73,6 +74,7 @@ pub struct Search<'a> {
     max_total_hits: Option<usize>,
     rtxn: &'a heed::RoTxn<'a>,
     index: &'a Index,
+    fields_ids_map: Cow<'a, FieldsIdsMap>,
```
dureuill: Please remove the Cow, replacing it with a reference. It looks like the only Owned case is a test case driver, where the fidmap reference could feasibly be propagated as a function parameter instead.

We shouldn't pay for the complexity and (small) runtime cost of Cow just for the tests

dureuill (commented):
Awesome, this is a long due change that will undoubtedly improve performance for many use cases 🌈 

reading the code, a lot of the changes are somewhat "mechanical" in nature, I did not see any potential issue.

Please look at inline comments, I request we remove the 🐄 from the Search object though.

Not for this PR, but I believe that the triple `(index, rtxn, fidmap)` is prevalent and coherent enough to warrant a dedicated object

crates/meilisearch/src/search/federated/perform.rs
dureuill: ah my reaction wasn't clear, I don't dislike the trim.empty, it's just that I'm sad it wasn't already there

dureuill (approved):
j'aime beaucoup 🌈

PR #6531, opened 2026-07-22
Make sure legacy attribute patterns works

Kerollmops:
This PR fixes a regression introduced in v1.50, where patterns like `"*"` no longer match filterable fields when they were declared using the old filterable syntax (not the `attributesPattern` syntax).

## Generative AI tools

### DIFF
--- a/crates/milli/src/filterable_attributes_rules.rs
+++ b/crates/milli/src/filterable_attributes_rules.rs
@@ -4,7 +4,9 @@ use deserr::{DeserializeError, Deserr, ValuePointerRef};
 use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
-use crate::attribute_patterns::{match_distinct_field, match_field_legacy, PatternMatch};
+use crate::attribute_patterns::{
+    match_distinct_field, match_field_legacy, match_pattern, PatternMatch,
+};
 use crate::constants::{RESERVED_GEOJSON_FIELD_NAME, RESERVED_GEO_FIELD_NAME};
 use crate::AttributePatterns;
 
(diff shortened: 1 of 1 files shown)


crates/milli/src/filterable_attributes_rules.rs
```diff
@@ -29,13 +31,23 @@ impl FilterableAttributesRule {
...
 
+    /// Takes two patterns and returns true if one pattern matches another.
     pub fn intersect_patterns(&self, other: &str) -> bool {
         match self {
             // If the rule is a field, match the field against the provided pattern using the legacy behavior
-            FilterableAttributesRule::Field(field) => matches!(
-                match_field_legacy(other, field),
-                PatternMatch::Parent | PatternMatch::Match
-            ),
+            FilterableAttributesRule::Field(field) => {
```
dureuill, line 40: perhaps we should document what we're doing here?


```suggestion
            // match using both the legacy and new syntaxes, so that we catch the old prefix based
            // syntax and the new wildcard behavior, for example in the `facets` field of the search
            FilterableAttributesRule::Field(field) => {
```

crates/milli/src/filterable_attributes_rules.rs
```diff
@@ -29,13 +31,23 @@ impl FilterableAttributesRule {
...
+                    match_field_legacy(other, field),
+                    PatternMatch::Parent | PatternMatch::Match
+                );
+
+                let new = matches!(
+                    match_pattern(other, field),
+                    PatternMatch::Parent | PatternMatch::Match
+                );
+
+                legacy | new
```
dureuill:
```suggestion
                legacy || new
```

isn't `||` more idiomatic for boolean operations? I checked that both work

dureuill (commented):
Small fixes otherwise we'll take it

dureuill (approved):
🚀

PR #6533, opened 2026-07-22
Introduce the `/tasks/stream` and `/batches/stream` SSE routes

Kerollmops:
This PR introduces two new [Server-sent Events](https://en.wikipedia.org/wiki/Server-sent_events) routes to expose the events related to batches and tasks. Both routes are only available once the `tasksStreamingRoute` is enabled through the `/experimental-features` route.

## To do

- [x] Introduce an SSE route on `/tasks/stream`
- [x] Introduce an SSE route on `/batches/stream`
- [x] Accept no parameters on the tasks and batches stream routes
- [x] Make both routes available through a CLI/env experimental feature
- [x] Make both routes available through a runtime feature.
- [x] Update [the experimental feature route](https://github.com/orgs/meilisearch/discussions/889)

## Generative AI tools


crates/index-scheduler/src/lib.rs
```diff
@@ -482,26 +483,35 @@ impl IndexScheduler {
...
-                        Ok(Ok(TickOutcome::WaitForSignal)) => run.scheduler.wake_up.wait(),
+                        Ok(Ok(TickOutcome::WaitForSignal)) => {
+                            match run.scheduler.wake_up.blocking_recv() {
+                                Ok(_) => (),
+                                Err(RecvError::Closed) => break,
+                                Err(RecvError::Lagged(_)) => {
+                                    // The receiver has been forcibly disconnected as we were lagging
+                                    // too far behind. We reconnect the receiver to handle new messages
+                                    // and break to analyse the changes (tick).
+                                    run.scheduler.wake_up = run.scheduler.wake_up.resubscribe();
```
dureuill: I don't really get this comment or what is being done here, because receiving a lagging message **does not** disconnect the receiver. Instead, it updates to the oldest message still available so that the next call to `recv` gets the oldest available message (or `Lagged` again, if a new message has been sent since then).

In my understanding a fairer description of what is being done here:

```
// The receiver lost messages as it was lagging behind the senders. 
// We definitely received a message in this case, so we will tick again.
// As the tick will catch up with any newly received tasks, it is fair to skip older messages by resubscribing.
```

I think we could probably do without the resubscribe, unless I'm missing some other reason why it is necessary.

crates/meilisearch/src/routes/batches.rs
```diff
@@ -203,3 +210,115 @@ async fn get_batches(
...
+            // change this loop is unblocked and fetches new tasks info.
+            let query = match tokio::time::timeout(refresh_rate, wake_up.recv()).await {
+                // We list all the tasks that were imported by a dump
+                Ok(Ok(ModifiedTasks::DumpImported)) => query.clone(),
+                Ok(Ok(ModifiedTasks::Some { ids })) => {
+                    Query { uids: Some(ids.into_iter().collect()), ..query.clone() }
+                }
+                Ok(Err(RecvError::Closed)) => break 'listener,
+                Ok(Err(RecvError::Lagged(_))) => {
+                    wake_up = wake_up.resubscribe();
```
dureuill: we should probably not do this, as this will lose more messages than necessary

crates/meilisearch/src/routes/tasks.rs
```diff
@@ -594,6 +598,90 @@ async fn get_tasks(
...
+            // change this loop is unblocked and fetches new tasks info.
+            let query = match wake_up.recv().await {
+                // We list all the tasks that were imported by a dump
+                Ok(ModifiedTasks::DumpImported) => query.clone(),
+                Ok(ModifiedTasks::Some { ids }) => {
+                    Query { uids: Some(ids.into_iter().collect()), ..query.clone() }
+                }
+                Err(RecvError::Closed) => break 'listener,
+                Err(RecvError::Lagged(_)) => {
+                    wake_up = wake_up.resubscribe();
```
dureuill: same as above

crates/index-scheduler/src/lib.rs
```diff
@@ -482,26 +483,35 @@ impl IndexScheduler {
...
-                run.scheduler.wake_up.wait_timeout(std::time::Duration::from_secs(60));
-
                 loop {
-                    let ret = catch_unwind(AssertUnwindSafe(|| run.tick()));
-                    match ret {
+                    match catch_unwind(AssertUnwindSafe(|| run.tick())) {
                         Ok(Ok(TickOutcome::TickAgain(_))) => (),
-                        Ok(Ok(TickOutcome::WaitForSignal)) => run.scheduler.wake_up.wait(),
+                        Ok(Ok(TickOutcome::WaitForSignal)) => {
+                            match run.scheduler.wake_up.blocking_recv() {
```
dureuill, line 505: as per blocking_recv()'s documentation, this would panic if started inside an async runtime.

I think we should comment on why this cannot happen:


```suggestion
                            // no panic: this function is called in a freshly spawned thread and so
                            // is not within an asynchronous execution context.
                            match run.scheduler.wake_up.blocking_recv() {
```

crates/index-scheduler/src/lib.rs
Kerollmops: [You are absolutely right](https://docs.rs/tokio/latest/tokio/sync/broadcast/error/enum.RecvError.html#variant.Lagged)! I'll simply remove the re-subscribe action and probably take your comment. My bad. I miss-understood the doc 🤦

crates/index-scheduler/src/lib.rs
```diff
@@ -852,7 +861,11 @@ impl IndexScheduler {
         }
 
         // notify the scheduler loop to execute a new tick
-        self.scheduler.wake_up.signal();
+        self.scheduler
+            .waker
+            .send(ModifiedTasks::Some { ids: RoaringBitmap::from([task.uid]) })
+            .unwrap();
```
dureuill, line 866: should we comment on why this `unwrap` will not happen? do we need the panicking unwrap, as opposed to ignoring the error with `let _ = ...`?

Kerollmops: I am wondering indeed. Because if there is no listener, it means that there is not even the index scheduler listening for new tasks to process. We could ignore the message or unwrap to immediately panic and log the issue 👀

crates/meilisearch/src/routes/tasks.rs
```diff
@@ -594,6 +598,87 @@ async fn get_tasks(
...
+    let (tx, rx) = tokio::sync::mpsc::channel(10);
+    let _join_handle = Handle::current().spawn(async move {
+        let mut wake_up = index_scheduler.as_ref().scheduler.wake_up.resubscribe();
+
+        'listener: loop {
+            // wait for new tasks to be available. Every time tasks statuses
+            // change this loop is unblocked and fetches new tasks info.
+            let query = match wake_up.recv().await {
+                // We list all the tasks that were imported by a dump
+                Ok(ModifiedTasks::DumpImported) => query.clone(),
```
dureuill: isn't that possibly quite a lot of tasks?

crates/index-scheduler/src/scheduler/mod.rs
```diff
@@ -127,10 +129,14 @@ impl Scheduler {
             dsr_fuel: _,
         } = options;
 
+        let (waker, wake_up) = tokio::sync::broadcast::channel(32);
+        // we want to start the loop right away in case meilisearch was ctrl+Ced while processing things
+        let _ = waker.send(ModifiedTasks::DumpImported).unwrap();
```
dureuill: I don't really get why we are resending all tasks on startup

crates/meilisearch/src/routes/tasks.rs
Kerollmops: Yes, but I need to send an event so the index scheduler can start processing tasks. Should I?

crates/meilisearch/src/routes/batches.rs
```diff
@@ -203,3 +210,112 @@ async fn get_batches(
...
+                match index_scheduler.get_batches_from_authorized_indexes(&query, &filters) {
+                    Ok((batches, _)) => batches,
+                    Err(e) => {
+                        tracing::error!("Impossible to get batches from authorized indexes: {e}");
+                        break 'listener;
+                    }
+                };
+
+            if batches.iter().any(|b| b.finished_at.is_some()) {
+                // If we find only finished batches let's wait
```
dureuill, line 300: find **only** finished batch, shouldn't the `any` be `all` then?

Kerollmops: In practice, there is only one processing batch at a time, but yeah, you're right! Changing that, my bad.

crates/index-scheduler/src/scheduler/mod.rs
Kerollmops: Let's discuss that this afternoon 👀

dureuill:
spooky label:

- changes the way we handle index scheduler ticks

crates/index-scheduler/src/lib.rs
dureuill: we can either unwrap with a justification or ignore the error, both will do

crates/index-scheduler/src/scheduler/mod.rs
```diff
@@ -24,7 +24,6 @@ mod test_failure;
 
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicU32, Ordering};
-use std::sync::Arc;
```
dureuill, line 27: great that we can remove the 🏹

crates/index-scheduler/src/scheduler/mod.rs
dureuill: nitpick, but I'm not a fan of naming this `DumpImported`, since this is not what happened here. Perhaps name around some initialization? `ModifiedTasks::StartProcessing` perhaps?

Not critical

crates/index-scheduler/src/lib.rs
```diff
@@ -482,26 +483,34 @@ impl IndexScheduler {
...
-                        Ok(Ok(TickOutcome::WaitForSignal)) => run.scheduler.wake_up.wait(),
+                        Ok(Ok(TickOutcome::WaitForSignal)) => {
+                            // no panic: this function is called in a freshly spawned thread and so
+                            // is not within an asynchronous execution context.
+                            match run.scheduler.wake_up.blocking_recv() {
+                                Ok(_) => (),
+                                Err(RecvError::Closed) => break,
+                                // The receiver lost messages as it was lagging behind the senders. We definitely
+                                // received a message in this case, so we will tick again. As the tick will catch
+                                // up with any newly received tasks, it is fair to skip older messages by resubscribing.
```
dureuill: we are no longer resubscribing in this case apparently 😅

dureuill (dismissed):
I have a few remaining nitpicks, feel free to land anyway

dureuill:
@Kerollmops  removing auto merge so you can review the nitpicks before deciding to merge

dureuill (approved):

dureuill:
ça part en prod

PR #6534, opened 2026-07-23
fix filter condition

dureuill:
Recently introduced filter conditions is broken on main due to a later commit that introduced a regression.

This PR fixes the filter conditions in DSRs and add an automated test to catch future regressions

## Generative AI tools


Kerollmops (approved):
Nice catch! Thanks 🙏

PR #6535, opened 2026-07-23
Fix a compilation issue with tests

Kerollmops:
Fix a compilation issue with tests and the new fields ids map changes introduced in #6528.


dureuill (approved):
👍

PR #6541, opened 2026-07-28
feat(search): add query memory guard for search condition evaluation

glatinone:
Introduce a thread-safe QueryMemoryGuard for search condition evaluation to prevent OOMs on deeply nested filters or massive OR-conjunctions.

- crates/milli/src/search/memory_guard.rs: atomic budget and estimators
- export from search::mod
- add UserError::VectorOrFilterBudgetExceeded variant

When unset, this is a no-op; returns a clear error instead of panicking.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


glatinone:
Closing this PR to avoid adding unnecessary review overhead for maintainers. Thank you for your time!

PR #6542, opened 2026-07-28
Speed up the search speed a bit more

Kerollmops:
This PR improves search speed slightly by preparing more items before formatting. It also improves the logging of search performances. The search speed goes from 48ms to 35ms in my tests.

## Generative AI tools

### DIFF
--- a/crates/meilisearch/src/search/federated/perform.rs
+++ b/crates/meilisearch/src/search/federated/perform.rs
@@ -1607,6 +1607,8 @@ impl SearchByIndex {
         let mut documents_seen = RoaringBitmap::new();
         let mut local_pinned_hits = Vec::new();
         for result_by_query in &mut results_by_query {
+            let _step = progress.update_progress_scoped(SearchStep::Format);
+
             let prev_documents_ids = std::mem::take(&mut result_by_query.documents_ids);
             let prev_scores = std::mem::take(&mut result_by_query.document_scores);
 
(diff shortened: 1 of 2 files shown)


crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -1663,7 +1665,10 @@ impl SearchByIndex {
                     }
 
                     let hit: Result<_, ResponseError> = (|| {
-                        let mut hit = hit_maker.make_hit(docid, &score, progress)?;
+                        let mut hit = {
+                            let _step = progress.update_progress_scoped(SearchStep::Format);
```
dureuill, line 1669: is this happening once per hit? I feel this could become significant?

crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -1607,14 +1607,16 @@ impl SearchByIndex {
         let mut documents_seen = RoaringBitmap::new();
         let mut local_pinned_hits = Vec::new();
         for result_by_query in &mut results_by_query {
+            let _step = progress.update_progress_scoped(SearchStep::Format);
```
dureuill, line 1610: is this happening once per query? I feel uneasy about having progress steps that are unbounded, at least while we don't have some fast time-tracking method

Kerollmops: Actually, they were already there. I just moved them out of the `make_hit` function and made them a bit more global than before to measure all the initialization of the `MakeHit` struct too.

dureuill: hmmm I'll accept it in the name of "it's not worse than before", but FTR not a fan (as long as looking at our watch is a costly operation)

dureuill (approved):
looks better than before, thank you

crates/meilisearch/src/search/federated/perform.rs
Kerollmops: Actually, it's even better than before as it is outside the make_hit calls loop now 😂

PR #6543, opened 2026-07-28
Fix duplicate pins in federated search

### LINKED ISSUE #6540: Duplicate pinned documents in federated search
**Describe the bug**
Pinned documents are not correctly deduplicated in federated search requests containing several queries hitting the same index.

**To Reproduce**
1. Enable the dsr experimental feature
2. Add a rule with an action to pin a document
3. Create a federated search request containing 2 queries that target the same index and firing the DSR
4. Observe the document being pinned twice.

**Expected behavior**
The document is only pinned once.

**Meilisearch version:**
The work on DSR as index did not modify the way pins are resolved, so this likely dates back from the original introduction of pinning in v1.41

dureuill:
## Related issue

Fixes #6540 

## API change

- Add `precedence` to `ScoreDetails::Pin` in ranking score details

## Implementation

- Introduce a trait to factor the behavior of deduplicating and sorting pins
- Use the trait to correctly deduplicate pins in federated search
- Fix existing tests
- Add tests:
    1. Checking that pins are no longer duplicated (checked that pins would be duplicated on main)
    2. Checking that the pin belonging to the rule with the earliest precedence wins

## Generative AI tools

### DIFF
--- a/crates/milli/src/dynamic_search_rules.rs
+++ b/crates/milli/src/dynamic_search_rules.rs
@@ -17,6 +17,7 @@ use crate::search::facet::ascending_facet_sort;
 use crate::search::facet::facet_range_search::find_docids_of_facet_within_bounds;
 use crate::search::facet::value_bounds::{evaluate_equal, to_str_bounds, ValueBounds};
 use crate::search::new::LocatedQueryTerm;
+use crate::search::{Pin, Precedence};
 use crate::update::new::document::DocumentFromDb;
 use crate::{
     AscDesc, DocumentId, FieldId, FieldsIdsMap, Index, IndexFilter, PinDoc, Result, SearchContext,
@@ -80,18 +81,17 @@ impl<'a> DynamicSearchRulesView<'a> {
         let active_rules =
             self.active_rules_for_query(query_terms, filter, search_context, fuel)?;
 
-        self.find_pins(self.rule_ids_sorted_by_precedence(active_rules)?, search_context, fuel)
-            .filter(
-                |pin| {
-                    if let Ok(pin) = pin.as_ref() {
-                        universe.remove(pin.doc_id)
-                    } else {
-                        true
-                    }
-                },
-            )
+        let pins: Result<Vec<PinDoc>> = self
+            .find_pins(self.rule_ids_sorted_by_precedence(active_rules)?, search_context, fuel)
+            .filter(|pin| if let Ok(pin) = pin.as_ref() { universe.remove(pin.id) } else { true })
             .take(fuel.max_pin_actions())
-            .collect()
+            .collect();
+
+        let mut pins = pins?;
+
+        Pin::dedup_and_sort(&mut pins);
+
+        Ok(pins)
     }
 
     pub fn rules_from_rule_ids<I>(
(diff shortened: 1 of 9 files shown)


crates/milli/src/dynamic_search_rules.rs
```diff
@@ -80,18 +81,17 @@ impl<'a> DynamicSearchRulesView<'a> {
...
-                    if let Ok(pin) = pin.as_ref() {
-                        universe.remove(pin.doc_id)
-                    } else {
-                        true
-                    }
-                },
-            )
+        let pins: Result<Vec<PinDoc>> = self
+            .find_pins(self.rule_ids_sorted_by_precedence(active_rules)?, search_context, fuel)
+            .filter(|pin| if let Ok(pin) = pin.as_ref() { universe.remove(pin.id) } else { true })
```
Kerollmops, line 86: Right ✅

crates/milli/src/search/mod.rs
```diff
@@ -51,8 +51,79 @@ pub struct SemanticSearch {
...
+
+impl Ord for Precedence {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        use std::cmp::Ordering::*;
+        match (self.0, other.0) {
+            (None, None) => Equal,
+            (None, Some(_)) => Greater,
+            (Some(_), None) => Less,
+            (Some(left), Some(right)) => left.cmp(&right),
+        }
```
Kerollmops, line 128: So, [you implement the reverse of the `<Option as Ord>::cmp`](https://doc.rust-lang.org/stable/src/core/option.rs.html#2465-2477)?

dureuill: Not exactly.

if we have `(Some, Some)`, it is the same as `Option::cmp`
It is only `None` which is handled as greater rather than less than `Some`

dureuill: Could be implemented as `Reverse<Option<Reverse<u64>>>`:
`None.cmp(Some)` yields `Greater`
`Some(left), Some(right)` yields `Reverse(Reverse(left.cmp(right))`

Not 100% on whether this is actually more readable though 😅

dureuill: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=78651201fbfecf919510c570d49c2c19

Kerollmops (changes requested):
Would you mind adding a comment, please?

crates/milli/src/search/mod.rs
```diff
@@ -51,8 +51,79 @@ pub struct SemanticSearch {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Precedence(pub Option<u64>);
+
+impl PartialOrd for Precedence {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for Precedence {
```
dureuill, line 120:
```suggestion
/// Overrides the natural order of Option such that None is considered greater than Some rather than Less
/// 
/// When both options are Some, use the regular order.
impl Ord for Precedence {
```

crates/milli/src/search/mod.rs
dureuill: added some comment

crates/milli/src/search/mod.rs
```diff
@@ -51,8 +51,82 @@ pub struct SemanticSearch {
...
+pub struct Precedence(pub Option<u64>);
+
+impl PartialOrd for Precedence {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+/// Overrides the natural order of Option such that None is considered greater than Some rather than Less
+///
```
dureuill:
```suggestion
///
```

Kerollmops (approved):
Nice, thanks for the fix!

PR #6544, opened 2026-07-29
Skip the settings when nothing changes

Kerollmops:
Small regression with the new settings indexer that made the engine compute and reindex a bunch of internal data structures even when nothing changes in the settings.

## Generative AI tools


dureuill (dismissed):

crates/milli/src/update/settings.rs
```diff
@@ -1629,6 +1631,16 @@ impl<'a, 't, 'i> Settings<'a, 't, 'i> {
             )
             .map(Some)
         } else {
+            update_index(
+                self.index,
+                self.wtxn,
+                inner_settings_diff.old.fields_ids_map.clone(),
```
dureuill: shouldn't that be `new` rather than `old`?

Kerollmops: Hum.. I suppose that would be better indeed 👀

dureuill (approved):
this time for good :-)

PR #6565, opened 2026-08-05
Fix code security issues in GitHub Actions workflows

curquiza:
## What this PR does

Fixes 53 high-severity SAST findings (Bastion security scan) in the CI/CD configuration:

### Pin all mutable action tags to full commit SHAs (49 findings, 16 workflow files)
Every `uses: action@tag` reference is now pinned to a full 40-character commit SHA with a trailing version comment (e.g. `actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6`). Tags and branches can be silently repointed by the action owner, enabling supply-chain attacks (as seen in the trivy-action and tj-actions compromises).

`dtolnay/rust-toolchain` also gets an explicit `toolchain:` input, since it normally infers the toolchain from the ref name — which is no longer possible when pinned to a SHA.

Dependabot already covers `github-actions` monthly, so pinned SHAs will keep being bumped automatically.

### Scope `MEILI_BOT_GH_PAT` to the steps that need it (2 findings)
In `update-cargo-toml-version.yml` and `bench-pr.yml`, `GH_TOKEN` was declared in the workflow-level `env:` block, exposing it to every step. It is now referenced only by the steps that actually use it.

### Remove dead `curl | sh` install path (1 finding)
The Go SDK test job still had the legacy `golang/dep` install (`curl … | sh`) behind `if [ -f Gopkg.toml ]`. meilisearch-go uses Go modules, so this branch never executed — removed.

### Add Dependabot cooldown (1 finding)
`cooldown: default-days: 7` on the `github-actions` ecosystem, so newly published action versions wait 7 days before update PRs (protects against compromised releases being pulled in immediately).

All 53 findings have been closed in Bastion with a reference to this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,8 @@ updates:
     directory: "/"
     schedule:
       interval: "monthly"
+    cooldown:
+      default-days: 7
     labels:
       - 'dependencies'
     rebase-strategy: disabled
--- a/.github/workflows/bench-manual.yml
+++ b/.github/workflows/bench-manual.yml
@@ -17,9 +17,10 @@ jobs:
     runs-on: benchmarks
     timeout-minutes: 180 # 3h
     steps:
-      - uses: actions/checkout@v6
-      - uses: dtolnay/rust-toolchain@1.91.1
+      - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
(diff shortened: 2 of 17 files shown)


Kerollmops (approved):
Looks clean but a lot to review! Merging as the CI passes.

Issue #6578, opened 2026-08-06
SearchQueue: Permit::drop(self) sends two release signals (explicit + implicit Drop), causing double capacity return and potential parallelism violation


gregormelhorn:
## Description

`Permit::drop(self)` sends one release signal, then Rust calls
`Drop for Permit` which sends a second signal. Every call to
`permit.drop().await` produces two signals. The scheduler grants
one permit per signal, so one logical release grants two permits.

## How to reproduce

A deterministic example:

1. Set `parallelism = 2`. Get two permits (all slots used).
2. Put two waiters in the queue.
3. Call `permit.drop().await` on one held permit.

The scheduler processes two signals:

- Signal 1: `searches_running` goes from 2 to 1. Waiter 1 gets a permit.
- Signal 2: `searches_running` goes from 1 to 0. Waiter 2 gets a permit.

After both signals, three searches are running. The configured limit
of 2 is exceeded. The admission gate at L154 sees 0 and lets another
request in.

## Root cause

`Permit::drop(self)` at L52-55 sends the first signal and uses
up `self`:

```rust
pub async fn drop(self) {
    let _ = self.sender.send(()).await;
}
```

After this method returns, `self` goes out of scope. Rust's `Drop`
trait fires. `Drop for Permit` at L58-67 clones the sender and
spawns a task that sends the second signal:

```rust
impl Drop for Permit {
    fn drop(&mut self) {
        let sender = self.sender.clone();
        std::mem::drop(tokio::spawn(async move { sender.send(()).await }));
    }
}
```

The release handler at L131-138 runs once per signal: it decrements
`searches_running` and grants a permit to the next waiter.

## Impact

The configured parallelism limit can be exceeded. One search finishing
grants permits to two waiters. The `saturating_sub` at L146 keeps
`searches_running` from going below zero, but does not stop the extra
grants.

Every route call site uses `permit.drop().await`:
`search.rs:600,666,900,967`, `multi_search.rs:195,307,379`,
`facet_search.rs:331`. The module doc recommends it.

## Expected behavior

A permit release must return exactly one unit of capacity. Options:

- The `Drop` impl checks a flag and skips the second signal when
  explicit drop already ran.
- The explicit `drop(self)` method prevents the implicit `Drop`
  from sending a signal.
- The implicit `Drop` is the only release path. The explicit method
  becomes a no-op wrapper or is removed.

## Environment

[...]

PR #6580, opened 2026-08-08
fix(search): keep searches_running accurate on refill grant (#6577)

### LINKED ISSUE #6577: SearchQueue: searches_running not re-incremented on refill grant, causing undercount and potential parallelism over-admission
## Description

When the scheduler grants a permit to a waiting request in the
release path, it does not increment `searches_running`. The
immediate-admission path increments; the refill path does not.
Over time, `searches_running` falls below the real count.

## Root cause

The immediate-admission path at L154-156 increments before
granting:

```rust
if searches_running < usize::from(parallelism) && queue.is_empty() {
    searches_running += 1;
    let _ = search_request.send(Permit { sender: sender.clone() });
    continue;
}
```

The refill path at L146-152 decrements on release, then grants
a permit to the next waiter — but never increments:

```rust
searches_running = searches_running.saturating_sub(1);
if !queue.is_empty() {
    let remove = rng.gen_range(0..queue.len());
    let channel = queue.swap_remove(remove);
    let _ = channel.send(Permit { sender: sender.clone() });
}
```

After N refill grants, `searches_running` is N below the real count.

## Impact

`searches_running()` undercounts live searches. The admission gate
at L154 reads the wrong value and can let in more requests than
configured. Combined with the duplicate release signals from
`Permit::drop(self)` + `Drop for Permit` (#6578), the error grows:
two signals → two refill grants without increment → gate sees 0 →
admits yet another request.

## Expected behavior

Every grant must increment `searches_running` by 1:

```rust
searches_running = searches_running.saturating_sub(1);
if !queue.is_empty() {
    let remove = rng.gen_range(0..queue.len());
    let channel = queue.swap_remove(remove);
    let _ = channel.send(Permit { sender: sender.clone() });
    searches_running += 1;  // ← add this line
}
```

## Environment

- Commit: `fff2ef5a42658b16a937d922aabc3fb7f89f2018`
- File: `crates/meilisearch/src/search_queue.rs`

abdulwaarith0:
fix(search): keep searches_running accurate on refill grant (#6577)

Fixes #6577.

In `crates/meilisearch/src/search_queue.rs`, when a permit is released the scheduler decrements `searches_running` and then refills the freed slot by granting it to a queued request. That granted request is a running search, but the refill path never re-incremented the counter, while the immediate-admission path does.

Over repeated release/refill cycles the counter drifts below the real number of running searches, so the admission gate undercounts and can admit more concurrent searches than the configured parallelism. (This compounds with the `Permit` double-release in #6578, which I'm handling separately.)

Fix: increment `searches_running` on the refill grant, same as the immediate-admission path.

Test: `searches_running_stays_accurate_after_refill` occupies the single slot at `parallelism = 1`, queues a second request, releases the first via `Drop`, and checks `searches_running()` stays at 1 after the refill. Fails before the change, passes after.

Scoped to #6577; #6578 will follow separately.

### DIFF
--- a/crates/meilisearch/src/search_queue.rs
+++ b/crates/meilisearch/src/search_queue.rs
@@ -135,6 +135,11 @@ impl SearchQueue {
                         let remove = rng.gen_range(0..queue.len());
                         let channel = queue.swap_remove(remove);
                         let _ = channel.send(Permit { sender: sender.clone() });
+                        // Granting a permit to a queued search starts a new running search,
+                        // so we must account for it here just like the immediate-admission path
+                        // below does. Otherwise `searches_running` drifts below the real count
+                        // and the admission gate over-admits.
+                        searches_running += 1;
                     }
                 },
 
--- a/crates/meilisearch/tests/search/search_queue.rs
+++ b/crates/meilisearch/tests/search/search_queue.rs
@@ -219,3 +219,41 @@ async fn works_with_capacity_of_zero() {
         .expect("I should get a permit straight away")
         .unwrap();
 }
+
+// Regression test for https://github.com/meilisearch/meilisearch/issues/6577
+// When a permit is released and the freed slot is refilled by granting it to a
+// queued request, that queued request becomes a running search. The refill path
+// must therefore keep `searches_running` accurate, exactly like the
+// immediate-admission path does. Before the fix it decremented on release but
+// never re-incremented on the refill grant, so the counter drifted below the
+// real number of running searches and the admission gate over-admitted.
+#[actix_rt::test]
+async fn searches_running_stays_accurate_after_refill() {
+    // parallelism = 1, so at most one search may run at a time.
+    let queue = Arc::new(SearchQueue::new(4, NonZeroUsize::new(1).unwrap()));
+
+    // Occupy the single slot.
+    let permit1 = queue.try_get_search_permit().await.unwrap();
+
+    // Queue a second request; with the slot busy it has to wait in the queue.
+    let q = queue.clone();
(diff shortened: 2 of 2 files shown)


abdulwaarith0:
Gentle nudge 🙏 — CI here is waiting on maintainer approval to run (the workflows show as `action_required` since this is my first PR to the repo). Whenever a maintainer has a moment to approve the workflow run and take a look, I'd appreciate it. Happy to address anything that comes up. Thanks!

ManyTheFish:
Hello, can you provide more information about the usage of an AI agent, please? If you used one, and for what?

Thanks!

abdulwaarith0:
Hi @ManyTheFish, sorry for not putting this in the description upfront, that's on me.

Yes, I used Claude Code on this one. I used it while reading through the scheduler loop in `search_queue.rs` to check my understanding of where the count drifts, and to help draft the regression test and write up the PR description. The change itself is the single `searches_running += 1` in the refill arm.

I do own it and I checked it rather than taking it on faith. I traced the release/refill path to confirm the counter is only re-incremented on the immediate-admission path and never when a freed slot gets handed to a queued request. I ran it locally with `cargo test -p meilisearch --test integration search_queue` (8 passed, fmt clean), and the new test fails without the one-line change and passes with it.

I also kept it to #6577 only. I ran into the separate `Permit` double-release behaviour while I was in there and deliberately left it out, which is #6578. That's also why the test drops the permit implicitly rather than calling `Permit::drop`, since the explicit path sends two release signals and would muddy what this test is actually checking. I followed the `#[actix_rt::test]` pattern already used in that file instead of introducing a different style.

Happy to walk through any of the reasoning, or to reshape the test if you'd prefer it done differently.

PR #6586, opened 2026-08-13
Refactor the progress object

ManyTheFish:
## Related issue

Related to #https://linear.app/meilisearch/issue/EXP-1326

## Generative AI tools

## Changelog

The ProgressTrace related to the performance details display has been reshaped.

1) The Progress object can now be built in a FAST or PRECISE mode. This mode changes how we compute time: CPU-based vs clock time.
2) The Step can now be created with a verbose level; this will allow us to increase the verbosity level of some steps that impact the process without having to remove them completely.
3) The Progress object is now built with a verbosity level; it will ignore all the steps where the verbosity level is higher. A quiet mode has been introduced for test and code parts where the performance trace is not used.
4) Some functions like `try_get_search_permit` now take a Progress.

Future steps:
- Add more steps under the relevant verbosity level
- Add a way to change the verbosity level dynamically (param? header?)


ManyTheFish:
closing in favor of #6631 and #6632

PR #6592, opened 2026-08-18
Constrain phrase search across stop word gaps

### LINKED ISSUE #5298: Phrase search doesn't work when it contains stop_words
**Describe the bug**
Searching a query containing a phrase search like `"the soup of the day"` when having stop words in the settings like `["of", "the"]` will never return any document, even if some documents exactly match the query.

**To Reproduce**
See the below discussion:
https://github.com/orgs/meilisearch/discussions/793#discussioncomment-11981864

**Meilisearch version:**
v.1.12.7

**Potential Fix**
See the below comment containing a part of the solution:
https://github.com/orgs/meilisearch/discussions/793#discussioncomment-11967148

serhiy-bzhezytskyy:
## Related issue

Fixes #5298

## Generative AI tools

    - *Claude Code — used for the investigation (reproduction, instrumentation of `compute_phrase_docids`, the measurements below), the implementation, and the tests. I own the contribution and verified every claim in this description against local runs.*

## What does this PR do?

With stop words configured, a phrase query degrades into a bare word intersection: with `stopWords: ["of", "the"]`, the phrase `"the soup of the day"` matches a document titled `soup day`.

`compute_phrase_docids` keeps stop words as `None` holes (positions preserved) but builds its proximity windows over the raw sequence, so a window of size 3 never contains two real words once two or more stop words separate them — the phrase gets no proximity constraint at all.

This PR:
- builds the windows over the real words, keeping their raw positions, so pairs are constrained by their true distance — the same "up to the gap" rule the `dist > 0` branch already applies (so `soup day` matching a shorter gap stays, as in #5491's test expectations);
- skips pairs wider than the indexed `MAX_DISTANCE` (unconstrainable; behavior unchanged there, covered by a test);
- adds three integration tests — the bare-intersection one fails on `main`; the other two cover the unconstrainable-pair path and the exactness ranking rule resolving a stop-word phrase.

No indexing change is needed: the index already preserves the gaps (a phrase `"soup day"` does not match `The soup of the day`), so the indexing-time concern from #5491 does not apply here. Median query time is unchanged (5 ms on both trees; 200 runs, 5k documents). I ran the full integration suite on both trees with `--no-default-features`: identical failure sets; I could not run the default-features leg locally, relying on CI for it.

## Requirements

- [ ] ⚠️ If there is any change in the DB: — *no DB change: the fix is entirely on the search path*
- [ ] If necessary, the feature have been tested in the Cloud production environment…
- [ ] If necessary, the documentation related to the implemented feature in the PR is ready.
- [ ] If necessary, the integrations related to the implemented feature in the PR are ready.

### DIFF
--- a/crates/meilisearch/tests/search/mod.rs
+++ b/crates/meilisearch/tests/search/mod.rs
@@ -2360,3 +2360,97 @@ async fn test_skip_on_creation() {
     let (resp, code) = index.get_document(2, None).await;
     assert_eq!(code, 404, "{resp}");
 }
+
+#[actix_rt::test]
+async fn phrase_search_with_many_consecutive_stop_words_still_returns_documents() {
+    // Consecutive stop words can push a word pair beyond the maximum indexed
+    // proximity; such a pair cannot be constrained and must not empty the
+    // results (the words-level intersection still applies).
+    let documents = json!([
+        {"id": 1, "title": "soup as well as of the day"},
+        {"id": 2, "title": "unrelated document"}
+    ]);
+
+    test_settings_documents_indexing_swapping_and_search(
+        &documents,
+        &json!({"stopWords": ["as", "well", "of", "the"]}),
+        &json!({"q": "\"soup as well as of the day\"", "attributesToRetrieve": ["id"]}),
+        |response, code| {
+            assert_eq!(code, 200, "{}", response);
+            snapshot!(json_string!(response["hits"]), @r###"
+            [
+              {
+                "id": 1
+              }
+            ]
+            "###);
+        },
+    )
+    .await;
+}
+
+#[actix_rt::test]
+async fn phrase_search_with_stop_words_is_not_a_bare_word_intersection() {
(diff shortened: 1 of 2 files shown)


serhiy-bzhezytskyy:
@ManyTheFish this fixes #5298 by constraining phrase search across stop-word gaps, following the semantics from your #5491 draft (its test expectations hold here too). Branch is now current with main. CI hasn't started on this PR yet -- could you launch it?

Kerollmops:
Hello @serhiy-bzhezytskyy 👋

Thanks for your contribution; However, the AGENTS.md is clear about generated code, and we do not accept contributions made by LLMs.

Thanks for your help anyway 🌵

serhiy-bzhezytskyy:
@Kerollmops thanks. sorry for noise. i tried to bring a value only. and LLM is just tool likes IDE for me with 25 years experience in this industry. See you in other repos and PRs :)

PR #6595, opened 2026-08-19
Make HitMaker TokenizerBuilder stop-words aware

### LINKED ISSUE #6594: Stop words get matched(and highlighted) when a differently-cased variant is indexed elsewhere
**Problem**

When you set stop words e.g. ["The", "the"], it is expected to be case-sensitive and it definitely is case-sensitive but if the document also contains a different cased variant of the stop word e.g. "THE", re-indexing of the document now adds the token(lemma) "the". The impact of this is that if your prefix search contains any case variant of the stop word, it will match(and highlight) the stop words(the actual variant you had set i.e. "The" or "the" here) too

**Steps to Reproduce**

1. Create an index with this data:
```json
[
  {
    "id": 1,
    "title": "The City",
  },
  {
    "id": 2,
    "title": "This is Their The City",
  },
  {
    "id": 3,
    "title": "New City",
  },
  {
    "id": 4,
    "title": "THE City",
  }
]
```

```curl
➜  meilisearch git:(main) ✗ curl -X POST 'http://localhost:7700/indexes/cities/documents' \
  --data-binary '[{"id":1,"title":"The City"}, {"id":2,"title":"This is Their The City"}, {"id":3,"title":"New City"}, {"id":4,"title":"THE City"}]'
```
2. Set stop words = `["The", "the"]`
```curl
➜  meilisearch git:(main) ✗ curl -X PUT 'http://localhost:7700/indexes/cities/settings/stop-words' \
  --data '["The", "the"]'
```
3. Search for "the"(note that search is case-insensitive) and watch the output:
```curl
➜  meilisearch git:(main) ✗ curl -s "$MEILI_URL/indexes/movies231/search" \
  --compressed \
  -X POST \
  -H 'content-type: application/json' \
  --data-raw '{"q":"the","attributesToHighlight":["*"],"limit":20}' | jq
{
  "hits": [
    {
      "id": 4,
      "title": "THE City",
      "_formatted": {
        "id": "4",
        "title": "<em>THE</em> City"
      }
    },
    {
      "id": 2,
      "title": "This is Their The City",
      "_formatted": {
        "id": "2",
        "title": "This is <em>The</em>ir <em>The</em> City"
      }
    }
  ],
  "query": "the",
  "processingTimeMs": 10,
  "limit": 20,
  "offset": 0,
  "estimatedTotalHits": 2,
  "requestUid": "01a015c2-23af-7822-8fa3-a45ffa5c90ea"
}
```
In this output, the document with id=2 has "The" highlighted as well but "The" is a stop word

**Expected behavior**
I fixed the issue locally and after the fix, i get the expected output from search(or multi-search) API. Here is the expected output:
```curl
➜  meilisearch git:(main) ✗ curl -s "$MEILI_URL/indexes/movies231/search" \
  --compressed \
  -X POST \
  -H 'content-type: application/json' \
  --data-raw '{"q":"the","attributesToHighlight":["*"],"limit":20}' | jq
{
  "hits": [
    {
      "id": 4,
      "title": "THE City",
      "_formatted": {
        "id": "4",
        "title": "<em>THE</em> City"
      }
    },
    {
      "id": 2,
      "title": "This is Their The City",
      "_formatted": {
        "id": "2",
        "title": "This is <em>The</em>ir The City"
      }
    }
  ],
  "query": "the",
  "processingTimeMs": 57,
  "limit": 20,
  "offset": 0,
  "estimatedTotalHits": 2,
  "requestUid": "01a015c8-909f-75e1-9c4f-4d43ec8ad2e3"
}
```

**Screenshots**
Current behaviour:

Image

Expected behaviour:

Image

**Meilisearch version:**
[e.g. v1.53.1]

**Additional context**
There are several related issues e.g. #5851 #4984 (apparently closed but seems not fixed)

harismuzaffer:
## Description

Hi. I have been debugging this issue since a few days and finally came up with this PR. Open to suggestions 

This PR fixes the behavior of stop words when prefix searching. The exact issue arises when your document contains different case variants of the stop word you have set.

## Related issue

Fixes #6594 

## What does this PR do?

It makes the `HitMaker` `TokenizerBuilder` stop-words aware i.e. the stop-words of the index are now passed to the tokenizer builder and the `tokenizer` now tokenizes so that the stop words are considered as `StopWord` and not `Word`. This allows `match_unique_words` to skip matching a stop word. I also added a test case for this change

The second change that i have done is reordering the match branches in `match_unique_words` so that exact-match check runs before the prefix-match check. In my opinion this is most natural. I came across this while debugging and i found that "THE" is evaluated as a prefix match rather than an exact match which is wrong. Because of this change, two insta snapshots were effected which i updated. However i am  unsure if this is breaching the scope of this PR. Happy to create a separate PR if that is what you prefer

## Generative AI tools

    - *Claude was used to assist in gathering knowledge of different crates but the code was not written by it*

## Requirements

    - [ ] Test that any impacted DB still works as expected after using `--upgrade-db` on a DB created with the last released Meilisearch

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4370,6 +4370,7 @@ dependencies = [
  "either",
  "file-store",
  "flate2",
+ "fst",
  "futures",
  "futures-util",
  "hex",
--- a/crates/meilisearch/Cargo.toml
+++ b/crates/meilisearch/Cargo.toml
@@ -111,6 +111,7 @@ humantime = { version = "2.3.0", default-features = false }
 cidr = { version = "0.3.2", features = ["serde"] }
 routes = { path = "../routes" }
 bumparaw-collections = "0.1.7"
+fst = "0.4.7"
 
 [dev-dependencies]
 actix-rt = "2.11.0"
--- a/crates/meilisearch/src/search/federated/perform.rs
+++ b/crates/meilisearch/src/search/federated/perform.rs
@@ -1360,6 +1360,7 @@ impl SearchByIndex {
         let separators = index.allowed_separators(&rtxn).without_index()?;
         let separators: Option<Vec<_>> =
             separators.as_ref().map(|x| x.iter().map(String::as_str).collect());
+        let stop_words = index.stop_words(&rtxn).without_index()?;
 
         let max_total_hits = index
             .pagination_max_total_hits(&rtxn)
@@ -1583,7 +1584,11 @@ impl SearchByIndex {
                 degraded |= query_degraded;
                 used_negative_operator |= query_used_negative_operator;
 
-                let tokenizer = HitMaker::tokenizer(dictionary.as_deref(), separators.as_deref());
+                let tokenizer = HitMaker::tokenizer(
+                    dictionary.as_deref(),
+                    separators.as_deref(),
+                    stop_words.as_ref(),
+                );
 
                 let formatter_builder = HitMaker::formatter_builder(matching_words, tokenizer);
 
--- a/crates/meilisearch/src/search/mod.rs
+++ b/crates/meilisearch/src/search/mod.rs
@@ -2242,8 +2242,9 @@ impl<'a> HitMaker<'a> {
     pub fn tokenizer<'b>(
         dictionary: Option<&'b [&'b str]>,
         separators: Option<&'b [&'b str]>,
+        stop_words: Option<&'b fst::Set<&'b [u8]>>,
     ) -> milli::tokenizer::Tokenizer<'b> {
-        let mut tokenizer_builder = TokenizerBuilder::default();
+        let mut tokenizer_builder = TokenizerBuilder::<&[u8]>::new();
         tokenizer_builder.create_char_map(true);
 
         if let Some(separators) = separators {
@@ -2252,6 +2253,10 @@ impl<'a> HitMaker<'a> {
 
         if let Some(dictionary) = dictionary {
             tokenizer_builder.words_dict(dictionary);
+        }
+
+        if let Some(stop_words) = stop_words {
+            tokenizer_builder.stop_words(stop_words);
         }
 
         tokenizer_builder.into_tokenizer()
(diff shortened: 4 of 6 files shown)


crates/meilisearch/tests/settings/tokenizer_customization.rs
```diff
@@ -172,18 +172,18 @@ async fn set_and_search() {
...
-            snapshot!(json_string!(response["hits"]), @r###"
+            snapshot!(json_string!(response["hits"]), @r#"
             [
               {
                 "id": 2,
                 "content": "G#D#G#D#G#C#D#G#C#",
                 "_formatted": {
                   "id": "2",
-                  "content": "<em>G</em>#<em>D#</em><em>G</em>#<em>D#</em><em>G</em>#<em>C#</em><em>D#</em><em>G</em>#<em>C#</em>"
+                  "content": "<em>G#</em><em>D#</em><em>G#</em><em>D#</em><em>G#</em><em>C#</em><em>D#</em><em>G#</em><em>C#</em>"
```
ManyTheFish: seems that prefix highlight is wrong

crates/meilisearch/tests/settings/tokenizer_customization.rs
```diff
@@ -457,11 +457,11 @@ async fn advanced_synergies() {
                 "content": "J. R. R. Tolkien",
                 "_formatted": {
                   "id": "2",
-                  "content": "<em>J. R.</em> R. Tolkien"
+                  "content": "<em>J. R. R.</em> Tolkien"
```
ManyTheFish: seems that prefix highlight is wrong

ManyTheFish (changes requested):
There is a small change in the test that I can't accept related to the prefix highlight.

That said, thank you for your contribution to the repository.
Happy to see some human-made contributions in the middle of this AI era. 🥵
Your usage of AI to understand the code is completely fine for us.

Let me know if you find a fix.

crates/meilisearch/tests/settings/tokenizer_customization.rs
harismuzaffer: Let me check this and the other one(line ~460)

harismuzaffer: The highlighting is indeed wrong after reordering the match branch. I revisited the `match_unique_words` and the loop over the candidate words must evaluate `starts_with` before the exact match branch otherwise a prefix match highlights whole token whenever the candidate equals it.

The reason i had reversed the match branch was that an exact-match query was getting evaluated in the prefix-match branch but highlighting itself remained unaffected.

Let me remove this commit and run tests once more. Let me know if you are fine with this

ManyTheFish (approved):
Approved!

Thank you for this bug fix! 🚀

PR #6599, opened 2026-08-24
fix(search_queue): prune cancelled waiters to prevent spurious TooManySearchRequests

### LINKED ISSUE #6508: Cancelled search requests remain in SearchQueue and can cause spurious TooManySearchRequests
### Describe the bug

`SearchQueue::try_get_search_permit` transfers a `oneshot::Sender<Permit>` to the search scheduler and then awaits the corresponding receiver. If the request future is cancelled after the scheduler has accepted that sender but before a permit is delivered, cancellation drops the receiver while the scheduler continues to retain the sender in its bounded `Vec` queue.

The scheduler does not prune closed oneshot senders before using `queue.len()` for capacity enforcement, random eviction, waiter selection, or the `searches_waiting` metric. A request that no longer exists can therefore continue to consume a queue slot and participate in admission decisions.

The vulnerable transition is:

```text
request creates (Sender<Permit>, Receiver<Permit>)
    -> Sender is transferred to SearchQueue::run
    -> scheduler stores Sender in queue
    -> request is cancelled while awaiting Receiver<Permit>
    -> Receiver is dropped
    -> scheduler retains the now-closed Sender
    -> stale entry remains counted in queue.len()
```

This can produce user-visible effects when the queue is under load:

- a live request can be randomly evicted and receive `TooManySearchRequests` while a cancelled waiter remains in the queue;
- a released permit can first be delivered to a closed receiver, adding an unnecessary scheduling turn before live work proceeds;
- `searches_waiting` can report cancelled requests as live pending work.

For example, with `parallelism = 1` and `capacity = 2`, request A can hold the running permit, cancelled request B can remain as a stale waiter, and live request C can occupy the second apparent slot. When request D arrives, the scheduler sees a full queue and randomly removes either B or C. If it removes C, a live request is rejected while the cancelled B remains admitted.

All source links below are pinned to commit `fff2ef5a42658b16a937d922aabc3fb7f89f2018`.

1. `try_get_search_permit` sends the waiter ticket to the scheduler before awaiting the receiver:
   [`crates/meilisearch/src/search_queue.rs#L177-L183`](https://github.com/meilisearch/meilisearch/blob/fff2ef5a42658b16a937d922aabc3fb7f89f2018/crates/meilisearch/src/search_queue.rs#L177-L183)

```rust
let (sender, receiver) = oneshot::channel();
self.sender.send(sender).await.map_err(|_| MeilisearchHttpError::SearchLimiterIsDown)?;
let permit = receiver
    .await
    .map_err(|_| MeilisearchHttpError::TooManySearchRequests(self.capacity))?;
```

Once `self.sender.send(sender).await` completes, the scheduler independently owns the sender. Cancelling the caller at or before `receiver.await` drops only the receiver; there is no cancellation guard that removes the transferred sender.

2. The scheduler retains these senders in a plain `Vec` and uses its raw length for capacity and metrics:
   [`crates/meilisearch/src/search_queue.rs#L121-L171`](https://github.com/meilisearch/meilisearch/blob/fff2ef5a42658b16a937d922aabc3fb7f89f2018/crates/meilisearch/src/search_queue.rs#L121-L171)

```rust
let mut queue: Vec<oneshot::Sender<Permit>> = Default::default();

// ...

} else if queue.len() >= capacity {
    let remove = rng.gen_range(0..queue.len());
    let thing = queue.swap_remove(remove);
    drop(thing);
}
queue.push(search_request);

// ...

metric_searches_waiting.store(queue.len(), Ordering::Relaxed);
```

There is no `Sender::is_closed()` check or other cancellation-aware removal before `queue.len()` controls admission.

3. Permit delivery also selects directly from the same population without skipping closed receivers in the same scheduler turn:
   [`crates/meilisearch/src/search_queue.rs#L131-L138`](https://github.com/meilisearch/meilisearch/blob/fff2ef5a42658b16a937d922aabc3fb7f89f2018/crates/meilisearch/src/search_queue.rs#L131-L138)

```rust
_ = search_finished.recv() => {
    searches_running = searches_running.saturating_sub(1);
    if !queue.is_empty() {
        let remove = rng.gen_range(0..queue.len());
        let channel = queue.swap_remove(remove);
        let _ = channel.send(Permit { sender: sender.clone() });
    }
},
```

If the selected receiver was cancelled, `send` fails. The stale entry has already affected queue length, eviction, selection, and metrics before that failure reveals it.

4. This wait occurs directly in HTTP request paths, including normal search, multi-search, and facet search:

- [`crates/meilisearch/src/routes/indexes/search.rs#L543-L559`](https://github.com/meilisearch/meilisearch/blob/fff2ef5a42658b16a937d922aabc3fb7f89f2018/crates/meilisearch/src/routes/indexes/search.rs#L543-L559)
- [`crates/meilisearch/src/routes/multi_search.rs#L157-L196`](https://github.com/meilisearch/meilisearch/blob/fff2ef5a42658b16a937d922aabc3fb7f89f2018/crates/meilisearch/src/routes/multi_search.rs#L157-L196)
- [`crates/meilisearch/src/routes/indexes/facet_search.rs#L292-L336`](https://github.com/meilisearch/meilisearch/blob/fff2ef5a42658b16a937d922aabc3fb7f89f2018/crates/meilisearch/src/routes/indexes/facet_search.rs#L292-L336)

A disconnected client, upstream timeout, or cancelled request handler can therefore reach this window while waiting for search admission.

### To Reproduce

The following focused whitebox test intentionally asserts the current bad state. Therefore, the test passing means the stale-waiter state is reachable. It does not use sleeps or depend on the scheduler's random eviction choice.

Steps:

1. Append the test-only module below to `crates/meilisearch/src/search_queue.rs`.
2. Run the targeted unit-test command.
3. Observe that the task waiting for a permit is successfully cancelled, but `searches_waiting()` remains `1` afterward.

<details>
<summary>Full whitebox test code</summary>

```rust
#[cfg(test)]
mod cancellation_tests {
    use super::*;

    #[tokio::test(flavor = "current_thread")]
    async fn cancelled_permit_waiter_remains_counted_in_search_queue() {
        let queue = Arc::new(SearchQueue::new(1, NonZeroUsize::new(1).unwrap()));

        // Saturate the only running-search slot so the next request must wait.
        let active_permit = queue
            .try_get_search_permit()
            .await
            .expect("the first request should receive a permit");
        assert_eq!(queue.searches_waiting(), 0);

        let waiting_queue = Arc::clone(&queue);
        let waiting_task = tokio::spawn(async move {
            waiting_queue.try_get_search_permit().await
        });

        // This is the acceptance point: the scheduler has stored the sender in
        // its private queue and published the corresponding queue length.
        tokio::time::timeout(Duration::from_secs(1), async {
            loop {
                if queue.searches_waiting() == 1 {
                    break;
                }
                tokio::task::yield_now().await;
            }
        })
        .await
        .expect("the waiter should be accepted by the scheduler");
        assert!(
            !waiting_task.is_finished(),
            "the accepted waiter should still be awaiting its permit"
        );

        // The request future is cancelled while it awaits the permit. This
        // drops its receiver, but the scheduler already owns the sender.
        waiting_task.abort();
        let cancellation = waiting_task
            .await
            .expect_err("aborting the waiter should cancel its task");
        assert!(cancellation.is_cancelled());

        // Bug-existence assertion: dropping the receiver does not remove the
        // sender from the scheduler queue, so the cancelled waiter remains a
        // counted capacity slot after cancellation.
        assert_eq!(queue.searches_waiting(), 1);

        drop(active_permit);
    }
}
```

</details>

Targeted command, run from the repository root:

```bash
cargo +1.91.1 test -p meilisearch --lib cancelled_permit_waiter_remains_counted_in_search_queue -- --nocapture
```

Observed result:

```text
Finished `test` profile [unoptimized + debuginfo] target(s) in 4m 04s
     Running unittests src/lib.rs (target/debug/deps/meilisearch-1dba1b807c17282b)

running 1 test
test search_queue::cancellation_tests::cancelled_permit_waiter_remains_counted_in_search_queue ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 14 filtered out; finished in 0.00s
```

Why a passing test reproduces the bug:

1. The first permit saturates the only running-search slot, establishing that the queue is initially healthy and empty.
2. The test waits until `searches_waiting() == 1`. In the current implementation this metric is published from `queue.len()`, so this confirms that the scheduler has received and stored the second request's sender.
3. `!waiting_task.is_finished()` confirms that the corresponding caller is still pending on permit delivery rather than having completed normally.
4. `waiting_task.abort()` cancels that real request future, and `JoinError::is_cancelled()` confirms the cancellation edge fired. Dropping the task also drops its oneshot receiver.
5. The final `searches_waiting() == 1` is the violated property: the request future no longer exists, but its transferred sender remains represented as a pending queue slot.

The timeout only guards against a deadlocked test. The transition itself is coordinated by the scheduler's published queue length, not by wall-clock sleeping or probabilistic scheduling.

### Expected behavior

Once a caller waiting for a permit is cancelled, its queue entry should no longer participate in live admission state. In particular:

- closed/cancelled waiters should be removed or ignored before `queue.len()` is used for capacity enforcement or eviction;
- `searches_waiting` should count only live waiters;
- permit delivery should skip closed receivers and continue selecting live waiters without making stale tickets compete with them;
- a live request should not receive `TooManySearchRequests` merely because a queue slot is occupied by a cancelled request.

A useful invariant is:

```text
queue capacity, eviction, metrics, and permit selection operate only on live waiters
```

### Meilisearch version

- Workspace version: `1.50.0`
- Audited and reproduced commit: `fff2ef5a42658b16a937d922aabc3fb7f89f2018`
- Branch: `main`

### Additional context

This is a cancellation-correctness issue in the admission protocol, not an issue with search execution after a permit has already been obtained. The normal later `permit.drop().await` path cannot clean up a request cancelled before permit delivery because that request never receives a `Permit`.

It is also separate from any potential double-notification concern involving `Permit::drop(self)` and `Drop for Permit`. This report requires neither obtaining a permit nor explicitly releasing one for the cancelled waiter. The defect is specifically:

```text
cancel before permit delivery
    -> receiver disappears
    -> scheduler retains the transferred wait ticket
```

A possible repair direction is to prune `oneshot::Sender`s whose receivers are closed before capacity and eviction decisions, and to skip closed senders while dispatching permits. The exact fix should preserve a stronger invariant across every scheduler turn: stale entries must not affect capacity, fairness, metrics, or liveness, even if they are retained briefly as implementation details.

The whitebox reproduction was run as a Rust unit test on Linux and requires no running Meilisearch server, index, client, or external service.

mmustafasenoglu:
Fixes #6508

## Summary
When a search request is cancelled while waiting for a permit, its oneshot receiver is dropped but the scheduler retains the sender in its queue. This caused stale waiters to be counted in queue.len().

## Changes
- Added prune_closed_senders helper in crates/meilisearch/src/search_queue.rs
- Called before selecting next waiter, before capacity checks, and before publishing metrics
- Added test cancelled_permit_waiter_removed_from_search_queue

## Testing
- New test verifies cancelled waiter no longer occupies queue slot
- Existing tests should still pass

### DIFF
--- a/crates/meilisearch/src/search_queue.rs
+++ b/crates/meilisearch/src/search_queue.rs
@@ -107,6 +107,12 @@ impl SearchQueue {
         self.searches_waiting_to_be_processed.load(Ordering::Relaxed)
     }
 
+    /// Remove closed senders from the queue.
+    /// A sender is closed when its receiver has been dropped (e.g., request cancelled).
+    fn prune_closed_senders(queue: &mut Vec<oneshot::Sender<Permit>>) {
+        queue.retain(|sender| !sender.is_closed());
+    }
+
     /// This function is the main loop, it's in charge on scheduling which search request should execute first and
     /// how many should executes at the same time.
     ///
@@ -130,6 +136,8 @@ impl SearchQueue {
                 biased;
                 _ = search_finished.recv() => {
                     searches_running = searches_running.saturating_sub(1);
+                    // Prune closed senders before selecting the next waiter
+                    Self::prune_closed_senders(&mut queue);
                     if !queue.is_empty() {
                         // Can't panic: the queue wasn't empty thus the range isn't empty.
                         let remove = rng.gen_range(0..queue.len());
@@ -146,6 +154,9 @@ impl SearchQueue {
                         None => continue,
                     };
 
+                    // Prune closed senders before making capacity decisions
+                    Self::prune_closed_senders(&mut queue);
+
                     if searches_running < usize::from(parallelism) && queue.is_empty() {
                         searches_running += 1;
                         // if the search requests die, it's not a hard error on our side
@@ -159,6 +170,7 @@ impl SearchQueue {
                         continue;
 
                     } else if queue.len() >= capacity {
+                        // Can't panic: queue.len() >= capacity > 0, so range is not empty.
                         let remove = rng.gen_range(0..queue.len());
                         let thing = queue.swap_remove(remove); // this will drop the channel and notify the search that it won't be processed
                         drop(thing);
@@ -166,6 +178,9 @@ impl SearchQueue {
                     queue.push(search_request);
                 },
             }
+
+            // Prune closed senders before publishing metrics
+            Self::prune_closed_senders(&mut queue);
 
             metric_searches_running.store(searches_running, Ordering::Relaxed);
             metric_searches_waiting.store(queue.len(), Ordering::Relaxed);
--- a/crates/meilisearch/tests/search/search_queue.rs
+++ b/crates/meilisearch/tests/search/search_queue.rs
@@ -219,3 +219,59 @@ async fn works_with_capacity_of_zero() {
         .expect("I should get a permit straight away")
         .unwrap();
 }
+
+#[actix_rt::test]
+async fn cancelled_permit_waiter_removed_from_search_queue() {
+    let queue = Arc::new(SearchQueue::new(1, NonZeroUsize::new(1).unwrap()));
+
+    // Saturate the only running-search slot so the next request must wait.
+    let active_permit = queue
+        .try_get_search_permit()
+        .await
+        .expect("the first request should receive a permit");
+    assert_eq!(queue.searches_waiting(), 0);
+
+    let waiting_queue = Arc::clone(&queue);
+    let waiting_task = tokio::spawn(async move {
+        waiting_queue.try_get_search_permit().await
+    });
+
+    // Wait until the scheduler has accepted the waiter and published the queue length.
+    tokio::time::timeout(Duration::from_secs(1), async {
+        loop {
+            if queue.searches_waiting() == 1 {
+                break;
+            }
+            tokio::task::yield_now().await;
+        }
+    })
+    .await
+    .expect("the waiter should be accepted by the scheduler");
+    assert!(
+        !waiting_task.is_finished(),
+        "the accepted waiter should still be awaiting its permit"
+    );
+
+    // The request future is cancelled while it awaits the permit.
+    // This drops its receiver, but the scheduler already owns the sender.
+    waiting_task.abort();
+    let cancellation = waiting_task
+        .await
+        .expect_err("aborting the waiter should cancel its task");
+    assert!(cancellation.is_cancelled());
+
+    // With the fix, the cancelled waiter's sender should be pruned from the queue.
+    // searches_waiting should return to 0.
+    tokio::time::timeout(Duration::from_secs(1), async {
+        loop {
+            if queue.searches_waiting() == 0 {
+                break;
+            }
+            tokio::task::yield_now().await;
+        }
+    })
+    .await
+    .expect("the cancelled waiter should be removed from the queue");
+
+    drop(active_permit);
+}


Kerollmops:
Hello @grep999 👋

Thanks for your help anyway 🌵

PR #6603, opened 2026-08-26
fix: return proper NDJSON from get task documents route

### LINKED ISSUE #6379: Get tasks documents, confusing `content-type`

waterWang:
### What does this PR do?

Fixes #6379. The `GET /tasks/{uid}/documents` route declared `Content-Type: application/x-ndjson` but the response body was not guaranteed to be valid NDJSON.

When documents were added as a JSON array or CSV, `read_json`/`read_csv` writes each object via `serde_json::to_writer` without a trailing newline, producing concatenated JSON (`{...}{...}{...}`). The route returned this raw file content with `application/x-ndjson` — a mismatch between the declared content type and the actual body.

### Fix

In `get_task_documents_file`, after reading the update file, re-serialize each JSON object followed by a newline character. This ensures the response body is always valid NDJSON regardless of how the documents were originally uploaded. Since `serde_json` is configured with `preserve_order`, the field order in each document is preserved.

### File changed

- `crates/meilisearch/src/routes/tasks.rs`: reformat the stored concatenated JSON as proper NDJSON before returning.

### DIFF
--- a/crates/meilisearch/src/routes/tasks.rs
+++ b/crates/meilisearch/src/routes/tasks.rs
@@ -12,7 +12,7 @@ use meilisearch_types::batches::BatchId;
 use meilisearch_types::deserr::query_params::Param;
 use meilisearch_types::deserr::DeserrQueryParamError;
 use meilisearch_types::error::deserr_codes::*;
-use meilisearch_types::error::{InvalidTaskDateError, ResponseError};
+use meilisearch_types::error::{Code, InvalidTaskDateError, ResponseError};
 use meilisearch_types::index_uid::IndexUid;
 use meilisearch_types::star_or::{OptionStarOr, OptionStarOrList};
 use meilisearch_types::task_view::TaskView;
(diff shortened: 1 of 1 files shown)


Kerollmops:
Hello @grep999 👋

Thanks for your help anyway 🌵

PR #6604, opened 2026-08-26
Flush fid_word_count for the last document of each thread in reindex

### LINKED ISSUE #6600: Reindexing after stop-words change leaves stale counts in `field_id_word_count_docids`
**Describe the bug**

Hi. On my localhost, i saw this issue when `curl -X PUT "$MEILI_URL/indexes/movies/settings/stop-words"` would fail continuously even after creating a new index. I debugged it, put some logs, and dumped the relevant databases and found that the issue is with `reindex` fn. Here is the high level description of the issue followed by some more details in Steps to Reproduce section.

When stop words are changed, the settings re-index fails to update `field_id_word_count_docids`(field_id, count) for some documents. In other words, if new stop words should change the count of words against a field X, this doesn't happen for some documents, leaving the old(now stale) count in the `field_id_word_count_docids` for some documents. This may probably affect the exactness ranking rule but i didn't test that

On development run, re-indexing fails when `merge_cbo_bitmaps` get called. The error is `del is not a subset of current, which must be impossible.` because of

```rust
            debug_assert!(
                del.is_subset(&current),
                "del is not a subset of current, which must be impossible."
            );

```
 On release run, re-indexing doesn't fail but `field_id_word_count_docids` does have stale entries.

**The exact issue is as follows:**

For the context, here is the code flow when re-indexing happens after stop words are changed:

```
reindex >
extract::extract_all_settings_changes >
WordDocidsExtractors::run_extraction_from_settings >
settings_change_extract > WordDocidsSettingsExtractorsData.process >
WordDocidsExtractors::extract_document_from_settings_change 
```

`process` runs a loop and spawns a thread, each thread handling bunch of documents

in `extract_document_from_settings_change`, the two tokenizer(old and new) runs(~808 line) and `insert_del_u32` and `insert_add_u32` is called for inserting the "to be deleted" entries and "to be added" entries in the `word_docid` related databases. In both these fns, `flush_fid_word_count` gets called for each word(?or document) that updates `field_id_word_count_docids`. `flush_fid_word_count` runs on **first** word of the next document. But for the last document in the thread, there is no next document, hence there is no `flush_fid_word_count` when the last document is processed by the thread. This leaves a stale entry for last documents processed by each thread. If there are N threads, then N document are affected.

So flushing `fid_word_count` at the end is the fix and i tested it locally. This fixes the issue for new indexes but for existing indexes, some kind of repair may be needed

I will create a PR after proper testing and possibly add a test case

**Steps to Reproduce**

- create an index for the movies.json file
```curl
curl -X POST "$MEILI_URL/indexes/movies16/documents" \
    -H "Authorization: Bearer $MEILI_KEY" \
    --data-binary @movies.json
```

- set stop words = ["The", "the", "a", "an", "to", "in", "of"]
```curl
curl -X PUT "$MEILI_URL/indexes/movies16/settings/stop-words" \
-H "Authorization: Bearer $MEILI_KEY" \
--data '["The", "the", "a", "an", "to", "in", "of"]'
```

- dump the `field_id_word_count_docids` database. I have written a script for this in meilitool. This is what i get(showing only the affected entries)
```
docid 15799 -> field_id=2 word_count=18
```

The document 15799 is this. Was able to find this by dumping `word_docids` and some tallying. Confirmed by actually printing in `tracing` :
```json
{
  "id": 60422,
  "title": "Another Happy Day",
  "overview": "A wedding at her parents' Annapolis estate hurls high-strung Lynn into the center of touchy family dynamics.",
  "genres": [
    "Comedy",
    "Drama"
  ],
  "poster": "https://image.tmdb.org/t/p/w500/yOS14Pr563QEUd3bath9LwhhId3.jpg",
  "release_date": 1321574400
}
```
The overview field(field 2) has 18 words but after setting stop words, should have 16

- now set stop words = ["The"]
```curl
curl -X PUT "$MEILI_URL/indexes/movies16/settings/stop-words" \
-H "Authorization: Bearer $MEILI_KEY" \
--data '["The"]'
```
The `field_id_word_count_docids` will still show word count 18 for field 2

```
docid 15799 -> field_id=2 word_count=18
```

- try different stop words. Sometimes the doc `15799` gets fixed but some other doc gets affected.

Apart from the stale entries, i also observed that `field_id_word_count_docids` will record duplicate entries for some documents e.g. from my dump:
```
repeated: repeated=[(15799, 2), (23399, 2)] -> (doc_id, field_id)
```

The above means that for doc 15799 and 23399 `field_id_word_count_docids` recorded two different word counts for field 2

**Expected behavior**

`field_id_word_count_docids` must have updated counts for every word of every document after reindexing and `field_id_word_count_docids` must have one count for a given (doc_id, field_id, word) combination

**Meilisearch version:**
[e.g. v1.53.1]

harismuzaffer:
## Description

When stop words are changed, during re-indexing, the fn `extract_document_from_settings_change` doesn't flush `fid_word_count` for the last document processed by each thread. On the other hand, `extract_document_change`(called by normal indexing) does it. This results in stale entries in `field_id_word_count_docids` database for last documents processed by each thread. This is because flush happens on the first word of the next document but for the last document in the thread, there is no next document and consequently this misses flush for the last document

## Related issue

Fixes #6600 

## What does this PR do?

This PR simply calls `flush_fid_word_count` in `extract_document_from_settings_change` at the end. This ensures that `field_id_word_count_docids` is properly updated for the current thread. I also added a test case that asserts the same by updating stop words that would change `field_id_word_count_docids`

## Generative AI tools

    - *Claude was used for knowledge purpose but not for generating the code(or tests)*

## Requirements

    - [ ] Test that any impacted DB still works as expected after using `--upgrade-db` on a DB created with the last released Meilisearch

### DIFF
--- a/crates/milli/src/update/new/extract/searchable/extract_word_docids.rs
+++ b/crates/milli/src/update/new/extract/searchable/extract_word_docids.rs
@@ -839,7 +839,10 @@ impl WordDocidsExtractors {
             }
         }
 
-        Ok(())
+        // Without this, `fid_word_count` for the last document of this thread is never written
+        let buffer_size = size_of::<FieldId>();
+        let mut buffer = BumpVec::with_capacity_in(buffer_size, &context.doc_alloc);
+        cached_sorter.flush_fid_word_count(&mut buffer)
     }
 }
 
--- a/crates/milli/src/update/test_settings.rs
+++ b/crates/milli/src/update/test_settings.rs
@@ -1,5 +1,6 @@
 use big_s::S;
 use heed::types::Bytes;
+use heed::RoTxn;
 use maplit::{btreemap, btreeset};
 use meili_snap::snapshot;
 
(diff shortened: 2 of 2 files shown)


ManyTheFish (approved):
Looks good! Thank you!

PR #6610, opened 2026-09-01
Tie authorization to index opening

ManyTheFish:
## Related issue

Related to https://linear.app/meilisearch/issue/ENGPROD-2699

## Generative AI tools

## Changelog

This PR forces the `IndexScheduler::user_index` method to take an `AuthFilter` and check if opening the index is allowed.
Introduce an `IndexScheduler::user_index_unsecured` method used in asynchronous processes where we can't provide the `AuthFilter`.

### Breaking change

When using the foreign filters experimental feature, it will no longer be possible to access the foreign index for filtering and hydration if the provided key is not allowed to access it, an `invalid_api_key` error will be returned instead.


crates/index-scheduler/src/lib.rs
```diff
@@ -581,7 +583,18 @@ impl IndexScheduler {
...
+            return Err(Error::AuthenticationError(AuthenticationError::InvalidToken));
+        }
+
+        self.user_index_unsecured(name)
+    }
+
+    /// Same as `user_index` but without checking authorization.
+    ///
+    /// This function is not meant to be used outside of the index scheduler.
+    pub(crate) fn user_index_unsecured(&self, name: &str) -> Result<Index> {
```
dureuill, line 597: nitpick: wondering if we want to keep `unsecured` as the modifier here.

perhaps `unchecked`?

`http-client` has `IpPolicy::danger_always_allow()`

so `danger_user_index_always_allow`? 

🤔 not sure it's better to be honest 😅 

unsecured correctly convey the sense of danger already

crates/meilisearch/src/routes/render.rs
```diff
@@ -120,22 +122,22 @@ pub async fn render_post(
...
+                if !auth_filter.is_index_authorized(template_index_uid) {
                     return Err(AuthenticationError::InvalidToken.into());
                 }
             }
             (Some(template_index_uid), Some(input_index_uid)) => {
                 // check both indexes
-                if !index_scheduler.filters().is_index_authorized(template_index_uid)
-                    || !index_scheduler.filters().is_index_authorized(input_index_uid)
+                if !auth_filter.is_index_authorized(template_index_uid)
+                    || !auth_filter.is_index_authorized(input_index_uid)
```
dureuill: are these checks made redundant by passing the auth filter on index opening?

crates/meilisearch/src/search/federated/perform.rs
```diff
@@ -135,8 +137,9 @@ pub async fn perform_federated_search(
 
     // 2.2. concurrently execute local queries
     progress.update_progress(FederatingResultsStep::ExecuteLocalSearch);
-    let params = SearchByIndexParams {
+    let params: SearchByIndexParams = SearchByIndexParams {
```
dureuill: I would imagine that explicitly specifying the type here is redudant since the right hand is a struct literal initialization?


```suggestion
    let params = SearchByIndexParams {
```

dureuill (dismissed):
I really like it ❤️ 

- I left a few comments that you are at liberty to take or not
- looks like the CI is red for some obscure reason?

crates/index-scheduler/src/lib.rs
ManyTheFish: I hesitated with unchecked, but unchecked relates more to safety than security to me (I may be wrong).

`danger_user_index_always_allow` eew! But I am ok to change it. The ugliness of the name can contribute to reducing usage

crates/meilisearch/src/routes/render.rs
ManyTheFish: done

crates/index-scheduler/src/lib.rs
dureuill: `unsecured` is probably fine

dureuill (approved):
looking good

PR #6622, opened 2026-09-08
remove HashMap transmute in indexer

dureuill:
## Generative AI tools

### DIFF
--- a/crates/milli/src/update/new/extract/cache.rs
+++ b/crates/milli/src/update/new/extract/cache.rs
@@ -492,6 +492,10 @@ where
         // Once we merged all of the spilled bitmaps we must also
         // fetch the entries from the non-spilled entries (the HashMaps).
         for (source_id, map) in maps.iter_mut() {
+            debug_assert!(
+                !(map.get_mut(first_key).is_some() && first_entry.source_id == *source_id),
+                "A thread should not have spiled a key that has been inserted in the cache"
+            );
             if first_entry.source_id != *source_id {
                 if let Some(new) = map.get_mut(first_key) {
                     output.union_and_clear_bbbul(new);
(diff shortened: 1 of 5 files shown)


crates/milli/src/update/new/extract/cache.rs
```diff
@@ -527,10 +492,6 @@ where
         // Once we merged all of the spilled bitmaps we must also
         // fetch the entries from the non-spilled entries (the HashMaps).
         for (source_id, map) in maps.iter_mut() {
-            debug_assert!(
-                !(map.get(first_key).is_some() && first_entry.source_id == *source_id),
-                "A thread should not have spiled a key that has been inserted in the cache"
-            );
             if first_entry.source_id != *source_id {
```
Kerollmops, line 499: Can we keep this debug assert by using a `get_mut`, instead?

crates/milli/src/update/new/extract/cache.rs
```diff
@@ -672,18 +645,18 @@ impl DelAddRoaringBitmap {
...
+        if let Some(ref mut bbbul) = del.take().filter(|del| !del.is_empty()) {
             let del = self.del.get_or_insert_with(RoaringBitmap::new);
             let mut iter = bbbul.iter_and_clear();
             while let Some(block) = iter.next_block() {
                 del.extend(block);
             }
         }
 
-        if let Some(ref mut bbbul) = add.take() {
+        if let Some(ref mut bbbul) = add.take().filter(|add| !add.is_empty()) {
```
Kerollmops, line 663: Do we still need to do the filter here as the Option will be taken and replaced by a None?

Kerollmops (changes requested):

crates/milli/src/update/new/extract/cache.rs
dureuill: yes it is needed otherwise we could be having a `Some(empty_bbbul)` leftover from a previous cleaning of the del_add_bbbul, which could create an empty del or an empty add

crates/milli/src/update/new/extract/cache.rs
dureuill: done

Kerollmops (approved):
LGTM, thanks 🇵🇹

PR #6624, opened 2026-09-09
Publish Docker images to GitHub Container Registry alongside Docker Hub

### LINKED ISSUE #6015: Add container image to ghcr
To reduce docker hub nasty limitation and to enjoy more option w/o using a mirror. :)

ximing:
Fixes #6015

## What does this PR do?

This PR extends the Docker publish workflow (`.github/workflows/publish-docker-images.yml`) so that every Meilisearch image (community and enterprise editions, all tags including `nightly`) is also published to the GitHub Container Registry, in addition to Docker Hub:

- `ghcr.io/meilisearch/meilisearch`
- `ghcr.io/meilisearch/meilisearch-enterprise`

## Motivation

Docker Hub is hard to reach from some networks (rate limits, restricted regions), while `ghcr.io` is often directly pullable. Publishing to GHCR gives users in those environments an official, trusted source for the images — no third-party mirror needed. See #6015.

## How it works

- The `build` job pushes the platform digests to **both registries in a single build step** (`docker/build-push-action` with both image names and `push-by-digest=true`), so there is **no extra build cost**: layers are built once and only uploaded twice.
- The `merge` job runs `docker/metadata-action` with both image names, so the tag list (`vX.Y.Z`, `vX.Y`, `vX`, `latest`, `nightly`) is generated for both registries, and the `imagetools create` manifest step tags both registries at once.
- Cosign signing covers both registries automatically, since it iterates over the tags produced by the metadata step.
- GHCR authentication uses the built-in `GITHUB_TOKEN` with a `packages: write` permission — **no new secrets are required**.
- The GHCR repository name is derived from `github.repository_owner` instead of being hardcoded, which keeps the workflow correct for forks (useful for testing).

## Verification

I tested this end-to-end on my fork ([run](https://github.com/ximing/meilisearch/actions/runs/34303888227), with the Docker Hub parts stripped since forks don't have the `DOCKERHUB_*` secrets):

- The single build step successfully pushed the same digests to two GHCR repositories
- The merge job created the multi-arch (`linux/amd64` + `linux/arm64`) manifest list on both registries with identical digests
- Both registries' tags were signed with Cosign in keyless mode

```
$ docker buildx imagetools inspect ghcr.io/ximing/meilisearch:nightly
Name:      ghcr.io/ximing/meilisearch:nightly
MediaType: application/vnd.oci.image.index.v1+json
Platform:  linux/amd64
Platform:  linux/arm64
```

## Note for maintainers

After the first push, the two packages (`meilisearch`, `meilisearch-enterprise`) will be created as **private** by default. They need to be made public once via the package settings (or org-wide for `ghcr.io/meilisearch/*`) so users can pull them anonymously.

### DIFF
--- a/.github/workflows/publish-docker-images.yml
+++ b/.github/workflows/publish-docker-images.yml
@@ -1,4 +1,4 @@
-name: Publish images to Docker Hub
+name: Publish images to Docker Hub and GitHub Container Registry
 
 on:
   push:
@@ -28,12 +28,15 @@ jobs:
             runner: ubuntu-24.04-arm
           - edition: community
             registry: getmeili/meilisearch
+            ghcr-registry: ghcr.io/${{ github.repository_owner }}/meilisearch
             feature-flag: ""
           - edition: enterprise
             registry: getmeili/meilisearch-enterprise
+            ghcr-registry: ghcr.io/${{ github.repository_owner }}/meilisearch-enterprise
             feature-flag: "--features enterprise"
 
-    permissions: {}
+    permissions:
+      packages: write # This is needed to push the images to GitHub Container Registry
     steps:
       - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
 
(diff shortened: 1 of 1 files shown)


Kerollmops:
Hello @ximing 👋

As we are receiving many contributions, we cannot consider those that do not follow the PR template. Additionally, modifying the publish workflow is quite tricky, so we would prefer not to.

Thanks for your help anyway 🌵

PR #6628, opened 2026-09-10
Fix concurrent geo-sort OOM by caching shared Arc<RTree>

### LINKED ISSUE #6596: Concurrent geo-sorted searches on an 11M-document index OOM-kill the process: each search deserializes the entire geo RTree
Concurrent geo-sorted searches on an 11M-document index OOM-kill the process: each search deserializes the entire geo RTree

## Summary

On an index of ~11.2M documents each carrying a `_geo` point, running several geo-sorted searches concurrently makes the meilisearch process allocate multiple GB of anonymous memory per query and get OOM-killed within seconds. A single such query survives (with a ~2–3 GB RSS spike and ~1.8 s latency). The process then restart-loops if the query mix continues.

Root cause appears to be in `milli`: for any geo-sorted search with ≥1000 candidates (the `GeoSortStrategy::Dynamic(1000)` default), `fill_cache` calls `index.geo_rtree(txn)`, which **bincode-deserializes the whole-corpus `RTree<GeoPoint>` from LMDB into anonymous heap memory, per search**. The only cache is a field on the per-search `GeoSort` struct (`search/new/geo_sort.rs`), so N concurrent geo-sorted searches hold N full copies of the tree.

## Environment

- meilisearch `v1.53.1` and `v1.52.3` (official Docker images `getmeili/meilisearch:v1.53.1` / `:v1.52.3`) — identical behavior on both
- Single node, container memory limit 4 GiB (`mem_limit`), host 16 GB RAM with ~7 GB free at incident time
- Index: 11,194,620 documents, every doc has `_geo`; searchable text fields; synonyms configured; all other settings default

## Observed behavior

Concurrency-10 benchmark of typeahead-style queries (3/6/10-char prefixes) with `"sort": ["_geoPoint(lat,lng):asc"]`:

- Process RSS climbs ~0.45 GiB → ~4 GiB within seconds, then the kernel kills it:
  ```
  Memory cgroup out of memory: Killed process 1692285 (meilisearch)
  total-vm:2175979948kB, anon-rss:4091448kB, file-rss:31200kB
  oom-kill:constraint=CONSTRAINT_MEMCG, ... task=meilisearch
  ```
  (10 such kills recorded across two versions and three query variants; `total-vm` is ~2.17 PB at each kill — presumably dominated by sparse LMDB maps, listed here for completeness.)
- Clients see dropped connections (`ConnectError` / `RemoteProtocolError`), not HTTP errors.
- A single worst-case query (3-char prefix, ~100k+ candidates) in isolation: HTTP 200 in ~1.76 s, RSS spike ~2–3 GiB, process survives.
- Bounding the candidate set does **not** help: adding `"filter": "_geoRadius(lat,lng,50000)"` still crashes under concurrency, as does restricting to 6/10-char prefixes — consistent with the whole-tree load being independent of the filtered candidate set.
- Baselines on the same index without geo sort: p95 61–77 ms at concurrency 10, zero errors, steady RSS ~100 MiB.

## Code pointers (current `main`)

- `crates/milli/src/documents/geo_sort.rs` — `fill_cache()` calls `index.geo_rtree(txn)?` and inserts the result into `self.rtree` (per-`GeoSort`-instance cache only).
- `crates/milli/src/index.rs` — `geo_rtree()` returns `SerdeBincode<RTree<GeoPoint>>` deserialized from the `main` DB: a full owned copy per call.
- `crates/milli/src/search/new/geo_sort.rs` — the `GeoSort` ranking rule (constructed per search) holds `rtree: Option<RTree<GeoPoint>>`; no cross-search sharing exists (no `Arc`/shared cache anywhere in the geo-sort path).

## Suspected fix direction

Share a single `Arc<RTree<GeoPoint>>` per index across concurrent searches (with invalidation on index update), or serve nearest-neighbor iteration from a zero-copy/memory-mapped representation, instead of one owned tree per search. Happy to test a patch against our corpus.

## Related

- #3581 (`_geoPoint sort is slow`) — the latency dimension of the same path (>1 s on 6.3M docs; we measure ~1.8 s at 11.2M). This report adds the memory/concurrency dimension: at ≥10 concurrent queries the per-query tree copies exhaust any practical container memory limit and the kernel kills the process.

## Reproduction sketch

1. Create an index with ~11M documents, each with a `_geo` field (our corpus is Canadian civic addresses; any uniform-ish point cloud should behave the same). Settings: `_geo` in `filterableAttributes` and `sortableAttributes`.
2. From 10 concurrent connections, issue searches with short `q` prefixes (2–4 chars, e.g. street-number prefixes that match 10k–1M docs) and `"sort": ["_geoPoint(<lat>,<lng>):asc"]`, `limit: 8`.
3. Watch RSS: it reaches a 4 GiB container limit in seconds; `journalctl -k` shows the memcg OOM kills.

I can share the exact benchmark harness (Python/httpx) if useful.

harriteja:
## Summary
Concurrent `_geoPoint` sorts each deserialized a full copy of the geo RTree into the per-search ranking rule. On large indexes that tree is huge, so parallel searches OOMed (see #6596).

## Approach
- Keep a content-addressed in-memory cache on `Index`: hash of the LMDB bytes → `Arc<RTree<GeoPoint>>`.
- `geo_rtree` now returns `Option<Arc<RTree<_>>>` so searches share one deserialized tree.
- Clear the cache on `put_geo_rtree` / `delete_geo_rtree` (and when the key is absent).
- Indexing paths that mutate the tree (`merger`, `typed_chunk`) clone out of the `Arc` before writing back.

## Test plan
- [x] `cargo test -p milli geo_rtree_cache_returns_shared_arc`
- [x] Existing milli geo_sort / geo filter tests
- [x] Manual: concurrent geo-sorts against a large geo index should no longer multiply RTree memory by concurrency

Fixes #6596

### DIFF
--- a/crates/milli/src/documents/geo_sort.rs
+++ b/crates/milli/src/documents/geo_sort.rs
@@ -1,4 +1,5 @@
 use std::collections::VecDeque;
+use std::sync::Arc;
 
 use heed::types::{Bytes, Unit};
 use heed::{RoPrefix, RoTxn};
@@ -70,7 +71,7 @@ pub fn fill_cache(
     ascending: bool,
     target_point: [f64; 2],
     field_ids: &Option<[u16; 2]>,
-    rtree: &mut Option<RTree<GeoPoint>>,
+    rtree: &mut Option<Arc<RTree<GeoPoint>>>,
     geo_candidates: &RoaringBitmap,
     cached_sorted_docids: &mut VecDeque<(u32, [f64; 2])>,
 ) -> crate::Result<()> {
@@ -79,7 +80,7 @@ pub fn fill_cache(
     // lazily initialize the rtree if needed by the strategy, and cache it in `self.rtree`
     let rtree = if strategy.use_rtree(geo_candidates.len() as usize) {
         if let Some(rtree) = rtree.as_ref() {
-            // get rtree from cache
+            // get rtree from per-search cache (already an Arc shared with Index)
             Some(rtree)
         } else {
             let rtree2 = index.geo_rtree(txn)?.expect("geo candidates but no rtree");
@@ -141,7 +142,7 @@ pub fn next_bucket(
     ascending: bool,
     target_point: [f64; 2],
     field_ids: &Option<[u16; 2]>,
-    rtree: &mut Option<RTree<GeoPoint>>,
+    rtree: &mut Option<Arc<RTree<GeoPoint>>>,
     cached_sorted_docids: &mut VecDeque<(u32, [f64; 2])>,
     geo_candidates: &RoaringBitmap,
     parameter: GeoSortParameter,
--- a/crates/milli/src/index.rs
+++ b/crates/milli/src/index.rs
@@ -761,8 +761,10 @@ impl Index {
     /// Returns the `rtree` which associates coordinates to documents ids.
     ///
     /// Concurrent callers share a single deserialized `Arc` when the LMDB payload
-    /// is unchanged (content-addressed cache). Writers that mutate the tree must
-    /// clone out of the `Arc` before modifying it.
+    /// is unchanged (content-addressed cache). Cold-cache misses are single-flighted:
+    /// the mutex is held across deserialize so only one thread materializes the tree
+    /// for a given content hash. Writers that mutate the tree must clone out of the
+    /// `Arc` before modifying it.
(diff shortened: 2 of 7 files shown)


Kerollmops:
Hello @harriteja 👋

However, your approach looks interesting, but it is too complex to use in Meilisearch and maintain. The best approach to avoid loading too much from disk would probably be to use LMDB to store the GEO nodes, avoiding the need to load the entire geo data structure for each geo query.

Thanks for your help anyway 🌵

PR #6631, opened 2026-09-14
Introduce a verbosity mode to the Progress trace

ManyTheFish:
Related to [linear.app/meilisearch/issue/EXP-1326](https://linear.app/meilisearch/issue/EXP-1326)

## Changelog

- Introduce an internal verbosity level to the progress trace, This verbosity mode allows:
  - creating a quiet progress trace that will not register any log when it's not used (tests, documents get)
  - (future) adding more steps with a debug verbosity mode allowing to dig deeper without impacting all the requests
- Change the search steps to be better split between multi-search queries
  - multi search trace will now be split by query id allowing to identify which query impacted the most the search time
  
  Example:
```json
"performanceDetails": {
  "wait in queue": "1.36ms",
  "preprocess filters > prepare": "985.00µs",
  "preprocess filters > send to remote": "220.00µs",
  "preprocess filters > execute local > evaluate filter": "139.00µs",
  "preprocess filters > execute local": "1.61ms",
  "preprocess filters > wait for remote": "35.97ms",
  "preprocess filters": "39.01ms",
  "process > prepare": "96.00µs",
  "process > send to remote": "118.00µs",
  "process > execute local > load field ids map": "63.00µs",
  "process > execute local > query[0] > evaluate filter": "84.00µs",
  "process > execute local > query[0] > tokenize query": "163.00µs",
  "process > execute local > query[0] > evaluate query": "116.00µs",
  "process > execute local > query[0] > keyword ranking": "634.00µs",
  "process > execute local > query[0]": "1.52ms",
  "process > execute local > pin hits": "29.00µs",
  "process > execute local > format": "36.00µs",
  "process > execute local": "2.90ms",
  "process > wait for remote": "37.24ms",
  "process > merge": "82.00µs",
  "process": "40.45ms",
  "hydrate > send to remote": "38.00µs",
  "hydrate > execute local": "35.00µs",
  "hydrate > wait for remote": "41.00µs",
  "hydrate": "295.00µs",
  "merge facets": "51.00µs"
}
```


crates/meilisearch/src/routes/indexes/search.rs
```diff
@@ -603,14 +604,14 @@ pub async fn search_with_url_query(
...
+        if let Ok((search_result, _)) = search_result.as_ref() {
             aggregate.succeed(search_result);
         }
         analytics.publish(aggregate, &req);
 
-        debug!(request_uid = ?request_uid, returns = ?&search_result, progress = ?progress.accumulated_durations(), "Search get");
+        let (search_result, progress) = search_result?;
 
-        let search_result = search_result?;
+        debug!(request_uid = ?request_uid, returns = ?&search_result, progress = ?progress, "Search get");
```
dureuill, line 614: it is OK to debug progress directly here? rather than displaying the accumulated durations?

crates/milli/src/progress.rs
```diff
@@ -50,9 +64,29 @@ struct InnerProgress {
...
+        Self { steps: Arc::new(RwLock::new(InnerProgress::default())), verbosity_mode }
+    }
+
+    /// Create a new progress with quiet verbosity mode.
+    /// This will not register any steps.
+    pub fn quiet() -> Self {
+        Self::new(ProgressVerbosityMode::Quiet)
+    }
+
+    /// Recreate the progress with the same verbosity and timestamp mode.
```
dureuill:
```suggestion
    /// Recreate the progress with the same verbosity mode.
```

crates/milli/src/steps.rs
```diff
@@ -0,0 +1,146 @@
...
+        WaitForRemote,
+        Merge,
+        Personalize,
+        Format,
+        PinHits,
+    }
+}
+
+#[derive(Default, Debug, Clone, Copy)]
+pub struct QueryStep(pub usize, pub usize);
```
dureuill: consider a named struct here?


```suggestion
pub struct QueryStep {
    pub current: usize,
    pub total: usize
}
```

crates/milli/src/search/steps.rs
```diff
@@ -0,0 +1,146 @@
...
+pub struct QueryStep(pub usize, pub usize);
+
+impl crate::progress::Step for QueryStep {
+    fn name(&self) -> std::borrow::Cow<'static, str> {
+        format!("query[{}]", self.0).into()
+    }
+
+    fn current(&self) -> u32 {
+        self.0 as u32
+    }
```
dureuill, line 80: I would expect `self.current` to be constrained to less than `self.total`.

To achieve this, I guess the `current` and `total` fields should not be public.

Instead you should have:

- a `fn new(total: usize) -> Self { current: 0, total: usize }`
- the existing `current` and `total`
- a `fn next(&mut self)` that increments `self.current` if it is lower than total

crates/milli/src/progress.rs
```diff
@@ -276,6 +319,9 @@ macro_rules! make_enum_progress {
             }
         }
     };
+    ($visibility:vis enum $name:ident { $($variant:ident,)+ }) => {
```
dureuill, line 322: this forces all variant to have a vis mode or all of them to not have it. Is that OK?

crates/meilisearch/src/documents_retrieval/mod.rs
```diff
@@ -93,18 +97,28 @@ impl DocumentSearch {
...
-            return Ok(DocumentSearchResult::Federated(Box::new(search_result)));
+            return Ok(DocumentSearchResult::Federated(
+                Box::new(search_result),
+                progress.accumulated_durations(),
+            ));
         }
 
         // Multi-search
         let search_results: Result<_, (ResponseError, _)> = async {
+            let mut multi_search_progress = Some(progress);
```
dureuill: I don't like that we're using an option here:

1. it is supposed to always be `Some`
2. it creates a codepath with a panic (the `unwrap`)

Can we instead use `let progress = std::mem::replace(&mut multi_search_progress, multi_search_progress.recreate());`?

crates/meilisearch/src/documents_retrieval/mod.rs
```diff
@@ -134,25 +148,29 @@ impl DocumentSearch {
...
-                    progress,
+                    &progress,
                     auth_filter,
                 )
                 .await
                 // Fixup the query index for the error
                 .map_err(|(err, _)| (err, Some(query_index)))?;
 
+                progress_by_query
+                    .insert(format!("query[{}]", query_index), progress.accumulated_durations());
```
dureuill, line 160: should we store directly the `query_index` rather than a string built from it in the `progress_by_query` map?

if not, can we comment on why?

crates/milli/src/search/steps.rs
```diff
@@ -0,0 +1,146 @@
+use crate::make_enum_progress;
+
```
dureuill, line 2: I don't know how I feel about moving most steps to the same place, can you comment on the rationale?

dureuill (commented):
Hello, thank you for this PR! I have a few comments and questions (see inline comments)

crates/meilisearch/src/routes/indexes/search.rs
ManyTheFish: Progress doesn't implement Debug or Display; what you have here are the accumulated durations returned by DocumentRetrieval

crates/milli/src/progress.rs
ManyTheFish: Yeah, I saw this behavior, and I was thinking that if you started providing the verbosity mode, specifying all of them would avoid confusion. So I kept it.

crates/milli/src/search/steps.rs
ManyTheFish: the rational I had is outdated, rolling back

crates/meilisearch/src/documents_retrieval/mod.rs
ManyTheFish: Oops, you are right, my bad. There is a simpler trick: Rust allows you to reassign variables without having to mem::replace.

see: [`d4550d5` (this PR)](https://github.com/meilisearch/meilisearch/pull/6631/commits/d4550d57e530947955bdf9921518b0b6881b24bd)

crates/milli/src/progress.rs
dureuill: ok for now

dureuill (approved):
let's roll 🛹

